Issue #123, opened 2015-06-26
jobStoreTest's tearDown method not called reliably


cket:
Our S3 console is full of test buckets from AWSjobStore, and when running fileJobStore temp files can be left behind.

cket:
According to the documentation, the teardown method will be called even if there is an exception in the test method as long as the exception was not in the setup method. It's possible that keyboard interrupt exceptions in the setup method are causing the leftover buckets, especially since the aws setup runs for several seconds while it waits for s3 buckets and simpleDB domains

hannes-ucsc:
Resolved by #190

Issue #70, opened 2015-05-19
Revisit job tree creation and destruction


hannes-ucsc:
Now that we have different storage backends, there should be a command to destroy a backend. I also think we should enforce that a --resume parameter is passed when jobTree is supposed to resume a tree. Otherwise it should fail if the persisted job tree already exists.

cket:
The deleteJobStore callback seems to do this, but it is only used in tests so far. We could call this at the end of MainLoop if the user passes a --clean-up flag.

hannes-ucsc:
I'd prefer a flag with an argument, e.g. with the following synopsis
`[--clean=never|onerror|always]`.

If you only do that flag, this issue wouldn't be fully taken care of. Ideally, I'd also want the `--resume` parameter mentioned above.

hannes-ucsc:
Resolved by #190

PR #190, opened 2015-07-30
Initial commit of "clean" and "resume" flags (resolves #154, #123 and #70)

### LINKED ISSUE #154: SortTest and variations don't call jobStore.delete()
Related to 123. They rely on shutil to clean up, which obviously doesn't delete AWS test artifacts. This means anytime anyone runs awsSortTest or awsSortTestWithMesos, buckets and simpleDB domains are left behind. Additionally, whenever a commit it pushed to jobTree, Jenkins runs both tests and subsequently leaves buckets and domains behind.

hannes-ucsc:
Continued from https://github.com/BD2KGenomics/toil-old/pull/13


hannes-ucsc:
Isn't there already a branch on this? So it should be in-progress? Assigning to cket for decision.

PR #198, opened 2015-08-03
Improve serialization efficiency. 

cket:
Job is pickled, compressed, and base64 encoded before being put into SDB

This has been successfully tested up to 1000 children jobs, and will likely max out around 8000.


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
I would like to see a unit test that proves that we can serialize a job with 5000 children and another one proving that we can't serialize 10000.

src/toil/jobStores/awsJobStore.py
```diff
@@ -169,7 +173,9 @@ def delete( self, jobStoreID ):
             log.debug( "Deleting %d file(s) associated with job %s", len( items ), jobStoreID )
             for attempt in retry_sdb( ):
                 with attempt:
-                    self.versions.batch_delete_attributes( { item.name: None for item in items } )
+                    loops = int(math.ceil(float(len(items))/batch_delete_limit))
```
hannes-ucsc: Don't use floating-point math for stuff like this. Integer division has the desired truncation properties. Besides, look at 

http://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks-in-python

for inspiration on how to partition a list. The general idea is to avoid division and multiplication and simple increment an index by the chunk size.

hannes-ucsc: Also, you shouldn't only retry the current batch. So the retry loop and the for-each batch loop need to be swapped.

hannes-ucsc:
The batch deletion definitely needs coverage. Remind me to show you tomorrow on how to determine test coverage. I would like to see the 5000 vs 10000 test I mentioned earlier but maybe that's overkill.

src/toil/jobStores/awsJobStore.py
```diff
@@ -829,35 +653,25 @@ def fromItem( cls, item, jobStoreID=None ):
...
         """
-        item = self.toDict( )
-        if parentJobStoreID is not None:
-            item[ 'parentJobStoreID' ] = parentJobStoreID
-        item = ((k, self.toItemTransform[ k ]( v )) for k, v in item.iteritems( ))
-        return { k: v for k, v in item if v is not None }
+        item = {}
+        serializedAndEncodedJob = base64.b64encode(bz2.compress(cPickle.dumps(self)))
+        # this convoluted expression splits the string into chunks of 1024 - the max value for an attribute in SDB
+        jobChunks = [serializedAndEncodedJob[i:i+1024] for i in range(0, len(serializedAndEncodedJob), 1024)]
```
hannes-ucsc: And this is the right way to partition a string. The same combination of slicing and range() with non-unit increments can be used to partition a list, too.

src/toil/jobStores/awsJobStore.py
```diff
@@ -169,7 +173,10 @@ def delete( self, jobStoreID ):
             log.debug( "Deleting %d file(s) associated with job %s", len( items ), jobStoreID )
             for attempt in retry_sdb( ):
                 with attempt:
-                    self.versions.batch_delete_attributes( { item.name: None for item in items } )
+                    batch_deletable_jobs = [items[i:i+batch_delete_limit] for i in range(0, len(items), batch_delete_limit)]
```
hannes-ucsc: Loop nesting is still inside out.

hannes-ucsc:
Jenkins, test this please.

PR #199, opened 2015-08-03
Encrypt AWS job store with SSE

cket:
Users can specify a file containing a key to use with S3's server side encryption on the AWS jobStore.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -85,7 +85,7 @@ def _addOptions(addGroupFn, defaultStr):
                 help="Absolute path to directory where temporary files generated during the Toil run should "
                      "be placed. default=%s" % defaultStr)
     addOptionFn("--sseKey", dest="sseKey", default=None,
-            help="Key to be used for encryption of files stored to cloud jobStore. Default=%s" % defaultStr)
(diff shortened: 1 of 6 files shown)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Needs unit test coverage.

hannes-ucsc:
According to [1] the key and key MD5 headers have to passed when the URL is retrieved, not when it is generated.  But we can't encumber hot deployment with encryption. This means that we need to 

1) Make encryption optional at least for the one job store API call that is used by hot deployment (i.e. add a encrypt=False keyword argument to that call.

2) Remove the addition of the encryption headers from getPublicUrl() such that the public URL does not require keys.

[1]: http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html

hannes-ucsc:
That should be encrypt=True since by default we want to encrypt if the store supports it.

hannes-ucsc:
I would also want early validation of the encryption key length. Either via argparse's 'type' parameter to add_argument() method or manually.

hannes-ucsc:
How do we address the chicken-egg problem with the config file? It contains the path to the key file but it is encrypted using the key in that key file.

PR #222, opened 2015-08-10
Specify memory reqs as string (resolves #172)

### LINKED ISSUE #172: Allow memory requirements to be specified as resource string

cket:
Byte-based resource requirements can be specified as various byte-based units in addition to just bytes. Adds this functionality to Job's resource requirements as well as the max_log_size and default/max memory/disk toil options. Adds bd2k-python-lib 1.7 as a dependency, and uses the human2byte method provided within that library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ setup(
     author='Benedict Paten',
     author_email='benedict@soe.usc.edu',
     url="https://github.com/BD2KGenomics/toil",
-    install_requires=[ ],
+    install_requires=['bd2k-python-lib>=1.7.dev1'],
     extras_require={
         'mesos': [
             'mesos.interface==0.22.0',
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Tests hung

cket:
Caused by what appears to be a bug in the human2bytes:

>>>sys.maxint
9223372036854775807

>>>str(human2bytes(str(sys.maxint)))
9223372036854775808L

>>>int(str(human2bytes(str(sys.maxint)))
9223372036854775808

In the job constructor, the default argument value for any resource req is sys.max int. I run human2bytes on memory/disk arguments,  and we later test if our resources equal sys.maxint to determine whether to instead set the parameter to it's configurable default. In this bug, they do not equal each other so we pass this huge int as the disk requirement for the job wrapper which mesos obviously cannot allocate, causing the test to hang.

cket:
This seems to be caused by float imprecision in the human2bytes method

hannes-ucsc:
Jenkins, test this please.

PR #235, opened 2015-08-14
Create entry point for Mesos executor (resolves #232)

### LINKED ISSUE #232: Make executor.py an entry point

hannes-ucsc:


hannes-ucsc:
Build hung. Probably because the Mesos slave was not able to find the executor entry point due to the fact that the venv wasn't active.

hannes-ucsc:
Yep.

```
sh: 1: mesos-toil-executor: not found
```

PR #267, opened 2015-08-20
Fix gridEngine (resolves #213)

### LINKED ISSUE #213: Fix GridEngine batch system

jvivian:
Added maxDisk to the gridEngineBatchSystem constructor

replaced undefined values in __str__ with self.values

Fixed typo (curline -> currline)

Added disk to issueBatchJob and self.checkResourceRequest

maxWait not defined in killBatchJobs method, replaced with standard killedJobsQueue.get()

Added shutdown method to gridEngine

Replaced getFromQueueSafely with standard .get() call

Changed "p=str(cpu)" to specify cores to:  ["-pe", "smp", str(cpu)]

CPU only accepts integers. Since TOIL handles floats, the values supplied to cpu will be rounded up.

Added debug messages for logging.

Added explanatory docstring for overridden shutdown() (+19 squashed commits)

### DIFF
--- a/src/toil/batchSystems/gridengine.py
+++ b/src/toil/batchSystems/gridengine.py
@@ -18,6 +18,7 @@ import logging
 import os
 import subprocess
 import time
+import math
 from Queue import Queue
 from threading import Thread
 
@@ -39,9 +40,9 @@ class MemoryString:
 
     def __str__(self):
         if self.unit != 'B':
-            return str(val) + unit
+            return str(self.val) + self.unit
         else:
-            return str(val)
+            return str(self.val)
 
     def byteVal(self):
         if self.unit == 'B':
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
Jenkins, test this please.

src/toil/batchSystems/gridengine.py
```diff
@@ -90,6 +91,8 @@ def getjobexitcode(sgeJobID):
 
         process = subprocess.Popen(args, stdout = subprocess.PIPE,stderr = subprocess.STDOUT)
         for line in process.stdout:
+            logger.debug(args)
```
hannes-ucsc: Are these really needed? If so, they need a little more contextual info in the log message.

jvivian: I would agree that they are not needed. I thought I deleted these commits when I rebased interactively but when I squashed I must have incorporated them.  Is the appropriate step to submit an additional commit that removes those lines?

hannes-ucsc: Yes

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
I rebased, resolved the conflicts and modified the commit message. Also amended commit with the removal of those two debug statements. Then force-pushed.

But I still have a few more comments which I will make as line notes.

src/toil/batchSystems/gridengine.py
```diff
@@ -255,14 +260,20 @@ def getRunningBatchJobIDs(self):
...
             return None
         jobID, retcode = i
         self.updatedJobsQueue.task_done()
         self.currentjobs.remove(jobID)
         return i
-    
+
+    def shutdown(self):
+        """
+        Because the gridEngine Worker is a daemon thread this fn is unnecessary
```
hannes-ucsc: It shouldn't be a daemon thread. We need shutdown() to work because we start and stop a batch system during the unit tests. I hope the same techniques we used in the single-machine batch system will work here.

hannes-ucsc:
Jenkins, test this please.

PR #269, opened 2015-08-20
Refactor cpu to cores (resolves #220)

### LINKED ISSUE #220: Change CPUs to cores for resource requests
It's confusing and ambiguous to use the term CPU, using "core" is slightly better

cket:


hannes-ucsc:
Failure doesn't look like it's related to this change. Transient?

Jenkins, test this please.

hannes-ucsc:
Some of the comments and strings were mangled by this change. A simple context-unaware search-and-replace just doesn't cut it in this case.

As it is, I would not want to merge this

README.md
```diff
@@ -170,7 +170,7 @@ The important arguments to **toilStats** are:
                         The default amount of memory to request for a job (in
                         bytes), by default is 2^31 = 2 gigabytes,
                         default=2147483648
-    --defaultCpu=DEFAULTCPU
+    --defaultCores=DEFAULTCORES
                         The default the number of cpus to dedicate a job.
                         default=1
     --maxCpus=MAXCPUS   The maximum number of cpus to request from the batch
```
hannes-ucsc: What about this one?

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -41,34 +41,34 @@ def supportsHotDeployment():
...
             raise InsufficientSystemResources('memory', memory, self.maxMemory)
         if disk > self.maxDisk:
             raise InsufficientSystemResources('disk', disk, self.maxDisk)
         
-    def issueBatchJob(self, command, memory, cpu, disk):
+    def issueBatchJob(self, command, memory, cores, disk):
         """Issues the following command returning a unique jobID. Command
         is the string to run, memory is an int giving
-        the number of bytes the job needs to run in and cpu is the number of cpus needed for
+        the number of bytes the job needs to run in and cores is the number of cpu cores needed for
```
hannes-ucsc: CPU

hannes-ucsc:
This is mostly plural vs. singular. Like it should be "core usage" and not "cores usage". 

I would also keep the changes to gridengine.py minimal. 1) because John and Alden are working on those modules and 2) we currently don't have good tests that would detect if this change breaks anything.

hannes-ucsc:
Jenkins, test this please.

PR #271, opened 2015-08-20
Allow job wrapping of more types of functions (resolves #264)

### LINKED ISSUE #264: Allow jobs to wrap class functions, nested functions and lambda functions
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-307)
┆Issue Number: TOIL-307

benedictpaten:
I am having issues with eclipse testing the utilTest, seeing if Jenkins will pass the tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ kwargs = dict(
     author='Benedict Paten',
     author_email='benedict@soe.usc.edu',
     url="https://github.com/BD2KGenomics/toil",
-    install_requires=[ 'bd2k-python-lib==1.7.dev1' ],
+    install_requires=[ 'bd2k-python-lib==1.7.dev1', 'dill==0.2.4' ],
     tests_require=[ 'mock==1.0.1', 'pytest==2.7.2' ],
     test_suite='toil',
     extras_require={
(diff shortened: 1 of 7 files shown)


benedictpaten:
Okay, I'm not understanding the output here - as it claims a test failed but lists 0 tests failed. I would expect the jobTest to fail until we make dill-0.2.4 is made a dependency. To do this do I just edit setup.py?

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil/237/console

utilsTest hung

hannes-ucsc:
> To do this do I just edit setup.py?

Yes

benedictpaten:
Still fiddling around with this, hoping to get at least a fully working prototype shortly. I think we'll still need the resource manager class for tracking the module from which the function derives.

cket:
I've been thinking about this issue. If I understand this correctly, the main problem is that these functions all could need information that isn't necessarily defined within them. Then when the function is pickled and run in a separate interpreter, this outside information is completely lost. Is that true?

hannes-ucsc:
Continued in #505

PR #284, opened 2015-08-21
Change behavior of Job.Runner.startToil (resolves #277)

### LINKED ISSUE #277: change return value behavior of Job.Runner.startToil
Currently Job.Runner.startToil returns the number of failed jobs. I think this is, on reflection, not very pythonic and easy to miss. I think it would be better if it threw some kind unfinished exception if there were failed jobs and instead returned None. As this is an interface changing I'm labelling it for the current release

benedictpaten:
Change behavior of Job.Runner.startToil to raise an exception if there are failed jobs at the end of a run.

Resolves #277


benedictpaten:
Resolved merge conflict

PR #286, opened 2015-08-22
Retry SDB domain creation on 500 errors (resolves #285)

### LINKED ISSUE #285: Sort test has slowed down
Runtime doubled to 13min. It started with my first stab at #274 (#275) and so I thought it was caused by it. Since then I made the existence check faster and more reliable (#283) but that didn't seem to have an effect. 

I did notice some "SDB is not available errors" today so maybe we should wait a bit.

hannes-ucsc:
Addresses #285


hannes-ucsc:
Jenkins, test this please.

PR #294, opened 2015-08-24
Verify jobs meet mesos' minimum resource requirements (resolves #208)

### LINKED ISSUE #208: Check mesos has minimum disk/mem resources
If you specify disk/memory less than 1mb, mesos will not run the task. We should check this early on.

cket:
If jobs don't meet mesos' minimum resource requirements, round up and log


src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -384,12 +384,21 @@ def _createTask(self, jt_job, offer):
         disk = task.resources.add()
         disk.name = "disk"
         disk.type = mesos_pb2.Value.SCALAR
-        disk.scalar.value = jt_job.resources.disk / 1000000
-
+        if jt_job.resources.disk > 1000000:
```
hannes-ucsc: Don't you want to use GiB instead of GB here?

PR #296, opened 2015-08-25
Azure job store (resolves #87)

### LINKED ISSUE #87: Write Azure job store

joelarmstrong:
Azure functionality seems to work OK (if slowly, from outside the azure datacenter anyway), and passes the existing tests.

Some caveats:

- Azure is very picky about table and container names, so the namePrefix probably has to be alphanumeric.
- For credential storage, right now it uses a supplemental file (~/.toilAzureCredentials) with a format that looks like this:

```
[AzureStorageCredentials]
accountName1=ACCOUNTKEY1==
accountName2=ACCOUNTKEY2==
```

because multiple Azure "storage accounts" can exist under a single Microsoft account. Right now the tests use the hardcoded "toiltestaccount" storage account.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,9 @@ kwargs = dict(
             'mesos.interface==0.22.0',
             'psutil==3.0.1' ],
         'aws': [
-            'boto==2.38.0' ] },
+            'boto==2.38.0' ],
+        'azure': [
+            'azure==0.11.1' ],},
     package_dir={ '': 'src' },
     packages=find_packages( 'src', exclude=[ '*.test' ] ),
     entry_points={
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -358,6 +358,10 @@ def loadJobStore( jobStoreString, config=None ):
         from toil.jobStores.awsJobStore import AWSJobStore
         region, namePrefix = jobStoreArgs.split( ':', 1 )
         return AWSJobStore( region, namePrefix, config=config )
+    elif jobStoreName == 'azure':
+        from toil.jobStores.azureJobStore import AzureJobStore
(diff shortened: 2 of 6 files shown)


hannes-ucsc:
Brilliant!

benedictpaten:
Does the the local file "toilAzureCredentials" have to be on the local disk, or could we make it a shared file in the jobStore, as we do for config, etc. ?

joelarmstrong:
I think that'd be putting the cart before the horse--to get to the jobStore, you have to have the credentials in the first place.

I think having to copy around a local file sucks, though--suggestions welcome.

joelarmstrong:
OK, the chunk size for jobs is now configurable (a parameter of AzureJobStore). I've run the tests using a 200-byte limit, and confirmed it's actually performing the chunking. But I left the test using the default parameter for the time being.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
I think the test failures are the result of the doctest collection.

joelarmstrong:
I think the 2 failures are due to a missing 'azure' package on the Jenkins host. I tried putting it in setup.py as an "extra" dependency, but I'm not sure how that works or if Jenkins will automatically use it.

hannes-ucsc:
Right, and the actual tests are being skipped by the needs_azure decorator. But when the test runner scans all modules for doctests, we get the import error, I think.

joelarmstrong:
Oh, I see what you mean. That explains the mesos test failures on my system (I don't have mesos installed).

I wonder if there is any good way around that.

hannes-ucsc:
Jenkins, test this please.

PR #303, opened 2015-08-25
Remove sysmaxint as a placeholder in Job.py (resolves #227)

### LINKED ISSUE #227: Replace sys.maxint with None

cket:
sys.maxint is also used as the default for the max resources flags in common, but those usages seem appropriate and were kept.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -59,8 +59,8 @@ class Job(object):
         require to run. Cores is the number of CPU cores required.
         """
         self.cores = cores
-        self.memory = human2bytes(str(memory)) if memory else memory
-        self.disk = human2bytes(str(disk)) if disk else disk
+        self.memory = human2bytes(str(memory)) if memory is not None else memory
+        self.disk = human2bytes(str(disk)) if disk is not None else disk
         #Private class variables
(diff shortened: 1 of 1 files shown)


src/toil/job.py
```diff
@@ -59,11 +59,8 @@ def __init__(self, memory=sys.maxint, cores=sys.maxint, disk=sys.maxint):
         require to run. Cores is the number of CPU cores required.
         """
         self.cores = cores
-        # passing sys.maxint to human2bytes seems to result in some float imprecision, and returns a value 1
-        # larger than sys.maxint. We later assume that any value here not equal to sys.maxint must be the user's value
-        # so it is passed to the batch system, which cannot allocate that many resources.
-        self.memory = human2bytes(str(memory)) if memory!=sys.maxint else sys.maxint
-        self.disk = human2bytes(str(disk)) if disk!=sys.maxint else sys.maxint
+        self.memory = human2bytes(str(memory)) if memory else memory
```
hannes-ucsc: I'd prefer "is not None" otherwise the empty string would sneak through. With "is not None" you'd get an exception.

PR #314, opened 2015-08-26
Clean util (resolves #298, resolves #313)

### LINKED ISSUE #298: Sort test leaves S3 buckets behind

### LINKED ISSUE #313: Create Toil clean util

cket:
I left the default value of the clean flag as 'never', I think it is safer to let people delete the jobStore themselves with the util. I implemented clean as part of the teardown in sort test, and it is tested itself in the utils unittest.


hannes-ucsc:
Let's discuss your decision to have `--clean` default to `never` in tomorrows scrum.

hannes-ucsc:
The conflicts are in sortTest which I recently streamlined. I think I would be the best person to resolve them. Ok, to go multi-author?

cket:
Yeah, fine by me!

src/toil/utils/toilClean.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+import os
+import sys
+
+from toil.lib.bioio import getBasicOptionParser
+from toil.lib.bioio import parseBasicOptions
+from toil.common import loadJobStore
+from toil.jobStores.abstractJobStore import JobStoreCreationException
+
+logger = logging.getLogger( __name__ )
```
hannes-ucsc: Missing a doc string? See if `toil --help` still works?

src/toil/utils/toilClean.py
```diff
@@ -0,0 +1,71 @@
...
+
+def main():
+    """Reports the state of the toil.
```
hannes-ucsc: Docstring is inaccurate.

src/toil/utils/toilClean.py
```diff
@@ -0,0 +1,71 @@
...
+    parser = getBasicOptionParser("usage: %prog [--jobStore] JOB_STORE", "%prog 0.1")
+
+    parser.add_option("--jobStore", dest="jobStore",
+                      help="Job store path. Can also be specified as the single argument to the script.\
+                       default=%default", default=os.path.abspath("./toil"))
+
+    options, args = parseBasicOptions(parser)
+    logger.info("Parsed arguments")
+
+    assert len(args) <= 1 #Only toil may be specified as argument
```
hannes-ucsc: I don't understand this. The jobStore can be passed in via the "--jobStore" option and as a positional argument?

hannes-ucsc:
Don't like the ambiguity between the positional and the option. What if a confused user specifies both?

Follow up in #329, #328 and #327

PR #323, opened 2015-08-28
Helpful message to warn user about module name conflicts (resolves #103)

### LINKED ISSUE #103: User scripts don't work if they are named test.py

cket:


hannes-ucsc:
There was one other place where a symbol is loaded. I think when the Job class is loaded and then pickled. Remember that user script can be function or class based. This just catches the function based ones.

PR #370, opened 2015-09-05
Client-side encryption for Azure job store (resolves #308)

### LINKED ISSUE #308: Add encryption to Azure job store

joelarmstrong:
Currently it uses only one key. Having a separate key for each file is certainly possible, but it involves some key distribution which would be slightly trickier.

Only the path to the key file is ever stored in the jobStore, because I was a bit paranoid about stuff getting pickled and passed around so often. If the job store never, ever gets pickled, maybe we could save a bit of I/O and read it in at job store creation.

This client side encryption should work fine for the file job store too.

### DIFF
--- a/src/toil/lib/encryption.py
+++ b/src/toil/lib/encryption.py
@@ -1,0 +1,89 @@
+# Copyright (C) 2015 UCSC Computational Genomics Lab
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import nacl
+from nacl.exceptions import CryptoError
+from nacl.secret import SecretBox
+
+# 16-byte MAC plus a nonce is added to every message.
+encryptionOverhead = 16 + SecretBox.NONCE_SIZE
+
+def encrypt(message, keyPath):
+    """
+    Encrypts a message given a path to a local file containing a key.
+
+    :param message: The message to be encrypted.
+    :param keyPath: A path to a file containing a 256-bit key (and nothing else).
+    :type message: str
+    :type keyPath: str
+    :rtype: str
+
+    A constant overhead is added to every encrypted message (for the nonce and MAC).
+    >>> import tempfile
+    >>> k = tempfile.mktemp()
+    >>> with open(k, 'w') as f:
+    ...     f.write(nacl.utils.random(SecretBox.KEY_SIZE))
+    >>> message = 'test'
+    >>> len(encrypt(message, k)) == encryptionOverhead + len(message)
+    True
+    """
+    with open(keyPath) as f:
+        key = f.read()
+    if len(key) != SecretBox.KEY_SIZE:
+        raise ValueError("Key is %d bytes, but must be exactly %d bytes" % (len(key),
+                                                                            SecretBox.KEY_SIZE))
+    sb = SecretBox(key)
+    # We generate the nonce using secure random bits. For long enough
+    # nonce size, the chance of a random nonce collision becomes
+    # *much* smaller than the chance of a subtle coding error causing
+    # a nonce reuse. Currently the nonce size is 192 bits--the chance
+    # of a collision is astronomically low. (This approach is
+    # recommended in the libsodium documentation.)
+    nonce = nacl.utils.random(SecretBox.NONCE_SIZE)
+    assert len(nonce) == SecretBox.NONCE_SIZE
+    return str(sb.encrypt(message, nonce))
+
+def decrypt(ciphertext, keyPath):
+    """
+    Decrypts a given message that was encrypted with the encrypt() method.
+
+    :param ciphertext: The encrypted message (as a string).
+    :param keyPath: A path to a file containing a 256-bit key (and nothing else).
+    :type keyPath: str
+    :rtype: str
+
+    Raises an error if ciphertext was modified
+    >>> import tempfile
+    >>> k = tempfile.mktemp()
+    >>> with open(k, 'w') as f:
+    ...     f.write(nacl.utils.random(SecretBox.KEY_SIZE))
+    >>> ciphertext = encrypt("testMessage", k)
+    >>> ciphertext = chr(ord(ciphertext[0]) ^ 1) + ciphertext[1:]
+    >>> decrypt(ciphertext, k)
+    Traceback (most recent call last):
+    ...
+    CryptoError: Decryption failed. Ciphertext failed verification
+
+    Otherwise works correctly
+    >>> decrypt(encrypt("testMessage", k), k)
+    'testMessage'
+    """
+    with open(keyPath) as f:
+        key = f.read()
+    if len(key) != SecretBox.KEY_SIZE:
+        raise ValueError("Key is %d bytes, but must be exactly %d bytes" % (len(key),
+                                                                            SecretBox.KEY_SIZE))
+    sb = SecretBox(key)
+    # The nonce is kept with the message.
+    return sb.decrypt(ciphertext)
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ kwargs = dict(
     author='Benedict Paten',
     author_email='benedict@soe.usc.edu',
     url="https://github.com/BD2KGenomics/toil",
-    install_requires=[ 'bd2k-python-lib==1.7.dev1' ],
+    install_requires=[ 'bd2k-python-lib==1.7.dev1', 'pynacl==0.3.0' ],
     tests_require=[ 'mock==1.0.1', 'pytest==2.7.2' ],
     test_suite='toil',
     extras_require={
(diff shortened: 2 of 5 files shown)


hannes-ucsc:
Jenkins, test this please.

joelarmstrong:
Hmm, looks like libffi headers are missing. I'm pretty sure the library itself is installed, since the `ctypes` package uses it, but Ubuntu/Debian often doesn't include headers. Maybe the libffi-dev package is required.

hannes-ucsc:
Also, this PR should have been set to failed status. Looks like the Jenkins user needs write perms to the repo for that to work. I'll fix that.

hannes-ucsc:
I'll add libffi-dev to toil-jenkins-slave.

Also: #371

hannes-ucsc:
https://github.com/BD2KGenomics/cgcloud/commit/a1ed642fbe34fa12c59a232fc435649525814be4

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Build was killed after 30min. We were already at 25 min so maybe the additional Azure test put it over the limit. I increased the timeout to 45.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Paranoia is good. For the AWS job store, DEBUG log level will leak the SSE-C key since at that level boto logs all HTTP request headers and the key is passed as a header.

We could wrap the key in a Secret class whose __str__ and __repr__ methods return "******" or something and that refuses to be pickled via copy_reg. To get at the key string from a Secret you'd have to call a specifc method. Just brainstorming here.

joelarmstrong:
Yeah, I think something unpickleable would be a great solution. Maybe something similar could be used for the AWS/Azure access credentials. Although obviously pickling/logging those isn't quite as bad, it would still suck for someone to accidentally pastebin a debug log or pickle file containing access keys. There are bots that scrape for AWS keys and run up your bill with bitcoin miners.

hannes-ucsc:
Hmm, looks like testToilSortOnAWS hangs?

joelarmstrong:
That's weird. Surely only the few option parsing lines I added could affect AWS, and they seem OK to me.

hannes-ucsc:
Jenkins, test this please.

joelarmstrong:
OK, it failed yet again, but in a different place. Let me investigate and try to reproduce. Maybe there is a subtle merge conflict.

src/toil/lib/encryption.py
```diff
@@ -0,0 +1,85 @@
...
+
+    A constant overhead is added to every encrypted message (for the nonce and MAC).
+    >>> import tempfile
+    >>> k = tempfile.mktemp()
+    >>> open(k, 'w').write(nacl.utils.random(SecretBox.KEY_SIZE))
+    >>> message = 'test'
+    >>> len(encrypt(message, k)) == encryptionOverhead + len(message)
+    True
+    """
+    key = open(keyPath).read()
```
hannes-ucsc: You're leaking a file handle here.

src/toil/lib/encryption.py
```diff
@@ -0,0 +1,85 @@
...
+    >>> decrypt(ciphertext, k)
+    Traceback (most recent call last):
+    ...
+    CryptoError: Decryption failed. Ciphertext failed verification
+
+    Otherwise works correctly
+    >>> decrypt(encrypt("testMessage", k), k)
+    'testMessage'
+    """
+    key = open(keyPath).read()
```
hannes-ucsc: Same here.

hannes-ucsc:
Maybe the file handle leaks are causing the test failures?

joelarmstrong:
I checked with 100000 tmp files a few hours ago, and python seemed to clean up the fds OK, but can't hurt to do it in the pythonic way. Updated it with a force push--not sure what the protocol is for when to/not to force push.

I think the gridengine test hanging is a fluke--it has only 1 second to find the job--but still can't figure out the AWS test issue, which is obviously not a fluke.

hannes-ucsc:
Jenkins, test this please.

joelarmstrong:
I can replicate the hang on the AWS sort tests on my laptop, even on master branch, but *only* when the test is run along with all the others with `make test`, not when it's run with `make test tests=src/toil/test/sort/sortTest.py::SortTest::testToilSortOnAWS`.

Looks like it might be a deadlock in the upload or download stream methods, 
since one end is blocking on a futex and the other is blocking on a read from a pipe:

```
Process 2556 attached with 2 threads
[pid  2579] futex(0xf156f0, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid  2556] read(45, 
```

I might try to get the python gdb extensions to work so I can get a backtrace. But they've never worked correctly for me.

hannes-ucsc:
I was surprised to find out that Jenkins builds not commit at the tip of the issues/308-azure-encryption branch, but a merge commit from the tip of that branch into master. I manually ran `make test` against both commits on the jenkins slave and both succeeded.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
In the past, Pyrasite worked for me:

http://stackoverflow.com/questions/132058/showing-the-stack-trace-from-a-running-python-application#answer-29881630

You may want to run the tests without capturing of stderr and out:

```
make test tests="-s src"
```

joelarmstrong:
I got gdb to work finally (just need to run with the python-dbg executable rather than python2.7).

Interestingly my backtraces for my AWS sort hang (not necessarily the same as the one on jenkins!) show a deadlock when reading the environment.pickle file in worker.py.

I have a hunch that reading the environment.pickle straight into the cPickle.load method may be a race. I'm not totally sure, but I *think* the GIL cannot ever be released when inside a C extension, and cPickle.load is definitely written in C. So if there is not already data written into the other end of the pipe, the worker will completely deadlock.

joelarmstrong:
OK, added my fix. Seems to work (n=3). I have no clue why the encryption code would trigger this race, but it is definitely a race.

hannes-ucsc:
It would be a bug for cPickle not to release the GIL before doing IO and it looks like the latest version does just that:

https://github.com/python-git/python/blob/master/Modules/cPickle.c#L555

joelarmstrong:
This is where mine was stalled:

https://github.com/python-git/python/blob/master/Modules/cPickle.c#L594

That getc can definitely block. It seems like a bug. Maybe it's relying on line-buffering?

hannes-ucsc:
The link is outdated. Trying to find the right one.

joelarmstrong:
I had a hard time finding the official repo too. But here it is (2.7 branch):

https://github.com/python/cpython/blob/2.7/Modules/cPickle.c#L591

Seems like the getc is still used without releasing the GIL (but I can't be 100% sure). Empirically, the patch seems to stop the AWS test from hanging for me, but maybe it's a coincidence.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Don't know why I have to keep nudging Jenkins to build. I added you to the whitelist explicitly and it should build every PR from you automatically at least from now on.

joelarmstrong:
OK, thanks. Wasn't sure if I had to say the magic words or what. I just logged into your Jenkins instance via Github OAuth. Maybe that was why it wouldn't automatically build.

hannes-ucsc:
There may be other places where we call cPickle.load() passing a pipe handle whose other end is written to by the AWS job store.

joelarmstrong:
I found 2 more:

https://github.com/BD2KGenomics/toil/blob/d6e89c9bdf376561366c02b56a7de472f4094cfb/src/toil/job.py#L1120

https://github.com/BD2KGenomics/toil/blob/d6e89c9bdf376561366c02b56a7de472f4094cfb/src/toil/jobStores/abstractJobStore.py#L57

hannes-ucsc:
`serialiseEnvironment()` in common.py does not specify a protocol when calling `cPickle.dump()`:

```
def serialiseEnvironment(jobStore):
    """
    Puts the environment in a globally accessible pickle file.
    """
    #Dump out the environment of this process in the environment pickle file.
    with jobStore.writeSharedFileStream("environment.pickle") as fileHandle:
        cPickle.dump(os.environ, fileHandle)
    logger.info("Written the environment for the jobs to the environment file")
```

According to https://docs.python.org/2/library/pickle.html#data-stream-format it uses the text-based protocol 0 in that case. Maybe that's why it's using the readline function that ultimately uses getc?

joelarmstrong:
Yep, looks like that's the reason.  The load_binint, load_binstring, etc. functions all use the fread code (which correctly releases the GIL), but the load_int, load_string, etc. functions all use the getc code.

hannes-ucsc:
It's already past the tests were it got stuck in previous builds.

hannes-ucsc:
Different, probably unrelated failure.

joelarmstrong:
Doh, looks like I will have to retry on that particular exception. I've never seen that before. Want me to fix it in this PR, or a different one?

But--phew! At least the mystery is solved. That was a tough one. Thanks for the help.

hannes-ucsc:
Excellent debugging! What do you think about retracting the workaround from b641a5a and instead making sure dump(...,HIGHEST_PROTOCOL) is always used?

joelarmstrong:
Sure, I think that makes sense. Force pushed that change instead of the other one.

We should also be a little paranoid about this if we eventually migrate pickling to dill.

joelarmstrong:
Blah, the gridengine test failed again, but I really think that's unrelated.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Aha, the curse has been lifted.

PR #386, opened 2015-09-10
azure stats strings were being passed as strings, not fds (resolves #385)

### LINKED ISSUE #385: Azure doesn't read stats files correctly
I completely misread the docstring and was passing strings to the stats callback rather than file handles (causing failure on #384).

joelarmstrong:
The leader ended up crashing when a worker happeend to log anything. Maybe we should add stats strings to the job store unit test?


joelarmstrong:
The connection got interrupted when pip installing "mock", which caused test failure; I'm force pushing a trivially amended commit to get Jenkins to rerun.

hannes-ucsc:
Re test coverage: #387

PR #397, opened 2015-09-16
Nested promises and immutable user file store (resolves #365, resolves #375, resolves #373, resolves #398)

benedictpaten:
This has not pulled in master for awhile, also the Azure and AWS jobStores are broken by a small number of changes. However, this pull request makes it possible to pass promises in an arbitrary, nested fashion, makes the user facing Job.FileStore interface essentially immutable and paves the way for robust caching. I am proposing assigning to Hannes/CJ to fix AWS and Azure support and to pull in master. Note this code really reorganises the serialisation process, so it is likely to heavily conflict with CJs changes.

PR #401, opened 2015-09-17
Make Parasol batch system create a separate batch for each set of ram and cpu requirements (resolves #396, resolves #429, resolves #456)

### LINKED ISSUE #396: Parasol batch system disregards cpu requirements
Using the following user script, with a Parasol batch system, I can get nodes with only 32 cores to run jobs that request 1000 cores:

```
#!/usr/bin/env python2.7
"""
simple_fail2.py: try to catch out Parasol Toin in mis-scheduling jobs

Theory: parasol tracks resource requirements *per batch*, not *per job*, so all
jobs will run with the CPU and memory requirements of the last job submitted.

"""

import argparse, sys, os, os.path, random, subprocess, shutil, itertools, glob
import socket, time, itertools

from toil.job import Job

def parse_args(args):
    """
    Takes in the command-line arguments list (args), and returns a nice argparse
    result with fields for all the options.
    
    Borrows heavily from the argparse documentation examples:
    <http://docs.python.org/library/argparse.html>
    """
    
    # Construct the parser (which is stored in parser)
    # Module docstring lives in __doc__
    # See http://python-forum.com/pythonforum/viewtopic.php?f=3&t=36847
    # And a formatter class so our examples in the docstring look good. Isn't it
    # convenient how we already wrapped it to 80 characters?
    # See http://docs.python.org/library/argparse.html#formatter-class
    parser = argparse.ArgumentParser(description=__doc__, 
        formatter_class=argparse.RawDescriptionHelpFormatter)
    
    # Add the Toil options so the job store is the first argument
    
    # General options
    
    # The command line arguments start with the program name, which we don't
    # want to treat as an argument for argparse. So we remove it.
    args = args[1:]
        
    return parser.parse_args(args)
   
def root_job(job):
    
    for i in xrange(10):
        # Start a huge child
        job.addChildJobFn(child_job, cores=1000, memory="4G", disk="1G")
    
    for i in xrange(10):
        # Start a smaller job bigger than this one, so it will go to the batch
        # system
        job.addChildJobFn(child_job, cores=1, memory="4G", disk="1G")
    
def child_job(job):
    
    job.fileStore.logToMaster("Job needing {} cores running on {}".format(
        job.cores, socket.gethostname()))
    
    time.sleep(10)
        
def main(args):
    """
    Parses command line arguments and do the work of the program.
    "args" specifies the program arguments, with args[0] being the executable
    name. The return value should be used as the program's exit code.
    """
    
    options = parse_args(args) # This holds the nicely-parsed options object
    
    # Make a root job
    root = Job.wrapJobFn(root_job, cores=1, memory="2G", disk=0)
    
    # Run it and see how many jobs fail
    failed_jobs = Job.Runner.startToil(root,  options)
    
    if failed_jobs > 0:
        raise Exception("{} jobs failed!".format(failed_jobs))
        
    print("All jobs completed successfully")
    
if __name__ == "__main__" :
    sys.exit(main(sys.argv))
```

The underlying problem is that, while Toil tracks resource requirements **per job**, Parasol tracks resource requirements **per batch**: all the jobs in a batch share the same per-job resource requirements. So if larger and smaller Toil jobs are in the same Parasol batch, they will be scheduled according to one set of resource requirements--the resource requirements of the most recently issued job.

Parasol commands look like this:

```
parasol -verbose -ram=257698037760 -cpu=32 -results=/hive/users/anovak/ga4gh/bake-off/hgvm/tree_test/results.txt add job '/cluster/home/anovak/python/bin/python2.7 -E /cluster/home/anovak/build/toil/src/toil/worker.py /hive/users/anovak/ga4gh/bake-off/hgvm/tree_test c/F/job1s9mH_'
```

The resource requirements show up before the batch file, and they **update the resource requirements for all jobs in the batch**. The Parasol batch system code in Toil treats them as if they set the resource requirements for just the job being added. So if you add a big job and then add a smaller job (as in my script above), the big job will have its resource requirements lowered to match those of the smaller job, and may schedule on a node that can't handle it.

The Parasol batch system code in Toil should probably be rewritten to keep jobs with different resource requirements in different batches.

### LINKED ISSUE #429: Toil hangs parasol when millions of jobs are running
I'm porting over an issue from the old jobTree repository as I don't think it got fixed in toil. If it did, then great! I was originally going to do it, but got a bit lazy and the jobTree fork happened. We may have to patch parasol to fix it, which shouldn't be a big deal.

----

This has been a problem for a while, but I'm just putting an issue up so I remember to fix this somehow.

When parasol has more than a million or so jobs queued, like now, the periodic "parasol -extended list jobs" command that jobTree runs hangs the entire parasol hub process for a couple minutes while it gets a listing of every job. This sucks, since it means that the cluster nodes start to go idle waiting for work, since the hub can't issue new jobs while it it's busy sending the list of queued jobs to jobTree. This gets even worse when there are a few jobTrees running; the cluster sometimes sits completely idle for several minutes.

We (read: I) should try to find some way around listing every job, maybe by looking to see if there's a way we can get the same information, but limited to just the jobTree batch rather than all batches. If there isn't a way currently, maybe modify parasol to include that functionality.

### LINKED ISSUE #456: ParasolBatchSytem.killBatchJobs doesn't update results file
Currently the parasol batch system uses "parasol remove job" in killBatchJobs, which completely deletes the job and prevents its status line from being written to the results file. Since the updated jobs worker gets its information from the results file, the CPU usage counter never gets decremented and getUpdatedJobs never returns anything about the job that was killed. The solution is probably to have killBatchJobs push the IDs of killed jobs to the updated jobs queue. 
This also creates a race condition because the job could end naturally before the "remove job" command is issued in killBatchJobs, which would alter the behavior of getUpdatedJob later on.

adderan:
- Create a new results file each time a job is issued with a new unique combination of ram and CPU requirements. 
- Switch the updated jobs worker to a thread instead of a daemon process.
- When shutting down, make the batch system clean up all the jobs it issued, and clear sick jobs.
- Use "parasol -results=<resultsForBatch> pstat2" on each batch to list issued jobs, instead of "parasol list jobs." This avoids listing all jobs on the cluster, which is slow when other users have created many jobs.
- Eliminate getFromQueueSafely in abstractBatchSystem because it was only necessary for multiprocessing.Queue, not Queue.Queue.

### DIFF
--- a/src/toil/batchSystems/parasol.py
+++ b/src/toil/batchSystems/parasol.py
@@ -23,18 +23,13 @@ import subprocess
 import time
 
 from Queue import Empty
-from multiprocessing import Process
-from multiprocessing import JoinableQueue as Queue
-
-#from threading import Thread
-#from Queue import Queue, Empty
+from Queue import Queue
+from threading import Thread
 
 from toil.batchSystems.abstractBatchSystem import AbstractBatchSystem
+from toil.lib.bioio import getTempFile, getTempDirectory
 
 logger = logging.getLogger( __name__ )
-
-def getParasolResultsFileName(toilPath):
-    return os.path.join(toilPath, "results.txt")
 
 def popenParasolCommand(command, runUntilSuccessful=True):
     """Issues a parasol command using popen to capture the output.
@@ -58,37 +53,7 @@ def popenParasolCommand(command, runUntilSuccessful=True):
         time.sleep(10)
         logger.warn("Waited for a few seconds, will try again")
 
-def getUpdatedJob(parasolResultsFile, outputQueue1, outputQueue2):
-    """We use the parasol results to update the status of jobs, adding them
-    to the list of updated jobs.
-    
-    Results have the following structure.. (thanks Mark D!)
-    
-    int status;    /* Job status - wait() return format. 0 is good. */
-    char *host;    /* Machine job ran on. */
-    char *jobId;    /* Job queuing system job ID */
-    char *exe;    /* Job executable file (no path) */
-    int usrTicks;    /* 'User' CPU time in ticks. */
-    int sysTicks;    /* 'System' CPU time in ticks. */
-    unsigned submitTime;    /* Job submission time in seconds since 1/1/1970 */
-    unsigned startTime;    /* Job start time in seconds since 1/1/1970 */
-    unsigned endTime;    /* Job end time in seconds since 1/1/1970 */
-    char *user;    /* User who ran job */
-    char *errFile;    /* Location of stderr file on host */
-    
-    plus you finally have the command name..
-    """
-    parasolResultsFileHandle = open(parasolResultsFile, 'r')
-    while True:
-        line = parasolResultsFileHandle.readline()
-        if line != '':
-            results = line.split()
-            result = int(results[0])
-            jobID = int(results[2])
-            outputQueue1.put(jobID)
-            outputQueue2.put((jobID, result))
-        else:
-            time.sleep(0.01) #Go to sleep to avoid churning
+
 
 class ParasolBatchSystem(AbstractBatchSystem):
     """The interface for Parasol.
@@ -100,31 +65,37 @@ class ParasolBatchSystem(AbstractBatchSystem):
                         "this batchsystem interface does not support such limiting" % maxMemory)
         #Keep the name of the results file for the pstat2 command..
         self.parasolCommand = config.parasolCommand
-        self.parasolResultsFile = getParasolResultsFileName(config.jobStore)
-        #Reset the job queue and results (initially, we do this again once we've killed the jobs)
+        self.parasolResultsDir = getTempDirectory(rootDir=config.jobStore)
+
         self.queuePattern = re.compile("q\s+([0-9]+)")
         self.runningPattern = re.compile("r\s+([0-9]+)\s+[\S]+\s+[\S]+\s+([0-9]+)\s+[\S]+")
-        self.killBatchJobs(self.getIssuedBatchJobIDs()) #Kill any jobs on the current stack
+
         logger.info("Going to sleep for a few seconds to kill any existing jobs")
         time.sleep(5) #Give batch system a second to sort itself out.
         logger.info("Removed any old jobs from the queue")
-        #Reset the job queue and results
-        exitValue = popenParasolCommand("%s -results=%s clear sick" % (self.parasolCommand, self.parasolResultsFile), False)[0]
-        if exitValue is not None:
-            logger.warn("Could not clear sick status of the parasol batch %s" % self.parasolResultsFile)
-        exitValue = popenParasolCommand("%s -results=%s flushResults" % (self.parasolCommand, self.parasolResultsFile), False)[0]
-        if exitValue is not None:
-            logger.warn("Could not flush the parasol batch %s" % self.parasolResultsFile)
-        open(self.parasolResultsFile, 'w').close()
-        logger.info("Reset the results queue")
-        #Stuff to allow max cpus to be work
-        self.outputQueue1 = Queue()
-        self.outputQueue2 = Queue()
-        #worker = Thread(target=getUpdatedJob, args=(self.parasolResultsFileHandle, self.outputQueue1, self.outputQueue2))
-        #worker.setDaemon(True)
-        worker = Process(target=getUpdatedJob, args=(self.parasolResultsFile, self.outputQueue1, self.outputQueue2))
-        worker.daemon = True
-        worker.start()
+
+        #In Parasol, each results file corresponds to a separate batch, and all
+        #jobs in a batch have the same cpu and memory requirements. The keys to this
+        #dictionary are the (cpu, memory) tuples for each batch. A new batch
+        #is created whenever a job has a new unique combination of cpu and memory
+        #requirements.
+        self.resultsFiles = dict()
+        self.maxBatches = config.maxParasolBatches
+
+        #Allows the worker process to send back the IDs of
+        #jobs that have finished, so the batch system can
+        #decrease its used cpus counter
+        self.cpuUsageQueue = Queue()
+
+        #Also stores finished job IDs, but is read
+        #by getUpdatedJobIDs().
+        self.updatedJobsQueue = Queue()
+
+        #Use this to stop the worker when shutting down
+        self.running = True
+
+        self.worker = Thread(target=self.updatedJobWorker, args=())
+        self.worker.start()
         self.usedCpus = 0
         self.jobIDsToCpu = {}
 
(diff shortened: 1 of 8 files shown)


hannes-ucsc:
The Jenkins build configuration was hosed because I had tweaked it temporarily for PR #397. In order to eliminate the need for these tweaks, there is now a "jenkins.sh" file in the source tree which will be run by Jenkins. I committed this change to master and modified all Toil build configurations to invoke jenkins.sh. This means that you should rebase onto master and push again. Otherwise the build will fail immediately.

I want to verify that the build does indeed fail if the jenkins.sh file is missing so I will trigger a build shortly.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Oh, I forgot that Jenkins builds not the last commit in the PR branch, but a temporary merge commit into the target branch, master in this case. Since the merge commit contains jenkins.sh from master.sh the build proceeded.

src/toil/batchSystems/parasol.py
```diff
@@ -132,23 +147,31 @@ def issueBatchJob(self, command, memory, cores, disk):
...
         pattern = re.compile("your job ([0-9]+).*")
-        parasolCommand = "%s -verbose -ram=%i -cpu=%i -results=%s add job '%s'" % (self.parasolCommand, memory, cores, self.parasolResultsFile, command)
+        parasolCommand = "%s -verbose -ram=%i -cpu=%i -results=%s add job '%s'" % (self.parasolCommand, memory, cores, batch.resultsFile, command)
 
         #Deal with the cpus
         self.usedCpus += cores
         while True: #Process finished results with no wait
             try:
-               jobID = self.outputQueue1.get_nowait()
+               jobID = batch.outputQueue1.get_nowait()
```
hannes-ucsc: Let's make this line the only line in the try: block and move the remaining three lines into an else: clause. That's safer and easier to read.

src/toil/batchSystems/parasol.py
```diff
@@ -201,7 +227,7 @@ def getRunningBatchJobIDs(self):
         #r 5410324 benedictpaten worker 1247030076 localhost
```
hannes-ucsc: Does this line make any sense?

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -218,8 +218,37 @@ def testIssueJob(self):
         self.batchSystem.issueBatchJob(jobCommand, memory=100e6, cores=1, disk=1000)
         self.wait_for_jobs(wait_for_completion=True)
         self.assertTrue(os.path.exists(test_path))
-
-
+    def testBatchResourceLimits(self):
```
hannes-ucsc: We use two empty lines between methods, IIRC.

hannes-ucsc: Sorry, make that one.

src/toil/batchSystems/parasol.py
```diff
@@ -90,6 +91,35 @@ def getUpdatedJob(parasolResultsFile, outputQueue1, outputQueue2):
         else:
             time.sleep(0.01) #Go to sleep to avoid churning
 
+
+class ParasolBatch():
+    def __init__(self, batchSystem, cores, memory):
```
hannes-ucsc: Some documentation would be nice.

src/toil/batchSystems/parasol.py
```diff
@@ -90,6 +91,35 @@ def getUpdatedJob(parasolResultsFile, outputQueue1, outputQueue2):
         else:
             time.sleep(0.01) #Go to sleep to avoid churning
 
+
+class ParasolBatch():
```
hannes-ucsc: A short doc string summarizing the purpose of this class would be nice.

src/toil/batchSystems/parasol.py
hannes-ucsc: I see. Never mind.

hannes-ucsc:
In addition to the nitpicky line nodes, I think the `self.batches` attribute should be a dictionary, where the keys are the requirement tuples. For the lookup in that dictionary, you may want to quantize the memory requirements in megabyte granularity, or whatever granularity makes sense with Parasol. We should also limit the size of the dictionary and raise an error if there are more than, say, 100 entries. Ideally, that limit should be configurable, either via command line or environment variable.

Oh, and I just caught a show-stopper: We can't have one process per batch. Maybe one thread per batch but not one process. And we don't use daemon threads/processes. If the original code used daemon threads then I might let that slide, but I don't want more daemon threads in Toil. I (or John) can explain why tomorrow during scrum. 

So ideally, we would want to have one pair of queues and a single worker. Have all result files be created in a single directory. In the worker thread/process, periodically list the directory and load all changed files. The queue items need to carry have a reference to the batch they came from. Later we may add inotify() support for more efficient change notifications. 

Lastly, `outputQueue1` and `outputQueue2` are bad names. I'm dubious as to why we need two queues, but if we really do, we need to give them more descriptive names.

joelarmstrong:
We usually use network filesystems to host the parasol batch directories, so sadly inotify wouldn't work (unless all processes were on the same host).

src/toil/batchSystems/parasol.py
```diff
@@ -78,17 +76,30 @@ def getUpdatedJob(parasolResultsFile, outputQueue1, outputQueue2):
...
+            line = fileHandle.readline()
+            if line != '':
+                results = line.split()
+                result = int(results[0])
+                jobID = int(results[2])
+                cpuUsageQueue.put(jobID)
+                updatedJobsQueue.put((jobID, result))
+
+        if not shutdownQueue.empty():
+            for fileHandle in resultsFileHandles:
```
hannes-ucsc: This should be in a finally clause outside of the loop.

src/toil/batchSystems/parasol.py
```diff
@@ -78,17 +76,30 @@ def getUpdatedJob(parasolResultsFile, outputQueue1, outputQueue2):
...
+        for fileHandle in resultsFileHandles:
+            line = fileHandle.readline()
+            if line != '':
+                results = line.split()
+                result = int(results[0])
+                jobID = int(results[2])
+                cpuUsageQueue.put(jobID)
+                updatedJobsQueue.put((jobID, result))
+
+        if not shutdownQueue.empty():
```
hannes-ucsc: I see. When I recommended using a None sentinel in the queue I thought that this thread consumed items from a queue. But since this thread doesn't have a queue to read from, it would make more sense to simply use a flag. Here's what I would do: I would make this function a method of the class. Then you wouldn't need to pass the queues in and you could change the while loop in the thread target function to ``while self.running``.

adderan:
So self.running should be a multiprocessing.Value object instead of a Queue?

hannes-ucsc:
Since the worker is now a threading.Thread, not a multiprocessing.Process there is no need to use anything from the multiprocessing module. With threads, thanks to Python's GIL you can just use a regular bool variable. If the worker were a multiprocessing.Process, I would use multiprocessing.Event for this. A multiprocessing.Value might work, too, but will probably have a higher overhead.

adderan:
It looks like the job store test that failed was still using getEmptyFileStoreID. Maybe that's the problem.

hannes-ucsc:
@joelarmstrong the last build [1] of this PR failed with 

```
HTTPError('Operation could not be completed within the specified time.',)
```

in the Azure job store. You already attempted to address this, am I correct?

[1]: http://jenkins.cgcloud.info/job/toil-pull-requests/467/testReport/junit/src.toil.test.jobStores.jobStoreTest/AzureJobStoreTest/testFileDeletion/

joelarmstrong:
No, I forgot about that particular bug, sorry. Should be fixed with #436.

hannes-ucsc:
#436 has been merged.

adderan:
I've left this as three commits for now, because there are three mostly separate changes being made. If we want, I can squash down to one commit.

src/toil/batchSystems/parasol.py
```diff
@@ -133,28 +104,45 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
...
+
     def issueBatchJob(self, command, memory, cores, disk):
         """Issues parasol with job commands.
         """
         self.checkResourceRequest(memory, cores, disk)
-        pattern = re.compile("your job ([0-9]+).*")
-        parasolCommand = "%s -verbose -ram=%i -cpu=%i -results=%s add job '%s'" % (self.parasolCommand, memory, cores, self.parasolResultsFile, command)
 
+        megabyte = 1e6
+        roundedMemory = int(memory/megabyte) * megabyte
```
hannes-ucsc: Rounding down is often referred to as truncation. The name of the variable should reflect that. Second question is if we want to truncate to MB or MiB, i,e, 10^6 or 2^20. Third, when I see floating point math in a situation were discrete values are desired, alarm bells go off inside my head. Floating point numbers are discrete, too, of course but the rules that govern truncation and rounding are more complicated than with integer math. You might just want to do

```
MiB = 1 << 20 # or 1024 * 1024
truncatedMemory = memory / MiB * MiB
```
In the old days, people did this

```
truncatedMemory = memory & -1 << 20
```
Depending on how familiar people are with bit manipulation, twos-complement and so on, people would prefer one or the other. I do like to stay in the discrete integer domain.

src/toil/batchSystems/parasol.py
```diff
@@ -238,9 +247,64 @@ def getRescueBatchJobFrequency(cls):
...
+            for fileHandle in resultsFileHandles:
+                line = fileHandle.readline()
+                if line != '':
+                    results = line.split()
+                    result = int(results[0])
+                    jobID = int(results[2])
+                    self.cpuUsageQueue.put(jobID)
+                    self.updatedJobsQueue.put((jobID, result))
+            time.sleep(0.01) #Go to sleep to avoid churning
+        for fileHandle in resultsFileHandles:
```
hannes-ucsc: This should to be in a `finally:` clause.

src/toil/batchSystems/parasol.py
```diff
@@ -238,9 +247,64 @@ def getRescueBatchJobFrequency(cls):
...
+        while self.running:
+            #Look for any new results files that have been created, and open them
+            newResultsFiles = set(os.listdir(self.parasolResultsDir)).difference(resultsFiles)
+            for newFile in newResultsFiles:
+                resultsFiles.add(newFile)
+                newFilePath = os.path.join(self.parasolResultsDir, newFile)
+                resultsFileHandles.append(open(newFilePath, 'r'))
+
+            for fileHandle in resultsFileHandles:
+                line = fileHandle.readline()
```
hannes-ucsc: I'm worried that you might get a partial line here. Lets add an assertion that the string end in '\n'.

src/toil/batchSystems/parasol.py
```diff
@@ -238,9 +247,64 @@ def getRescueBatchJobFrequency(cls):
...
+                if line != '':
```
hannes-ucsc: `if line:` is equivalent

src/toil/batchSystems/parasol.py
hannes-ucsc: Actually, this code is sensitive to the buffering pattern on the writing side:

```python
>>> w=open('foo','w')
>>> r=open('foo')
>>> r.readline()
''
>>> w.write('bar\n')
>>> r.readline()
''
>>> w.flush()
>>> r.readline()
'bar\n'
>>> w.write('bar2') # no newline at the end
>>> w.flush()
>>> w.write('\n') # now write newline
>>> r.readline() # returns partial line, not good
'bar2'
>>> w.write('bar3\n')
>>> w.flush()
>>> r.readline()
'\n'
>>> r.readline()
'bar3\n'
>>> w.close()
>>> r.close()
```

Not quite sure what the best way to handle this is. Probably just the assertion so we know early when this happens.

PR #431, opened 2015-09-29
Caching (resolves #376, resolves #171, resolves #379, resolves #443, resolves #444, resolves #445, resolves #450, resolves #402)

### LINKED ISSUE #171: Add optional file caching behaviour to jobStore

### LINKED ISSUE #376: Implement cache to work with job files
Depends on #375 working

### LINKED ISSUE #379: Create caching implementation

### LINKED ISSUE #402: Change log to master behavior
the FileStore.logToMaster does not result in a log message within the worker, rather the message is transmitted back. In the case the job crashes you therefore don't see these messages in the worker log. I think logToMaster should therefore both create a log message within the worker, as well as transmitting back the log message. I also think we should be able to specify the log level of the log to master message.

### LINKED ISSUE #443: sortTest and utilsTest can hang indefinitely trying to restart
This is not so worrying for interactive testing, but really annoying with Jenkins

### LINKED ISSUE #444: Encapsulate does not support return Values
Type Error (TypeError: 'NoneType' object has no attribute '__getitem__') when attempting to use the return value from an encapsulated job.

### LINKED ISSUE #445: static addChild definition of a node does what Encapsulate claims to do by default
If `A` is a job with dynamically added children, if

```
a, b = A(), B()
a.addChild(b)
```

then `b` should run in parallel with the dynamically added children unless `A().encapsulate` is specified. `b` appears to run only after all of `a`'s dynamically added children have completed.

### LINKED ISSUE #450: Does toil allow return values to be PromisedJobReturnValue instances
I noticed this while reading the Docstring of Job.run
```
Note: We disallow return values to be PromisedJobReturnValue instances
(generated by the Job.rv() function - see below).
A check is made that will result in a runtime error if you attempt to do this.
Allowing PromisedJobReturnValue instances to be returned does not work because
the mechanism to pass the promise uses a jobStoreFileID that will be deleted once
the current job and its descendants have been completed. This is similar to
scope rules in a language like C, where returning a reference to memory allocated
on the stack within a function will produce an undefined reference. 
Disallowing this also avoids nested promises (PromisedJobReturnValue
instances that contain other PromisedJobReturnValue).
```

I have such an instance in my pipeline so i decided to test it out on a toy example.
```
from toil.job import Job

import argparse

def f(job):
    '''
    DOCSTRING
    '''
    G = job.addChildJobFn(g)
    return G.rv()

def g(job):
    '''
    DOCSTRING
    '''
    return 'test.foo'


def i(job, work_dir):
    '''
    DOCSTRING
    '''
    pass

def test_1():
    '''
    DOCSTRING
    '''
    parser.add_argument('-f', dest='txt', default='txt')
    params = parser.parse_args()
    F = Job.wrapJobFn(f)
    I = Job.wrapJobFn(i, F.rv())
    F.addFollowOn(I)

if __name__ == '__main__':
    test_1() 
```
f is returning a PromisedJobReturnValue object of G and this should result in a runtime error, right?

benedictpaten:
Not ready for incorporation, just want to see if the azure/aws tests are passing.


benedictpaten:
Something is up with Jenkins (or the Jenkins/git hub link), I can't see the errors.

hannes-ucsc:
Have a look at the end of the build's console output. It will say that the build is cancelled because it took to long. You have two options:

a) Run all tests locally.

b) Modify jenkins.sh, append `tests="-s src"` to the `make` invocation. That will disable log [capturing] and print the log to the Jenkins console page in realtime. That will tell you which tests is hanging. You can then change `src` to a specific source file or even a specific test method [usage].

[capturing]: https://pytest.org/latest/capture.html
[usage]: https://pytest.org/latest/usage.html

When the hang is resolved, revert the changes to `jenkins.sh` because we shouldn't merge a PR with changes to it.

joelarmstrong:
For whatever reason, lately you have to go to http://jenkins.cgcloud.info, and log in. After that you can click the jenkins github link.

hannes-ucsc:
@joelarmstrong Yeah, I noticed that, too. I've been playing with authentication and authorization in Jenkins. At the least, I want to increase the session duration. Jenkins seems to log you out after like 30 min

benedictpaten:
This is now ready to merge from a feature point of view, once the tests pass.

benedictpaten:
Okay, all tests pass apart from the parasol sort test, which I have put a "skip" by. The parasol test just appears to hang. I think possibly it is hanging because jobs are being lost. I'd appreciate @adderan taking a look.

benedictpaten:
I'm fixing Arjun's errors in this PR, because otherwise I'm going to have a merge issue, I think.

jenkins.sh
```diff
@@ -9,4 +9,4 @@ virtualenv --system-site-packages venv
 make develop extras=[aws,mesos,azure,encryption]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
-make $make_targets
+make $make_targets tests="-s src"
```
hannes-ucsc: This needs to be undone before we can merge.

src/toil/test/sort/sortTest.py
```diff
@@ -184,6 +190,7 @@ def testFileGridEngine(self):
         self._toilSort(jobStore=self._getTestJobStorePath(), batchSystem='gridengine')
 
     @needs_parasol
+    @unittest.skip("skipping until parasol support is less flaky")
```
hannes-ucsc: Not critical, but this could mention the corresponding GH issue.

hannes-ucsc: Which is #449.

hannes-ucsc:
Almost there. Just the two line notes above should be addressed.

benedictpaten:
Okay, fixed both of Arjun's issues. However, I think I may have introduced a hang bug with daemon threads in worker. Adding back the "needs work" flag, hopefully fix this evening.

benedictpaten:
Okay, once the tests pass I think this is all ready to merge. I've fixed the issues @hannes-ucsc listed and the thread support is now pretty robust.

hannes-ucsc:
The tests turn out to be flappy. I rewound the merge on master and opened PR #453 instead.

benedictpaten:
Hmm, so I set it up to run over and over on my box and capture the errors when they occur.  I note this interesting failure:

======================================================================
ERROR: testFileSingle (toil.test.sort.sortTest.SortTest)
----------------------------------------------------------------------
  File "/Users/benedictpaten/git/toil/src/toil/test/sort/sortTest.py", line 182, in testFileSingle
    self._toilSort(jobStore=self._getTestJobStorePath(), batchSystem='singleMachine')
  File "/Users/benedictpaten/git/toil/src/toil/test/sort/sortTest.py", line 83, in _toilSort
    makeFileToSort(tempSortFile, lines=lines, lineLen=lineLen)
  File "/Users/benedictpaten/git/toil/src/toil/test/sort/sortTest.py", line 277, in makeFileToSort
    with open(fileName, 'w') as fileHandle:
IOError: [Errno 13] Permission denied: '/var/folders/tq/ssptxjwn2mg8ln95gh6mzq_w0000gn/T/toil-test-toil.test.sort.sortTest.SortTest.testFileSingle-tempDir-H3oES1/fileToSort.txt'

===============================

It looks like something is going on with the temporary files created by toil test.

hannes-ucsc:
Please continue discussion in #453.

Issue #421, opened 2015-09-24
Incorrect catch for missing .toilAzureConfig


adamnovak:
The Azure job store tries to give a useful message when it can't read your .toilAzureConfig for the key for the Azure job store storage account. However, it is looking for a `NoOptionError`, and if the entire file is missing instead a `NoSectionError` is raised, producing an error like this:

```


DEBUG:toil.batchSystems.mesos.executor:Starting executor
I0924 22:47:08.587648 43191 exec.cpp:132] Version: 0.22.0
I0924 22:47:08.592193 43192 exec.cpp:206] Executor registered on slave 20150924-214557-83886090-5050-44087-S8
INFO:toil.batchSystems.mesos.executor:Registered with framework
DEBUG:toil.batchSystems.mesos.executor:Running task 0
DEBUG:toil.batchSystems.mesos.executor:Sending status update ...
DEBUG:toil.batchSystems.mesos.executor:Sent status update
DEBUG:toil.batchSystems.mesos.executor:Invoking command: '/usr/bin/python2.7 -E /usr/local/lib/python2.7/dist-packages/toil/worker.py azure:hgvm:tree2 683b1010_55c3_4978_a7e7_853f17ed1803'
DEBUG:toil.batchSystems.mesos.executor:Sent stats message
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 408, in <module>
    main()
  File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 89, in main
    jobStore = loadJobStore(jobStoreString)
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 366, in loadJobStore
    return AzureJobStore( account, namePrefix, config=config )
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/azureJobStore.py", line 78, in __init__
    account_key = _fetchAzureAccountKey(accountName)
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/azureJobStore.py", line 63, in _fetchAzureAccountKey
    return configParser.get('AzureStorageCredentials', accountName)
  File "/usr/lib/python2.7/ConfigParser.py", line 330, in get
    raise NoSectionError(section)
ConfigParser.NoSectionError: No section: 'AzureStorageCredentials'
DEBUG:toil.batchSystems.mesos.executor:Sending status update ...
DEBUG:toil.batchSystems.mesos.executor:Sent status update
[...]

PR #437, opened 2015-09-29
Ship the environment in the Mesos task data

adamnovak:
This allows the `azureJobStore` to work on Mesos even when `.toilAzureCredentials` and/or `AZURE_ACCOUNT_KEY` haven't been pre-distributed to all the nodes. Just set `AZURE_ACCOUNT_KEY` on the master and it will be available on the nodes when the worker attempts to connect to the job store.

Works around the secondary issues described in #421 that require storage key pre-distribution or a shared file system.

### DIFF
--- a/src/toil/batchSystems/mesos/__init__.py
+++ b/src/toil/batchSystems/mesos/__init__.py
@@ -40,4 +40,6 @@ ToilJob = namedtuple('ToilJob', (
     # The resource object representing the user script
     'userScript',
     # The resource object representing the toil source tarball
-    'toilDistribution'))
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
This changes the entire environment of the executor. The executor typically outlast the worker processes and so this is scoped wrong with respect to both time and the fact that all variable are set. The environment changes should only affect the launched worker process, not the executor process and it should only modify specific variables. The general data path (mesos scheduler => serialized job => mesos executor => worker process) is correct though and this PR proves that. I will close it and reassign for a more surgical implementation of this mechanism.

Issue #125, opened 2015-06-27
Change stats reporting to use JSON instead of XML, or some better binary format


benedictpaten:

hannes-ucsc:
I'd prefer a pickled dict subclass that exposes its keys as attributes. John has used these before. This is what argparse's namespace class does. Then I can do 

config.foo

instead of 

config.attrib['foo']

hannes-ucsc:
Sorry wrong issue.

hannes-ucsc:
@cket, check out Expando and MagicExpando:

https://github.com/BD2KGenomics/bd2k-python-lib/blob/master/src/bd2k/util/expando.py

PR #438, opened 2015-09-29
Unicode-encode log messages (resolves #427)

### LINKED ISSUE #427: Character values >=128 in job output crash Toil log retrieval
My user script invokes wget, which likes to output "smart quotes" to its standard output.

This standard output gets collected by toil, put into XML, and sent back to the master (at least when I run in debug mode).

Unfortunately, `ElementTree` assumes all its strings are encoded in ASCII, when it tries to encode characters with XML entities (like &amp; and &quot;), and it panics when it sees bytes >128, because it doesn't know what entities to use:

```
          File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 408, in <module>
            main()
          File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 392, in main
            jobStore.writeStatsAndLogging(ET.tostring(elementNode))
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1126, in tostring
            ElementTree(element).write(file, encoding, method=method)
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 820, in write
            serialize(write, self._root, encoding, qnames, namespaces)
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 939, in _serialize_xml
            _serialize_xml(write, e, encoding, qnames, None)
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 939, in _serialize_xml
            _serialize_xml(write, e, encoding, qnames, None)
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 937, in _serialize_xml
            write(_escape_cdata(text, encoding))
          File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1073, in _escape_cdata
            return text.encode(encoding, "xmlcharrefreplace")
        UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 48: ordinal not in range(128)
```

Toil should assume an encoding for the standard output and standard error of the user script (probably UTF-8), and should strip or replace characters that can't later be encoded in the XML serialization format.

adamnovak:
ElementTree tries to encode the log messages with XML entities, and it
can be unable to do this if bytes >128 are present in `str` values in
the tree.

I have solved this by converting all the relevant strings to `unicode`
objects, assuming that the user script output being collected is
actually UTF-8. In the event of a UTF-8 decoding error, the offending
characters are replaced with placeholders, but the Toil code should keep
going (and ElementTree should still be able to serialize the logs).

Fixes #427, while we wait for #125.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -896,11 +896,11 @@ class Job(object):
         #Finish up the stats
         if stats != None:
             stats = ET.SubElement(stats, "job")
-            stats.attrib["time"] = str(time.time() - startTime)
+            stats.attrib["time"] = unicode(time.time() - startTime)
             totalCpuTime, totalMemoryUsage = getTotalCpuTimeAndMemoryUsage()
-            stats.attrib["clock"] = str(totalCpuTime - startClock)
-            stats.attrib["class"] = self._jobName()
-            stats.attrib["memory"] = str(totalMemoryUsage)
+            stats.attrib["clock"] = unicode(totalCpuTime - startClock)
+            stats.attrib["class"] = unicode(self._jobName())
+            stats.attrib["memory"] = unicode(totalMemoryUsage)
         #Return any logToMaster logging messages + the files that should be deleted
         #from the job store once the job has been registered as complete
         return fileStore.loggingMessages, fileStore.deletedJobStoreFileIDs.union(promiseFilesToDelete)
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Just talked to @adamnovak. It would be better to set the encoding to utf-8 1) when the file with the redirected worker log is opened for the purpose of reading it into a string here: 

https://github.com/BD2KGenomics/toil/blob/master/src/toil/worker.py#L386

and 2) on the ElementTree object here (I think)

https://github.com/BD2KGenomics/toil/blob/master/src/toil/worker.py#L174

I will close this PR and reassign the issue.

Issue #352, opened 2015-09-02
Implement provisioning of Toil nodes on Azure


hannes-ucsc:

hannes-ucsc:
See #439

ejacox:
Closed in favor of #1501

PR #439, opened 2015-09-29
Add an Azure deployment template

adamnovak:
This doesn't really have an associated issue, but these are the files I have been using to get Toil Mesos clusters up on Azure while I wait for @joelarmstrong to write his cgcloud-alike tool. This is apparently the Right Way to stamp out Azure app deployments, and I think if we want we could even put it in the Azure store.

This template is based on the official example Mesos template (which is
under an included open source license). It provides a "Deploy to Azure"
button that you can click on to configure and deploy a Mesos-based Toil
cluster on Azure.

Assuming you have the `azure` command-line tool, you can start up a
cluster from the command line like so:

```
cp azuredeploy.parameters.json mycluster.json
nano mycluster.json # Set your password, unique names, and SSH key
azure group create toilcluster westus
azure group deployment create --template-file azuredeploy.json -e mycluster.json toilcluster toilcluster
```

Make sure that you don't use any illegal characters (like underscores or
upper-case letters) in the globally unique storage account name you need
to define. Also make sure you have supplied a unique DNS name component
for your cluster's firewall load banalcer.

When the setup is complete, it will output an ssh command to use to
connect to your cluster's first master. It should look something like:

```
ssh azureuser@toilcluster.westus.cloudapp.azure.com -p 2211
```

If you elected to create additional masters, they will show up on
subsequent ports.

The cluster runs Mesos 0.23 (as 0.22 misinterpreted the version of
Docker) and has Toil, Python 2.7.10, and the required version 0.22.0
`mesos.interface` and `mesos.native` Python modules installed.

Chronos and Marathon can both optionally be installed on the cluster.
They are disabled by default, but some Docker stuff is installed to
support them. I kept that around since it seems that many pipelines use
Docker.

The cluster nodes have relatively small system disks, but on Azure they
get big ephemeral disks mounted at /mmt. Mesos has been configured to
use this space for Mesos scratch work, but if Docker is going to be used
heavily, the Docker container filesystem data ought to be stored on
these ephemeral disks as well. This has not been done yet.


joelarmstrong:
Hey cool, thanks! Maybe this saves me from having to actually do any work on cgcloud.

hannes-ucsc:
Impressive!

I've assigned this PR to you, @joelarmstrong for you review and test drive. Once I get your OK, I will merge this.

adamnovak:
Note that there are several references here to BD2kGenomics/toil/master that Azure uses to retrieve the scripts that run on the instances to set them up. So if you want to test it, you probably have to grep through and point them at your versions of the scripts.

adamnovak:
@hannes-ucsc & @joelarmstrong, what work, exactly, is needed? Should I just rebase this on master?

hannes-ucsc:
I'll merge this. @joelarmstrong, Ideally, I would like a second pair of eyes on this but we can track this as part of #352.

PR #453, opened 2015-10-04
Caching (continued from #431, resolves #376, resolves #171, resolves #379, resolves #443, resolves #444, resolves #445, resolves #450, resolves #402)

### LINKED ISSUE #171: Add optional file caching behaviour to jobStore

### LINKED ISSUE #376: Implement cache to work with job files

### LINKED ISSUE #379: Create caching implementation

### LINKED ISSUE #402: Change log to master behavior

### LINKED ISSUE #443: sortTest and utilsTest can hang indefinitely trying to restart

### LINKED ISSUE #444: Encapsulate does not support return Values

### LINKED ISSUE #445: static addChild definition of a node does what Encapsulate claims to do by default

### LINKED ISSUE #450: Does toil allow return values to be PromisedJobReturnValue instances

hannes-ucsc:
Continued from #431.


hannes-ucsc:
@benedictpaten Re your https://github.com/BD2KGenomics/toil/pull/431#issuecomment-145589969:

This happens every time for me on the second repetition of the test. 

The file is initially created with `rw-r--r--` but when the second repetition attempts to create the file again, the perms have  changed to `r--r--r--`. That's why `open(...,'w')` fails with permission denied.

So, each repetition should clean up after itself by removing the file to sort.

This is not a race though, and the other permission denied issues may be unrelated.

I'm assuming that it's the caching logic that removes the `w` bit from the file's mode. You mentioned sth along those lines.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

PR #475, opened 2015-10-08
Fix to readGlobalFile (resolves #468)

### LINKED ISSUE #468: job.fileStore.readGlobalFile() should have an option to write to a specific temp dir obtained using job.fileStore.getLocalTempDir()
if a tool requires n inputs, we need them all to be in the same dir so the dir can be mounted to /data in a docker image. We cannot assume that job.fileStore.readGlobalFile() as it is right now will read all n files into the same directory

benedictpaten:
This fix follows the discussion with Arjun and Hannes.

Resolves #468

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -478,22 +478,49 @@ class Job(object):
             #TODO: Make this work with the caching??
             return self.jobStore.writeFileStream(None if not cleanup else self.jobWrapper.jobStoreID)
         
-        def readGlobalFile(self, fileStoreID):
+        def readGlobalFile(self, fileStoreID, userPath=None):
             """
-            Returns a path to a local copy of the file keyed by fileStoreID. 
-            The file will be read only. 
+            Returns an absolute path to a local, temporary copy of the file 
+            keyed by fileStoreID. 
+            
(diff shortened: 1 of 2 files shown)


src/toil/job.py
```diff
@@ -478,22 +478,49 @@ def writeGlobalFileStream(self, cleanup=False):
...
+        def readGlobalFile(self, fileStoreID, userPath=None):
             """
-            Returns a path to a local copy of the file keyed by fileStoreID. 
-            The file will be read only. 
+            Returns an absolute path to a local, temporary copy of the file 
+            keyed by fileStoreID. 
+            
+            *The returned file will be read only (have permissions 444).* 
+            
+            userPath is a path to the name of file to which the global file will be
```
hannes-ucsc: Love the extensive documentation but we should use `:param:` tags if possible

hannes-ucsc:
With this change we should also roll back my workaround in the CWL runner.

hannes-ucsc:
And by my workaround I mean #466.

benedictpaten:
Closing in favor of PR #477, which use a properly named branch.

PR #476, opened 2015-10-08
Prevent batch systems from returning killed jobs in getUpdatedBatchJobs (resolves #460)

### LINKED ISSUE #460: Make killBatchJobs consistent between batch systems
Currently, the leader treats killed batch jobs separately from jobs that ended by themselves. JobBatcher.killJobs kills jobs and then sends them to JobBatcher.processFinishedJob, with the exit status manually set to 1. Also, in each iteration of the main loop, processFinishedJobs is called on the latest updated job. This means that if batchSystem.killBatchJobs adds the killed jobs to the updated queue, they will be processed twice. (single machine currently does this)
I think JobBatcher.killJobs shouldn't call processFinishedJobs, but all batch systems should add the killed jobs to the updated jobs queue with exit status 1. This means that all finished jobs are processed in the same place.

adderan:
- In all batch systems, getUpdatedBatchJob now returns (jobID, exitValue) tuples only for jobs on which killBatchJobs was not called. This prevents killed jobs from being processed twice in the leader.
- batchSystemTest.testKillJobs updated to enforce this behavior.

### DIFF
--- a/src/toil/batchSystems/gridengine.py
+++ b/src/toil/batchSystems/gridengine.py
@@ -302,7 +302,7 @@ class GridengineBatchSystem(AbstractBatchSystem):
         return self.worker.getRunningJobIDs()
 
     def getUpdatedBatchJob(self, maxWait):
-        i = self.updatedJobsQueue.get()
+        i = self.getFromQueueSafely(self.updatedJobsQueue, maxWait)
         logger.debug('UpdatedJobsQueue Item: %s', i)
         if i is None:
             return None
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -90,7 +90,7 @@ class AbstractBatchSystem:
         """Gets a job that has updated its status,
         according to the job manager. Max wait gives the number of seconds to pause
         waiting for a result. If a result is available returns (jobID, exitValue)
-        else it returns None.
+        else it returns None. Does not return anything for jobs that were killed.
         """
         raise NotImplementedError('Abstract method: getUpdatedBatchJob')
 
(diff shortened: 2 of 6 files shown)


src/toil/batchSystems/gridengine.py
```diff
@@ -299,7 +299,7 @@ def getRunningBatchJobIDs(self):
         return self.worker.getRunningJobIDs()
 
     def getUpdatedBatchJob(self, maxWait):
-        i = self.updatedJobsQueue.get()
+        i = self.getFromQueueSafely(self.updatedJobsQueue, maxWait)
```
hannes-ucsc: AFAIK, `getFromQueueSafely` was only necessary for `multiprocessing.Queue`, but not `Queue.Queue`. There is no need for these mostly IO bound threads in the batch systems to be multiprocessing processes. The the source to getFromQueueSafely does not specify the exact issue it is trying to work around. I am inclined to remove all mention of multiprocessing. GridEngine and Parasol are so slow that a little bit of GIL contention isn't going to make a difference.

src/toil/job.py
```diff
@@ -601,7 +601,7 @@ def asyncUpdate():
                     
                     #Indicate any files that should be deleted once the update of 
                     #the job wrapper is completed.
-                    self.jobWrapper.filesToDelete = len(self.filesToDelete)
+                    self.jobWrapper.filesToDelete = self.filesToDelete
```
benedictpaten: This should be self.jobWrapper.filesToDelete = list(self.filesToDelete), as, intentionally, the fileStore uses a set, whereas the jobWrapper uses a list. I'm fixing this in my PR on the fileStore, suggest dropping this change.

adderan:
I think I'll get rid of getFromQueueSafely in the other PR, because in this one parasol still uses a process for the worker.

hannes-ucsc:
@adderan, This looks like it can be merged. However, I hesitate because you removed the `in progress` label and added the `ready` label. These two labels aren't exactly intuitive (they are used by Waffle). An issue is `ready` if it is ready to be picked up by someone to work on. So the `ready` state precedes the `in progress` state. IOW, `ready` != `ready to be merged`. To indicate `ready to be merged`, we use the absence of the `needs work` label.

Let me know if you would like me to merge this.

adderan:
Oops, yes I do mean that it's ready to be merged.

PR #477, opened 2015-10-08
Issues/467 fix file store methods (resolves #467, resolves #468)

### LINKED ISSUE #467: Why does job.fileStore.writeGlobalFile() delete  local copy of file?
I just updated to the latest build of toil (34) and writeGlobalFile now "safely removed (the local file) at the end of the job by placing it in a location returned by getLocalTempDir". This would not be acceptable if the user is running the  pipeline on on files not downloaded to a local temp dir but, from their original locations.

### LINKED ISSUE #468: job.fileStore.readGlobalFile() should have an option to write to a specific temp dir obtained using job.fileStore.getLocalTempDir()

benedictpaten:
Fixed Job.FileStore.readGlobalFile to allow a user specified location with the job's temporary directory structure. Fixed Job.FileStore.writeGlobalFile to not put files in the cache and subsequently delete files which are not within the job's temporary directory structure.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -478,22 +478,49 @@ class Job(object):
             #TODO: Make this work with the caching??
             return self.jobStore.writeFileStream(None if not cleanup else self.jobWrapper.jobStoreID)
         
-        def readGlobalFile(self, fileStoreID):
+        def readGlobalFile(self, fileStoreID, userPath=None):
             """
-            Returns a path to a local copy of the file keyed by fileStoreID. 
-            The file will be read only. 
+            Returns an absolute path to a local, temporary copy of the file 
+            keyed by fileStoreID. 
+            
+            *The returned file will be read only (have permissions 444).* 
+            
+            userPath is a path to the name of file to which the global file will be 
+            copied or hard-linked (see below). userPath must either be: (1) a 
+            file path contained within a directory or, recursively, a subdirectory 
+            of a temporary directory returned by Job.FileStore.getLocalTempDir(), 
+            or (2) a file path returned by Job.FileStore.getLocalTempFile(). 
+            If userPath is specified and this is not true a RuntimeError exception 
+            will be raised. If userPath is specified and the file is already cached, 
+            the userPath file will be a hard link to the actual location, else it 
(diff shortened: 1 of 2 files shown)


src/toil/job.py
```diff
@@ -429,41 +429,45 @@ def asyncWrite():
...
-            Get a local temporary file. This file will exist for the duration of the job only, and
+            Get an absolute path to a local temporary file. 
+            This file will exist for the duration of the job only, and
             is guaranteed to be deleted once the job terminates.
             """
             handle, tmpFile = tempfile.mkstemp(prefix="tmp", 
                                                suffix=".tmp", dir=self.localTempDir)
             os.close(handle)
-            return tmpFile
+            return os.path.abspath(tmpFile)
```
hannes-ucsc: Isn't this sensitive to the current working directory? If so, that would be bad.

src/toil/job.py
```diff
@@ -429,41 +429,45 @@ def asyncWrite():
...
         def getLocalTempDir(self):
             """
-            Get a new local temporary directory. This directory will exist for the 
+            Get the absolute path to a new local temporary directory. 
+            This directory will exist for the 
             duration of the job only, and is guaranteed to be deleted once
             the job terminates, removing all files it contains recursively. 
             """
-            return tempfile.mkdtemp(prefix="t", dir=self.localTempDir)
+            return os.path.abspath(tempfile.mkdtemp(prefix="t", dir=self.localTempDir))
```
hannes-ucsc: This shouldn't be necessary. We should just make sure that locaTempDir is absolute.

hannes-ucsc: Toil should be insensitive to the working directory and we should block/convert relative paths from entering the system.

benedictpaten: Please change!

hannes-ucsc: Actually, I think it's better this way.

Issue #489, opened 2015-10-12
Caching logic will fault if file_size > cacheSize


arkal:
Lines 680-686 in _cleanLocalTempDir 
```
            while totalCachedFileSizes > cacheSize:
                fileSize, fileStoreID =  cachedFileSizes.pop()
                filePath = self.jobStoreFileIDToCacheLocation[fileStoreID]
                self.jobStoreFileIDToCacheLocation.pop(fileStoreID)
                os.remove(filePath)
                totalCachedFileSizes -= fileSize
                assert totalCachedFileSizes >= 0
```
show that if totalCachedFileSizes > cacheSize the filePath is removed.

If a file larger than cacheSize is created, it is added to jobStoreFileIDToCacheLocation and it starts getting ```os.link```ed however the _cleanLocalTempDir will delete the file before the link is complete. If the first thing in the next job is to retrieve this file from filestore ( which it probably will be), it will attempt to get it from filestore since it is NOT in cache.

However, depending on the file size, I/O, etc, if the file is still being written asynchronously to filestore, readGlobalFile starts processing an incomplete file. 

Either _cleanLocalTempDir needs to wat for the write to complete before deleting the file (since it is the rate limiting step), of readGlobalFile should have a way of asserting the integrity of the input file before reading it (Do not read a file till it is fully in filestore)

arkal:
#486

benedictpaten:
Hi Arjun,

You are incorrect, I think, in thinking that this code will cause an error.
It is only run after a job is run, and before a new job is run. So it can
never occur that a file is being deleted while a link is being created. As
to the asynchronous write, it does not matter in a unix file system if the
file being read from through a file stream is deleted, as the file handle
will keep it valid until it is closed. I am not going to claim there are no
bugs, but I really need to see a working regression test (and please look
at tests/src/jobFileStoreTest.py) before you assign me an issue!

arkal:
My code ran fine when i increased my cacheSize >> totalCachedFileSizes but kept crashing before. I tested with smaller fastqs and managed to recreate it by reducing cache size appropriately. I didn't post my code since it required a 4.5G bwa index that you don't want in a test. I'll look at the reg tests and make a new one in a similar format to yours that picks up this issue.
Sorry about that.

benedictpaten:
Hmm, that's odd. I thought the tests were testing this case. Something is
definitely up, but not sure what. What platform were you were working on?

arkal:
centos
```
[...]

PR #493, opened 2015-10-14
489 cachesize less than filesize

arkal:
Added some code to src/toil/test/src/jobFileStoreTest.py so that it tests the case where a file is created within a localTempDir using writeGlobalFile.

Wrote a test jobCacheLessThanFileSizeTest.py that tests 2 conditions when a file is created by writeGlobalFile in a localTempDir
a) The file size is greater than the cache size
b) The file size is less than the cache size

There is a weird Error case that is seen in both tests at a semi-frequent rate. Needs more debugging to identify the cause. Test b) almost always fails in absence of an Error.

This is supplementary to issue #489


hannes-ucsc:
Continued with fixed branch name in #506.

PR #501, opened 2015-10-20
Asynchronous write caching fix (resolves #500, resolves #488)

benedictpaten:
Fixes bug, improves Job.FileStore tests. Thanks Arjun for finding and trapping bug!

Resolves #500 and #488

PR #506, opened 2015-10-20
Handle cache size less than file size (resolves #489)

### LINKED ISSUE #489: Caching logic will fault if file_size > cacheSize

hannes-ucsc:
Resolves #489


benedictpaten:
Confused by this, seems redundant to #501 ? Can we close?

hannes-ucsc:
This contains no commits by me so closing is ok with me. @arkal?

PR #515, opened 2015-10-26
Add my real-time logger code (resolves #359)

### LINKED ISSUE #359: Real-time logging to master
I want to see some output from my jobs as they are running, so I can get an idea of what's up with them and guess at whether (and when) they are going to finish. `--logDebug` does not appear to accomplish this.

It would be nice if this also worked over the network somehow.

adamnovak:
This isn't really done yet, and it doesn't have any tests to verify that it actually works, but with this, you ought to be able to:

```
from toil.realtimeLogger import RealtimeLogger

RealtimeLogger.get().info("This goes straight to master")
```

Probably will address #359, even though this doesn't use TCP.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -419,9 +419,12 @@ class Job(object):
             :returns: return value of job's run function
             """
             setLoggingFromOptions(options)
-            # Set up real-time log message reception, and put the details in the
-            # environment to be sent out to workers.
-            RealtimeLogger.start_master()
+            
+            if options.realTimeLogging:
+                # Set up real-time log message reception, and put the details in the
+                # environment to be sent out to workers.
+                RealtimeLogger.start_master(level=options.logLevel)
+                
             with setupToil(options, userScript=job.getUserScript()) as (config, batchSystem, jobStore):
                 if options.restart:
                     jobStore.clean(job._loadRootJob(jobStore)) #This cleans up any half written jobs after a restart
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -41,6 +41,7 @@ from toil.lib.bioio import (setLoggingFromOptions,
                                getTotalCpuTimeAndMemoryUsage, getTotalCpuTime)
 from toil.common import setupToil, addOptions
 from toil.leader import mainLoop
+from toil.realtimeLogger import RealtimeLogger
 
 class Job(object):
     """
@@ -418,6 +419,9 @@ class Job(object):
             :returns: return value of job's run function
             """
             setLoggingFromOptions(options)
+            # Set up real-time log message reception, and put the details in the
+            # environment to be sent out to workers.
+            RealtimeLogger.start_master()
             with setupToil(options, userScript=job.getUserScript()) as (config, batchSystem, jobStore):
                 if options.restart:
                     jobStore.clean(job._loadRootJob(jobStore)) #This cleans up any half written jobs after a restart
@@ -435,6 +439,9 @@ class Job(object):
                     #Setup the first wrapper.
                     rootJob = job._serialiseFirstJob(jobStore)
                 return mainLoop(config, batchSystem, jobStore, rootJob)
+                
+            # Stop reporting real-time log messages.
+            RealtimeLogger.stop_master()
 
     class FileStore( object ):
         """
--- a/src/toil/realtimeLogger.py
+++ b/src/toil/realtimeLogger.py
@@ -32,77 +32,56 @@ import collections
 import logging
 import logging.handlers
 import SocketServer
-import struct
 import socket
 import threading
 
-class LoggingDatagramHandler(SocketServer.DatagramRequestHandler):
+class LoggingDatagramHandler(SocketServer.BaseRequestHandler):
     """
     Receive logging messages from the jobs and display them on the master.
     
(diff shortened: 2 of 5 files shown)


hannes-ucsc:
Thank you for the PR. This will need a few changes. Are you throwing this over the fence or do you want to keep contributing? I am perfectly fine with either answer, it just affects our internal planning and resource allocation.

src/toil/job.py
```diff
@@ -362,6 +363,9 @@ def startToil(job, options):
     failed jobs
             """
             setLoggingFromOptions(options)
+            # Set up real-time log message reception, and put the details in the
+            # environment to be sent out to workers.
+            RealtimeLogger.start_master()
```
hannes-ucsc: This should be made optional via a option. The port should also be configurable (if it isn't already).

hannes-ucsc: And by option I mean command line option.

src/toil/realtimeLogger.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Start up the master server and put its details into the options
+        namespace.
+        
+        Python logging should have already been configured.
+        
+        """
+        
+        # Start up the logging server
+        cls.logging_server = SocketServer.ThreadingUDPServer(("0.0.0.0", 0),
```
hannes-ucsc: Ah, I see. Ephemeral port. That should work. Disregards my above comment about configuring a port.

src/toil/realtimeLogger.py
```diff
@@ -0,0 +1,190 @@
...
+        
+        Note that if the master logs here, you will see the message twice,
+        since it still goes to the normal log handlers too.
+        """
+        
+        if cls.logger is None:
+            # Only do the setup once, so we don't add a handler every time we
+            # log
+            cls.logger = logging.getLogger('realtime')
+            cls.logger.setLevel(logging.INFO)
```
hannes-ucsc: Should the level be configurable?

src/toil/realtimeLogger.py
```diff
@@ -0,0 +1,190 @@
...
+        # Set up a thread to do all the serving in the background and exit when
+        # we do
+        cls.server_thread = threading.Thread(
+            target=cls.logging_server.serve_forever)
+        cls.server_thread.daemon = True
+        cls.server_thread.start()
+        
+        # Set options for logging in the class and the options namespace
+        # Save them in the environment so they get sent out to jobs
+        os.environ["RT_LOGGING_HOST"] = socket.getfqdn()
```
hannes-ucsc: Every environment variable introduced by Toil should be prefixed with `TOIL_`. So I think `TOIL_RT_LOG_HOST` would be more appropriate.

hannes-ucsc:
Also, UDP packets can and will arrive in any order. The kernel will also drop the if too many arrive at a time. I've observed both. As currently written, the while loop in the server would go haywire if any of those two things happened. While I don't think we should reimplement TCP, we should recover from dropped or out of order packets and print a warning if either of those happen. This is typically solved by adding a sequence number. Alternatively, let's just use TCP.

Also, I think it is safe to assume that you get an entire datagram with every call to read or at least with socket.recvfrom. So the length prefixing of packets isn't really necessary for UDP.

adamnovak:
I can keep working on this when I can find the time. Your suggestions all make sense to me, except the out-of-order packet handling. Since you may have many workers, and they may be overloading the master with log messages, I don't want to use TCP. TCP forces the master to keep an open connection to every worker, can potentially bog down the network retransmitting dropped packets, and, if the master still can't keep up, will eventually make the workers block writing log messages if their buffers fill up. What I want is for logging messages to be immediately and unreliably fired off to the master, and for the master to just display them as they come in and silently drop them if it can't keep up.

Seeing messages out of order could potentially be confusing, but I don't think it's a good idea for a "real time" logging system to wait around overly long to put them back in order, or to wait around while requesting retransmission of log messages that dropped. Maybe I could put in a second or so of delay and some sequence numbers, to let it fix up minor ordering issues. On the other hand, that might take a disproportionate amount of code.

The only time that out-of-order or dropped datagrams can *break* the code that I have now is if a message manages to be split over multiple datagrams. What I think I should do is (somehow) work out the max packet size, keep all messages below that size, and remove my multiple-read-call loop. Logging messages shouldn't be kilobytes long; the ones that are can be either split (and reassembled out of order) or maybe truncated.

hannes-ucsc:
Completely agree about not using TCP. This should be a best effort only. I also don't want to reimplement TCP. So no reassembly, reordering or retransmissions. But I *think* that the main server loop will get out of sync if packets are truncated, dropped or swapped and it will never recover. We should talk in person.

adamnovak:
OK, I have addressed the potential issues with the server getting out of sync with the messages. It turns out that SocketServer just delivers every datagram to a separate handler instance. I've switched to just putting bare JSON in the datagram and not even using the stream-based interface to read it, but instead just taking the datagram contents as a string on the server side.

I also prefixed the environment variables used for communicating logging server info to the workers.

Finally, I've added a unit test, and sugared up the syntax so you can just do `RealtimeLogger.info("message")`.

hannes-ucsc:
Continued in #633

PR #529, opened 2015-11-06
Ability to communicate environment variables from leader to worker (resolves #441, resolves #421)

### LINKED ISSUE #421: Incorrect catch for missing .toilAzureConfig

### LINKED ISSUE #441: Implement AbstractBatchSystem.setEnv
For details see 

https://github.com/BD2KGenomics/toil/pull/437#issuecomment-144546077

and 

https://github.com/BD2KGenomics/toil/issues/421#issuecomment-143406075

JakeNarkizian:
Implemented setEnv in mesos, and single machine batch systems. Added command line option for setEnv. Added getEnv in abstractJobStore and called it from leader.py to iterate and set variables in batch system,

Resolves #441 and #421

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -99,6 +99,14 @@ class AbstractBatchSystem:
         """
         raise NotImplementedError('Abstract Method: shutdown')
 
+    def setEnv(self, name, value=None):
+        """
+        Sets environment variables for the process of a given job so that an arbitrary
+        environment variable can be passed to a worker node running a job. If no value
+        is provided it will be looked up from the current environment.
+        """
+        pass
+
     @classmethod
     def getRescueBatchJobFrequency(cls):
         """Gets the period of time to wait (floating point, in seconds) between checking for 
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -77,10 +77,13 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
         # Dict of launched jobIDs to TaskData named tuple. Contains start time, executorID, and slaveID.
         self.runningJobMap = {}
 
+        # Dictionary of environment variables
+        self.environment = {}
+
         # Queue of jobs whose status has been updated, according to mesos. Req'd by toil
         self.updatedJobsQueue = Queue()
 
-        # Wether to use implicit/explicit acknowledgments
+        # Whether to use implicit/explicit acknowledgments
         self.implicitAcknowledgements = self.getImplicit()
 
         # Reference to the Mesos driver used by this scheduler, to be instantiated in run()
@@ -262,6 +265,14 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
         if driver_result != mesos_pb2.DRIVER_STOPPED:
             raise RuntimeError("Mesos driver failed with %i", driver_result)
 
+    def setEnv(self, name, value=None):
+        if value is None:
+            try:
+                value = os.environ[name]
+            except KeyError:
+                raise RuntimeError("%s does not exist in current environment", name)
+        self.environment[name] = value
+
     def registered(self, driver, frameworkId, masterInfo):
         """
         Invoked when the scheduler successfully registers with a Mesos master
@@ -296,6 +307,7 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
 
     def _prepareToRun(self, job_type, offer, index):
         jt_job = self.jobQueueList[job_type][index]  # get the first element to insure FIFO
+        jt_job.environment = self.environment
         task = self._createTask(jt_job, offer)
         return task
 
--- a/src/toil/batchSystems/mesos/executor.py
+++ b/src/toil/batchSystems/mesos/executor.py
@@ -124,7 +124,7 @@ class MesosExecutor(mesos.interface.Executor):
                 job.userScript.register()
             log.debug("Invoking command: '%s'", job.command)
             with self.popenLock:
-                return subprocess.Popen(job.command, shell=True)
+                return subprocess.Popen(job.command, shell=True, env=dict(os.environ, **job.environment))
 
         def sendUpdate(taskState, message=''):
             log.debug("Sending status update ...")
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -85,6 +85,8 @@ class SingleMachineBatchSystem(AbstractBatchSystem):
         self.coreFractions = ResourcePool(self.numWorkers)
         # A lock to work around the lack of thread-safety in Python's subprocess module
         self.popenLock = Lock()
+        # A dictionary of environment variables to be passed to the job subprocess
+        self.environment = {}
         # A pool representing available memory in bytes
         self.memory = ResourcePool(self.maxMemory)
         log.info('Setting up the thread pool with %i workers, '
@@ -114,7 +116,9 @@ class SingleMachineBatchSystem(AbstractBatchSystem):
                     with self.coreFractions.acquisitionOf(coreFractions):
                         log.info("Executing command: '%s'.", jobCommand)
                         with self.popenLock:
-                            popen = subprocess.Popen(jobCommand, shell=True)
+                            envDict= dict(os.environ, **self.environment)
+                            popen = subprocess.Popen(jobCommand, shell=True, env=envDict)
+
                         info = Info(time.time(), popen, killIntended=False)
                         self.runningJobs[jobID] = info
                         try:
@@ -192,6 +196,14 @@ class SingleMachineBatchSystem(AbstractBatchSystem):
 
         for thread in self.workerThreads:
             thread.join()
+
+    def setEnv(self, name, value=None):
+        if value is None:
+            try:
+                value = os.environ[name]
+            except KeyError:
+                raise RuntimeError("%s does not exist in current environment", name)
+        self.environment[name] = value
 
     def getUpdatedBatchJob(self, maxWait):
         """
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -53,6 +53,7 @@ class Config(object):
         self.masterIP = '127.0.0.1:5050'
         self.parasolCommand = "parasol"
         self.maxParasolBatches = 10000
+        self.environment = {}
         
         #Resource requirements
         self.defaultMemory = 2147483648
@@ -132,6 +133,21 @@ class Config(object):
         setOption("masterIP") 
         setOption("parasolCommand")
         setOption("maxParasolBatches", int, iC(1))
+
+        def splitEnvVarsToDict(envVars):
+            """
+            Parses string of environment variables and values into a dictionary
+            """
+            d = dict()
+            for item in envVars:
+                try:
+                    name,val = item.split('=')
+                    d[name] = val
+                except ValueError:
+                    d[item] = None
+            return d
+
+        setOption("environment", splitEnvVarsToDict)
         
         #Resource requirements
         setOption("defaultMemory", h2b, iC(1))
@@ -213,6 +229,10 @@ def _addOptions(addGroupFn, config):
     addOptionFn("--maxParasolBatches", dest="maxParasolBatches", default=None,
                 help="Maximum number of batches Parasol is allowed to create - a batch \
                 is created for each job that has a unique set of resource requirements. Default=%i" % config.maxParasolBatches)
+
+
+    addOptionFn("--setEnv", dest="environment", default=[], action="append",
+                help="Set of arbitrary environment variables for workers. Default=%s" % config.environment)
 
     #
     #Resource requirements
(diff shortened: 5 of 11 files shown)


src/toil/jobStores/abstractJobStore.py
```diff
@@ -223,6 +223,9 @@ def exists(self, jobStoreID):
     # One year should be sufficient to finish any pipeline ;-)
     publicUrlExpiration = timedelta(days=365)
 
+    def getEnv(self):
+        return ['AZURE_ACCOUNT_KEY'] #not sure what other env variables to ask for
```
hannes-ucsc: This is an Azure-specific detail that has no place in the abstract job store. Override this method in the Azure job store and change this implementation to return an empty collection.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -99,6 +99,12 @@ def shutdown(self):
         """
         raise NotImplementedError('Abstract Method: shutdown')
 
+    def setEnv(self,name, value=None):
+        """Sets environment variables. If no value is provided it will be looked up
```
hannes-ucsc: Newline after """

hannes-ucsc: Needs more substance. Describe what the effect of this method on the running jobs is.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -99,6 +99,12 @@ def shutdown(self):
...
+        from the current environment.
+        """
+        raise NotImplementedError('Abstract method: setEnv')
```
hannes-ucsc: No need for the message argument to the exception constructor. It will be clear from the trace back which method was left unimplemented. Also, the ABC mechanism will catch unimplemented methods much earlier, at instantiation time.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -77,6 +77,9 @@ def __init__(self, config, maxCores, maxMemory, maxDisk, masterIP,
         # Dict of launched jobIDs to TaskData named tuple. Contains start time, executorID, and slaveID.
         self.runningJobMap = {}
 
+        # Dict of envirnment variables
```
hannes-ucsc: Spelling. And lets change "dict" to "dictionary"

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -262,6 +265,15 @@ def shutdown(self):
         if driver_result != mesos_pb2.DRIVER_STOPPED:
             raise RuntimeError("Mesos driver failed with %i", driver_result)
 
+    def setEnv(self,name, value=None):
```
hannes-ucsc: space after comma

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -262,6 +265,15 @@ def shutdown(self):
...
+        """
+        Sets environment variables into dictionary for each job.
```
hannes-ucsc: Only need to provide a docstring if the docstring in the overridden method needs to be amended, e.g. if you are modifying the contract defined in that method signature/docstring.

src/toil/batchSystems/mesos/executor.py
```diff
@@ -124,7 +124,7 @@ def runJob(job):
                 job.userScript.register()
             log.debug("Invoking command: '%s'", job.command)
             with self.popenLock:
-                return subprocess.Popen(job.command, shell=True)
+                return subprocess.Popen(job.command, shell=True, env=job.environment)
```
hannes-ucsc: Here you need to mix the current environment with job.environment. Otherwise, the current environment would be lost and the only variables requested via setEnv would be present. But we want the current environment variables *plus* the ones requested via setEnv.

src/toil/batchSystems/singleMachine.py
```diff
@@ -193,6 +195,14 @@ def shutdown(self):
         for thread in self.workerThreads:
             thread.join()
 
+    def setEnv(self,name, value=None):
+        """
+        Sets environment variables into dictionary for each job.
```
hannes-ucsc: Same here: no docstring necessary.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -262,6 +265,15 @@ def shutdown(self):
...
+        """
+        if value is None and name not in os.environ:
+            raise RuntimeError("%s does not exist in current environment", name)
+
+        self.environment[name] = value or os.environ[name]
```
hannes-ucsc: The way this method is implemented it does two lookups into the os.environ dictionary. This can introduce race conditions and is not optimal. It also checks `value` twice, but differently, causing this method to fail in certain cases if the value is "". Rewrite this method to only use one lookup and one conditional on value. Hint: use a try-except statement. The last line of the method will read `self.environment[name] = value`

src/toil/common.py
```diff
@@ -213,6 +215,7 @@ def _addOptions(addGroupFn, config):
     addOptionFn("--maxParasolBatches", dest="maxParasolBatches", default=None,
                 help="Maximum number of batches Parasol is allowed to create - a batch \
                 is created for each job that has a unique set of resource requirements. Default=%i" % config.maxParasolBatches)
+    addOptionFn("--setEnv", dest="setEnv", default={}, help="Set of arbitrary environment variables for workers. Default=%i" % config.setEnv)
```
hannes-ucsc: Did you actually try this? How do I specify multiple variables on the same script invocation?

src/toil/leader.py
```diff
@@ -335,6 +336,10 @@ def mainLoop(config, batchSystem, jobStore, rootJobWrapper):
     logger.info("Found %s jobs to start and %i jobs with successors to run",
                 len(toilState.updatedJobs), len(toilState.successorCounts))
 
+    #Set proper environment variables
+    for v in getEnv():
```
hannes-ucsc: getEnv() should be a method of the batch system class. It can't be a statically defined function in the module. Look at the other methods in AbstractBatchSystem and model your new method after those.

hannes-ucsc:
Aside from the line nodes, this functionality needs a unit test. The test should prove for every supported batch system that a variable injected via setEnv is indeed present when a job runs.

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: The return value must be a dictionary, since you need to support entries with fixed value that is not looked up from the environment. Most importantly, the AzureJobStore implementation of this method needs to return a value for AZURE_ACCOUNT_KEY since that value is read from a config file, not the current directory. I outlined this in my original email.

hannes-ucsc:
The tests failed with an obvious import error. This indicates to me that you never ran the the unit tests on your box before committing. But you should. At least the basic ones.

src/toil/leader.py
JakeNarkizian: I'm slightly confused, should getEnv() be a function in both the batch system and jobstore? if so what is the purpose of AbstractBatchSystem.getEnv()?

hannes-ucsc: Yep, I mixed up batch system and job store. Here's what the line not should have read:

> getEnv() should be a method of the job store class. It can't be a statically defined function in a module. Look at the other methods in AbstractJobStore and model your new method after those.

hannes-ucsc:
I won't be able to review this until the tests pass.

The commit history is a mess: the commits from the master branch (e.g. 8c646c6) should not appear in this branch and the "reverted changes that were accidentally committed" commit should not be there. You should probably squash all commits and rebase onto the latest master (as you always should with single-author PRs).

src/toil/batchSystems/singleMachine.py
```diff
@@ -85,6 +85,8 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
         self.coreFractions = ResourcePool(self.numWorkers)
         # A lock to work around the lack of thread-safety in Python's subprocess module
         self.popenLock = Lock()
+        # A dictionary of environment variables to be passed to the job subprocess
+        self.evironment = {}
```
hannes-ucsc: Spelling

src/toil/batchSystems/singleMachine.py
```diff
@@ -193,6 +195,16 @@ def shutdown(self):
         for thread in self.workerThreads:
             thread.join()
 
+    def setEnv(self, name, value=None):
+        try:
+            if name is "":
```
hannes-ucsc: Why is this necessary?

Also, you must use == for strings.

src/toil/batchSystems/singleMachine.py
```diff
@@ -193,6 +195,16 @@ def shutdown(self):
...
+                return
+            if value is None and name is not "":
```
hannes-ucsc: 2nd term is a dupe.

src/toil/batchSystems/singleMachine.py
hannes-ucsc: Also, `if` and `try` should be swapped.

src/toil/common.py
```diff
@@ -132,6 +133,24 @@ def f(x):
         setOption("masterIP") 
         setOption("parasolCommand")
         setOption("maxParasolBatches", int, iC(1))
+
+        def splitEnvVarsToDict(envVars):
+            """
+            Parses of list of environment variables and values. Names and their values are to be
+            separated by a space, an entirely different variable is to be separated by a comma.
+            e.g. --setEnv var1 val1, var2, var3 val3
```
hannes-ucsc: 1) This is an unconventional syntax. For variables, the convention is `name=value`
2) This syntax does not support spaces in values, which are more common than you might think.
3) argparse has the ability to mention an option multiple times and it will collect all occurrences in a list (see https://docs.python.org/2.7/library/argparse.html#action action='append')

hannes-ucsc: If we were to go with this method (which I don't think we need to), I would want to see comprehensive doctests on it.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +114,13 @@ def deleteJobStore(self):
         """
         raise NotImplementedError()
 
+    def getEnv(self):
+        """
+        Returns a list of relevant environment variables so that the leader can iterate over the list
```
hannes-ucsc: It returns a dict not a list

hannes-ucsc: And don't mention what a potential caller would want to do with the return value. That is their business. Take a step back and ask yourself "Why are we doing this?". The docstring should describe the *purpose* of the method. What is the intent behind it? Why would a subclass want to implement this method?

hannes-ucsc:
Time is running out on this. We need to get the release out the door. I will have to take this over sometime tomorrow. If I do, don't consider your attempt a failure since I will be building on top of what you did.

hannes-ucsc:
Tests still fail (there are some failures, and overall the build hangs), commits are not squashed and include commits from other branch.

hannes-ucsc:
You should focus on finding out why `src/toil/test/mesos/mesosTest.py::MesosTest::test_hello_world` hangs. The hang prevents us from getting a proper test report from Jenkins. The test report will enable us to figure out why the other tests are failing.

hannes-ucsc:
OK, I will take it from here.

PR #573, opened 2015-11-23
Azure Patches

### LINKED ISSUE #570: Azure File Store Very Slow Download Speed
My jobs that run on the Azure file store are hanging when downloading large files (where "large" is about 30 or 40 megabytes).

Here is a log on a singleMachine batch system with the Azure job store. It can run 2 jobs at a time (since I've set max cores to 32 and set it to use 16 cores in all my jobs), and right now those two jobs are both stuck trying to pull files out of the global file store: the cactus.tar.gz file should be 33 MB, as it was when generated using the file job store, and the snp1000g.tar.gz file should be 36 MB. It is taking far longer to download these files from the file store than it took to upload them; it's been waiting 11 minutes so far, for an overall download speed (on Kolossus, from Azure) of 0.10 MB/s or less. The line it seems to be hung in in my code is:

```
local_path = job.fileStore.readGlobalFile(index_dir_id)
```

The log from my code is here:
[logAzureStoreDebug.txt](https://github.com/BD2KGenomics/toil/files/41182/logAzureStoreDebug.txt)

It seems to work fine on the indexes for my smaller graphs, about 1-2 megabytes in size.

Unless this can get fixed quickly (or is not present in the current Toil release; I'm using master from a couple of days ago), I don't think I will be able to do my scale-up that I have planned for tomorrow.

adamnovak:
I've patched up the Azure deployment template to work again, and to let you deploy any Github-hosted Toil instead of just BD2KGenomics/master (which is still the default). This makes debugging on Azure significantly easier.

I've also fixed the Azure jobStore to read large files to disk correctly, to fix #570.

### DIFF
--- a/contrib/azure/README.md
+++ b/contrib/azure/README.md
@@ -20,7 +20,7 @@ The following image is an example of a cluster with 1 jumpbox, 3 masters, and 3 
 
 You can see the following parts:
 
-1. **Mesos on port 5050** - Mesos is the distributed systems kernel that abstracts cpu, memory and other resources, and offers these to services named "frameworks" for scheduling of workloads. **Note that Mesos masters only listen on the 10.0.0.0/18 subnet**. In particular, **note that Mesos does not listen on localhost**. If you run a Toil job on the master, you will want to pass `--batchSystem mesos --masterAddress 10.0.0.5:5050`.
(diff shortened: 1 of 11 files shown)


hannes-ucsc:
I want to add a test that exposes this and I don't think I want to take the modification that tweaks _downloadStream(). However, I want to take the contrib changes and the readFile fix. I'll create another PR and close this one. Thank you very much for helping with this.

PR #578, opened 2015-11-24
Damp down the Azure chunk size

adamnovak:
I tested this and I'm not getting the same errors from Azure. I can now save large jobs to the Azure job store.

It's noticeably slow to actually do so: it looks like writes to the table service aren't very fast, so maybe we should consider keeping the job data itself in blobs and just point to them from the tables (and handle the case where a worker needs to wait on a promised blob actually becoming available under the blob consistency model).

Closes 577.


hannes-ucsc:
Closing. I prefer the approach in #580 because it allows us to utilize the full 64k and we don't have to UTF-16 encode a base64 encoded binary.

Issue #579, opened 2015-11-24
Conflict errors on Azure after retrying `insert_entity` (because original insert really succeeded)


adamnovak:
I've noticed Toil producing an occasional Conflict error when trying to write things to the Azure JobStore (with my #578 patch) when creating a new job:

```
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:       INFO:toil.jobStores.azureJobStore:Exception WindowsAzureError passed 
predicate, trying again
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:       Traceback (most recent call last):
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:         File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 
282, in main
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:           fileStore=fileStore)
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:         File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 128
6, in _execute
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:           self._serialiseJobGraph(jobWrapper, jobStore, returnValues, False
)
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:         File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 121
4, in _serialiseJobGraph
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:           jobsToJobWrappers = self._makeJobWrappers(jobWrapper, jobStore)
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:         File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 112
9, in _makeJobWrappers
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:           successor._makeJobWrappers2(jobStore, jobsToJobWrappers), successors)
WARNING:toil.leader:9eca3f3b_fafc_4b1a_b060_f16a72308246:         File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 1129, in <lambda>
[...]

PR #585, opened 2015-11-25
Fix slow restart against the AzureJobStore (resolves #584)

### LINKED ISSUE #584: AzureJobStore.clean() is very slow
I'm attempting to restart a run with somewhere between several tens of thousands and hundreds of jobs in an AzureJobStore.

The first step of the restart process appears to be to "clean" the job store, which involves walking the entire job graph, in serial, before scheduling any jobs. With 100,000 jobs, and one (or more, with chunking) round-trip HTTP request to the strongly-consistent-but-slow Azure Tables system per job, all happening in serial, this could take a long time. It's already taken 25 minutes so far for me, which is time my cluster is just sitting there not doing useful work.

Ideally the cleaning process (and, if possible, all the scheduling) should be able to be done on-line, without deserializing all the jobs. It could just be implicit: if the modification timestamp on a job is earlier than the start time of the Toil run, we just fix up its retry count when we load it. At the very least the traversal should be able to be done in parallel, so we don't need to wait for the round-trip latency to the table service for every single job.

adamnovak:
Basically, we now download all the jobs from the jobStore before trying
to "clean" the jobStore or rebuild the ToilState. This means we aren't
running individual per-job operations against every job in the jobStore,
which can be slow if the jobStore is Azure or AWS and each of those
operations requires a round-trip HTTP request.

To make this work, I also fixed the Azure job store's `jobs()` method,
which was only returning the first 1000 jobs.

The `clean()` method and the restart process in general is now much
noisier, with info messages letting you know how many jobs have been
loaded so far.

Fix logger naming in AzureJobStore

This still needs unit tests, but will close #584.

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -123,45 +123,21 @@ class AzureJobStore(AbstractJobStore):
         return self.namePrefix + self.nameSeparator + name
 
     def jobs(self):
-        # We need to page through the results, since we only get some of them at
-        # a time. Just like in the BlobService. See the only documentation
-        # available: the API bindings source code, at:
-        # https://github.com/Azure/azure-storage-python/blob/09e9f186740407672777d6cb6646c33a2273e1a8/azure/storage/table/tableservice.py#L385
         
-        # These two together constitute the primary key for an item.
-        next_partition_key = None
-        next_row_key = None
-    
         # How many jobs have we done?
         total_processed = 0
-    
-        while True:
-            # Get a page (up to 1000 items)
-            page = self.jobItems.query_entities(
-                next_partition_key=next_partition_key,
-                next_row_key=next_row_key)
             
-            for jobEntity in page:
-                # Process the items in the page
-                yield AzureJob.fromEntity(jobEntity)
-                total_processed += 1
-                
-            if hasattr(page, 'x_ms_continuation'):
-                # Next time ask for the next page. If you use .get() you need
-                # the lower-case evrsions, but this is some kind of fancy case-
-                # insensitive dictionary.
-                next_partition_key = page.x_ms_continuation['NextPartitionKey']
-                next_row_key = page.x_ms_continuation['NextRowKey']
-            else:
-                # No continuation to check
-                next_partition_key = None
-                next_row_key = None
+        for jobEntity in self.jobItems.query_entities_auto():
+            # Process the items in the page
+            yield AzureJob.fromEntity(jobEntity)
+            total_processed += 1
             
-            if not next_partition_key and not next_row_key:
-                # If we run out of pages, stop
-                break
-                
-            logger.info("Processed %d total jobs" % total_processed)
+            if total_processed % 1000 == 0:
+                # Produce some feedback for the user, because this can take
+                # a long time on, for example, Azure
+                logger.info("Processed %d total jobs" % total_processed)
+            
+        logger.info("Processed %d total jobs" % total_processed)
 
     def create(self, command, memory, cores, disk,
                predecessorNumber=0):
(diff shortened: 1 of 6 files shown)


src/toil/jobStores/azureJobStore.py
```diff
@@ -120,8 +120,45 @@ def qualify(self, name):
...
+        # These two together constitute the primary key for an item.
+        next_partition_key = None
+        next_row_key = None
+    
+        # How many jobs have we done?
+        total_processed = 0
+    
+        while True:
+            # Get a page (up to 1000 items)
+            page = self.jobItems.query_entities(
```
hannes-ucsc: Can the paging functionality be placed in a reusable method of the AzureTable class?

adamnovak:
OK I added some tests. I'm not sure how to make a test log output (I tried print). It seems like running clean with the cache should always be faster than without the cache, so I've added that as an assert in the test of the clean function.

Still to do is making the paging reusable, and adding a test for toil state construction with the cache.

Also, in order to test the Azure jobs method yielding >1000 jobs, I need to actually make >1000 jobs. I do that in this, but it's going to be slow making so many trips to the job store and back.

hannes-ucsc:
That's ok. We'll probably skip this test for regular builds. But it would be great to expose the AWS store to the same test.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -423,6 +424,58 @@ def assertUrl(self, url):
...
+            # See how long it takes to clean with cache
+            jobCache = {jobWrapper.jobStoreID: jobWrapper
+                        for jobWrapper in master.jobs()}
+            cacheStart = time.time()
+            master.clean(rootJob, jobCache=jobCache)
+            cacheEnd = time.time()
+            
+            cacheTime = cacheEnd - cacheStart
+            
+            print("Without cache: %s With cache: %s" % (str(noCacheTime), str(cacheTime)))
```
hannes-ucsc: To use print as a function you would need to add `from __future__ import print_function` at the top.

adamnovak: It seems perfectly happy with the syntax without that, though it might be parsed as a print statement and then a parenthesized string.

Do we use that elsewhere in Toil?

hannes-ucsc: The syntax is valid but the parentheses are redundant and will be eliminated. It just looks as if you are using the print function. I would just remove the parentheses. I would only use the `print()` function if wanted access to its `file=` keyword argument. Choosing the output stream is the only feature that the `print` statement doesn't give you. I think we mix and match print statement and function in Toil, and I don't see a problem with that as long as we don't use visual ambiguity introduced by redundant parentheses.

adamnovak:
I'm not sure what's up with the tests. Jenkins failed it on `src/toil/test/sort/sortTest.py::SortTest::testFileSingle10000` with an error from the FileJobStore saying that a job that was supposed to exist didn't. Unfortunately, that test works just fine for me locally.

I've thrown out the print function; maybe that will magically make the tests pass.

adamnovak:
OK, I've successfully gotten it to fail. It looks like there are bad/corrupt jobs getting into the FileJobStore, and when I go through to load all of them to make the cache, the `.jobs()` iterator explodes because it can't deserialize some jobs.

It looks like either `.jobs()` needs to be modified to skip over jobs that can't be deserialized, or a custom override or version of `.clean()` has to somehow run first and clean up any damaged jobs.

hannes-ucsc:
Looks good. I think the correct procedure would be to merge this as is and open a new bug ticket against the FileJobStore. If you have a stack trace from the failure you mention in https://github.com/BD2KGenomics/toil/pull/585#issuecomment-160811287, can you post that here or add that to the new ticket.

For posterity, can you confirm that you saw `testManyJobs` pass at least once against `AzureJobStore`?

adamnovak:
The testManyJobs Azure test was (occasionally?) failing due to a conflict error on insert (probably after a retry). I've told it to just log those and proceed, and I can get it to pass on my machine. It takes about 10 minutes.

adamnovak:
OK, I think this is ready on my end.

src/toil/jobStores/azureJobStore.py
```diff
@@ -135,7 +148,13 @@ def create(self, command, memory, cores, disk,
...
         entity = job.toItem(chunkSize=self.jobChunkSize)
         entity['RowKey'] = jobStoreID
-        self.jobItems.insert_entity(entity=entity)
+        try:
+            self.jobItems.insert_entity(entity=entity)
+        except WindowsAzureConflictError as e:
+            # A conflict on insert probably means we retried but already
+            # succeeded. Just ignore it. We made a _newJobID() ourselves, so
+            # unless that's broken we can't be re-using a job ID.
+            logger.warning("Conflict when inserting job %s" % jobStoreId)
```
hannes-ucsc: Although useful, this is unrelated to the issue that this PR is solving. It seems to partially address #579, partially because this here isn't the only call of insert_entity.

hannes-ucsc:
I'm inclined to merge this without 5b33d1cbf2d452b3bb52545ee34bb3152b0be5c6 (the commit that partially resolves #579) even though that commit is required to make `testManyJobs` pass. I think there may be a more general solution to #579.

The 3.1.x branch and master are virtually identical at this point so I don't think I need to redo the PR on that branch.

hannes-ucsc:
Continued in #628.

adamnovak:
It definitely might make sense to take care of insert_entity all at once
(maybe in the table wrapper class?).

PR #590, opened 2015-12-01
Support propagating Azure Storage keys for multiple accounts

adamnovak:
This adds support for propagating multiple Azure accounts to your nodes, in the absence of a shared filesystem for `.toilAzureConfig`. You just set `AZURE_ACCOUNT_KEY_<accountname>` for each account, and pass `-e AZURE_ACCOUNT_KEY_<accountname>` options for each account to your Toil script invocation.

Come to think of it, this is really only actually *useful* in combination with my own personal code which piggy-backs on top of the Toil Azure credential management code in order to get the keys for the storage accounts holding input and output data for my jobs.

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -48,15 +48,18 @@ def _fetchAzureAccountKey(accountName):
     """
     Find the account key for a given Azure storage account.
 
-    The account key is taken from the AZURE_ACCOUNT_KEY environment
-    variable if it exists, or from looking in the file
-    "~/.toilAzureCredentials". That file has format:
+    The account key is taken from the AZURE_ACCOUNT_KEY_<account> environment
+    variable if it exists, then from plain AZURE_ACCOUNT_KEY, and then from
+    looking in the file "~/.toilAzureCredentials". That file has format:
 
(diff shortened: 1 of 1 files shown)


src/toil/jobStores/azureJobStore.py
```diff
@@ -48,15 +48,18 @@ def _fetchAzureAccountKey(accountName):
...
+    variable if it exists, then from plain AZURE_ACCOUNT_KEY, and then from
+    looking in the file "~/.toilAzureCredentials". That file has format:
 
     [AzureStorageCredentials]
     accountName1=ACCOUNTKEY1==
     accountName2=ACCOUNTKEY2==
     """
-    if 'AZURE_ACCOUNT_KEY' in os.environ:
+    
+    if "AZURE_ACCOUNT_KEY_{}".format(accountName) in os.environ:
```
hannes-ucsc: I prefer try-except for detecting presence of keys in dictionaries. It is DRY and more efficient. Barring that, I use get() with the 2nd argument.

Also prefer string concatenation for suffixes in favor of of format() or %.

I'll change these things in a continuation PR.

hannes-ucsc:
Continued in #637.

PR #592, opened 2015-12-01
Azure cluster config

adamnovak:
This PR updates the install scripts for setting up an Azure cluster to be more robust.


hannes-ucsc:
Also merged into releases/3.1.1

PR #610, opened 2015-12-04
Provide better error message handling in rv() for function names

dgaston:
Return self.userFunctionName in rv() RuntimeError message to ease debugging of toil scripts

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -274,7 +274,8 @@ class Job(object):
             #Returns the jobStoreFileID and jobStore string
             if self._promiseJobStore == None:
                 raise RuntimeError("Trying to pass a promise from a promising job "
-                                   "that is not predecessor of the job receiving the promise")
+                                   "that is not predecessor of the job receiving the promise. Receiving job"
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Hmm, this won't work because the userFunctionName is introduced by a subclass. Given that this is a error message, I would go for a light weight approach of `getattr( self, 'userFunctionName', self.__class__.__name__ )`. Alternatively, we could introduce a `Job.name()` method that defaults to returning `self.__class__.__name__` but can be overridden by subclasses.

Also, the fact that the tests pass indicates that either all of our test jobs are functions (which I don't think is the case) or that this error condition isn't covered. IOW, we also need a test for this.

@dgaston, thank you for the contribution. Don't feel obliged to make the above changes. We will schedule them internally for the 3.2.0 release.

dgaston:
Makes sense, happy to have helped in any way.

hannes-ucsc:
This will be addressed by #677

PR #614, opened 2015-12-05
Fix NPE in cwltoil entry point

kellrott:
If no args are passed to cwltoil tool main method, use system args (the default None will cause exception on the [workdir] + args)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
@scottfrazer looks like this is the same thing you ran into.

PR #622, opened 2015-12-10
Issue 621 cwl schema (resolves #621)

### LINKED ISSUE #621: cwltoil raises schema-related ValidationException
Tests started failing just now.

http://jenkins.cgcloud.info/job/toil-pull-requests/lastCompletedBuild/testReport/src.toil.test.cwl.cwlTest/CWLTest/test_run_revsort/

It would be nice if the schema didn't change underneath a particular cwl version. IOW, if I pin my cwl version, the schema used by it should never change. But that's just my hunch. I don't understand the intricacies involved here.

@tetron would you have time to take a look?

tetron:
Fix for issue #621 

This was due to the cwltool module having an open-ended dependency on the schema-salad module.  I made a incompatible change to schema-salad which broke cwltool which was pulling in the latest version.

cwltool now pins a specific version of schema salad, but this means toil needs to pin a new version of cwltool.

Also includes a small fix to the cwltoil main method.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ kwargs = dict(
         'encryption': [
             'pynacl==0.3.0'],
         'cwl': [
-            'cwltool==1.0.20151122025918']},
+            'cwltool==1.0.20151210154014']},
     package_dir={'': 'src'},
     packages=find_packages('src', exclude=['*.test']),
     entry_points={
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ kwargs = dict(
         'encryption': [
             'pynacl==0.3.0'],
         'cwl': [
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Weird pickling error in tests:

http://jenkins.cgcloud.info/job/toil-pull-requests/721/testReport/src.toil.test.cwl.cwlTest/CWLTest/test_run_revsort/

tetron:
@hannes-ucsc this seems to be fixed now, also pinned schema salad version so it won't start failing due to future updates.

hannes-ucsc:
Thanks, Peter! If at all possible, and for a small PR like this one, I'd like to see commits squashed, and rebased onto latest master instead of merged from master.

PR #623, opened 2015-12-10
Eliminate cache requirement (resolves #586)

### LINKED ISSUE #586: Eliminate cache requirement
Resulting from a conversation with @arkal and @jvivian.

@arkal can explain in more detail but he's running into issues where cached files are evicted prematurely. All input, output and intermediate files involved in a single sample run of his pipeline can fit onto the disk space of a single beefy EC2 instance, yet the current caching implementation doesn't utilize that.

The cache should be shared by all concurrent workers on a machine. It should fill up the disk and evict files as late as possible. A simple heuristic for this would be to persistently track the disk requirements of the set of all concurrently running jobs on a machine and ensuring that a job's required disk space is available before it is added to that set, i.e. shortly before it is started. If not enough free space is available, files need to be evicted from the cache until it is. The persistent tracking could be done in a file with locking. The overhead of managing that file and the contention introduced by the file locking should both be negligible for the job throughput Toil currently achieves.

arkal:
Resolves #586

### DIFF
--- a/src/toil/cache.py
+++ b/src/toil/cache.py
@@ -21,6 +21,7 @@ import hashlib
 import shutil
 import time
 from contextlib import contextmanager
+from struct import pack, unpack
 from fcntl import flock, LOCK_EX
 
 logger = logging.getLogger( __name__ )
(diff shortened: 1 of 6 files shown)


hannes-ucsc:
Continued in #642

PR #633, opened 2016-01-04
Add real-time logging facility (resolves #359)

### LINKED ISSUE #359: Real-time logging to master

hannes-ucsc:
Continued from #515
Resolves #359


hannes-ucsc:
Jenkins, test this please.

PR #635, opened 2016-01-04
Remove unnecessary assert that restricts toil commands names (resolves #631)

### LINKED ISSUE #631: Toil script as a program generates assertion
Take and example Toil program from the documentation, turn it into a python program `tryIt', with 
 
    #!/usr/bin/env python

as the first line.
  
     ./tryIt   
      File "./tryIt", line 10, in <module>  
        j1 = Job.wrapJobFn(helloWorld, "first")  
      File "/cluster/home/markd/opt/centos6.0/x86_64/lib/python2.7/site-packages/toil/job.py", line 238, in wrapJobFn  
      File "/cluster/home/markd/opt/centos6.0/x86_64/lib/python2.7/site-packages/toil/job.py", line 1350, in __init__  
      File "/cluster/home/markd/opt/centos6.0/x86_64/lib/python2.7/site-packages/toil/resource.py", line 331, in forModule  
        assert extension in ('.py', '.pyc')  
    AssertionError

diekhans:
remove unnecessary assert that prevents toil commands from following UNIX naming conventions.

Fixes https://github.com/BD2KGenomics/toil/issues/631

Did not add tests for this, as I don't understand the test structure enough to do this.  Happy to add them if advised on the best approach.

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -328,7 +328,6 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name'))):
         filePath = os.path.abspath(module.__file__)
         filePath = filePath.split(os.path.sep)
         filePath[-1], extension = os.path.splitext(filePath[-1])
-        assert extension in ('.py', '.pyc')
         if name == '__main__':
             if module.__package__:
                 # invoked via python -m foo.bar


hannes-ucsc:
Unfortunately, the assert serves a function. This has to do with how the hot-deployment code (resource.py) ships the userscript (and its siblings) around on a Mesos cluster, or any cluster that does not have a shared FS.

diekhans:
If it's truly a user error, an assert is an incorrect way to
report it.   Asserts are intended to catch bugs in code.

However, if it's a program and not a package, it should not be a
requirement to use a non-standard naming convention.

Tell me more as to why this is is a requirement for shipping
code.

hannes-ucsc:
I agree. Feel free to change the assert to a conditional.

Reading the comments in resource.py and the documentation to `PyZipFile` should clarify the need for the extension.

diekhans:
The documentation does not explain why a program is not ship
as a file resource.

hannes-ucsc:
Don't quite understand what you mean by that.

diekhans:
did you pick up my change or fix this independently.  It was straight forward, but i could never find the
time to learn enough to test in a cloud.

hannes-ucsc:
We won't take this fix because .py[c] is actually required by Toil's hot-deployment mechanism.

But I reopened the issue this PR intends to address (#631). It was closed in error.

diekhans:
Internally, .py[c] is required by the system, however this
requirement doesn't have to be responsibility of the user.

There is the (untested) patch to address this:

  https://github.com/BD2KGenomics/toil/compare/master...diekhans:noMsDosNames

While this isn't tested in the context of Toil, the code that
does the actually work was tested in a stand-alone  program.

It's only ~15 lines of code. Much less work than explain to Unix
programmers why toil violates command line abstraction.

I can rebase and test it next month or have at it.

hannes-ucsc:
Thank you, that'd be great. I've opened #858 to track this.

PR #642, opened 2016-01-11
Eliminate cache requirement (resolves #586)

### LINKED ISSUE #586: Eliminate cache requirement

arkal:
Resolves #586

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -60,7 +60,7 @@ class Config(object):
         self.defaultMemory = 2147483648
         self.defaultCores = 1
         self.defaultDisk = 2147483648
-        self.defaultCache = self.defaultDisk
+        self.defaultCache = 0.8
         self.maxCores = sys.maxint
         self.maxMemory = sys.maxint
         self.maxDisk = sys.maxint
@@ -150,7 +150,14 @@ class Config(object):
         setOption("defaultMemory", h2b, iC(1))
         setOption("defaultCores", float, fC(1.0))
         setOption("defaultDisk", h2b, iC(1))
-        setOption("defaultCache", h2b, iC(0))
+        #  defaultCache needs to be handled a tad differently because it can be either a float b/w [0.0, 1.0] or an int.
+        x = getattr(options, 'defaultCache', None)
+        if x is None or float(x) > 1.0:
+            #  Is not a fraction of free space.  Is an absolute size.
+            setOption("defaultCache", h2b, iC(0))
+        else:
+            #  Is a fraction of free space.
+            setOption("defaultCache", float, fC(0.0, 1.0))
         setOption("maxCores", int, iC(1))
         setOption("maxMemory", h2b, iC(1))
         setOption("maxDisk", h2b, iC(1))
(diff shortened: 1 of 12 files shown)


src/toil/common.py
```diff
@@ -150,7 +150,14 @@ def fC(minValue, maxValue=None):
         setOption("defaultMemory", h2b, iC(1))
         setOption("defaultCores", float, fC(1.0))
         setOption("defaultDisk", h2b, iC(1))
-        setOption("defaultCache", h2b, iC(0))
+        #  defaultCache needs to be handled a tad differently because it can be either a float b/w [0.0, 1.0] or an int.
+        x = getattr(options, 'defaultCache', None)
```
hannes-ucsc: options.defaultCache

arkal:
The error was trivial. Need to invoke Job.CachedFileStore.terminateEvent.set() in case of failed cached run instead of Job.FileStore.terminateEvent.set(). It's fixed in the new commit. TOIL no longer hangs on raised error.

hannes-ucsc:
I filed a PR against your fork with the changes that we did together.

Other things that should be addressed, in order of importance:

- Rebase now (and keep rebasing) to latest master
- Fix the Jenkins build
- The tests should be methods of a subclass of toil.test.ToilTest.

hannes-ucsc:
The tests leave the working copy dirty. See end of build output. And keep this branch rebased onto master.

hannes-ucsc:
bd2k-python-lib 1.12.dev8 now has an alternative to shutil.copyfileobj that support a limit:

https://github.com/BD2KGenomics/bd2k-python-lib/commit/7bba82990deb7dd65776bfb172c788e9310ee0c8

You'll want to update setup.py to refer to that version of bd2k-python-lib.

arkal:
Jenkins, test this please.

hannes-ucsc:
The history of this branch is still a mess. Master has progress a bit and there a lot of "back-and-forth" commits in this branch. It also seems that this branch was never rebased? I did a trial rebase and hit a lot of conflicts. At this stage I recommend that you start a fresh branch off of master. Before you do that, make sure your master reflects the very latest upstream master (I think in your local clone that would be "bd2k/master"). Then check out your fresh branch, compare it (with PyCharm) with your issues/586a-eliminate-cache-requirement branch and apply the changes interactively in the split-pane diff editor.

From then on, be very eager to keep your fresh branch rebased on top of master. Do not merge from master. Use rebase instead.

arkal:
Jenkins, test this please.

arkal:
moved to #656 with a cleaner history.

PR #653, opened 2016-01-22
Add support for worker clean-up to single-machine and Mesos batch systems (resolves #651)

### LINKED ISSUE #651: Add support for worker clean-up to single-machine and Mesos batch systems
1. Put config.workDir and config.workflowID into a namedtuple in AbstractBatchSystem.
2. Pass this to executor
3. add a cleanup method in AbstractBatchSystem that uses this info.

arkal:
Resolves #651 

1. Implemented abstract methods in AbstractBatchSystem for this.
2. SingleMachine and Mesos call the worker cleanup function on shutdown
3. All batch systems now have a static method for indicating whether they
    support worker cleanup.

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -18,6 +18,12 @@ from collections import namedtuple
 from Queue import Empty
 import os
 
+# A class containing the information required for worker cleanup on shutdown of the batch system.
+WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
+    # A path to the value of config.workDir (where the cache would go)
+    'workDir',
+    # The value of config.workflowID (used to identify files specific to this workflow)
+    'workflowID'))
 
 class AbstractBatchSystem:
     """An abstract (as far as python currently allows) base class
@@ -48,11 +54,8 @@ class AbstractBatchSystem:
         """
         :type dict[str,str]
         """
-        self.workerCleanupInfo = namedtuple('workerCleanupInfo', (
-            # A path to the value of config.workDir (where the cache would go)
-            'workDir',
-            # The value of config.workflowID (used to identify files specific to this workflow)
-            'workflowID'))(self.config.workDir, self.config.workflowID)
+        self.workerCleanupInfo = WorkerCleanupInfo(workDir=self.config.workDir,
+                                                   workflowID=self.config.workflowID)
 
 
     def checkResourceRequest(self, memory, cores, disk):
@@ -147,13 +150,20 @@ class AbstractBatchSystem:
         return os.path.join(toilPath, "results.txt")
 
     @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
+        """
+        raise NotImplementedError('Abstract method')
+
+    @staticmethod
     def workerCleanup(workerCleanupInfo):
         '''
         Cleans up the worker node on batch system shutdown. For now it does nothing.
         :param collections.namedtuple workerCleanupInfo: A named tuple consisting of all the
         relevant information for cleaning up the worker.
         '''
-        pass
+        assert workerCleanupInfo.__class__.__name__ == 'WorkerCleanupInfo'
 
 
 class InsufficientSystemResources(Exception):
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -14,6 +14,7 @@
 
 
 from __future__ import absolute_import
+from collections import namedtuple
 from Queue import Empty
 import os
 
@@ -47,6 +48,12 @@ class AbstractBatchSystem:
         """
         :type dict[str,str]
         """
+        self.workerCleanupInfo = namedtuple('workerCleanupInfo', (
+            # A path to the value of config.workDir (where the cache would go)
+            'workDir',
+            # The value of config.workflowID (used to identify files specific to this workflow)
+            'workflowID'))(self.config.workDir, self.config.workflowID)
+
 
     def checkResourceRequest(self, memory, cores, disk):
         """Check resource request is not greater than that available.
@@ -138,6 +145,15 @@ class AbstractBatchSystem:
         and LSF currently use this.
         """
         return os.path.join(toilPath, "results.txt")
+
+    @staticmethod
+    def workerCleanup(workerCleanupInfo):
+        '''
+        Cleans up the worker node on batch system shutdown. For now it does nothing.
+        :param collections.namedtuple workerCleanupInfo: A named tuple consisting of all the
+        relevant information for cleaning up the worker.
+        '''
+        pass
 
 
 class InsufficientSystemResources(Exception):
--- a/src/toil/batchSystems/gridengine.py
+++ b/src/toil/batchSystems/gridengine.py
@@ -366,3 +366,7 @@ class GridengineBatchSystem(AbstractBatchSystem):
             raise ValueError("GridEngine does not support commata in environment variable values")
         return AbstractBatchSystem.setEnv(self, name, value)
 
+    @staticmethod
+    def supportsWorkerCleanup():
+        return False
+
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -298,3 +298,7 @@ class LSFBatchSystem(AbstractBatchSystem):
         if self.maxCPU is 0 or self.maxMEM is 0:
                 RuntimeError("lshosts returns null ncpus or maxmem info")
         logger.info("Got the maxCPU: %s" % (self.maxMEM))
+
+    @staticmethod
+    def supportsWorkerCleanup():
+        return False
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -115,7 +115,8 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
                       command=command,
                       userScript=self.userScript,
                       toilDistribution=self.toilDistribution,
-                      environment=self.environment.copy())
+                      environment=self.environment.copy(),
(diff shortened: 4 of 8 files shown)


arkal:
Haha much lol. Terminated the jenkins instance by mistake while trying to terminate a personal one

arkal:
Jenkins, test this please.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -139,6 +146,22 @@ def _getResultsFileName(self, toilPath):
         """
         return os.path.join(toilPath, "results.txt")
 
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
+        """
+        raise NotImplementedError('Abstract method: supportsWorkerCleanup')
```
hannes-ucsc: Can you remove the message argument to the exception constructor? It really is redundant and bound to become stale (I've seen this happen again and again). The stack trace will make it clear which method was not implemented.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -47,6 +48,12 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
         """
         :type dict[str,str]
         """
+        self.workerCleanupInfo = {
```
hannes-ucsc: I'd prefer a namedtuple here. Any reason you changed it back to dict?

Also take a look at the dict() built-in with keyword arguments. It makes for more consise dicts literals if their keys resemble Python identifiers. But again, I'd prefer a namedtuple.

src/toil/batchSystems/abstractBatchSystem.py
hannes-ucsc: The message argument to the exception constructor is redundant (the stacktrace will make it obvious which method is affected) and bound to get inconsistent with the actual name of the method.

src/toil/batchSystems/gridengine.py
```diff
@@ -366,3 +366,10 @@ def setEnv(self, name, value=None):
             raise ValueError("GridEngine does not support commata in environment variable values")
         return AbstractBatchSystem.setEnv(self, name, value)
 
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
```
hannes-ucsc: No need to duplicate the overridden method's docstring (unless you promise to keep these copies consistent from now until all eternity).

src/toil/batchSystems/lsf.py
```diff
@@ -298,3 +298,10 @@ def obtainSystemConstants(self):
         if self.maxCPU is 0 or self.maxMEM is 0:
                 RuntimeError("lshosts returns null ncpus or maxmem info")
         logger.info("Got the maxCPU: %s" % (self.maxMEM))
+
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
```
hannes-ucsc: No need to duplicate the overridden method's docstring (unless you promise to keep these copies consistent from now until all eternity).

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -509,3 +510,10 @@ def __mbToBytes(mem):
         used when converting Mesos reqs to Toil reqs
         """
         return mem * 1024 * 1024
+
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
```
hannes-ucsc: No need to duplicate the overridden method's docstring (unless you promise to keep these copies consistent from now until all eternity).

src/toil/batchSystems/parasol.py
```diff
@@ -346,3 +346,10 @@ def shutdown(self):
         for results in self.resultsFiles.values():
             os.remove(results)
         os.rmdir(self.parasolResultsDir)
+
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
```
hannes-ucsc: No need to duplicate the overridden method's docstring (unless you promise to keep these copies consistent from now until all eternity).

src/toil/batchSystems/singleMachine.py
```diff
@@ -225,6 +226,13 @@ def getRescueBatchJobFrequency(cls):
         """
         return 5400
 
+    @staticmethod
+    def supportsWorkerCleanup():
+        """
+        Indicates whether the batch system supports the cleanup of workers on shutdown
```
hannes-ucsc: No need to duplicate the overridden method's docstring (unless you promise to keep these copies consistent from now until all eternity).

src/toil/batchSystems/mesos/executor.py
```diff
@@ -98,7 +100,10 @@ def launchTask(self, driver, task):
         def runTask():
             log.debug("Running task %s", task.task_id.value)
             sendUpdate(mesos_pb2.TASK_RUNNING)
-            popen = runJob(pickle.loads(task.data))
+            # This is where task.data is first invoked. Using this position to setup cleanupInfo
+            taskData = pickle.loads(task.data)
+            self.workerCleanupInfo = taskData.workerCleanupInfo
```
hannes-ucsc: Make the assignment conditional (so it only happens once) and assert that subsequent tasks don't change it.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -14,6 +14,7 @@
 
 
 from __future__ import absolute_import
+from collections import namedtuple
```
hannes-ucsc: Unused import? But see below.

hannes-ucsc:
Please squash the commits and change the commit message to be consistent with the title of this PR (what I changed it to).

src/toil/batchSystems/abstractBatchSystem.py
arkal: Yes, if the named tuple is defined and setup here, It can't be easily pickled and unpickled outside this scope.
http://stackoverflow.com/questions/16377215/how-to-pickle-a-namedtuple-instance-correctly

easier to keep it a dict that is easily pickle-able

src/toil/batchSystems/abstractBatchSystem.py
arkal: will fix

src/toil/batchSystems/gridengine.py
arkal: will fix

src/toil/batchSystems/lsf.py
arkal: will fix

src/toil/batchSystems/mesos/batchSystem.py
arkal: will fix

src/toil/batchSystems/parasol.py
arkal: will fix

src/toil/batchSystems/singleMachine.py
arkal: will fix

src/toil/batchSystems/mesos/executor.py
arkal: are we expecting the values to change during a run for some reason?

src/toil/batchSystems/abstractBatchSystem.py
arkal: Sorry, this is an artifact of having attempted a namedtuple and failed before moving to a  dict.

src/toil/batchSystems/mesos/executor.py
hannes-ucsc: No. And that's what I want to assert rather than assume.

src/toil/batchSystems/abstractBatchSystem.py
hannes-ucsc: Easy isn't always better. I prefer a namedtuple because it makes two things obvious: 1) the immutability and 2) the fact that the set of keys is static across al instances. It also opens the door for adding behavior via methods. The fix is easy, just move the definition to the global scope. If you are worried about proximity between definition and use, I recommend moving the clean-up stuff towards the end of the class and have the namedtuple definition immediately below the class.

src/toil/batchSystems/mesos/executor.py
arkal: right!

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -14,9 +14,15 @@
...
 from Queue import Empty
 import os
 
+cleanupInfoTuple = namedtuple('cleanupInfoTuple', (
```
hannes-ucsc: CleanupInfoTuple since it is a class.

hannes-ucsc: Actuall that should be `WorkerCleanupInfo`.

src/toil/batchSystems/mesos/executor.py
```diff
@@ -98,7 +101,13 @@ def launchTask(self, driver, task):
         def runTask():
             log.debug("Running task %s", task.task_id.value)
             sendUpdate(mesos_pb2.TASK_RUNNING)
-            popen = runJob(pickle.loads(task.data))
+            # This is where task.data is first invoked. Using this position to setup cleanupInfo
+            taskData = pickle.loads(task.data)
+            if hasattr(self, 'workerCleanupInfo'):
```
hannes-ucsc: Too complicated. Why not just `self.workerCleanupInfo is not None`? All object attributes should be initialized in the constructor. It's bad style to introduce new attributes after construction. I can explain why face to face. If you don't have a value for an attribute yet, use None. PyCharm should warn you about this, actually.

hannes-ucsc:
Sorry for being such a stickler. Obviously, I could have done this "my way" myself in 5 minutes but the true purpose of this PR is not the code but the transfer of my experience and knowledge.

arkal:
That's fine :) It's a learning experience for me!

PR #655, opened 2016-01-24
Added multipart copy function and unit tests (resolves #645)

### LINKED ISSUE #645: Write S3 multipart copy function
Write a standalone Python function (and unit tests of that function) that copies large files from one bucket to another or within a bucket. The function should have the following signature:

```
def copy_key_multipart( src, dst ):
    """
    :param str src: a string containing a URL of the form s3://BUCKET_NAME/KEY_NAME pointing at
           the source object/key

    :param str dst: a string containing a URL of the form s3://BUCKET_NAME/KEY_NAME pointing at
           the destination object/key
    """
    pass
```

Use Python's built-in urlparse [1] module to parse the url. Use boto [2] to interface with S3. Use [3] as a reference.

[1]: https://docs.python.org/2/library/urlparse.html
[2]: http://boto.cloudhackers.com/en/latest/
[3]: http://docs.aws.amazon.com/AmazonS3/latest/dev/CopyingObjctsMPUapi.html

JakeNarkizian:
Function copy_key_multipart copies a key from a source bucket to a destination bucket in chunks of size 5MiB.

### DIFF
--- a/src/toil/test/src/mpCopyTest.py
+++ b/src/toil/test/src/mpCopyTest.py
@@ -1,0 +1,105 @@
+import boto
+import uuid
+import math
+import itertools
+import sys
+import unittest
+import io
+
+from bd2k.util.expando import Expando
+from toil.mpcopy import copy_key_multipart
+
+
+class CopyKeyMultipartTest(unittest.TestCase):
+    def setUp(self):
+        super(CopyKeyMultipartTest, self).setUp()
+        self.s3 = boto.connect_s3()
+
+        srcName = 'mp_copy_test_src_bucket_' + str(uuid.uuid1())
+        self.srcBucket = self.s3.create_bucket(srcName)
+        dstName = 'mp_copy_test_dst_bucket_' + str(uuid.uuid1())
+        self.dstBucket = self.s3.create_bucket(dstName)
+
+    def tearDown(self):
+        super(CopyKeyMultipartTest, self).tearDown()
+        for bkt in [self.srcBucket, self.dstBucket]:
+            for key in bkt.list():
+                key.delete()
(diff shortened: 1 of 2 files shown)


arkal:
Jenkins, test this please.

src/toil/test/src/mpCopyTest.py
```diff
@@ -0,0 +1,79 @@
...
+from toil.mpcopy import copy_key_multipart
+
+
+chunk_size = int(2**20 * 5)  # if this value is changed the etags will not match
+test_upload_size = int(2**20 * 5.3)
+
+
+class CopyKeyMultipartTest(unittest.TestCase):
+    def setUp(self):
+        self.fileid = 'file_'+str(uuid.uuid1())
```
hannes-ucsc: Must always call overridden method in super class.

src/toil/test/src/mpCopyTest.py
```diff
@@ -0,0 +1,79 @@
...
+                assert start == self.totalSize
+            except:
+                upload.cancel_upload()
+                raise
+            else:
+                upload.complete_upload()
+
+
+    def teardown(self):
+        for bkt in [self.src['bucket'], self.dst['bucket']]:
```
hannes-ucsc: Isn't it `tearDown()`, not `teardown()`?

Assuming I'm right, this method was never called. So it's disconcerting that the lack of clean up wasn't noticed earlier.

hannes-ucsc:
Please split into separate modules like I asked in my email.

JakeNarkizian:
I apologize I thought you meant to split the copy function as the code under test into its own module. Which code were you referring to?

JakeNarkizian:
Resolved with #709

PR #659, opened 2016-01-26
Import and Export functions added to AWS JobStore and FileJobStore (resolves #618)

### LINKED ISSUE #618: Add jobStore.importFile() and .exportFile()
Signatures should be 

```
importFile(sourceUrl:str) -> jobStoreFileId:str
exportFile(jobStoreFileId:str, destUrl:str)
```

Every concrete job store would be required to support `file:` URLs and URLs pointing to its native storage mechanism, e.g. `s3:` for the AWSJobStore.

JakeNarkizian:
These methods are exposed to the user in job.FileStore. The test will fail because the multipart copy function has not beed added.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -711,6 +711,12 @@ class Job(object):
                 #This will result in the files removal from the cache at the end of the current job
                 self._jobStoreFileIDToCacheLocation.pop(fileStoreID)
 
+        def importFile(self, sourceUrl):
+            return self.jobStore.importFile(sourceUrl)
+
+        def exportFile(self, jobStoreFileID, destUrl):
+            self.jobStore.exportFile(jobStoreFileID, destUrl)
+
         def logToMaster(self, text, level=logging.INFO):
             """
             Send a logging message to the leader. The message will also be \
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -422,6 +422,30 @@ class AbstractJobStore(object):
         """
         raise NotImplementedError()
 
+    fileUrlRegex = re.compile("[fF]ile://.+$")
+
+    @abstractmethod
+    def importFile(self, sourceUrl):
+        """
+        Imports file pointed at by sourceUrl into jobstore. The jobStoreFileId of the new file is
+        returned.
+
+        :param sourceUrl: URL of scheme 'file:' pointing to local files and urls pointing to the
+            local file scheme of the jobstore, e.g. 's3:' for AWS
+        :return jobStoreFileId
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def exportFile(self, jobStoreFileId, destUrl):
+        """
+        Exports file to destination pointed at by destUrl.
+
+        :param destUrl: URL of scheme 'file:' pointing to local files and urls pointing to the
+            local file scheme of the jobstore, e.g. 's3:' for AWS
+        """
+        raise NotImplementedError()
+
     ##########################################
     # The following methods deal with shared files, i.e. files not associated
     # with specific jobs.
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -24,6 +24,8 @@ import base64
 import hashlib
 import itertools
 import repr as reprlib
+import urlparse
+import shutil
 
 from bd2k.util import strict_bool
 from bd2k.util.objects import InnerClass
@@ -46,6 +48,44 @@ from toil.jobWrapper import JobWrapper
 import toil.lib.encryption as encryption
 
 log = logging.getLogger(__name__)
+
+
+def copy_key_multipart(srcBucket, srcKey, dstBucket, dstKeyName, part_size=None, headers={}):
+    """
+    Copies a key pointed at by the src URL to the dst URL in multiple parts.
+
+    :param srcBucket: the bucket containing the key to be copied from
+    :param srcKey: the key to be copied from
+    :param dstBucket: the bucket containing the key to be copied to
+    :param str dstKeyName: the name of the key that will be copied to
+    :param int part_size: override the default part_size of 50MiB in bytes
+    :param headers: provide enc
+    """
+    # set part_size
+    partSize = part_size or 2**20 * 50  # default size of 50MiB
+    assert partSize > int(2**20 * 5) - 1  # part_size too small
+    assert partSize < int(2**30 * 5) + 1  # part_size too big
+    totalSize = srcKey.size
+
+    # initiate copy
+    upload = dstBucket.initiate_multipart_upload(dstKeyName, headers=headers)
+    try:
+        start = 0
+        partIndex = itertools.count()
+        while start < totalSize:
+            end = min(start + partSize, totalSize)
+            upload.copy_part_from_key(src_bucket_name=srcBucket.name,
+                                      src_key_name=srcKey.name,
+                                      part_num=next(partIndex)+1,
+                                      start=start,
+                                      end=end-1,  # end is inclusive
+                                      headers=headers)
+            start += partSize
+    except:
+        upload.cancel_upload()
+        raise
+    else:
+        return upload.complete_upload()
 
 
 class AWSJobStore(AbstractJobStore):
@@ -298,6 +338,74 @@ class AWSJobStore(AbstractJobStore):
             log.debug("File %s does not exist, skipping deletion.", jobStoreFileID)
         else:
             info.delete()
+
+    nativeUrlRegex = re.compile(r"""
+        # url scheme
+        [sS]3://
+
+        # bucket name
+        #   (1) Must have length 3-255
+        #   (2) Contain only letters, numbers, dashes(-), periods(.), and underscores(_)
+        #   (3) successive periods are not allowed
+        #   (4) bucket name must start with a letter or number
+        [a-z0-9]   (?=(?:.{2,254}/))   ([a-z0-9-_] | ([.][a-z0-9-_]))*/
+
+        # key name
+        .+$
+    """, re.VERBOSE)
+
+    def importFile(self, sourceUrl):
+        url = urlparse.urlparse(sourceUrl)
+        if self.nativeUrlRegex.match(sourceUrl):
+            srcBucketName = url.netloc
+            srcKeyName = (url.path)[1:]
+            srcBucket = self._loadBucketOrFail(srcBucketName)
+
+            # create new file
+            info = self.FileInfo.create(srcKeyName)
+            info.copyFrom(srcBucket, srcKeyName)
+            info.save()
+
+        elif self.fileUrlRegex.match(sourceUrl):
+            localFilePath = url.netloc + url.path
+            if not os.path.exists(localFilePath):
+                raise RuntimeError("File pointed at by url '%s' does not exist" % sourceUrl)
+
+            # create new file
+            info = self.FileInfo.create(localFilePath.split('/')[-1])
+            info.upload(localFilePath)
+            info.save()
+
+        else:
+            raise RuntimeError("The url '%s' is not a valid file or s3 url" % sourceUrl)
+
+        return info.fileID
+
+    def exportFile(self, jobStoreFileID, destUrl):
+        url = urlparse.urlparse(destUrl)
+        info = self.FileInfo.loadOrFail(jobStoreFileID)
+
+        if self.nativeUrlRegex.match(destUrl):
+            dstBucket = self._loadBucketOrFail(url.netloc)
+            info.copyTo(dstBucket, url.path[1:])  # url path is the dest key name
+
+        elif self.fileUrlRegex.match(destUrl):
+            localFilePath = url.netloc + url.path
+            open(localFilePath, 'w+').close()  # NOTE: if the file already exists it will be overwritten here
+            self.readFile(jobStoreFileID, localFilePath)
+
+        else:
+            raise RuntimeError("The url '%s' is not a valid file or s3 url" % destUrl)
+
+    def _loadBucketOrFail(self, bucketName):
+        s3 = boto.connect_s3()
+        try:
+            b = s3.get_bucket(bucketName)
+        except S3ResponseError as e:
+            if e.error_code == 404:
+                raise RuntimeError("Bucket '%s' does not exist" % bucketName)
+            raise
+        return b
 
     def writeStatsAndLogging(self, statsAndLoggingString):
         info = self.FileInfo.create(str(self.statsFileOwnerID))
@@ -675,6 +783,71 @@ class AWSJobStore(AbstractJobStore):
                 assert key.size == file_size
                 # Make resonably sure that the file wasn't touched during the upload
                 assert self._fileSizeAndTime(localFilePath) == (file_size, file_time)
+
+        def copyFrom(self, srcBucket, srcKeyName):
+            srcKey = srcBucket.get_key(srcKeyName)
+            if srcKey is None:
+                raise RuntimeError("key '%s' does not exists in bucket '%s'" % (srcKeyName, srcBucket))
+
+            headers = self._s3EncryptionHeaders()
+            if srcKey.size <= self._maxInlinedSize():
+                self.content = srcKey.get_contents_as_string()
+
+            elif srcKey.size < (2**20 * 5):
+                self.version = self.outer.filesBucket.copy_key(new_key_name=self.fileID,
+                                                               src_bucket_name=srcBucket.name,
+                                                               src_key_name=srcKey.name,
+                                                               headers=headers).version_id
+            else:
+                self.version = copy_key_multipart(srcBucket=srcBucket,
+                                                  srcKey=srcKey,
+                                                  dstBucket=self.outer.filesBucket,
+                                                  dstKeyName=self.fileID,
+                                                  headers=headers).version_id
+            # compare hashes
+            if self.encrypted:
+                log.debug("MD5 hash of imported or exported files cannot be checked."
+                          "It is possible that the file '%s' does not match the imported or exported file"
+                          % self._fileID)
+            elif srcKey.size <= self._maxInlinedSize():
+                assert hashlib.md5(self.content).hexdigest() == srcKey.etag[1:-1]
+            else:
+                assert self.outer.filesBucket.get_key(self._fileID).etag[1:-1] == srcKey.etag[1:-1]
+
+        def copyTo(self, dstBucket, dstKeyName):
+            headers = self._s3EncryptionHeaders()
+            if self.content is not None:
+                dstKey = dstBucket.new_key(dstKeyName)
+                dstKey.set_contents_from_string(self.content)
+                srcHash = hashlib.md5(self.content).hexdigest()
+
+            elif self.version:
+                copyHeaders = {}
+                for key in headers.keys():
+                    copyHeaders["amz-copy-source".join(key.split('amz'))] = headers[key]
+
+                if self.outer.filesBucket.get_key(self._fileID, headers=headers).size < (2**20 * 5):
+                    dstBucket.copy_key(new_key_name=dstKeyName,
+                                       src_bucket_name=self.outer.filesBucket.name,
+                                       src_key_name=self.fileID,
+                                       headers=copyHeaders)
+                else:
+                    copy_key_multipart(srcBucket=self.outer.filesBucket,
+                                       srcKey=self.outer.filesBucket.get_key(self.fileID, headers=headers),
+                                       dstBucket=dstBucket,
+                                       dstKeyName=dstKeyName,
+                                       headers=copyHeaders).version_id
+                srcHash = self.outer.filesBucket.get_key(self._fileID, headers=headers).etag[1:-1]
+
+            else:
+                assert False
+
+            if self.encrypted:
+                log.debug("MD5 hash of imported or exported files cannot be checked."
+                          "It is possible that the file '%s' does not match the imported or exported file"
+                          % self._fileID)
+            else:
+                assert dstBucket.get_key(dstKeyName).etag[1:-1] == srcHash
 
         @contextmanager
         def uploadStream(self, multipart=True, allowInlining=True):
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+
+import hashlib
 from contextlib import contextmanager
 import logging
 import marshal as pickler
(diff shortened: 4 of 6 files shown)


JakeNarkizian:
Jenkins, test this please.

JakeNarkizian:
Jenkins, retest this please.

hannes-ucsc:
Set needs_work label because encryption support needs to be added.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +303,65 @@ def deleteFile(self, jobStoreFileID):
         else:
             info.delete()
 
+    def importFile(self, sourceUrl):
+        url = urlparse.urlparse(sourceUrl)
+        if url.scheme.lower() == 's3':
+            if len(sourceUrl.split('/')) != 4:
```
JakeNarkizian: It would be better to use a regex to check if the URL is syntactically valid in the actual if statements e.g. if re.match(<s3 pattern>, url) is not None: ... elif re.match(<s3 pattern>, url) is not None: ... else: runtime error

src/toil/jobStores/fileJobStore.py
```diff
@@ -187,6 +191,43 @@ def fileExists(self, jobStoreFileID):
             raise NoSuchFileException("Path %s is not a file in the jobStore" % jobStoreFileID)
         return True
 
+    def importFile(self, sourceUrl, jobStoreID=None):
+        url = urlparse.urlparse(sourceUrl)
+        if url.scheme.lower() == 'file':
```
JakeNarkizian: not as big of a deal here but these checks should also use regex's for consistency

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +303,65 @@ def deleteFile(self, jobStoreFileID):
...
+
+        else:
+            raise RuntimeError("url scheme '%s' is not supported" % url.scheme)
+        return info.fileID
+
+    def exportFile(self, jobStoreFileID, destUrl):
+        url = urlparse.urlparse(destUrl)
+        info = self.FileInfo.loadOrFail(jobStoreFileID)
+
+        if url.scheme.lower() == 's3':
```
JakeNarkizian: likewise for regex's here

src/toil/jobStores/fileJobStore.py
```diff
@@ -187,6 +191,43 @@ def fileExists(self, jobStoreFileID):
...
+
+            JobStoreFileId = jobStoreID or self._getRelativePath(absPath)
+            self._assertFileMatchesJobStoreFileID(JobStoreFileId, localFilePath)
+            return JobStoreFileId
+        else:
+            raise RuntimeError("url scheme '%s' is not supported" % url.scheme)
+
+    def exportFile(self, jobStoreFileId, destUrl):
+        url = urlparse.urlparse(destUrl)
+        if url.scheme.lower() == 'file':
```
JakeNarkizian: not as big of a deal here but these checks should also use regex's for consistency

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    # set part_size
+    part_size = part_size or int(2**20 * 50)  # default size of 5MiB
+    assert part_size > int(2**20 * 5) - 1  # part_size too small
+    assert part_size < int(2**30 * 5) + 1  # part_size too big
+
+    # parse URLs
+    srcUrl = urlparse(src)
+    dstUrl = urlparse(dst)
+    assert dstUrl.scheme == srcUrl.scheme == 's3'
```
JakeNarkizian: redundant check here. should be eliminated.

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+import boto
+import math
+import itertools
+
+
+class S3File:
+    def __init__(self, url):
+        self.s3 = boto.connect_s3()
+        self.url = urlparse(url)
+        assert self.url.scheme == 's3'
```
JakeNarkizian: should also use regex

src/toil/jobStores/aws/jobStore.py
JakeNarkizian: note also that the regex's should be precompiled and stored in a higher level class variable above the import and export functions so that they do not need to be compiled on each function call.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -422,6 +422,28 @@ def updateFileStream(self, jobStoreFileID):
...
         raise NotImplementedError()
 
+    @abstractmethod
+    def importFile(self, sourceUrl):
+        """
+        Imports file pointed at by sourceUrl into filesBucket of jobstore. The jobStoreFileId of
+        the new file is returned.
+
+        :param sourceUrl: URL of scheme 'file:' pointing to local files and urls pointing to the
+        local file scheme of the jobstore, e.g. 's3:' for AWS
```
cket: I think that Sphinx will complain about this docstring when we build the docs. If the :param description contains multiple lines it should have a \ to delimitate each line. To confirm this, you can run 'make docs' and see what output is generated. See the Job class for examples on properly formatted multiline descriptions.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -422,6 +422,28 @@ def updateFileStream(self, jobStoreFileID):
...
+        :return jobStoreFileId
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def exportFile(self, jobStoreFileId, destUrl):
+        """
+        Exports file to destination pointed at by destUrl.
+
+        :param destUrl: URL of scheme 'file:' pointing to local files and urls pointing to the
```
cket: Same here

src/toil/job.py
```diff
@@ -711,6 +711,12 @@ def deleteGlobalFile(self, fileStoreID):
                 #This will result in the files removal from the cache at the end of the current job
                 self._jobStoreFileIDToCacheLocation.pop(fileStoreID)
 
+        def importFile(self, sourceUrl):
+            return self.jobStore.importFile(sourceUrl)
+
+        def exportFile(self, jobStoreFileId, destUrl):
+            self.jobStore.exportFile(jobStoreFileId, destUrl)
```
cket: Should be jobStoreFileID (vs. Id) to be consistent with the rest of the file

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +303,65 @@ def deleteFile(self, jobStoreFileID):
         else:
             info.delete()
 
+    def importFile(self, sourceUrl):
+        url = urlparse.urlparse(sourceUrl)
+        if url.scheme.lower() == 's3':
+            if len(sourceUrl.split('/')) != 4:
+                raise RuntimeError("The url '%s' is not a valid s3 url" % sourceUrl)
+            srcBucket = self._loadBucketOrFail(url.netloc)
```
cket: Consider refactoring url.netloc and url.path[1:] into more informative variable names like bucketName and keyName

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +303,65 @@ def deleteFile(self, jobStoreFileID):
...
+
+            try:
+                srcKey = srcBucket.get_key((url.path)[1:])
+                assert srcKey is not None
+            except AssertionError:
+                raise RuntimeError("key '%s' does not exists in bucket '%s'" % ((url.path)[1:], url.netloc))
```
cket: I would refactor this into an if statement that checks if the key is none. try/except can be more efficient than if statements in cases like 
```python
if key in dictionary:
    return dictionary[key]
```
where we are doing a redundant key lookup, but in this case the if statement would be just as efficient as well as more concise.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -676,6 +739,69 @@ def upload(self, localFilePath):
                 # Make resonably sure that the file wasn't touched during the upload
                 assert self._fileSizeAndTime(localFilePath) == (file_size, file_time)
 
+        def copyFrom(self, srcBucket, srcKeyName):
+            srcKey = srcBucket.get_key(srcKeyName)
+            assert srcKey is not None
+
+            headers = self._s3EncryptionHeaders()
+            dstHash = None
```
cket: This variable seems to be unused

src/toil/jobStores/aws/jobStore.py
```diff
@@ -676,6 +739,69 @@ def upload(self, localFilePath):
...
+                                       src_key_name=self.fileID,
+                                       headers=copyHeaders)
+                else:
+                    copy_key_multipart(src="s3://%s/%s" % (self.outer.filesBucket.name, self.fileID),
+                                       dst="s3://%s/%s" % (dstBucket.name, dstKeyName),
+                                       totalSize=self.outer.filesBucket.lookup(self.fileID, headers=headers).size,
+                                       headers=copyHeaders)
+                srcHash = self.outer.filesBucket.get_key(self._fileID, headers=headers).etag[1:-1]
+            else:
+                assert False
```
cket: I would raise a more descriptive runtime error

src/toil/jobStores/aws/jobStore.py
```diff
@@ -676,6 +739,69 @@ def upload(self, localFilePath):
...
+
+            if self.encrypted:
+                log.debug("MD5 hash of imported or exported files cannot be checked."
+                          "It is possible that the file '%s' does not match the imported or exported file"
+                          % self._fileID)
+            else:
+                assert dstBucket.get_key(dstKeyName).etag[1:-1] == srcHash
+
+
+
```
cket: I think PEP8 will complain about too many empty lines here

src/toil/jobStores/fileJobStore.py
```diff
@@ -187,6 +191,43 @@ def fileExists(self, jobStoreFileID):
...
+            localFilePath = url.netloc + url.path
+
+            fd, absPath = self._getTempFile(jobStoreID)
+            shutil.copyfile(localFilePath, absPath)
+            os.close(fd)
+
+            JobStoreFileId = jobStoreID or self._getRelativePath(absPath)
+            self._assertFileMatchesJobStoreFileID(JobStoreFileId, localFilePath)
+            return JobStoreFileId
```
cket: Should be lower case 'j' and upper case 'D' (jobStoreFileID vs. JobStoreFileId)

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+    def copyTo(self, dest, totalSize, part_size=None, headers={}):
+        partSize = part_size or 2**20 * 50  # 50MiB
+
+        # initiate copy
+        upload = dest.bucket.initiate_multipart_upload(dest.key, headers=headers)
+        try:
+            start = 0
+            partIndex = itertools.count()
+            while start < totalSize:
+                end = min(start + partSize-1, totalSize-1)  # bytes are indexed form zero
```
cket: Why is partSize decremented by 1? Doesn't this mean we will always be uploading (partSize-1) bytes at a time instead of the specified partSize? Also I think it should be "indexed from zero" vs. "indexed form zero"

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+            upload.cancel_upload()
+            raise
+        else:
+            return upload.complete_upload()
+
+
+def copy_key_multipart(src, dst, totalSize, part_size=None, headers={}):
+    """
+    :param str src: a string containing a URL of the form s3://BUCKET_NAME/KEY_NAME pointing at
+           the source object/key
```
cket: I like the type annotations here, and the indented second line is a good alternative to using the \ as a delimiter

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+    :param str dst: a string containing a URL of the form s3://BUCKET_NAME/KEY_NAME pointing at
+           the destination object/key
+    :param int part_size: override the default part_size of 50MiB
+    :param headers: provide enc
```
cket: Can you expand on this description?

src/toil/mpcopy.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # parse URLs
+    srcUrl = urlparse(src)
+    dstUrl = urlparse(dst)
+    assert dstUrl.scheme == srcUrl.scheme == 's3'
+
+    srcFile = S3File(src)
+    dstFile = S3File(dst)
+
+    return srcFile.copyTo(dstFile, totalSize, part_size=part_size, headers=headers)
```
cket: This file will be eliminated in the final PR but for future reference every file should end with a newline

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -494,6 +558,35 @@ def _batchDeletionSize(self):
...
             return 5 * 1024 * 1024
 
+        def _hashFile(self, localFilePath):
+            with open(localFilePath, 'r') as f:
+                return hashlib.md5(f.read()).hexdigest()
+
+        def _hashJobStoreFileID(self, JobStoreFileID):
+            pass
+
+        def _runTestMethodWithParams(self, testFunc, params):
```
cket: Very cool way to generate this tests!

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -483,6 +500,53 @@ def testManyJobs(self):
...
+        def testExportFileWithNonExistantJobStoreFileID(self):
+            url = 'file://./thisfileshouldntexist%s' % uuid.uuid1()
+            self.master.exportFile(uuid.uuid1(), url)
+
+        def _importFileFromFileOfSize(self, size):
+            """
+            General function that creates a file on disk and imports it into the job store.
+            This functionality is needed to export files so it has been made its own function.
+            """
+            path = './testFileOfSize_%d_%s' % (size, uuid.uuid1())
```
cket: Is there a reason you are using uuid1? I think uuid4 is the standard

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -529,11 +622,44 @@ class FileJobStoreTest(hidden.AbstractJobStoreTest):
...
+            return hashlib.md5(f.read()).hexdigest()
+
 
 @needs_aws
 class AWSJobStoreTest(hidden.AbstractJobStoreTest):
     testRegion = 'us-west-2'
 
+    def setUp(self):
+        self.s3 = boto.connect_s3()
+        self.srcBucket = self.s3.create_bucket('im-ex-portfile_src%s' % uuid.uuid1())
```
cket: Another uuid1

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -564,6 +690,109 @@ def _batchDeletionSize(self):
...
+        keyName = self.__uploadToBucketKeyOfSize(size).name
+        srcUrl = 's3://%s/%s' % (self.srcBucket.name, keyName)
+        JobStoreFileID = self.master.importFile(srcUrl)
+        self.assertEqual(self._hashKey(keyName), self._hashJobStoreFileID(JobStoreFileID))
+        return JobStoreFileID
+
+    def _hashFile(self, localFilePath):
+        totalSize = os.path.getsize(localFilePath)
+        with open(localFilePath, 'r') as f:
+            if totalSize <= 2**20 * 5:
```
cket: Isn't this equivalent to the partSize variable? If you define it at the beginning of the function then you can avoid this expression. Also, it may make sense to pull the default partSize out into a class variable so you can avoid redefining it over and over

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -564,6 +690,109 @@ def _batchDeletionSize(self):
...
+        :return: a pointer to the key that was uploaded
+        """
+        partSize = 2**20 * 50
+
+        bucket = self.s3.get_bucket(bucketName) if bucketName is not None else self.srcBucket
+        keyName = '%.0fMiB_file_%s' % (float(file_size/(2**20)), str(uuid.uuid1()))
+        key = Key(bucket)
+        key.key = keyName
+        with open('/dev/urandom', 'r') as f:
+            if file_size < 2**20*5:
```
cket: equivalent to partSize

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -564,6 +690,109 @@ def _batchDeletionSize(self):
...
+        return key.etag[1:-1]
+    
+    def _hashKey(self, keyName, bucketName=None):
+        bucket = self.s3.get_bucket(bucketName) if bucketName is not None else self.srcBucket
+        key = bucket.get_key(keyName)
+        if key is None:
+            raise RuntimeError("key '%s' does not exist in bucket '%s'" % (keyName, bucketName))
+        return key.etag[1:-1]
+
+    def __uploadToBucketKeyOfSize(self, file_size, bucketName=None):
```
cket: file_size should be camelcase

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -564,6 +690,109 @@ def _batchDeletionSize(self):
...
+        with open('/dev/urandom', 'r') as f:
+            if file_size < 2**20*5:
+                key.set_contents_from_string(f.read(file_size))
+            else:  # Multipart
+                mp = bucket.initiate_multipart_upload(key_name=keyName)
+                start = 0
+                partNum = itertools.count()
+                try:
+                    while start <= file_size:
+                        end = min(start + partSize, file_size)
```
cket: Why isn't it file_size-1 like in the copyTo function?

cket:
Done reviewing for now. Looks good, mostly stylistic issues. Also needs to be rebased at some point

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: I'd prefer indention instead of backslashes, if possible. I.e.

```
:param sourceUrl: URL of scheme 'file:' pointing to local files and urls pointing to the
       local file scheme of the jobstore, e.g. 's3:' for AWS
```

src/toil/mpcopy.py
JakeNarkizian: partSize is decremented to ensure that each part is only of size partSize. end is inclusive thus end - start = partSize -1

hannes-ucsc: I always think and code with half-open intervals (end exclusive) and zero-based indices. When I interface with code that doesn't follow that convention I do the conversion as late as possible, e.g. ` copy_part_from_key(...,end=end-1)`.

src/toil/test/jobStores/jobStoreTest.py
JakeNarkizian: no

src/toil/test/jobStores/jobStoreTest.py
JakeNarkizian: because the upload_part_from_file we are specifying size, its irrelevant now anyway because of Hannes's earlier suggestion to treat end as exclusive until the last moment.

JakeNarkizian:
jenkins, test this please.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,43 @@
 
 log = logging.getLogger(__name__)
 
+def copy_key_multipart(srcBucket, srcKey, dstBucket, dstKeyName, part_size=None, headers={}):
```
cket: part_size is should be camel case here too

JakeNarkizian: I was told to make that variable in that function snake case. Is that still the case @hannes-ucsc ?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +339,73 @@ def deleteFile(self, jobStoreFileID):
         else:
             info.delete()
 
+    nativeUrlRegex = re.compile(r"""
+        # url scheme
+        [sS]3://
+
+        # bucket name
+        #   (1) Must have length 3-255
+        #   (2) Contain only letters, numbers, dashes(-), periods(.), and underscores(_)
```
cket: I don't think this is true, by default our toil buckets are appended with "--files" and "--stats"

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +339,73 @@ def deleteFile(self, jobStoreFileID):
...
+        #   (3) Dashes and periods must be preceded and succeeded by a letter or number
```
cket: I don't think this is true, by default our toil buckets are appended with "--files" and "--stats"

src/toil/jobStores/aws/jobStore.py
```diff
@@ -299,6 +339,73 @@ def deleteFile(self, jobStoreFileID):
         else:
             info.delete()
 
+    nativeUrlRegex = re.compile(r"""
+        # url scheme
+        [sS]3://
+
+        # bucket name
+        #   (1) Must have length 3-255
```
cket: The allowed length was actually 3-63, check out http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

JakeNarkizian: "The rules for bucket names in the US East (N. Virginia) region allow bucket names to be as long as 255" Just covering all the bases

src/toil/jobStores/fileJobStore.py
```diff
@@ -187,6 +193,44 @@ def fileExists(self, jobStoreFileID):
...
+
+            fd, absPath = self._getTempFile(jobStoreID)
+            shutil.copyfile(localFilePath, absPath)
+            os.close(fd)
+
+            jobStoreFileID = jobStoreID or self._getRelativePath(absPath)
+            self._assertFileMatchesJobStoreFileID(jobStoreFileID, localFilePath)
+            return jobStoreFileID
+        else:
+            raise RuntimeError("The url '%s' is not a valid file or s3 url" % sourceUrl)
```
cket: Can you import from s3 to the fileStore? If not I don't think it should be mentioned in the error

src/toil/jobStores/fileJobStore.py
```diff
@@ -187,6 +193,44 @@ def fileExists(self, jobStoreFileID):
...
+
+    def exportFile(self, jobStoreFileId, destUrl):
+        url = urlparse.urlparse(destUrl)
+        if self.fileUrlRegex.match(destUrl):
+            localFilePath = url.netloc + url.path
+            self.readFile(jobStoreFileId, localFilePath)
+            self._assertFileMatchesJobStoreFileID(jobStoreFileId, localFilePath)
+        else:
+            raise RuntimeError("The url '%s' is not a valid file or s3 url" % destUrl)
```
cket: Same question here

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -483,6 +497,53 @@ def testManyJobs(self):
             # Make sure we have the right number of jobs
             self.assertEquals(len(allJobs), 3001)
 
+        def testImportFileVaryingFileSizes(self):
+            self._runTestMethodWithParams(self._importFileFromFileOfSize, [0, 1, 2**20, 2**20*5])
```
cket: I would make the larger sizes variables and/or include comments so it's more obvious why these sizes are being chosen

hannes-ucsc:
Continued in #709

PR #670, opened 2016-02-09
Add logging when rejecting a resource offer (resolves #669)

### LINKED ISSUE #669: Print warning message if Mesos offers are repeatedly declined
When offers are declined a DEBUG level log message is generated. That should be a WARNING level message. I've seen several cases where user's are dumbfounded as to why their Toil script is hanging. However, the warning needs to not be printed repeatedly, or at least not too frequently. Otherwise, it becomes noise that users ignore.

fnothaft:
Currently, when we reject a resource offer because it does not satisfy the resource requirements for any of the currently offered tasks, we only log at debug level. Additionally, we don't explain why the offer was rejected, and we only explain for the job with the fewest cores requested.

This commit checks to see if we are rejecting a resource offer while we do not have any jobs currently running (i.e., we are hung/not making progress), and logs the reason that the currently offered jobs are not being scheduled at the WARNING logging level.

### DIFF
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -379,6 +379,11 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
                         remainingCores -= job_type.cores
                         remainingMem -= self.__bytesToMB(job_type.memory)
                         remainingStor -= job_type.disk
+
+                    else:
+                        # if task is already running (odd corner case above), log this
+                        log.warn("Skipping task %s, which is already in running job map." % task.task_id.value)
+
                     nextToLaunchIndex += 1
 
             # If we put the launch call inside the while loop, multiple accepts are used on the same offer.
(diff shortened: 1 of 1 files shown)


src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -388,9 +393,47 @@ def resourceOffers(self, driver, offers):
...
+                        resource_limits.append("insufficient disk space (%d MB offered, %d MB requested)" % (offerStor, self.__bytesToMB(job_type.disk)))
+
+                    # log the resource request as well as the explanation
+                    task_explanations.append("%s: %s" % (job_type, ", ".join(resource_limits)))
+
+                log.warn("Although no jobs are running, offer %s with memory %dMB, %d cores, and disk %dMB is not large enough to run any tasks." % 
+                          (offer.id.value, offerMem, offerCores, offerStor))
+                log.warn("Pending job types are:\n%s" % '\n'.join(task_explanations))
+
+            else:
```
fnothaft: Note: this should be `elif len(tasks) == 0:` not `else:`. Let me fix this.

hannes-ucsc: Might want to nest an `if` instead of evaluating len(tasks) twice. Also `if len(foo) ==  0:` is equivalent to `if not foo:`. That plus inverting the condition and swapping the `if` and `else` bodies will usually make for more readable code.

hannes-ucsc: Also, could this be extracted into a method? I feel like it distracts from the main flow if left inline.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -379,6 +379,11 @@ def resourceOffers(self, driver, offers):
                         remainingCores -= job_type.cores
                         remainingMem -= self.__bytesToMB(job_type.memory)
                         remainingStor -= job_type.disk
+
```
hannes-ucsc: In Toil, we usually don't have empty lines before `else`.

hannes-ucsc:
Ok, I stared at both PRs (#670 and #680) for a while now and talked to @arkal. I think @arkal, @cket, @fnothaft and I should meet and discuss the consolidation of these two PRs. Main points:

* #670 makes the log message very verbose but conditional on whether jobs are running
* #680 hides useful information in a debug() level message but it can't be made warn() because it would be printed redundantly for all offers until a matching offer occurs
* I'm concerned about the lookup in the "running jobs map" I can't seem to think of a single case where this can happen. #680 raises an exception, #670 logs a warning
* #680 calls declineOffer() and I think that's actually a big deal since it fixes what I can only consider a giant bug in the original code

hannes-ucsc:
I reworked #680 into #753 and merged it. I still want to compare the two approaches.

PR #680, opened 2016-02-12
Fixed Mesos inadequate offer log (resolves #669)

### LINKED ISSUE #669: Print warning message if Mesos offers are repeatedly declined

### LINKED ISSUE #679: Mesos batch system reporting wrong requirements when offer is inadequate
When an offer is received, all job_types (including those with no available jobs) are iterated through to search for an available (unscheduled), compatible job. If no hits are produced, the log reports the resources of the one with the lowest cpu requirements (based on how the edict keys are sorted), whether it had an available job or not.

arkal:
Resolves #679 which duplicates #669

If the following is run on a m3.medium instance (effective resources: 1cpu, 3.7G mem, 3.7G disk, however Mesos only offers 1.9G disk and 2.7G mem),
```
   F = Job.wrapJobFn(f, memory='1G', cores=0.5, disk='1G')
    G = Job.wrapJobFn(f, memory='1G', cores=0.5, disk='1.5G')
    H = Job.wrapJobFn(f, memory='1G', cores=0.5, disk='1.5G')
    I = Job.wrapJobFn(f, memory='1G', cores=0.5, disk='3.5G')
    F.addChild(G)
    F.addChild(H)
    G.addChild(I)
    H.addChild(I)
```
Then the error while attempting to run I (requires 3.5G, Mesos will only offer 1.9) is
```
Received offer e8272c64-bf86-4aef-bff5-5a2c82c205a6-O435 with cores: 1.0, disk: 1981.0, and mem: 2740.0
Offer e8272c64-bf86-4aef-bff5-5a2c82c205a6-O435 not large enough to run any tasks. Required: ResourceRequirement(memory=1073741824, cores=0.5, disk=1073741824) Offered: (2873098240.0, 1.0, 2077229056.0)
```
This stems from `MesosBatchSystem.resourceOffers` using at `job_types[-1]` in the error instead of the `job_type`s actually probed.

Regarding why mesos is offering less that the actual amount, 
```
(The slave will reserve 1 GB or 50% of detected memory, whichever is
smaller, in order to run itself and other operating system services. Likewise, it
will reserve 5 GB or 50% of detected disk, whichever is smaller.)
        - Greenberg, David. "Building Applications on Mesos." O'Reilly, 2015.
```

### DIFF
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -352,8 +352,8 @@ class MesosBatchSystem(AbstractBatchSystem, mesos.interface.Scheduler):
 
         # Right now, gives priority to largest jobs
         for offer in offers:
-            tasks = []
             # TODO: In an offer, can there ever be more than one resource with the same name?
+            tasks = []
             offerCores, offerMem, offerStor = self._determineOfferResources(offer)
             log.debug("Received offer %s with cores: %s, disk: %s, and mem: %s" \
                       % (offer.id.value, offerCores, offerStor, offerMem))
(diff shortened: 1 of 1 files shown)


src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -362,6 +363,9 @@ def resourceOffers(self, driver, offers):
             remainingStor = offerStor
 
             for job_type in job_types:
+                if len(self.jobQueueList[job_type]) == 0:
```
hannes-ucsc: `if not self.jobQueueList[job_type]:`

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -373,25 +377,25 @@ def resourceOffers(self, driver, offers):
...
                         remainingCores -= job_type.cores
                         remainingMem -= self.__bytesToMB(job_type.memory)
                         remainingStor -= job_type.disk
                     nextToLaunchIndex += 1
-
+                if len(subTasks[job_type]) == 0:
+                    log.debug("Offer %s not large enough to run any tasks. Required: %s Offered: %s"
+                          % (offer.id.value, job_type, (self.__mbToBytes(offerMem),
+                                                        offerCores, self.__mbToBytes(offerStor))))
+                allTasks += subTasks[job_type]
```
hannes-ucsc: Use `list.extend()` instead

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -352,8 +352,9 @@ def resourceOffers(self, driver, offers):
 
         # Right now, gives priority to largest jobs
         for offer in offers:
-            tasks = []
             # TODO: In an offer, can there ever be more than one resource with the same name?
+            allTasks = []
```
hannes-ucsc: Leave name as `tasks`

src/toil/batchSystems/mesos/batchSystem.py
hannes-ucsc: Actually, invert the if and eliminate the continue. Leave the comment, though.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -352,8 +352,9 @@ def resourceOffers(self, driver, offers):
...
+            subTasks = defaultdict(list)
```
hannes-ucsc: Using the term sub-task is confusing here. A sub task is a task that needs to be finished, before its super task can be considered finished. This does not apply here. Rename to `tasksByType`.

hannes-ucsc: In fact I don't see the need for a dictionary here. You loop over the job types once, which means that you set each entry once and in the same iteration you access that entry. You might as well use a simple list that get overwritten in each iteration. Correct me if I am wrong.

hannes-ucsc:
@arkal could you take a look at #670 and see if this intersects with your effort?

arkal:
yes, it does. @	fnothaft mentions specifically which of the 3 requirements fails. I'm just printing all 3 (similar to before)

Do you want me to implement his changes as well when i'm fixing up the changes youp roposed?

hannes-ucsc:
Let's proceed as if that PR didn't exist. Both need work and we can reconcile them when they are both sound.

hannes-ucsc:
... unless you think either one obsoletes the other one, that is.

arkal:
They are both sort-of targeting the same problem hence I thought it would make sense to do them together.

src/toil/batchSystems/mesos/batchSystem.py
arkal: Very true. I used a dict just in case we decided to use the info in the data structure for some logging or other operations in the future

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -362,35 +362,47 @@ def resourceOffers(self, driver, offers):
...
+                        # check to make sure task isn't already running (possibly in very unlikely
+                        # edge case)
+                        tasksByJobType.append(task)
+                        log.info("Preparing to launch Mesos task %s using " % task.task_id.value +
+                                 "offer %s..." % offer.id.value)
                         remainingCores -= job_type.cores
                         remainingMem -= self.__bytesToMB(job_type.memory)
                         remainingStor -= job_type.disk
+                    else:
+                        raise RuntimeError
```
hannes-ucsc: Can we change this to 

```
assert int(task.task_id.value) not in self.runningJobMap
```

arkal: will rebase, squash and add this.

arkal:
@hannes-ucsc can i close this PR?

hannes-ucsc:
No. See my lengthy comment 

https://github.com/BD2KGenomics/toil/pull/670#issuecomment-188552453

fnothaft:
Ah, thanks for pinging this @hannes-ucsc and @arkal. I think I had missed Hannes' comment about combining the two. Let's definitely meet and discuss. Would you like me to jump on the main toil scrum (that'd be somewhat convenient but I'm not sure it's the right venue), or we could just chat on Slack sometime, or would you prefer to arrange a different meeting time? Let me know...

hannes-ucsc:
@fnothaft Are you going to come to UCSC anytime soon? This not high priority.

fnothaft:
> @fnothaft Are you going to come to UCSC anytime soon? This not high priority.

Yeah, I'll come down as soon as things unblock on the ADAM recompute. Hopefully two weeks from now...? ;)

hannes-ucsc:
Continued in #753

Issue #683, opened 2016-02-16
logToMaster() doesn't work


hannes-ucsc:
This is an example from the docs. It does not print the logToMaster() message, neither on master nor on 3.1.x.

```
from toil.job import Job

class HelloWorld(Job):
    def __init__(self, message):
        Job.__init__(self,  memory="2G", cores=2, disk="3G")
        self.message = message

    def run(self, fileStore):
        fileStore.logToMaster("Hello, world!, I have a message: %s"
                              % self.message)

if __name__=="__main__":
    options = Job.Runner.getDefaultOptions("./toilWorkflowRun")
    options.logLevel = "INFO"
    Job.Runner.startToil(HelloWorld("woot"), options)
```

hannes-ucsc:
Needs to be ported to master.

PR #687, opened 2016-02-17
Add cleanWorkDir flag to control retention of temporary files (resolves #686, resolves #690)

### LINKED ISSUE #686: Add option for retaining files in work directory (--workDir)

### LINKED ISSUE #690: A failing root job causes EOFError
```python
options.clean = "always"
```
Where tempFileTestErrorJob is
```python
    raise RuntimeError()
```

            jobStoreFileID = fH.read()

cket:


cket:
First commit is the test only

cket:
The first build passed but should have failed. It failed as expected on my local machine, I'm not sure what is going on

hannes-ucsc:
You pushed both commits. Is that intentional?

src/toil/test/src/retainTempDirTest.py
```diff
@@ -0,0 +1,37 @@
...
+        testDir = self._createTempDir()
+        options = Job.Runner.getDefaultOptions(self._getTestJobStorePath())
+        options.workDir = testDir
+        options.clean = "always"
+        options.cleanWorkDir = "never"
+        A = Job.wrapJobFn(tempFileTestJob)
+        Job.Runner.startToil(A, options)
+        retainedTempData = os.listdir(testDir)
+        assert retainedTempData != []
+
```
hannes-ucsc: Now that you implemented all four cases, shouldn't there be a test for each one of them?

fnothaft:
Also, you _need_ to make https://github.com/BD2KGenomics/toil/pull/687/files#diff-92b64e753eea47c35c71afa81657c763R455 conditional, or remove it. It is pretty br0000000tal to log that a job failed, but delete the log file that contains all of the info about _why_ the job failed. Right now, toil takes a snippet of the end and prints that from the leader, but that's not sufficient if you have log files with huge error messages. E.g., I'm debugging a contig mismatch issue in the GATK right now, and the error message is >4000 lines long.

fnothaft:
Or, actually, if you remove the truncate file call before writing back the log file to the global file store and print a user accessible URL to where the log file is written in the file store, that'd work too. It is much easier to debug if toil does not destructively modify the logs generated by failed jobs.

cket:
I changed the behavior so that the full log file is saved in the work directory and its path is logged and reported back to the user. However, if the new cleanWorkDir flag is not set that work directory can be deleted and your log file will be lost regardless

fnothaft:
I still think that deleting the log file by default is really bad behavior in general. At the very least, it should be saved back to a jobstore or etc.

hannes-ucsc:
The truncation is @benedictpaten's idea and it seems reasonable to me. We can't just upload arbitrarily large log files to the job store. We have to draw a line somewhere, for the sake of time and space. Otherwise people will I complain that their job store is filled up with garbage logs.  50k may be too aggressive. If all else fails, you still have `--cleanWorkDir` as a last resort.

src/toil/leader.py
```diff
@@ -521,7 +521,10 @@ def mainLoop(config, batchSystem, jobStore, rootJobWrapper, jobCache=None):
     with jobStore.readSharedFileStream("rootJobReturnValue") as fH:
         jobStoreFileID = fH.read()
     with jobStore.readFileStream(jobStoreFileID) as fH:
-        rootJobReturnValue = cPickle.load(fH)
+        try:
+            rootJobReturnValue = cPickle.load(fH)
+        except EOFError:
+            raise FailedJobsException(jobStoreFileID, totalFailedJobs)  # root job failed, cannot be unpickled
```
hannes-ucsc: Can we log the original exception including it's stack trace here?

```
log.error('Failed to unpicke root job', exc_info=True)
```

hannes-ucsc:
Just this one thing, otherwise great.

hannes-ucsc:
One more thing: could you please file a PR against master that cherry-picks your fixes for #683 and #686?

fnothaft:
> The truncation is @benedictpaten's idea and it seems reasonable to me. We can't just upload arbitrarily large log files to the job store. We have to draw a line somewhere, for the sake of time and space. Otherwise people will I complain that their job store is filled up with garbage logs. 50k may be too aggressive. If all else fails, you still have --cleanWorkDir as a last resort.

I strenuously disagree. This log truncation thing is the single most frustrating thing I've run across in a batch scheduling system (considering YARN, multiple LSF installs, SGE, etc.). From my perspective, in an analysis pipeline, logs _are_ data. You should be checking logs for errors and warnings at the end of the pipeline.

Anyways, we're talking about a batch system that is meant for processing files that are multiple hundred gigabytes and we're cribbing over uploading log files that are typically <10MB. At the very least, can you expose this truncation setting as a user modifiable flag? As a user of the system, I would disable log truncation 100% of the time.

hannes-ucsc:
The reason the log is saved in the job store is that it can be logged on the master. It would be problematic if the leader had to print a 10MB log from every node in a 1000 node cluster.

I do see your point about logs being data and I am open to discussing a change in Toil that accommodates that use case. I feel like this PR/issue is not the place to discuss that change, though.

fnothaft:
> I do see your point about logs being data and I am open to discussing a change in Toil that accommodates that use case. I feel like this PR/issue is not the place to discuss that change, though.

That's fair. If you don't have someone from your side with bandwidth to implement that change, I would be game to contribute that change after we're done with the ADAM recompute. I'll drop the logging bit on this PR for now, open a new issue, and we can move on.

benedictpaten:
I thought there was a parameter for this? Anyhow, it's trivial to add or
turnoff rotating logging if we want. I've definitely been in situations
where a firestorm of job failures, each with large log files bright a
network to its knees, so by default I'd favor the switch turned on, but
perhaps that's my hpc/network filesystem bias.

Benedict
(from phone, forgive extra typos)

fnothaft:
> I thought there was a parameter for this? Anyhow, it's trivial to add or turnoff rotating logging if we want.

Not to my knowledge, but if I'm just being ignorant, then I'd be glad to be educated...

hannes-ucsc:
Please continue discussion of log truncation in #697. This PR is about retaining work dir.

hannes-ucsc:
First commit message has # in the wrong place.

PR #689, opened 2016-02-18
Add debug logging around link/chmod calls.

fnothaft:
Related to https://github.com/BD2KGenomics/toil-scripts/issues/106.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -579,7 +579,22 @@ class Job(object):
                 fileHandle = open(absLocalFileName, 'r')
                 if os.stat(absLocalFileName).st_uid == os.getuid():
                     #Chmod if permitted to make file read only to try to prevent accidental user modification
-                    os.chmod(absLocalFileName, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+                    try:
+                        os.chmod(absLocalFileName, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+                    except OSError:
+                        success = True
+                        try:
+                            alfStat = os.stat(absLocalFileName)
+                            log.warning("Chmod of %s failed. File is owned by user %s with projection bits %o." % (
+                                    absLocalFileName,
+                                    alfStats.st_uid,
+                                    alfStats.st_mode))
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Hesitant to merge this. Too much code without test exposure. I'm afraid it will just complicate things and lead to more obscure failures. The upcoming `--cleanWorkDir` flag will make diagnosing this situation much easier (#686 and PR #687).

If you truly believe that calling stat() on a file on which chmod() just failed is a good idea, the panic() context manager should be used to preserve the original exception:

https://github.com/BD2KGenomics/bd2k-python-lib/blob/master/src/bd2k/util/exceptions.py#L6

fnothaft:
> Hesitant to merge this. Too much code without test exposure.

Definitely; I will add tests. This is a quick example of what I'd like to have; not a definitive implementation.

> I'm afraid it will just complicate things and lead to more obscure failures.

The current failures are pretty obscure... Which is why it's taken several days to debug them. The code wraps the `stat` calls in `try/except` blocks, but I will gladly look further into the `panic` context manager.

hannes-ucsc:
If an exception is raised in an `except:` block, `sys.exc_info` is overwritten and the argument-less raise statement that intends to reraise the original exception will instead raise the secondary one. However, capturing the original exception with `as e` and reraising it with `raise e` will print the wrong stack trace, that of the `raise e` not the original one. That's what the `panic()` CM fixes.  

Had `--cleanWorkDir` been available, debugging your issue would have taken much less time. In my experience it has never helped being too smart in recovering from or producing diagnostics about exceptions. You know what to log here if chmod fails only for the reasons you saw it fail. There are plenty of other reasons it could fail and in those situation stat() might also fail, therefore obscuring the problem. Lastly, if we merged this we would need to do similar thing many other system calls.

hannes-ucsc:
This adds too much complexity in a rarely executed branch in the control flow. Without tests that provide coverage of that added complexity, I feel like I can't merge this.

hannes-ucsc:
I should also add that I think that this is overkill which is why I closed as opposed to leaving it open in hope of finding someone to add those tests.

PR #692, opened 2016-02-19
Use the env variable TOIL_WORKDIR  (resolves #691)

### LINKED ISSUE #691: Use env variable TOIL_WORKDIR preferentially over temp

arkal:
Hierarchy for selection of a workdir is

config.workDir >> $TOIL_WORKDIR >> $TMPDIR

This will help with lazy setup of the toil-disk flag in mesos in future versions.

Resolves #691


src/toil/worker.py
```diff
@@ -133,7 +133,10 @@ def badWorker():
 
     setLogLevel(config.logLevel)
 
-    tempRootDir = config.workDir
+    # Hierarchy for workdir is config.workDir >> $TOIL_WORKDIR >> $TMPDIR
+    # If config.workDir is None, try setting it to $TOIL_WORKDIR. If that is not set, os.getenv
+    # returns None.
+    tempRootDir = config.workDir if config.workDir else os.getenv('TOIL_WORKDIR')
```
hannes-ucsc: This can be shortened to `config.workDir or os.getenv('TOIL_WORKDIR')`

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Please rebase and squash.

arkal:
@hannes-ucsc If the user hasn't provided workDir and  TOIL_WORKDIR is not set, then the fallback is TMPDIR. We were talking about eliminating fallbacks yesterday. How do you want to proceed? I think fallbacks are sort-of necessary here.

arkal:
@hannes-ucsc I have this in 2 separate commits so we can roll back to the earlier one if you prefer. I can rebase to a single commit if this is ok.

hannes-ucsc:
Jenkins, test this please.

src/toil/worker.py
```diff
@@ -133,9 +134,7 @@ def badWorker():
 
     setLogLevel(config.logLevel)
 
-    tempRootDir = config.workDir
-    if tempRootDir is not None and not os.path.exists(tempRootDir):
-        raise RuntimeError("The temporary directory specified by workDir: %s does not exist" % tempRootDir)
+    tempRootDir = getToilWorkDir(config.workDir)
```
hannes-ucsc: Pass entire config. Extracting a specific attribute is an implementation detail of `getToilWorkDir()`

src/toil/common.py
```diff
@@ -524,3 +525,33 @@ def parseSetEnv(l):
...
+    This is particularly useful if the user hasn't provided a value for --workDir (passed to this
+    function as configWorkDir) and the TOIL_WORKDIR environment variable is unset.  In this case,
+    the value returned points to a directory 'toil' located somewhere on the system in a location
+    reserved for temporary files (obtained from the tempfile module).  The variable is created if
+    required.
+    Hierarchy for workdir is config.workDir >> $TOIL_WORKDIR >> $TMPDIR
+    :param configWorkDir: Value passed to the program using the workDir flag
+    :return: Path to the work dir
+    """
+    configWorkDir = configWorkDir or os.getenv('TOIL_WORKDIR')
```
hannes-ucsc: rename to workDir

src/toil/common.py
```diff
@@ -524,3 +525,33 @@ def parseSetEnv(l):
...
+    if configWorkDir is not None and not os.path.exists(configWorkDir):
```
hannes-ucsc: Move to end of function, remove first term in condition.

src/toil/common.py
```diff
@@ -524,3 +525,33 @@ def parseSetEnv(l):
...
+        raise RuntimeError("The temporary directory specified by workDir: %s does not "
+                           "exist" % tempRootDir)
+    elif configWorkDir is None:
+        #  Neither --workDir not TOIL_WORKDIR is set
+        configWorkDir = os.path.join(tempfile.gettempdir(), 'toil')
```
hannes-ucsc: This would give every run of every user script the same working directory. Clearly not what we want, or?

hannes-ucsc:
Don't keep obsolete commits in a PR. If you want to hold on to a likely-to-be obsolete commit for backup purposes, just have a local branch point to it.

src/toil/common.py
arkal: The way it works now, worker temp directories will just get dumped into the directory pointed to by $TMPDIR (or other temp identifier). This way, it gets dumped instead into a directory called toil that may, or may not exist. We want to be able to do a folder walk on the toil workdir and if it is exactly $TMPDIR, we are going to have a lot of files messing with our estimations. This reduces that aspect to some extent.

What do you think of 'toil' + workFlowID ?

src/toil/worker.py
arkal: This function will be used by multiple things. One of them will be executor.py that knows only config.workDIR, but doesn't know the rest of config.

src/toil/common.py
hannes-ucsc: That might work.

Then the question is if the the workflowID should also be appended to an explicitly configured --workDir.

arkal: yes, it's worthy of discussion. A user *can* really specify a non-empty directory as the workdir.

hannes-ucsc:
Discussed with @arkal. Decided to implement TOIL_WORKDIR and --workDir and TMPDIR all the same in that they point to a parent dir (the work dir) that contain per-workflow subdirs (the workflow dirs).

arkal:
Jenkins, test this please.

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
...
+    the value returned points to a directory 'toil' located somewhere on the system in a location
+    reserved for temporary files (obtained from the tempfile module).  The variable is created if
+    required.
+    Hierarchy for workdir is config.workDir >> $TOIL_WORKDIR >> $TMPDIR
+    :param workflowID: Unique identifier for the workflow
+    :param workDir: Value passed to the program using the workDir flag
+    :return: Path to the work dir
+    """
+    workDir = workDir or os.getenv('TOIL_WORKDIR')
+    if workDir is None:
```
hannes-ucsc: Looks like this doesn't do what we agreed on: https://github.com/BD2KGenomics/toil/pull/692#issuecomment-193533704

We want the workflow dir to always be created, right?

arkal: whoops. Fixed that

arkal:
@hannes-ucsc this is ready for review

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
...
+    (obtained from the tempfile module).  The workflow directory is created one level below workDir
+    and is named toil-<workFlowID>.
+    Hierarchy for workdir is config.workDir >> $TOIL_WORKDIR >> $TMPDIR
+    :param workflowID: Unique identifier for the workflow
+    :param configWorkDir: Value passed to the program using the --workDir flag
+    :return: Path to the workflow directory
+    """
+    workDir = configWorkDir or os.getenv('TOIL_WORKDIR') or tempfile.gettempdir()
+    if not os.path.exists(workDir):
+        # This will never happen if tempfile.tempdir() was used (which hapens only if the other 2
```
hannes-ucsc: remove this comment

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
...
+        # were not set)
+        raise RuntimeError("The directory specified by workDir/TOIL_TEMPDIR: %s " % workDir +
```
hannes-ucsc: --workDir or TOIL_WORKDIR

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
...
+    workflowDir = os.path.join(workDir, 'toil-%s' % workflowID)
+    try:
+        # Directory creation is atomic
+        os.mkdir(workflowDir)
+    except OSError as err:
+        if err.errno != 17:
+            # The directory exists if a previous worker set it up.
+            raise
+    else:
+        logger.info('Created the workFlow directory at %s' % workflowDir)
```
hannes-ucsc: workflow

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
             raise ValueError('Empty name')
         d[k] = v
     return d
+
+
+def getToilWorkflowDir(workflowID, configWorkDir=None):
+    """
+    Returns a path to the directory where worker directories and the cache will be located for this
+    workflow.  This is particularly useful if the user hasn't provided a value for --workDir (passed
```
hannes-ucsc: cut everything after the first sentence. A docstring should describe what a function does, not how.

src/toil/common.py
```diff
@@ -524,3 +530,36 @@ def parseSetEnv(l):
...
+    to this function as workDir) and the TOIL_WORKDIR environment variable is unset.  In this case,
+    the value used for workDir is somewhere on the system in a location reserved for temporary files
+    (obtained from the tempfile module).  The workflow directory is created one level below workDir
+    and is named toil-<workFlowID>.
+    Hierarchy for workdir is config.workDir >> $TOIL_WORKDIR >> $TMPDIR
+    :param workflowID: Unique identifier for the workflow
```
hannes-ucsc: Add type annotations, e.g `:param str workflowID`

arkal:
Thanks @hannes-ucsc, I've made the changes.

hannes-ucsc:
Jenkins, test this please.

src/toil/common.py
```diff
@@ -524,3 +530,30 @@ def parseSetEnv(l):
             raise ValueError('Empty name')
         d[k] = v
     return d
+
+
+def getToilWorkflowDir(workflowID, configWorkDir=None):
+    """
+    Returns a path to the directory where worker directories and the cache will be located for this
+    workflow.
+    :param str workflowID: Unique identifier for the workflow
```
hannes-ucsc: insert empty line here

src/toil/common.py
```diff
@@ -524,3 +530,30 @@ def parseSetEnv(l):
...
+    :param str configWorkDir: Value passed to the program using the --workDir flag
+    :return: Path to the workflow directory
+    :rtype: str
+    """
+    workDir = configWorkDir or os.getenv('TOIL_WORKDIR') or tempfile.gettempdir()
+    if not os.path.exists(workDir):
+        raise RuntimeError("The directory specified by --workDir or TOIL_WORKDIR: %s " % workDir +
```
hannes-ucsc: use either string concatenation or interpolation. Not both.

hannes-ucsc: Also, a grammar purist would want the message to be `The directory specified by --workDir or TOIL_WORKDIR (%s) does not exist`. It's good to have delimiters around an interpolated string so as to better identify empty strings.

src/toil/common.py
```diff
@@ -524,3 +530,30 @@ def parseSetEnv(l):
...
+                           "does not exist.")
+    # create the workflow dir
```
hannes-ucsc: # Create

src/toil/jobStores/abstractJobStore.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, config=None):
             config.workflowID = str(uuid4())
             self.__config = config
             self.writeConfigToStore()
+        logger.info('The workflow ID is: %s' %self.__config.workflowID)
```
hannes-ucsc: Should this be DEBUG level? And again `"The workflow ID is '%s'."`

hannes-ucsc: And space between % and self.

src/toil/worker.py
```diff
@@ -133,9 +134,7 @@ def badWorker():
 
     setLogLevel(config.logLevel)
 
-    tempRootDir = config.workDir
-    if tempRootDir is not None and not os.path.exists(tempRootDir):
-        raise RuntimeError("The temporary directory specified by workDir: %s does not exist" % tempRootDir)
+    tempRootDir = getToilWorkflowDir(config.workflowID, config.workDir)
```
hannes-ucsc: Should we rename tempRootDir or would that be too invasive, i.e. affect many lines in the code?

src/toil/jobStores/abstractJobStore.py
arkal: No i believe this should be INFO such that the user is always told what the workflow ID is. They will need it to access their data in the `toil-<workflowID>` folder

src/toil/worker.py
arkal: Just the next statement. Renaming to toilWorkflowDir

src/toil/jobStores/abstractJobStore.py
arkal: Unless you feel the only reason one would want to know the workflow dir is in a debugging situation

arkal:
@hannes-ucsc this is ready for review

arkal:
Also, I checked. The job store is loaded only once when toil is being setup on the leader. Thus the "The workflow ID is: XYZ" message is only printed once. I say we just keep this in INFO

hannes-ucsc:
Then you didn't check thoroughly enough:

https://github.com/arkal/toil/blob/6901c6ec7af964f4d005a6389e3ce672fac1393f/src/toil/worker.py#L102

arkal:
Correction* The job store is loaded only once with config!=None when toil is being setup on the leader. So the workflow ID is created only once.

I moved the logging into the if block so it prints only once.

arkal:
This also means it won't be printed on a restart, though

src/toil/jobStores/abstractJobStore.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, config=None):
         else:
             assert config.workflowID is None
             config.workflowID = str(uuid4())
+            logger.info('The workflow ID is: \'%s\'' % config.workflowID)
```
hannes-ucsc: `logger.info("The workflow ID is: '%s'" % config.workflowID)`

arkal:
Do you prefer double quotes for this codebase? I default to singles but I can try and stick to doubles from now if you'd like.

hannes-ucsc:
See bullet 3 in https://github.com/BD2KGenomics/toil/wiki/Python-Dos-and-Don'ts

My rule of thumb is to use the least noisy syntax. Backslashes are noisier that double quotes. Double quotes are noisier than single quotes.

arkal:
Got it!

PR #699, opened 2016-02-24
Issues/3.1.x/686 clean work dir 690 improper exception (resolves #686, resolves #690)

### LINKED ISSUE #686: Add option for retaining files in work directory (--workDir)

### LINKED ISSUE #690: A failing root job causes EOFError

cket:
Resolves #686, #690

Continued from https://github.com/BD2KGenomics/toil/pull/687, I initially selected the wrong target branch. The tests in this PR are cleaner but the functionality is the same.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -44,6 +44,7 @@ class Config(object):
         # Because the stats option needs the jobStore to persist past the end of the run,
         # the clean default value depends the specified stats option and is determined in setOptions
         self.clean = None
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
Beware: In violation of the protocol, I rebased, resolved the conflicts and force pushed.

PR #705, opened 2016-03-02
Use ephemeral for running Toil tests on Jenkins (resolves #704)

### LINKED ISSUE #704: Make jenkins toil tests run on /mnt/ephemeral
Set this such that future tests that may require larger disk space have a bigger sandbox to work in.

arkal:
resolves #704

The test framework uses the system TMPDIR to create temp working directories for running toil tests. If we
set it to a directory on /mnt/ephemeral while running tests on jenkins, then the tests can run in a larger volume
allowing for tests to use more space if need be.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -10,4 +10,5 @@ pip2.7 install sphinx
 make develop extras=[aws,mesos,azure,encryption,cwl]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
+mkdir /mnt/ephemeral/tmp && export TMPDIR=/mnt/ephemeral/temp
 make $make_targets


jenkins.sh
```diff
@@ -10,4 +10,5 @@ pip2.7 install sphinx
 make develop extras=[aws,mesos,azure,encryption,cwl]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
+mkdir -p /mnt/ephemeral/temp && export TMPDIR="/mnt/ephemeral/temp"
```
hannes-ucsc: s/temp/tmp/g

hannes-ucsc: And no need for quotes in the assignment

jenkins.sh
```diff
@@ -10,4 +10,5 @@ pip2.7 install sphinx
 make develop extras=[aws,mesos,azure,encryption,cwl]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
+mkdir -p /mnt/ephemeral/tmp && export TMPDIR=/mnt/ephemeral/temp
```
hannes-ucsc: Sorry, changed my mind about `-p`. Lets remove it. Can you guess why?

hannes-ucsc:
Almost there ...

jenkins.sh
arkal: Haha we're assuming that there's a mountpoint called /mnt/ephemeral. If there isn't, the directory will still be created but it'll still be on the root volume.

hannes-ucsc:
Will merge once tests pass.

hannes-ucsc:
And remember to take the `needs work` label off when you're done.

PR #708, opened 2016-03-07
Fix services (resolves #706, resolves #703, resolves #646, resolves #644, resolves #557)

### LINKED ISSUE #557: Service jobs cause excessive polling on file store
The polling loop in ServiceJob.run() polls for a file in the job store once per second and service job. I feel that this is too frequent. We should consider switching to a capped, exponential back- strategy for the polling interval. It wouldn't be too bad for a 1 hour job to be artificially extended by 30 seconds on average if the polling interval had increased to one minute.

### LINKED ISSUE #644: Services are not stopped on Toil script failure

### LINKED ISSUE #646: Service class start method should have Job.FileStore argument
To allow services to have temporary files in a consistent manner we should add Job.FileStore instance argument to Job.Service.start.

### LINKED ISSUE #703: Service jobs have no way of checking on the processes they represent
_From @hannes-ucsc on March 1, 2016 21:44_

We should add a check() method to the Service class signature that is repeatedly called by the loop that waits for the latch file in the job store. The check method should raise an exception if the process is dead/unusable.

_Copied from original issue: BD2KGenomics/toil-scripts#127_

### LINKED ISSUE #706: Link fate of jobs in a service subcluster
Depends on #703

Related to #644

1. Assume there is a Toil cluster with one or more Spark/HDFS subclusters. If one of the Spark/HDFS master processes fails, the service job representing it should also fail (#703).

2. When that service job fails, the service jobs that represent the Spark/HDFS slaves connected to that Spark/HDFS master should fail. As it is now I think the slave processes and their representative services will just keep running (this issue).

3. When the master service job fails, the driver job using it will fail.

4. On a subcluster with replication factor R, if the right combination of R Spark/HDFS workers die, a block of HDFS data is lost. This leads to driver failure but it may take a while before that situation is discovered.

5. If the driver job fails, the entire cluster should be taken down (#644).

benedictpaten:


src/toil/job.py
```diff
@@ -1480,14 +1499,14 @@ def run(self, fileStore):
...
+        if fileStore.jobStore.fileExists(self.startJobStoreID):
+            fileStore.jobStore.deleteFile(self.startJobStoreID)
+        assert not fileStore.jobStore.fileExists(self.startJobStoreID)
         #Now block until we are told to stop, which is indicated by the removal
         #of a file
-        assert self.stopFileStoreID != None
-        while fileStore.jobStore.fileExists(self.stopFileStoreID):
+        assert self.terminateJobStoreID != None
+        while fileStore.jobStore.fileExists(self.terminateJobStoreID):
             time.sleep(1) #Avoid excessive polling
```
hannes-ucsc: This should use exponential back-off, see #557. We will hit the request rate limit if we are not careful!!!

src/toil/leader.py
```diff
@@ -322,6 +352,126 @@ def __init__( self, jobStoreString, numberOfFailedJobs ):
...
+                for serviceJobStoreID, memory, cores, disk, startJobStoreID, terminateJobStoreID in serviceJobList:                
+                    assert jobStore.exists(startJobStoreID)
+                    assert jobStore.exists(terminateJobStoreID)
+                    serviceJobsToStart.put((serviceJobStoreID, memory, cores, disk))
+                
+                # Wait until all the services of the batch are running
+                for serviceTuple in serviceJobList:
+                    while jobStore.exists(serviceTuple[4]):
+                        # Sleep to avoid thrashing
+                        time.sleep(1.0)
```
hannes-ucsc: This should use exponential back-off, see #557. We will hit the request rate limit if we are not careful!!!

hannes-ucsc:
What issues exactly does this address. For a list of candidates:

https://waffle.io/BD2KGenomics/toil?search=service

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Unfortunately, the word "resolves" has to appear before every ticket number in the title. I fixed that just now.

benedictpaten:
Thanks, I knew that, sorry.

Benedict
(from phone, forgive extra typos)

benedictpaten:
@hannes-ucsc  can you review and or merge?

src/toil/batchSystems/singleMachine.py
```diff
@@ -87,6 +87,8 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
         """
         :type list[Thread]
         """
+        # An event used to tell the workers to quit
+        self.terminate = Event()
```
hannes-ucsc: https://github.com/BD2KGenomics/toil/wiki/Python-Dos-and-Don'ts#dont-use-threadingevent-to-signal-boolean-conditions-between-threads

benedictpaten: Not sure I agree with this - as if we switch the thread out for a process with multiprocessing then the boolean doesn't work, whereas there exist process safe events. However, I just backed out that addition.

hannes-ucsc:
Realistically, I won't have time to review this in-depth but I don't think a full review is necessary considering that the author of the PR is the author of Toil ;-)

I think this should be squashed because the commit history looks a bit messy.

Other than that, feel free to merge this at your discretion, @benedictpaten.

benedictpaten:
Okay, squashed commits, will merge once I see tests all passed.

PR #709, opened 2016-03-08
Add import/export file methods to AbstractJobStore (resolves #645, resolves #618, resolves #660)

### LINKED ISSUE #618: Add jobStore.importFile() and .exportFile()

### LINKED ISSUE #645: Write S3 multipart copy function

### LINKED ISSUE #660: Implement jobStore.importFile() and .exportFile() for Azure
The methods are exposed to user scripts in Job.FileStores. Specifications for the methods are described in #618

JakeNarkizian:
Every concrete jobStore now supports importing and exporting files to other jobStores. These methods are exposed to the toil user in job.py. There is also now a standalone multipart copy function for S3, and a dynamic test method generator function in toil test.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -725,6 +725,12 @@ class Job(object):
                 #This will result in the files removal from the cache at the end of the current job
                 self._jobStoreFileIDToCacheLocation.pop(fileStoreID)
 
+        def importFile(self, srcUrl):
+            return self.jobStore.importFile(srcUrl)
+
+        def exportFile(self, jobStoreFileID, dstUrl):
+            self.jobStore.exportFile(jobStoreFileID, dstUrl)
+
         def logToMaster(self, text, level=logging.INFO):
             """
             Send a logging message to the leader. The message will also be \
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -13,12 +13,16 @@
 # limitations under the License.
 from __future__ import absolute_import
 
+import urlparse
+
 import re
 from abc import ABCMeta, abstractmethod
 from contextlib import contextmanager
 from datetime import timedelta
 from uuid import uuid4
 from toil.job import JobException
+from bd2k.util import memoize
+from bd2k.util.objects import abstractstaticmethod, abstractclassmethod
 
 try:
     import cPickle
@@ -53,6 +57,46 @@ class NoSuchFileException(Exception):
 class JobStoreCreationException(Exception):
     def __init__(self, message):
         super(JobStoreCreationException, self).__init__(message)
+
+
+@memoize
+def getJobStoreClasses():
+    """
+    Compiles as list of the classes of all jobStores whose dependencies
+    are installed.
+
+    Note that job store names must be manually added.
+    """
+    jobStoreClassNames = (
+        "toil.jobStores.azureJobStore.AzureJobStore",
+        "toil.jobStores.fileJobStore.FileJobStore",
+        "toil.jobStores.aws.jobStore.AWSJobStore"
+    )
+
+    jobStoreClasses = []
+    for className in jobStoreClassNames:
+        moduleName, className = className.rsplit('.', 1)
+        from importlib import import_module
+        try:
+            module = import_module(moduleName)
+        except ImportError:
+            logger.info("Unable to import '%s'" % moduleName)
+        else:
+            jobStoreClasses.append(getattr(module, className))
+    return jobStoreClasses
+
+
+def findJobStoreForUrl(url):
+    """
+    Returns the AbstractJobStore subclass that supports the given URL.
+
+    :param urlparse.ParseResult url: The given URL
+    :rtype: toil.jobStore.AbstractJobStore
+    """
+    for cls in getJobStoreClasses():
+        if cls._supportsUrl(url):
+            return cls
+    raise RuntimeError("No existing job store supports the URL '%s'" % url)
 
 
 class AbstractJobStore(object):
@@ -158,10 +202,118 @@ class AbstractJobStore(object):
             raise JobStoreCreationException("The job store '%s' does not exist, so there "
                                             "is nothing to restart." % jobStoreString)
 
+    def importFile(self, srcUrl):
+        """
+        Imports the file pointed at by sourceUrl into job store. The jobStoreFileId of the new
+        file is returned.
+
+        Note that the helper method _importFile is used to read from the source and write to
+        destination (which is the current job store in this case). To implement any optimizations that
+        circumvent this, the _importFile method should be overridden by subclasses of AbstractJobStore.
+
+        Currently supported schemes are:
+            - 's3' for objects in Amazon S3
+                e.g. s3://bucket/key
+
+            - 'wasb' for blobs in Azure Blob Storage
+                e.g. wasb://container/blob
+
+            - 'file' for local files
+                e.g. file://local/file/path
+
+        :param str srcUrl: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :return The jobStoreFileId of imported file.
+        :rtype: str
+        """
+        url = urlparse.urlparse(srcUrl)
+        return self._importFile(findJobStoreForUrl(url), url)
+
+    def _importFile(self, otherCls, url):
+        """
+        Refer to importFile docstring for information about this method.
+
+        :param type otherCls: The concrete subclass of AbstractJobStore that supports exporting to the given URL.
+        :param urlparse.ParseResult url: The parsed url given to importFile.
+        :return: The job store file ID of the imported file
+        :rtype: str
+        """
+        with self.writeFileStream() as (writable, jobStoreFileID):
+            otherCls._readFromUrl(url, writable)
+            return jobStoreFileID
+
+    def exportFile(self, jobStoreFileID, dstUrl):
+        """
+        Exports file to destination pointed at by the destination URL.
+
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        Note that the helper method _exportFile is used to read from the source and write to
+        destination. To implement any optimizations that circumvent this, the _exportFile method
+        should be overridden by subclasses of AbstractJobStore.
+
+        :param str jobStoreFileID: The id of the file in the job store that should be exported.
+        :param str dstUrl: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        """
+        url = urlparse.urlparse(dstUrl)
+        return self._exportFile(findJobStoreForUrl(url), jobStoreFileID, url)
+
+    def _exportFile(self, otherCls, jobStoreFileID, url):
+        """
+        Refer to exportFile docstring for information about this method.
+
+        :param type otherCls: The concrete subclass of AbstractJobStore that supports exporting to the given URL.
+        :param str jobStoreFileID: The id of the file that will be exported.
+        :param urlparse.ParseResult url: The parsed url given to importFile.
+        """
+        with self.readFileStream(jobStoreFileID) as readable:
+            otherCls._writeToUrl(readable, url)
+
+    @abstractclassmethod
+    def _readFromUrl(cls, url, writable):
+        """
+        Reads the contents of the object at the specified location and writes it to the given
+        writable stream.
+
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        :param str url: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :param writable: a writable stream
+        """
+        raise NotImplementedError()
+
+    @abstractclassmethod
+    def _writeToUrl(cls, readable, url):
+        """
+        Reads the contents of the given readable stream and writes it to the object at the
+        specified location.
+
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        :param str url: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :param readable: a readable stream
+        """
+        raise NotImplementedError()
+
+    @abstractclassmethod
+    def _supportsUrl(cls, url):
+        """
+        Returns True if the job store supports the URL's scheme.
+
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        :param urlparse.ParseResult url: a parsed URL that may be supported
+        :return bool: returns true if the cls supports the URL
+        """
+        raise NotImplementedError()
+
     @abstractmethod
     def deleteJobStore(self):
         """
-        Removes the jobStore from the disk/store. Careful!
+        Removes the job store from the disk/store. Careful!
         """
         raise NotImplementedError()
 
@@ -435,7 +587,7 @@ class AbstractJobStore(object):
         
         jobStoreID is the id of a job, or None. If specified, when delete(job) 
         is called all files written with the given job.jobStoreID will be 
-        removed from the jobStore.
+        removed from the job store.
         """
         raise NotImplementedError()
 
@@ -457,7 +609,7 @@ class AbstractJobStore(object):
         
         jobStoreID is the id of a job, or None. If specified, when delete(job) 
         is called all files written with the given job.jobStoreID will be 
-        removed from the jobStore.
+        removed from the job store.
         
         Call to fileExists(getEmptyFileStoreID(jobStoreID)) will return True.
         """
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+
 from StringIO import StringIO
 from contextlib import contextmanager
 import logging
@@ -47,6 +48,48 @@ import toil.lib.encryption as encryption
 
 log = logging.getLogger(__name__)
 
+defaultPartSize = 2 ** 20 * 50
+
+
+def copyKeyMultipart(srcKey, dstBucketName, dstKeyName, headers=None):
+    """
+    Copies a key from a source key to a destination key in multiple parts. Note that if the
+    destination key exists it will be overwritten implicitly, and if it does not exist a new
+    key will be created.
+
+    :param boto.s3.key.Key srcKey: The source key to be copied from.
+    :param str dstBucketName: The name of the destination bucket for the copy.
+    :param str dstKeyName: The name of the destination key that will be created or overwritten.
+    :param dict headers: Any headers that should be passed.
+
+    :rtype: boto.s3.multipart.CompletedMultiPartUpload
+    :return: An object representing the completed upload.
+    """
+    partSize = defaultPartSize
+    s3 = boto.connect_s3()
+    headers = headers or {}
+    totalSize = srcKey.size
+
+    # initiate copy
+    upload = s3.get_bucket(dstBucketName).initiate_multipart_upload(dstKeyName)
+    try:
+        start = 0
+        partIndex = itertools.count()
+        while start < totalSize:
+            end = min(start + partSize, totalSize)
+            upload.copy_part_from_key(src_bucket_name=srcKey.bucket.name,
+                                      src_key_name=srcKey.name,
+                                      part_num=next(partIndex)+1,
+                                      start=start,
+                                      end=end-1,
+                                      headers=headers)
+            start += partSize
+    except:
+        upload.cancel_upload()
+        raise
+    else:
+        return upload.complete_upload()
+
 
 class AWSJobStore(AbstractJobStore):
     """
@@ -58,7 +101,7 @@ class AWSJobStore(AbstractJobStore):
     """
 
     @classmethod
-    def createJobStore(cls, jobStoreString, config=None):
+    def loadOrCreateJobStore(cls, jobStoreString, config=None, **kwargs):
         region, namePrefix = jobStoreString.split(':')
         if not cls.bucketNameRe.match(namePrefix):
             raise ValueError("Invalid name prefix '%s'. Name prefixes must contain only digits, "
@@ -71,7 +114,7 @@ class AWSJobStore(AbstractJobStore):
         if '--' in namePrefix:
             raise ValueError("Invalid name prefix '%s'. Name prefixes may not contain "
                              "%s." % (namePrefix, cls.nameSeparator))
-        return cls(region, namePrefix, config=config)
+        return cls(region, namePrefix, config=config, **kwargs)
 
     # Dots in bucket names should be avoided because bucket names are used in HTTPS bucket
     # URLs where the may interfere with the certificate common name. We use a double
@@ -88,7 +131,7 @@ class AWSJobStore(AbstractJobStore):
 
     # Do not invoke the constructor, use the factory method above.
 
-    def __init__(self, region, namePrefix, config=None):
+    def __init__(self, region, namePrefix, config=None, partSize=defaultPartSize):
         """
         Create a new job store in AWS or load an existing one from there.
 
@@ -108,6 +151,7 @@ class AWSJobStore(AbstractJobStore):
         self.filesBucket = None
         self.db = self._connectSimpleDB()
         self.s3 = self._connectS3()
+        self.partSize = partSize
 
         # Check global registry domain for existence of this job store. The first time this is
         # being executed in an AWS account, the registry domain will be created on the fly.
@@ -227,6 +271,48 @@ class AWSJobStore(AbstractJobStore):
         info.save()
         log.debug("Created %r.", info)
         return info.fileID
+
+    def _importFile(self, otherCls, url):
+        if issubclass(otherCls, AWSJobStore):
+            srcBucket, srcKey = self._extractKeyInfoFromUrl(url)
+            info = self.FileInfo.create(srcKey.name)
+            info.copyFrom(srcKey)
+            info.save()
+            return info.fileID
+        else:
+            return super(AWSJobStore, self)._importFile(otherCls, url)
+
+    def _exportFile(self, otherCls, jobStoreFileID, url):
+        if issubclass(otherCls, AWSJobStore):
+            dstBucket, newKey = self._extractKeyInfoFromUrl(url)
+            info = self.FileInfo.loadOrFail(jobStoreFileID)
+            info.copyTo(newKey.name, dstBucket)
+        else:
+            super(AWSJobStore, self)._exportFile(otherCls, jobStoreFileID, url)
+
+    @classmethod
+    def _readFromUrl(cls, url, writable):
+        srcBucket, srcKey = cls._extractKeyInfoFromUrl(url)
+        srcKey.get_contents_to_file(writable)
+
+    @classmethod
+    def _writeToUrl(cls, readable, url):
+        dstBucket, dstKey = cls._extractKeyInfoFromUrl(url)
+        dstKey.set_contents_from_string(readable.read())
+
+    @staticmethod
+    def _extractKeyInfoFromUrl(url):
+        """
+        :return: (bucket, key)
+        """
+        s3 = boto.connect_s3()
+        bucket = s3.get_bucket(url.netloc)
+        key = bucket.new_key(url.path[1:])
+        return bucket, key
+
+    @classmethod
+    def _supportsUrl(cls, url):
+        return url.scheme.lower() == 's3'
 
     def writeFile(self, localFilePath, jobStoreID=None):
         info = self.FileInfo.create(jobStoreID)
@@ -633,8 +719,6 @@ class AWSJobStore(AbstractJobStore):
                 else:
                     raise
 
-        s3PartSize = 50 * 1024 * 1024
-
         def upload(self, localFilePath):
             file_size, file_time = self._fileSizeAndTime(localFilePath)
             if file_size <= self._maxInlinedSize():
@@ -642,7 +726,7 @@ class AWSJobStore(AbstractJobStore):
                     self.content = f.read()
             else:
                 headers = self._s3EncryptionHeaders()
-                if file_size <= self.s3PartSize:
+                if file_size <= self.outer.partSize:
                     key = self.outer.filesBucket.new_key(key_name=self.fileID)
                     key.name = self.fileID
                     key.set_contents_from_filename(localFilePath, headers=headers)
@@ -656,7 +740,7 @@ class AWSJobStore(AbstractJobStore):
                             start = 0
                             part_num = itertools.count()
                             while start < file_size:
-                                end = min(start + self.s3PartSize, file_size)
+                                end = min(start + self.outer.partSize, file_size)
                                 assert f.tell() == start
                                 upload.upload_part_from_file(fp=f,
                                                              part_num=next(part_num) + 1,
@@ -683,7 +767,7 @@ class AWSJobStore(AbstractJobStore):
             with os.fdopen(readable_fh, 'r') as readable:
                 with os.fdopen(writable_fh, 'w') as writable:
                     def multipartReader():
-                        buf = readable.read(self.s3PartSize)
+                        buf = readable.read(self.outer.partSize)
                         if allowInlining and len(buf) <= self._maxInlinedSize():
                             self.content = buf
                         else:
@@ -700,7 +784,7 @@ class AWSJobStore(AbstractJobStore):
                                                                  part_num=part_num + 1,
                                                                  headers=headers)
                                     if len(buf) == 0: break
-                                    buf = readable.read(self.s3PartSize)
+                                    buf = readable.read(self.outer.partSize)
                             except:
                                 upload.cancel_upload()
                                 raise
@@ -725,6 +809,68 @@ class AWSJobStore(AbstractJobStore):
                 # thread to finish.
                 thread.join()
                 assert bool(self.version) == (self.content is None)
+
+        def copyFrom(self, srcKey):
+            """
+            Copies contents of source key into file.
+
+            :param srcKey: The key that will be copied from
+            """
+            size = srcKey.bucket.get_key(srcKey.name).size
+            if size <= self._maxInlinedSize():
+                self.content = srcKey.get_contents_as_string()
+            else:
+                self.version = self._copyKey(srcKey=srcKey,
+                                             dstBucketName=self.outer.filesBucket.name,
+                                             dstKeyName=self._fileID,
+                                             headers=self._s3EncryptionHeaders()).version_id
+
+        def copyTo(self, dstKeyName, dstBucket):
+            """
+            Copies contents of file to destination key.
+
+            :param str newKeyName: The name of the new key to be created or copied to from the file.
+            :param Bucket dstBucket: The bucket that contains or will contain the destination key.
+            """
+            dstKey = dstBucket.new_key(dstKeyName)
+
+            if self.content is not None:
+                dstKey.set_contents_from_string(self.content)
+
+            elif self.version:
+                srcKey = self.outer.filesBucket.get_key(self.fileID, headers=self._s3EncryptionHeaders())
+                self._copyKey(srcKey=srcKey,
+                              dstBucketName=dstBucket.name,
+                              dstKeyName=dstKeyName,
+                              headers={"amz-copy-source".join(k.split('amz')): v
+                                       for k, v in self._s3EncryptionHeaders().items()})
+            else:
+                assert False
+
+        def _copyKey(self, srcKey, dstBucketName, dstKeyName, headers=None):
+            headers = headers or {}
+            s3 = boto.connect_s3()
+
+            if srcKey.size > self.outer.partSize:
+                return copyKeyMultipart(srcKey=srcKey,
+                                        dstBucketName=dstBucketName,
+                                        dstKeyName=dstKeyName,
+                                        headers=headers)
+            else:
+                dstBucket = s3.get_bucket(dstBucketName)
+                return dstBucket.copy_key(new_key_name=dstKeyName,
+                                          src_bucket_name=srcKey.bucket.name,
+                                          src_key_name=srcKey.name,
+                                          metadata=srcKey.metadata,
+                                          headers=headers)
+
+        def _hashesAreEqual(self, srcHash, dstHash):
+            if self.encrypted:
+                log.debug("MD5 hash of imported or exported files cannot be checked because it is encrypted."
+                          % self._fileID)
+                return True
+            else:
+                return dstHash == srcHash
 
         def download(self, localFilePath):
             if self.content is not None:
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import urlparse
+import re
 import os
 import uuid
 import logging
@@ -29,7 +31,7 @@ from azure import WindowsAzureMissingResourceError, WindowsAzureError
 
 from azure.storage import (TableService, BlobService, SharedAccessPolicy, AccessPolicy,
                            BlobSharedAccessPermissions, EntityProperty)
-from bd2k.util import strict_bool
+from bd2k.util import strict_bool, memoize
 
 from bd2k.util.threading import ExceptionalThread
 
@@ -73,6 +75,7 @@ def _fetchAzureAccountKey(accountName):
 
 maxAzureTablePropertySize = 64 * 1024
 
+
 class AzureJobStore(AbstractJobStore):
     """
     A job store that uses Azure's blob store for file storage and
@@ -83,7 +86,7 @@ class AzureJobStore(AbstractJobStore):
         self.keyPath = None
 
         self.account_key = _fetchAzureAccountKey(accountName)
-
+        self.accountName = accountName
         # Table names have strict requirements in Azure
         self.namePrefix = self._sanitizeTableName(namePrefix)
         logger.debug("Creating job store with name prefix '%s'" % self.namePrefix)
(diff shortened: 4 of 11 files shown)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Title was wrong. I fixed it. Tests are failing.

JakeNarkizian:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

JakeNarkizian:
All the test failures were due to too many buckets on s3.

JakeNarkizian:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Previous test runs failed for reasons unrelated to these changes (ran out of S3 buckets). The most recent run fails with a utf8 encoding error. Seems like something is trying to utf8-encode a string that's already utf8-encoded.

tetron:
#711 is blocked on this

hannes-ucsc:
@tetron, despite the test failures I think this is in pretty good shape. Good enough for you to base work on it, I think. The signatures of the added AbstractJobStore methods won't change.

JakeNarkizian:
@hannes-ucsc this is ready for review

src/toil/jobStores/abstractJobStore.py
```diff
@@ -54,6 +56,48 @@ def __init__(self, message):
...
+        else:
+            jobStoreClasses.append(getattr(module, className))
+    return jobStoreClasses
+
+
+@memoize
+def findJobStoreForUrl(url):
+    """
+    This method returns the class of the jobStore which supports exporting to and from the given
+    url.
```
hannes-ucsc: s/url/URL globally, in all strings and comments.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
             raise JobStoreCreationException("The job store '%s' does not exist, so there "
                                             "is nothing to restart." % jobStoreString)
 
+    def importFile(self, srcUrl):
+        """
+        Imports the file pointed at by sourceUrl into jobStore. The jobStoreFileId of the new
```
hannes-ucsc: s/jobStore/job store/ globally in all strings and comments.

hannes-ucsc:
Don't squash the commits that address my comments. We'll squash until after my review passes.

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Do that as a single commit.

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Do that as a single commit.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+                jobStore e.g. 's3' for AWS.
+        :return The jobStoreFileId of imported file.
+        """
+        with self.writeFileStream() as (writable, jobStoreFileID):
+            findJobStoreForUrl(srcUrl).writeFromUrl(srcUrl, writable)
+            return jobStoreFileID
+
+    def exportFile(self, jobStoreFileID, dstUrl):
+        """
+        Exports file to destination pointed at by the destUrl.
```
hannes-ucsc: `Export the given file from this job store to the specified location`

Speak about what URL schemes might be supported, depending on which job stores are installed.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        """
+        Imports the file pointed at by sourceUrl into jobStore. The jobStoreFileId of the new
+        file is returned.
+
+        :param str sourceUrl: URL that points to either a local file or to a file in a supported
+                jobStore e.g. 's3' for AWS.
+        :return The jobStoreFileId of imported file.
+        """
+        with self.writeFileStream() as (writable, jobStoreFileID):
+            findJobStoreForUrl(srcUrl).writeFromUrl(srcUrl, writable)
```
hannes-ucsc: Make sure that support for http:// URLs is tracked as an issue.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+            return jobStoreFileID
+
+    def exportFile(self, jobStoreFileID, dstUrl):
+        """
+        Exports file to destination pointed at by the destUrl.
+
+        :param str jobStoreFileID: The id of the file in the jobStore that should be exported.
+        :param str destUrl: URL that points to either a local file or to a file in a supported
+                jobStore e.g. 's3' for AWS.
```
hannes-ucsc: Mention what a URL referencing a local file would look like.

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: I think you want to be explicit and mention each scheme that we currently support. Give example URLs for each scheme.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        """
+        with self.readFileStream(jobStoreFileID) as readable:
+            findJobStoreForUrl(dstUrl).writeToUrl(readable, dstUrl)
+
+    @abstractstaticmethod
+    def writeFromUrl(url, writable):
```
hannes-ucsc: `readFromUrl`

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        """
+        This method takes a url and a writable stream and writes the contents of the file
```
hannes-ucsc: Remove `This method` from all docstrings.

hannes-ucsc: `Reads the contents of the object at the specified location and writes it to the given writable stream.``

hannes-ucsc: This whole "takes a" leads to awkward wording. Using `given` and `specified` is more concise.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        """
+        raise NotImplementedError()
+
+    @abstractstaticmethod
+    def writeToUrl(readable, url):
+        """
+        This method takes a url and a readable stream and writes the contents from the stream
+        to the file pointed at by the url.
+
+        :param str url: url pointing to a file in a jobStore
```
hannes-ucsc: Not true. The URL does not point at the job store.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+
+        :param str url: a url string that may be supported
+        :return bool: returns true if the cls supports the url
+        """
+        return cls.nativeUrlRegex().match(url) is not None
+
+    @abstractclassmethod
+    def nativeUrlRegex(cls):
+        """
+        This method returns a compiled regex to match with a url.
```
hannes-ucsc: `Returns a compiled regex that matches any URL this job store support imports and exports of.`

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Do the removal of `This method` as one commit.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +158,81 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+
+    @classmethod
+    def supportsUrl(cls, url):
+        """
+        This method returns True if a JobStore supports the scheme of the given url.
+
+        :param str url: a url string that may be supported
+        :return bool: returns true if the cls supports the url
+        """
+        return cls.nativeUrlRegex().match(url) is not None
```
hannes-ucsc: Seems like a good idea but now subclasses are forced to use regexes to match their URL. This greatly restricts their options which is a too big a price to pay for the DRYness of this approach. If you really think about it, the only code you deduplicated is `.match(url) is not None` not a very big savings if you compare it to the price you paying for the limited flexibility.

Also you are conflating two different concerns: determining if a URL is supported (which should just look at the scheme) and wether a supported URL is actually valid. These concerns should be separate. I want the AWSJobStore to claim any URL starting with `s3://`. If then that URL turns out to be invalid I want an explicit exception to be raised saying that. In the current implementation I'm just getting a generic `No existing jobStore supports the url ` exception. I'd rather get a 'The URL ... is invalid`. 

Last but not least, don't parse URLs with regexes. That's reinventing the wheel, use the `urlparse` module for that.

hannes-ucsc:
Note to self: Stopped at copyKeyMultipart

JakeNarkizian:
Jenkins, retest this please.

src/toil/jobStores/abstractJobStore.py
JakeNarkizian: Just realized I forgot to s/jobStore/job store/ ... sorry about that,  I'll get it on the next round.

JakeNarkizian:
@hannes-ucsc

src/toil/jobStores/abstractJobStore.py
```diff
@@ -54,6 +58,48 @@ def __init__(self, message):
...
+    This method returns the class of the jobStore which supports exporting to and from the given
+    URL.
+
+    :param URLparse.ParseResult URL
+    :return: The class of the jobStore which supports this URL.
+    """
+    for cls in getJobStoreClasses():
+        if cls.supportsUrl(url):
+            return cls
+    raise RuntimeError("No existing jobStore supports the url '%s'" % url)
```
hannes-ucsc: URL

src/toil/jobStores/abstractJobStore.py
```diff
@@ -54,6 +58,48 @@ def __init__(self, message):
...
 
 
+@memoize
+def getJobStoreClasses():
+    """
+    Compiles as list of the classes of all jobStores whose dependencies
+    are installed.
+
+    Note that jobStore names must be manually added to the 'jobStoreClassNames'
+    tuple below.
```
hannes-ucsc: Delete `to the 'jobStoreClassNames' tuple below.`

src/toil/jobStores/abstractJobStore.py
```diff
@@ -54,6 +58,48 @@ def __init__(self, message):
...
+            jobStoreClasses.append(getattr(module, className))
+    return jobStoreClasses
+
+
+def findJobStoreForUrl(url):
+    """
+    This method returns the class of the jobStore which supports exporting to and from the given
+    URL.
+
+    :param URLparse.ParseResult URL
```
hannes-ucsc: urlparse

hannes-ucsc: url

hannes-ucsc: Come on!

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Why hasn't this been addressed?

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Why hasn't this been addressed?
 Add a line note

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,6 +160,86 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        """
+        Returns True if the jobStore supports the URL's scheme.
+
+        :param urlparse.ParseResult url: a parsed URL that may be supported
+        :return bool: returns true if the cls supports the URL
+        """
+        raise NotImplementedError()
+
+    @abstractstaticmethod
+    def _realizeUrl(url):
```
hannes-ucsc: Remove. This is an implementation detail that is not part of the API. You don't want to restrict subclasses to a specific way of implementing this detail unless you get something in return for paying that price.

src/toil/jobStores/abstractJobStore.py
JakeNarkizian: I added all the sup[ported schemes and examples for each under the docstring for importFile() should I add them to both import and export?

src/toil/jobStores/abstractJobStore.py
JakeNarkizian: same as above... I did this in import file, should I have two copies of it for both methods?

JakeNarkizian:
@hannes-ucsc ready

src/toil/jobStores/abstractJobStore.py
```diff
@@ -54,6 +58,47 @@ def __init__(self, message):
...
+    return jobStoreClasses
+
+
+def findJobStoreForUrl(url):
+    """
+    This method returns the class of the job store which supports exporting to and from the given
+    URL.
+
+    :param urlparse.ParseResult url:
+    :return: The class of the job store which supports this URL.
```
hannes-ucsc: "The concrete AbstractJobStore subclass that supports the given URL"

Add :rtype: annotation

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +159,88 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
             raise JobStoreCreationException("The job store '%s' does not exist, so there "
                                             "is nothing to restart." % jobStoreString)
 
+    def importFile(self, srcUrl):
+        """
+        Imports the file pointed at by sourceUrl into job store. The jobStoreFileId of the new
+        file is returned.
+
+        Currently supported schemes include:
```
hannes-ucsc: s/include/are

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +159,88 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+
+            - 'wasb' for blobs in Azure Blob Storage
+                e.g. wasb://container/blob
+
+            - 'file' for local files
+                e.g. file://local/file/path
+
+        :param str sourceUrl: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :return The jobStoreFileId of imported file.
```
hannes-ucsc: Add `:rtype:" str`

src/toil/jobStores/abstractJobStore.py
hannes-ucsc: Note to self: This has been addressed.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,47 @@
 
 log = logging.getLogger(__name__)
 
+s3partSize = 2**20*50 
+
+
+def copyKeyMultipart(newKeyName, dstBucketName, srcBucketName, srcKey, headers={}):
+    """
+    Copies a key pointed at by the src URL to the dst URL in multiple parts.
```
hannes-ucsc: Don't abbreviate src and dst in the comment.

src/toil/job.py
```diff
@@ -711,6 +711,12 @@ def deleteGlobalFile(self, fileStoreID):
                 #This will result in the files removal from the cache at the end of the current job
                 self._jobStoreFileIDToCacheLocation.pop(fileStoreID)
 
+        def importFile(self, sourceUrl):
```
hannes-ucsc: srcUrl

src/toil/job.py
```diff
@@ -711,6 +711,12 @@ def deleteGlobalFile(self, fileStoreID):
...
+            return self.jobStore.importFile(sourceUrl)
+
+        def exportFile(self, jobStoreFileID, destUrl):
```
hannes-ucsc: dstUrl

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,50 @@ def getEmptyFileStoreID(self, jobStoreID=None):
         log.debug("Created %r.", info)
         return info.fileID
 
+    def importFile(self, srcUrl):
+        url = urlparse.urlparse(srcUrl)
```
hannes-ucsc: reuse srcUrl

hannes-ucsc: ignore previous comment

hannes-ucsc:
Consider this:

```diff
(bd2k:/__me__/)hannes-mbpr:toil hannes $ git diff src/toil/jobStores/abstractJobStore.py src/toil/jobStores/aws/jobStore.py
diff --git a/src/toil/jobStores/abstractJobStore.py b/src/toil/jobStores/abstractJobStore.py
index 84577ee..939fcd9 100644
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -179,8 +179,12 @@ class AbstractJobStore(object):
         :return The jobStoreFileId of imported file.
         """
         url = urlparse.urlparse(srcUrl)
+        other = findJobStoreForUrl(url)
+        return self._importFile(other, url)
+
+    def _importFile(self, other, url):
         with self.writeFileStream() as (writable, jobStoreFileID):
-            findJobStoreForUrl(url).readFromUrl(url, writable)
+            other.readFromUrl(url, writable)
             return jobStoreFileID
 
     def exportFile(self, jobStoreFileID, dstUrl):
diff --git a/src/toil/jobStores/aws/jobStore.py b/src/toil/jobStores/aws/jobStore.py
index 07d0343..6f2cae7 100644
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -272,16 +272,15 @@ class AWSJobStore(AbstractJobStore):
         log.debug("Created %r.", info)
         return info.fileID
 
-    def importFile(self, srcUrl):
-        url = urlparse.urlparse(srcUrl)
-        if AWSJobStore.supportsUrl(url):
+    def _importFile(self, other, url):
+        if self == other:
             srcBucket, srcKey = self._extractKeyInfoFromUrl(url)
             info = self.FileInfo.create(srcKey.name)
             info.copyFrom(srcKey)
             info.save()
             return info.fileID
         else:
-            return super(AWSJobStore, self).importFile(srcUrl)
+            return super(AWSJobStore, self)._importFile(other, url)
 
     def exportFile(self, jobStoreFileID, dstUrl):
         url = urlparse.urlparse(dstUrl)
(bd2k:/__me__/)hannes-mbpr:toil hannes $ 
```

hannes-ucsc:
One change to above proposal:

`self == other` should be `isinstance(other, type(self))`

JakeNarkizian:
@hannes-ucsc hit me again

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +158,118 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        :rtype: str
+        """
+        url = urlparse.urlparse(srcUrl)
+        return self._importFile(findJobStoreForUrl(url), url)
+
+    def _importFile(self, otherCls, url):
+        """
+        Refer to importFile docstring for information about this method.
+
+        :param toil.jobStores.abstractJobStore.AbstractJobStore otherCls: The class of job store to import from.
```
hannes-ucsc: I don't think you need to fully qualify the class name here.

And the type hit is incorrect because it currently specifies that an instance of AbstractJobStore is passed. I looked around and couldn't find a type hinting syntax for an argument that is a class rather than an instance of a class. A safe bet would be to use `type` as the type hint:

```
>>> class Foo(object):
...     pass
...     
>>> foo = Foo()
>>> isinstance(foo,Foo)
True
>>> isinstance(Foo,type)
True
>>> type(Foo)
<type 'type'>
>>> type(foo)
<class 'Foo'>
```

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +158,118 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        """
+        url = urlparse.urlparse(dstUrl)
+        return self._exportFile(findJobStoreForUrl(url), jobStoreFileID, url)
+
+    def _exportFile(self, otherCls, jobStoreFileID, url):
+        """
+        Refer to exportFile docstring for information about this method.
+
+        :param toil.jobStores.abstractJobStore.AbstractJobStore otherCls: The class of job store to import from.
```
hannes-ucsc:
```
:param type otherCls: the concrete subclass of AbstractJobStore that supports exporting to the given URL
```

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +158,118 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        :param str jobStoreFileID: The id of the file that will be exported.
+        :param urlparse.ParseResult url: The parsed url given to importFile.
+        """
+        with self.readFileStream(jobStoreFileID) as readable:
+            otherCls.writeToUrl(readable, url)
+
+    @abstractstaticmethod
+    def readFromUrl(url, writable):
```
hannes-ucsc: Should we make this protected?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +158,118 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        :param str url: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :param writable: a writable stream
+        """
+        raise NotImplementedError()
+
+    @abstractstaticmethod
+    def writeToUrl(readable, url):
```
hannes-ucsc: Should we make this protected?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -114,10 +158,118 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
...
+        Refer to AbstractJobStore.importFile documentation for currently supported URL schemes.
+
+        :param str url: URL that points to a file or object in the storage mechanism of a
+                supported URL scheme e.g. a blob in an Azure Blob Storage container.
+        :param readable: a readable stream
+        """
+        raise NotImplementedError()
+
+    @abstractclassmethod
+    def supportsUrl(cls, url):
```
hannes-ucsc: Should we make this protected?

hannes-ucsc: If only methods of AbstractJobStore or a subclass thereof call these methods, I think we should.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,47 @@
...
 log = logging.getLogger(__name__)
 
+s3partSize = 2**20*50 
+
+
+def copyKeyMultipart(newKeyName, dstBucketName, srcBucketName, srcKey, headers={}):
+    """
+    Copies a key pointed at by the source URL to the destination URL in multiple parts.
+
+    :param str newKeyName: The name of the new key that will be created.
```
hannes-ucsc: You don't know if this is a new key. It could be an existing key that you are overwriting.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,47 @@
 
 log = logging.getLogger(__name__)
 
+s3partSize = 2**20*50 
+
+
+def copyKeyMultipart(newKeyName, dstBucketName, srcBucketName, srcKey, headers={}):
```
hannes-ucsc: srcBucketName, srcKeyName, dstBucketName, dstKey

hannes-ucsc: And you can't use instances of mutable types as default arguments.

https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,47 @@
...
+    """
+    Copies a key pointed at by the source URL to the destination URL in multiple parts.
```
hannes-ucsc: What source URL and what destination URL?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -47,6 +49,47 @@
 
 log = logging.getLogger(__name__)
 
+s3partSize = 2**20*50
```
hannes-ucsc: Lets call it defaultS3PartSize to indicate that it can be overriden. Also makes it S3 instead of s3.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -108,6 +151,7 @@ def __init__(self, region, namePrefix, config=None):
         self.filesBucket = None
         self.db = self._connectSimpleDB()
         self.s3 = self._connectS3()
+        self.partSize = partSize
```
hannes-ucsc: This should be s3partSize or defaultS3PartSize should be defaultPartSize. I'm ok with either way, as long as it's consistent. defaultPartSize and partSize is probably easier.

src/toil/jobStores/aws/jobStore.py
hannes-ucsc: But see below ...

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
         log.debug("Created %r.", info)
         return info.fileID
 
+    def _importFile(self, otherCls, url):
+        if issubclass(otherCls, type(self)):
```
hannes-ucsc: `if issubclass(otherCls, AWSJobStore):`

because self could be an instance of a sibling or cousin of otherCls.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+        srcBucket, srcKey = AWSJobStore._extractKeyInfoFromUrl(url)
+        srcKey.get_contents_to_file(writable)
+
+    @staticmethod
+    def writeToUrl(readable, url):
+        dstBucket, dstKey = AWSJobStore._extractKeyInfoFromUrl(url)
+        dstKey.set_contents_from_string(readable.read())
+
+    @staticmethod
+    def _extractKeyInfoFromUrl(url):
```
hannes-ucsc: Why not just `_parseS3Url()`

hannes-ucsc: Scratch that.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+            srcBucket, srcKey = self._extractKeyInfoFromUrl(url)
+            info = self.FileInfo.create(srcKey.name)
+            info.copyFrom(srcKey)
+            info.save()
+            return info.fileID
+        else:
+            return super(AWSJobStore, self)._importFile(otherCls, url)
+
+    def _exportFile(self, otherCls, jobStoreFileID, url):
+        if issubclass(otherCls, type(self)):
```
hannes-ucsc: See above

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+            dstBucket, newKey = self._extractKeyInfoFromUrl(url)
+            info = self.FileInfo.loadOrFail(jobStoreFileID)
+            info.copyTo(newKey.name, dstBucket)
+        else:
+            super(AWSJobStore, self)._exportFile(otherCls, jobStoreFileID, url)
+
+    @staticmethod
```
hannes-ucsc: @classmethod

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+    def readFromUrl(url, writable):
+        srcBucket, srcKey = AWSJobStore._extractKeyInfoFromUrl(url)
```
hannes-ucsc: `cls._extractKeyInfoFromUrl`

hannes-ucsc: ... because that would allow subclasses to override _extractKeyInfoFromUrl

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+        srcKey.get_contents_to_file(writable)
+
+    @staticmethod
```
hannes-ucsc: see above

src/toil/jobStores/aws/jobStore.py
```diff
@@ -228,6 +272,48 @@ def getEmptyFileStoreID(self, jobStoreID=None):
...
+        :return: (bucket, key)
+        """
+        s3 = boto.connect_s3()
+        bucket = s3.get_bucket(url.netloc)
+        key = bucket.new_key(url.path[1:])
+        return bucket, key
+
+    @classmethod
+    def supportsUrl(cls, url):
+        return url.scheme.lower() == 's3'
```
hannes-ucsc: Ahh, good catch:

http://tools.ietf.org/html/rfc3986#section-3.1

src/toil/jobStores/aws/jobStore.py
```diff
@@ -726,6 +810,70 @@ def reader():
...
+                dstBucket = boto.connect_s3().get_bucket(dstBucketName)
+                return dstBucket.copy_key(new_key_name=newKeyName,
+                                          src_bucket_name=srcBucketName,
+                                          src_key_name=srcKey.name,
+                                          metadata=srcKey.metadata,
+                                          headers=headers)
+
+        def _hashesAreEqual(self, srcHash, dstHash):
+            if self.encrypted:
+                log.debug("MD5 hash of imported or exported files cannot be checked "
```
hannes-ucsc: This is a little verbose for a debug level message.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -726,6 +810,70 @@ def reader():
...
+                                             srcBucketName=srcKey.bucket.name,
+                                             srcKey=srcKey,
+                                             headers=self._s3EncryptionHeaders()).version_id
+
+        def copyTo(self, newKeyName, dstBucket):
+            """
+            Copies contents of file to destination key.
+
+            :param str newKeyName: The name of the new key to be created or copied to from the file.
+            :param dstBucket: The bucket that contains or will contain the destination key.
```
hannes-ucsc: `from boto.s3.bucket import Bucket` at the top and `:param Bucket dstBucket:` here.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -726,6 +810,70 @@ def reader():
...
+                self._copyKey(newKeyName=newKeyName,
+                              dstBucketName=dstBucket.name,
+                              srcBucketName=self.outer.filesBucket.name,
+                              srcKey=srcKey,
+                              headers={"amz-copy-source".join(k.split('amz')): v
+                                       for k, v in self._s3EncryptionHeaders().items()})
+            else:
+                assert False
+
+        def _copyKey(self, newKeyName, dstBucketName, srcBucketName, srcKey, headers={}):
```
hannes-ucsc: arg order, naming and mutable default arg, see copyKeyMultipart

src/toil/jobStores/azureJobStore.py
```diff
@@ -192,6 +195,33 @@ def deleteJobStore(self):
...
+        blobService.put_block_blob_from_file(containerName, blobName, readable)
+        blobService.get_blob(containerName, blobName)
+
+    @staticmethod
+    def _extractBlobInfoFromUrl(url):
+        """
+        :return: (blobService, containerName, blobName)
+        """
+        netloc = url.netloc.split('@')
+        containerName, accountName = netloc[0], netloc[1].split('.blob.core.windows.net')[0]
```
hannes-ucsc: Think about what happens if the URL doesn't contain `blob.core.windows.net`. And what should happen?

src/toil/jobStores/azureJobStore.py
```diff
@@ -192,6 +195,33 @@ def deleteJobStore(self):
...
+        blobName = url.path.split('/')[1]
```
hannes-ucsc: Are you doing this because path starts with a '/'. If so that this is a convoluted way to get rid of the leading slash.

src/toil/jobStores/fileJobStore.py
```diff
@@ -144,8 +149,44 @@ def jobs(self):
                         pass  
 
     ##########################################
-    #Functions that deal with temporary files associated with jobs
-    ##########################################    
+    # Functions that deal with temporary files associated with jobs
+    ##########################################
+
+    def _importFile(self, otherCls, url):
+        if issubclass(otherCls, type(self)):
```
hannes-ucsc: See AWSFileStore

src/toil/jobStores/fileJobStore.py
```diff
@@ -144,8 +149,44 @@ def jobs(self):
...
+    def writeToUrl(readable, url):
+        with open(FileJobStore._extractPathFromUrl(url), 'w') as f:
+            f.write(readable.read())
+
+    @staticmethod
+    def _extractPathFromUrl(url):
+        """
+        :return: local file path of file pointed at by the given URL
+        """
+        return url.netloc + url.path
```
hannes-ucsc: That can't be right ...

`file:///foo/bar.txt` is equivalent to `file://localhost/foo/bar.txt`

You should verify in `supportsUrl()` that netloc is `''` or `'localhost'`.

Note that `file://some-remote-host/foo/bar.txt` is in fact a valid file URL but not one that FileJobStore can support.

`file://foo/bar.txt` is not a valid URL but urlparse() parses it. You should catch that, too.

src/toil/jobStores/fileJobStore.py
```diff
@@ -144,8 +149,44 @@ def jobs(self):
...
+
+    @classmethod
+    def supportsUrl(cls, url):
+        return url.scheme.lower() == 'file'
```
hannes-ucsc: see above

src/toil/test/__init__.py
```diff
@@ -253,3 +253,139 @@ def needs_cwl(test_item):
         raise
     else:
         return test_item
+
+MethodNameRegex = re.compile('[a-zA-Z_0-9]+$')
```
hannes-ucsc: methodNameRegex

hannes-ucsc: Why do you anchor only at the end? Because you use match() instead of find()? That would be brittle.

hannes-ucsc: And method names can't start with a digit ...

src/toil/test/__init__.py
```diff
@@ -253,3 +253,139 @@ def needs_cwl(test_item):
...
+
+
+def make_tests(generalMethod, targetClass=None, **kwargs):
```
hannes-ucsc: Add FIXME: move to bd2k-python-lib

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -14,26 +14,37 @@
...
                                              NoSuchFileException)
-from toil.jobStores.fileJobStore import FileJobStore
-from toil.test import ToilTest, needs_aws, needs_azure, needs_encryption
 
+from bd2k.util.objects import abstractstaticmethod, abstractclassmethod
+from toil.jobStores.fileJobStore import FileJobStore
+from toil.lib import encryption
+from toil.test import ToilTest, needs_aws, needs_azure, needs_encryption, make_tests
 logger = logging.getLogger(__name__)
+mpTestPartSize = 2**20 * 5
```
hannes-ucsc: Can we make this a class attribute of AJST and initialize right before makeImportExportTests?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -292,6 +303,66 @@ def callback(f2):
...
+                self.assertEqual(self._hashJobStoreFileID(jobStoreFileID), srcHash)
+
+                dstUrl = foreignJobStore._getUrlForTestFile()
+                self.addCleanup(foreignJobStore._cleanUpExternalStore, dstUrl)
+                self.master.exportFile(jobStoreFileID, dstUrl)
+                dstHash = foreignJobStore._hashUrl(dstUrl)
+                self.assertEqual(self._hashJobStoreFileID(jobStoreFileID), dstHash)
+
+            jobStoreTestClasses = [FileJobStoreTest, AWSJobStoreTest, AzureJobStoreTest]
+            make_tests(importExportFile, targetClass=cls, foreignJobStore={jsCls.__name__:
```
hannes-ucsc: This would look better if each keyword argument were on a line of its own, no?

hannes-ucsc: The fact that the dict comprehension wraps irks me ...

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -605,6 +778,46 @@ def testLargeJob(self):
...
+
+    def _hashJobStoreFileID(self, jobStoreFileID):
+        with self.master.readFileStream(jobStoreFileID) as f:
+            return hashlib.md5(f.read()).hexdigest()
+
+    @staticmethod
+    def _hashUrl(url):
+        from toil.jobStores.azureJobStore import AzureJobStore
+        blobService, containerName, blobName = AzureJobStore._extractBlobInfoFromUrl(urlparse.urlparse(url))
+        content = str(blobService.get_blob_to_bytes(containerName, blobName))
```
hannes-ucsc: Is `str` used to convert from unicode? If so, this should be explained in a short comment.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -605,6 +778,46 @@ def testLargeJob(self):
...
+        return hashlib.md5(content).hexdigest()
+
+    @staticmethod
+    def _createExternalStore():
+        from toil.jobStores.azureJobStore import _fetchAzureAccountKey
+        blobService = BlobService(account_key=_fetchAzureAccountKey(AzureJobStoreTest.accountName),
+                                  account_name=AzureJobStoreTest.accountName)
+        containerName = 'import-export-test-%s' % (uuid.uuid4())
```
hannes-ucsc: `% uuid.uuid4()`

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -292,6 +303,66 @@ def callback(f2):
...
+        def _createExternalStore():
+            raise NotImplementedError()
+
+        @abstractstaticmethod
+        def _cleanUpExternalStore(url):
+            raise NotImplementedError()
+
+        @classmethod
+        def makeImportExportTests(cls):
+            def importExportFile(self, foreignJobStore, size):
```
hannes-ucsc: otherJobStore? Since you agreed with me on self vs. other yesterday...

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -535,6 +606,41 @@ class FileJobStoreTest(hidden.AbstractJobStoreTest):
...
         return FileJobStore(self.namePrefix, config=config)
 
+    @classmethod
+    def _getUrlForTestFile(cls, size=None):
+        fileName = 'testfile_%s' % uuid.uuid4()
+        dirPath = cls._externalStore()
+        localFilePath = dirPath + fileName
+        url = 'file://%s' % localFilePath
+
+        if size is None:
```
hannes-ucsc: There are a too many empty lines here for my taste. And despite an `else:` clause isn't exactly necessary because of the `return` in the `if` branch, I like to throw it in for clarity and symmetry.

hannes-ucsc:
I'm very happy with this. Great job on the generative tests. Your next push should squash all commits. I think Friday is the day this gets merged ... :-)

src/toil/test/__init__.py
JakeNarkizian: this regex checks if a string could be part of a method name, and every method name used int he make_tests function is prepended with "test_" which is why digits are included

hannes-ucsc: I see. methodNamePartRegex then? And my anchoring remark still applies, right?

JakeNarkizian:
jenkins, retest this please.

JakeNarkizian:
jenkins, retest this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

JakeNarkizian:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
PR #723 will be merged soon. Once it does, please rebase and add the test for the staging.

hannes-ucsc:
#723 has been merged. @JakeNarkizian, can you rebase this on master and see if you can fix the test failures? They look to be straight-forward to fix.

PR #715, opened 2016-03-12
logLevel in logToMaster is not a valid parameter anymore

brainstorm:
The [promises documentation example](http://toil.readthedocs.org/en/releases-3.1.x/tutorial.html#promises) raises the following error when tested:

```
2016-03-11 17:44:37,186 WARNING:toil.leader: o/u/job3P9XQk:       File "promises_test.py", line 6, in fn
2016-03-11 17:44:37,186 WARNING:toil.leader: o/u/job3P9XQk:         job.fileStore.logToMaster("i is: %s" % i, logLevel=100)
2016-03-11 17:44:37,186 WARNING:toil.leader: o/u/job3P9XQk:     TypeError: logToMaster() got an unexpected keyword argument 'logLevel'
```

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -305,7 +305,7 @@ example::
     from toil.job import Job
     
     def fn(job, i):
-        job.fileStore.logToMaster("i is: %s" % i, logLevel=100)
+        job.fileStore.logToMaster("i is: %s" % i, level=100)
         return i+1
         
     j1 = Job.wrapJobFn(fn, 1)


brainstorm:
Instead, it should be passed as `options.logLevel` to Job.Runner, as far as I understand:

```
if __name__=="__main__":
## Toil options
    options = Job.Runner.getDefaultOptions("./toilWorkflowRun")
    options.batchSystem = 'slurm'
    options.logLevel = 'DEBUG'
```

arkal:
I believe the parameter has been renamed `level`

brainstorm:
Should I change the docs accordingly or perhaps it makes more sense to drop it in favor of the global `options.logLevel`?

hannes-ucsc:
Thank you for contributing, @brainstorm. I would prefer just fixing the docs to refer to `level=`. Just because the method already has "log" in its name and prepending it to the keyword argument name would be redundant. Python's built-in logging module also calls that argument `level`.

hannes-ucsc:
@brainstorm could you possibly squash the two commits? If you don't have time, never mind. It would look better in the history.

brainstorm:
Sure, does it look good now?

hannes-ucsc:
Thank you!

PR #721, opened 2016-03-23
Reset TMPDIR with every call of jenkins.sh (resolves #719)

### LINKED ISSUE #719: Reset $TMPDIR with every call of jenkins.sh
Delete the tmpdir, create and use it, then clean it up after the test.

If a PR on a jenkins slave is queued before another finishes on a slave, and the queued one is run on the same slave as the previous, then mkdir /mnt/ephemeral/tmp fails and $TMPDIR is not set up to a location on the ephemeral.  This PR addresses this issue.

arkal:
resolves #719

`/mnt/ephemeral/tmp` is now deleted before and after the jenkins tests. This prevents the testing temp directory from having other files in it when the test starts.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -10,5 +10,7 @@ pip2.7 install sphinx
 make develop extras=[aws,mesos,azure,encryption,cwl]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
+rm -rf /mnt/ephemeral/tmp
 mkdir /mnt/ephemeral/tmp && export TMPDIR=/mnt/ephemeral/tmp
 make $make_targets
+rm -rf /mnt/ephemeral/tmp


jenkins.sh
```diff
@@ -10,5 +10,7 @@ pip2.7 install sphinx
 make develop extras=[aws,mesos,azure,encryption,cwl]
 export LIBPROCESS_IP=127.0.0.1
 export PYTEST_ADDOPTS="--junitxml=test-report.xml"
+rm -rf /mnt/ephemeral/tmp
```
hannes-ucsc: Doesn't this fail if the directory doesn't exist? The way jenkins.sh is run is that any failed command leads to build failure.

arkal: no it doesn't! unless it varies on different platforms. My build just started on a previous slave and the log doesn't show an error.

hannes-ucsc: Ahh yes, the -f does that.

However, to prove that it doesn't fail if /mnt/ephemeral/tmp is absent, you would need to run it on a fresh slave.

hannes-ucsc:
Jenkins, test this please.

arkal:
@hannes-ucsc this is ready!

PR #725, opened 2016-03-29
Port commits from 3.1.x onto master (resolves #683, resolves #686, resolves #690)

### LINKED ISSUE #683: logToMaster() doesn't work

### LINKED ISSUE #686: Add option for retaining files in work directory (--workDir)

### LINKED ISSUE #690: A failing root job causes EOFError

cket:
Cherry picked already merged commits from 3.1.x branch. Resolves a logToMaster issue, EOF error upon root job failure, and adds support for optionally retaining the workDir on workers.

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -15,7 +15,6 @@
 
 from __future__ import absolute_import
 from collections import namedtuple
-from Queue import Empty
 from toil.common import getToilWorkflowDir
 import os
 import shutil
@@ -25,7 +24,10 @@ WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
     # A path to the value of config.workDir (where the cache would go)
     'workDir',
     # The value of config.workflowID (used to identify files specific to this workflow)
-    'workflowID'))
+    'workflowID',
+    # The value of the cleanWorkDir flag
+    'cleanWorkDir'))
+
 
 class AbstractBatchSystem:
     """An abstract (as far as python currently allows) base class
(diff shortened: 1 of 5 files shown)


src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -160,16 +162,17 @@ def supportsWorkerCleanup():
...
         :param collections.namedtuple workerCleanupInfo: A named tuple consisting of all the
         relevant information for cleaning up the worker.
-        '''
+        """
         assert workerCleanupInfo.__class__.__name__ == 'WorkerCleanupInfo'
         workflowDir = getToilWorkflowDir(workerCleanupInfo.workflowID, workerCleanupInfo.workDir)
-        # TODO: logic here for cleanup flag
-        shutil.rmtree(workflowDir)
-
+        dirIsEmpty = os.listdir(workflowDir) == []
```
cket: @hannes-ucsc I checked and did not find a method that simplifies checking if a directory is empty

cket:
@hannes-ucsc I will merge this PR later today per your instructions yesterday

src/toil/batchSystems/abstractBatchSystem.py
hannes-ucsc: Yes, I was wrong. `os.stat(path).n_link >2` only works to check if a directory contains sub-directories.

hannes-ucsc:
> I will merge this

SGTM

PR #743, opened 2016-04-04
Move documentation build logic into conf.py (resolves #611)

### LINKED ISSUE #611: Module Index link is broken on Sphinx docs index page
http://toil.readthedocs.org/en/latest/py-modindex.html

This link is broken. We should have a way of browsing all the modules in the docs.

cket:


hannes-ucsc:
@cket and me agreed that he would try to build this on RTD before it is merged to master. If it works, the `needs work` label can be taken off.

cket:
Removing needs work — module index builds correctly on RTD for this branch

PR #755, opened 2016-04-06
Enforce venv in conf.py (resolves #752)

### LINKED ISSUE #752: Enforce venv'ed Sphinx when building docs
Add check to config.py. See Makefile for how to ~~enforce~~ check for a venv.

cket:


hannes-ucsc:
Tests failed because @arkal was messing with the slave at the same time. I will merge this anyhow because the build log shows that it made it through the `docs` target.

hannes-ucsc:
I didn't catch that the target branch is wrong and that the commit message is missing the ticket number. I will revert this. Please file a new PR that fixes these issues.

hannes-ucsc:
Never mind. Too much work.

PR #791, opened 2016-04-13
Implement promised requirements (resolves #729)

### LINKED ISSUE #729: Fix Promises return value
Our group would like to broaden the scope of promises (e.g. return both a fileStoreID and file-size that can be resolved and used in the same function).

jpfeil:
Resolves #729

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -18,6 +18,7 @@ import sys
 import importlib
 from argparse import ArgumentParser
 from abc import ABCMeta, abstractmethod
+import dill
 import tempfile
 import uuid
 import time
@@ -60,13 +61,35 @@ class Job(object):
         exhausting all their retries, remove any successor jobs and rerun this job to restart the subtree. \
         Job must be a leaf vertex in the job graph when initially defined, \
         see :func:`toil.job.Job.checkNewCheckpointsAreCutVertices`.
-        :type cores: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type disk: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type cache: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type memory: int or string convertable by bd2k.util.humanize.human2bytes to an int
+        :param int|PromisedRequirement cores:
+        :param int|str|PromisedRequirement disk:        
+        :param int|str|PromisedRequirement cache:
+        :param int|str|PromisedRequirement memory:
         """
-        self.cores = cores
+
+        # If resource requirement is a PromisedRequirement,
+        # then set the parameter to the default until the
+        # promise can be fulfilled.
+        # See toil.job.PromisedRequirement
+        if isinstance(cores, PromisedRequirement):
+            self.cores_promise = cores
+            cores = None 
+
+        if isinstance(disk, PromisedRequirement):
+            self.disk_promise = disk
+            disk = None
+
+        if isinstance(memory, PromisedRequirement):
+            self.memory_promise = memory
+            memory = None
+
+        if isinstance(cache, PromisedRequirement):
+            self.cache_promise = cache
+            cache = None
+
         parse = lambda x : x if x is None else human2bytes(str(x))
+
+        self.cores = cores
         self.memory = parse(memory)
         self.disk = parse(disk)
         self.cache = parse(cache)
@@ -176,6 +199,9 @@ class Job(object):
         :return: The new child job that wraps fn.
         :rtype: toil.job.FunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return self.addChild(PromisedRequirementFunctionWrappingJob(fn, *args, **kwargs))
         return self.addChild(FunctionWrappingJob(fn, *args, **kwargs))
 
     def addFollowOnFn(self, fn, *args, **kwargs):
@@ -188,6 +214,9 @@ class Job(object):
         :return: The new follow-on job that wraps fn.
         :rtype: toil.job.FunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return self.addFollowOn(PromisedRequirementFunctionWrappingJob(fn, *args, **kwargs))
         return self.addFollowOn(FunctionWrappingJob(fn, *args, **kwargs))
 
     def addChildJobFn(self, fn, *args, **kwargs):
@@ -201,6 +230,9 @@ class Job(object):
         :return: The new child job that wraps fn.
         :rtype: toil.job.JobFunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return self.addChild(PromisedRequirementJobFunctionWrappingJob(fn, *args, **kwargs))
         return self.addChild(JobFunctionWrappingJob(fn, *args, **kwargs))
 
     def addFollowOnJobFn(self, fn, *args, **kwargs):
@@ -214,6 +246,9 @@ class Job(object):
         :return: The new follow-on job that wraps fn.
         :rtype: toil.job.JobFunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return self.addFollowOn(PromisedRequirementJobFunctionWrappingJob(fn, *args, **kwargs))
         return self.addFollowOn(JobFunctionWrappingJob(fn, *args, **kwargs))
 
     @staticmethod
@@ -228,6 +263,9 @@ class Job(object):
         :return: The new function that wraps fn.
         :rtype: toil.job.FunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return PromisedRequirementFunctionWrappingJob(fn, *args, **kwargs)
         return FunctionWrappingJob(fn, *args, **kwargs)
 
     @staticmethod
@@ -242,6 +280,9 @@ class Job(object):
         :return: The new job function that wraps fn.
         :rtype: toil.job.JobFunctionWrappingJob
         """
+        kwargs = Job.update_kwargs(kwargs)
+        if Job.has_promised_requirement(kwargs):
+            return PromisedRequirementJobFunctionWrappingJob(fn, *args, **kwargs)
         return JobFunctionWrappingJob(fn, *args, **kwargs)
 
     def encapsulate(self):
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
Commit message is wrong.

src/toil/job.py
```diff
@@ -60,13 +61,38 @@ def __init__(self, memory=None, cores=None, disk=None, cache=None, checkpoint=Fa
         exhausting all their retries, remove any successor jobs and rerun this job to restart the subtree. \
         Job must be a leaf vertex in the job graph when initially defined, \
         see :func:`toil.job.Job.checkNewCheckpointsAreCutVertices`.
-        :type cores: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type disk: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type cache: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type memory: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        """
-        self.cores = cores
+        :type cores: int; Can also accept a PromisedRequirement object
```
hannes-ucsc: `:param int|str cores: ...`

hannes-ucsc: Read PEP 0484 for a specification . The PyCharm support web site has a page about type hinting, too.

src/toil/job.py
```diff
@@ -60,13 +61,38 @@ def __init__(self, memory=None, cores=None, disk=None, cache=None, checkpoint=Fa
...
+        :type disk: int or string convertable by bd2k.util.humanize.human2bytes to an int;
+                    Can also accept a PromisedRequirement object
+        :type cache: int or string convertable by bd2k.util.humanize.human2bytes to an int;
+                     Can also accept a PromisedRequirement object
+        :type memory: int or string convertable by bd2k.util.humanize.human2bytes to an int;
+                      Can also accept a PromisedRequirement object
+        """
+
+        # If resource requirement is a PromisedRequirement,
```
hannes-ucsc: Wrap comments at 100.

src/toil/job.py
```diff
@@ -60,13 +61,38 @@ def __init__(self, memory=None, cores=None, disk=None, cache=None, checkpoint=Fa
...
+        # then set the parameter to the default until the
+        # promise can be fulfilled.
+        # See toil.job.PromisedRequirement
+        if isinstance(cores, PromisedRequirement):
+            self.cores_promise = cores
+            cores = 1
```
hannes-ucsc: I'd be happier if we used None. Here

hannes-ucsc:
You need to modify setup.py and mention dill as a requirement.

src/toil/job.py
```diff
@@ -254,6 +298,32 @@ def encapsulate(self):
         """
         return EncapsulatedJob(self)
 
+
+    @staticmethod
+    def update_kwargs(kwargs):
```
hannes-ucsc: If you modify kwargs in place you shouldn't return it. Returning it would indicate to the caller that you return a copy and that you didn't touch the argument, but the copy.

hannes-ucsc: The name of this method is too unspecific.

src/toil/job.py
```diff
@@ -1470,6 +1540,39 @@ def run(self, fileStore):
...
+class PromisedRequirementJobFunctionWrappingJob(FunctionWrappingJob):
+    """
+    Creates a child job function using parent function parameters and promised resource requirements.
+
+    (see :class:`toil.job.JobFunctionWrappingJob and :class:`toil.job.Job`)
+    """
+
+    def run(self, fileStore):
+        userFunction = self._getUserFunction()
+        self._kwargs["cores"] = self.cores_promise() if hasattr(self, "cores_promise") else self.cores
```
hannes-ucsc: hasattr(self, ...) is a bad smell. Attributes should always be initialized in the constructor. Assigning None is ok and `is None` can be used to test wether they are set.

src/toil/job.py
```diff
@@ -1654,3 +1757,29 @@ def promisedJobReturnValueUnpickleFunction(jobStoreString, jobStoreFileID):
 copy_reg.pickle(PromisedJobReturnValue,
                 promisedJobReturnValuePickleFunction,
                 promisedJobReturnValueUnpickleFunction)
+
+
+class PromisedRequirement(object):
```
hannes-ucsc: Document the class, including its use as a callable.

src/toil/job.py
```diff
@@ -1654,3 +1757,29 @@ def promisedJobReturnValueUnpickleFunction(jobStoreString, jobStoreFileID):
...
+    def __init__(self, *args):
+        """
+        Class for generating function resource requirement values using promised return values.
+        (see toil.job.PromisedJobReturnValue)
+        :param args[0]: function for calculating job requirement
+                        default: lambda x: x
+        :param args[1:]: list of args, ordered with respect to function parameters
+        """
+        if hasattr(args[0], '__call__'):
```
hannes-ucsc: This catches `disk=PromisedReturn(child.rv())`. What about `disk=child.rv()`. Is that handled somewhere else?

src/toil/job.py
hannes-ucsc: Also, this whole block is repeated above. Any way to make it more DRY?

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+    for _ in range(cpu_count):
+        A.addFollowOn(Job.wrapFn(increment_counter, filename,
+                                 cores=PromisedRequirement(lambda x: scale * x, A.rv()),
+                                 memory='1M'))
+
+def one():
+    return 1
+
+
+def increment_counter(filepath):
```
hannes-ucsc: Bad name. This doesn't just increment the counter.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+class MaxCoresPromisedRequirementTest(ToilTest):
+    """
+    This test ensures that the PromisedRequirement method properly assigns
+    resource requirements to the single machine batch system.
+    """
+
+    def setUp(self):
+        super(MaxCoresPromisedRequirementTest, self).setUp()
+        logging.basicConfig(level=logging.INFO)
+
```
hannes-ucsc: get rid of empty line

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+        self.counterPath = writeTempFile('0,0')
```
hannes-ucsc: I feel like we talked about this already. Maybe you should take notes ;-) 

Use the functionality in ToilTest for getting a temporary directory.

jpfeil:
Jenkins, test this please.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+                                 cores=PromisedRequirement(lambda x: scale * x, A.rv()),
+                                 memory='1M'))
+
+def one():
+    return 1
+
+
+def increment_counter(filepath):
+    """
+    Without the second argument, increment counter, sleep one second and decrement.
```
hannes-ucsc: What second argument? This comment is redundant. The code below says the same thing, and is actually more accurate since here you say that it waits for 1s which isn't true.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+    def setUp(self):
+        super(MaxCoresPromisedRequirementTest, self).setUp()
+        logging.basicConfig(level=logging.INFO)
+
+        self.counterPath = writeTempFile('0,0')
+        self.cpu_count = multiprocessing.cpu_count()
+
+    def test_dynamic(self):
+        """
+        Tests using a PromisedRequirement object to allocate the number of cores per function
```
hannes-ucsc: Remember what a hard time you had with understanding how the test for concurrency worked? You need to explain this better so the next person doesn't end up in the same place.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        options = Job.Runner.getDefaultOptions(self._getTestJobStorePath())
+        options.logLevel = "INFO"
+        Job.Runner.startToil(Job.wrapJobFn(start, self.cpu_count, self.counterPath, 1), options)
+        n, m = getCounters(self.counterPath)
```
hannes-ucsc: Again, we talked about this. The jobs should return maxValue so you don't have to read the file again.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+    This test ensures that the PromisedRequirement method properly assigns
+    resource requirements to the single machine batch system.
+    """
+
+    def setUp(self):
+        super(MaxCoresPromisedRequirementTest, self).setUp()
+        logging.basicConfig(level=logging.INFO)
+
+        self.counterPath = writeTempFile('0,0')
+        self.cpu_count = multiprocessing.cpu_count()
```
hannes-ucsc: camel case

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+        Job.Runner.startToil(Job.wrapJobFn(start, self.cpu_count, self.counterPath, 1), options)
+        n, m = getCounters(self.counterPath)
+        self.assertEqual((n, m), (0, self.cpu_count))
+
+        resetCounters(self.counterPath)
+        Job.Runner.startToil(Job.wrapJobFn(start, self.cpu_count, self.counterPath, 2), options)
+        n, m = getCounters(self.counterPath)
+        self.assertEqual((n, m), (0, self.cpu_count / 2))
+
+    def test_static(self):
```
hannes-ucsc: camel case

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Tests using a PromisedRequirement to allocate number of cores in a static DAG.
```
hannes-ucsc: This isn't a static DAG.

src/toil/test/src/promisedRequirementTest.py
```diff
@@ -0,0 +1,157 @@
...
+    Tests dynamic job function creation using a PromisedRequirement.
+    """
+    child = job.addChildFn(one)
+    for _ in range(cpu_count):
+        job.addFollowOnFn(increment_counter, filename,
+                          cores=PromisedRequirement(lambda x: scale * x, child.rv()),
+                          memory='1M')
+
+
+def static_start(job, cpu_count, filename, scale):
```
hannes-ucsc: Besides this being a misnomer, these two functions are nearly identical and should be refactored for DRYness.

We do need a test for static DAGs.

hannes-ucsc:
Alright, so the main things are: 

I don't like how you handle promise_disk etc, i.e. the shadow attributes. Ideally, I would want to get rid of them.

I'm worried that resetting the requirements to the default values will have unintended consequences. The single machine batch system implements a strict FIFO regime and looks at the requirements very late. The Mesos batch system puts jobs into separate queues, based on their requirements, so it looks at them earlier.

Now  that I am saying it, your tests should be executed against every batch system. You can either leave your tests in a separate module and replicate the structure of the existing batch system tests. Or you add your tests to the existing batch system tests. I think I'd prefer the former.

hannes-ucsc:
We might not be able to support static DAGs. And we might have to defer the submission of a job with promised requirements to until after the promises have been resolved.

hannes-ucsc:
And keep this branch rebased on master, once per day.

hannes-ucsc:
@jpfeil, PR #823 is under review by @benedictpaten, so we might not get to merging it today. It failed for unrelated reasons but passes your test and all other tests that could be affected by promises. You may just want to rebase this branch onto the branch for #823 avoid being blocked that way. Also, if I were you, I would ditch your personal fork and start pushing directly to the upstream. The fact that the branch for #818 was in your fork—to which I can't push—made it necessary for me to close #818 and start #823 in its place.

jpfeil:
Continued in #828

PR #795, opened 2016-04-15
Removes Job.Service.addChild as a public method  (resolves #794)

### LINKED ISSUE #794: Fix Job.Service.addChild
Making Job.Service.addChild public is dangerous, as a user may call it after the service system is setup. It is better to modify Job.addService to allow service graphs to be defined.

benedictpaten:
Replaces functionality by adding an argument Job.addService.

Job.Service.addChild is dangerous, because the user might call addChild
after the Service has already been serialised, e.g. in
Job.Service.start.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -139,7 +139,7 @@ class Job(object):
         followOnJob._addPredecessor(self)
         return followOnJob
 
-    def addService(self, service):
+    def addService(self, service, parentService=None):
         """
         Add a service. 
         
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Commit title lacks issue reference.

hannes-ucsc:
Branch needs to be rebased and commit message changed. Thanks.

benedictpaten:
@hannes-ucsc Done, is this message okay?

hannes-ucsc:
Unfortunately. It's too long. I don't subscribe to the 50 character limit for the title line, but GitHub wraps at around 70 characters and the ticket reference should occur before that.

Also, we prefer a rebased branch rather than branches with merge commits, especially if it's a single commit branch.

http://chris.beams.io/posts/git-commit/

hannes-ucsc:
Jenkins, test this please.

PR #806, opened 2016-04-25
Ensure default are properly staged and unstaged. (resolves #601)

### LINKED ISSUE #601: Test and scale-test support for draft-3 of the CWL 1.0 specification
Due end of December 2015

tetron:
Passes CWL conformance tests with AWS/Mesos batch system.  Calls
setLoggingFromOptions(options) to set up toil logging.  Bump cwltool version
dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ kwargs = dict(
         'encryption': [
             'pynacl==0.3.0'],
         'cwl': [
-            'cwltool==1.0.20160413143011']},
+            'cwltool==1.0.20160425140546']},
     package_dir={'': 'src'},
     packages=find_packages('src', exclude=['*.test']),
     entry_points={
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -18,14 +18,15 @@
 from toil.job import Job
 from toil.common import Toil
 from toil.version import version
+from toil.lib.bioio import setLoggingFromOptions
 
 from argparse import ArgumentParser
 import cwltool.main
 import cwltool.workflow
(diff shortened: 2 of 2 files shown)


hannes-ucsc:
Commit title is too long and lacking issue reference. And please no periods at the end of commit messages.

It appears that the commit message is actually just one long line.

http://chris.beams.io/posts/git-commit/

tetron:
Hold off on this PR for the moment, I have a few S3 released fixes to add.

tetron:
Added fixes for handling S3 URLs.  Rebased to latest master.

tetron:
@hannes-ucsc please review & merge when you get a chance

setup.py
```diff
@@ -40,7 +40,7 @@
         'encryption': [
             'pynacl==0.3.0'],
         'cwl': [
-            'cwltool==1.0.20160413143011']},
+            'cwltool==1.0.20160425140546']},
```
guillermo-carrasco: @tetron would it be possible to have here a `>=` better than `==`? As I did in #827 , otherwise we will run into the same problem again. What do you think?

tetron:
The pin is on purpose, this allows us to change cwltool without having to worry about breaking production toil code that relies on it.  An open ended dependency just means more headaches for everyone involved when incompatiblies are introduced whether by accident or on purpose.

hannes-ucsc:
Re the pin: basically what he (@tetron) said. Let's continue discussing that on #827. I want to understand why @guillermo-carrasco sees the need for removing the pin.

guillermo-carrasco:
👍  thanks guys

PR #812, opened 2016-04-25
Deprecate toil restart (resolves #807)

### LINKED ISSUE #807: Remove toil restart (resolves #807, resolves #805)
And update docs to reflect this. Issue #805 is a ticket to fix toil restart so this PR also indirectly resolves that ticket.

cket:
I didn't change any documentation in the attic directory, I assumed we are not trying to keep those files updated.


hannes-ucsc:
> I didn't change any documentation in the attic directory, I assumed we are not trying to keep those files updated.

Your assumption is correct. Attic is like a trash can that gets never emptied. Even though git lets you recover deleted files, obviously, finding the commit where a file is deleted can be hard, AFAIK. That's why I tend to us an "attic" instead.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -224,8 +224,8 @@ def _checkJobStoreCreation(create, exists, jobStoreString):
         :raise JobStoreCreationException:  if create == exists
         """
         if create and exists:
-            raise JobStoreCreationException("The job store '%s' already exists. Use --restart or "
-                                            "'toil restart' to resume the workflow, or remove the "
+            raise JobStoreCreationException("The job store '%s' already exists. Use --restart "
```
hannes-ucsc: You might want to rewrap this string literal. I use PyCharm's `Edit —Fill Paragraph` for that. It magically works within string literals and comments. Make sure you have wrap column set to 100.

hannes-ucsc:
Do we mention `toil restart` anywhere in the Sphinx documentation?

cket:
We mentioned it in cli.rst but I think thats it. I looked each result returned by searching 'toil restart', 'toilrestart' (case insensitive), and just plain 'restart' so I'm fairly sure I got all references to it.

PR #813, opened 2016-04-25
Fix logging functionality (resolves #803, resolves #836)

### LINKED ISSUE #803: Logging is not being setup
Log to master messages don't appear when running sortTest. We need a test for logToMaster functionality since we tend to keep breaking it.

### LINKED ISSUE #836: logToMaster messages overwritten
When jobs are chained only the latest job's logToMaster messages are shown

cket:
Resolves #803

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -462,9 +462,9 @@ class Job(object):
         @staticmethod
         def startToil(job, options):
             """
-            Runs the toil workflow using the given options \
-            (see Job.Runner.getDefaultOptions and Job.Runner.addToilOptions) \
-            starting with this job.
+            Deprecated by toil.common.Toil.run. Runs the toil workflow using the given options
+            (see Job.Runner.getDefaultOptions and Job.Runner.addToilOptions) starting with this
+            job.
             :param toil.job.Job job: root job of the workflow
             :raises: toil.leader.FailedJobsException if at the end of function \
             their remain failed jobs.
--- a/src/toil/test/mesos/helloWorld.py
+++ b/src/toil/test/mesos/helloWorld.py
@@ -21,10 +21,12 @@ import os
 from toil.common import Toil
 from toil.job import Job
 
-logToMasterMessage = "The child job is now running!"
+childMessage = "The child job is now running!"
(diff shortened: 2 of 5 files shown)


src/toil/test/mesos/helloWorld.py
```diff
@@ -34,7 +35,7 @@ def hello_world(job):
 def hello_world_child(job, hw):
 
     path = job.fileStore.readGlobalFile(hw)
-
+    job.fileStore.logToMaster("the child job is now running!")
```
hannes-ucsc: Probably makes sense to factor this magic string out into a global variable …

src/toil/test/src/regularLogTest.py
```diff
@@ -0,0 +1,36 @@
...
+
+class RegularLogTest(ToilTest):
+
+    def testLogToMaster(self):
+        toilOutput = subprocess.check_output([sys.executable,
+                                              '-m', helloWorld.__name__,
+                                              './toilTest',
+                                              '--clean=always',
+                                              '--logLevel=info'], stderr=subprocess.STDOUT)
+        assert "the child job is now running!" in toilOutput
```
hannes-ucsc: … and use that variable here. That makes the connection clear. Without it, an unsuspecting user might change the string in helloWorld.py without changing it here.

hannes-ucsc:
Build is failing because of left-over files in the source tree. They don't look like they are yours, @cket. Hmmm…

src/toil/test/src/regularLogTest.py
hannes-ucsc: And `'The child ...` while we're at it.

src/toil/job.py
```diff
@@ -458,6 +458,7 @@ def addToilOptions(parser):
 
         @staticmethod
         def startToil(job, options):
+            # FIXME: deprecate this function in favor of the Toil context manager
```
hannes-ucsc: You know what, why don't you just rewrite the comment instead of this FIXME. I think the Sphinx doc is reasonably explicit about the deprecation, it's just that this docstring isn't clear about it.

PR #818, opened 2016-04-26
Fix stale promise job store cache (resolves #817)

### LINKED ISSUE #817: Consecutive workflows in same process use wrong cached promised job store
Running a second workflow in the same python process as another workflow will use the cached job store instance from the first workflow even if the second workflow uses a different job store. This results in the exception below. It is a secondary exception whose backtrace is misleading, which I haven't figured out. Probably caused by the fact that the exception is raised during unpickling.

```
self = <toil.jobStores.fileJobStore.FileJobStore object at 0x7f6a3ed44a50>
jobStoreFileID = 's/g/tmp3K726R.tmp'

    def _checkJobStoreFileID(self, jobStoreFileID):
        """
            Raises NoSuchFileException if the jobStoreFileID does not exist or is not a file.
            """
        if not self.fileExists(jobStoreFileID):
           raise NoSuchFileException("File %s does not exist in jobStore" % jobStoreFileID)
           TypeError: __init__() takes at most 3 arguments (4 given)
```

jpfeil:
Resolves #817

### DIFF
--- a/src/toil/test/src/promisesTest.py
+++ b/src/toil/test/src/promisesTest.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2015 UCSC Computational Genomics Lab
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Build failed for unrelated reasons, e.g. me switching instance types. Build should still fail to prove the bug but it has to fail in the test that @jpfeil added.

Jenkins, test this please.

hannes-ucsc:
Build failed again, for unrelated reasons but the test @jpfeil added also failed as expected:

http://jenkins.cgcloud.info/job/toil-pull-requests/1151/testReport/junit/src.toil.test.src.promisesTest/CachedUnpicklingJobStoreTest/test/

hannes-ucsc:
Continued in #823.

PR #821, opened 2016-04-27
Fix imports of files > 5GiB from S3 (resolves #816, resolves #829)

### LINKED ISSUE #816: Unable to import large files from S3 buckets
I'm trying to use 1000 Genomes data which is already hosted in an AWS bucket.  This is a raw reads file of about ~13 GiB.  When calling `toil.importFile` on `s3://1000genomes/phase3/data/HG01977/sequence_read/SRR360135_1.filt.fastq.gz` I get the following error:

```
2016-04-26 14:13:37,623 ERROR:root: Got exception 'S3ResponseError: 400 Bad Request
<Error><Code>InvalidRequest</Code><Message>The specified copy source is larger than the maximum allowable size for a copy source: 5368709120</Message><RequestId>CCB50A66B881B563</RequestId><HostId>1dAVKee/hXEaLCvisKyxUHlWtLbZwqcedEXOsKVSzAoE5ISrsuykzg9teW9vzidQaIqImJxbNgI=</HostId></Error>' while writing 's3://1000genomes/phase3/data/HG01977/sequence_read/SRR360135_1.filt.fastq.gz'
  File "/usr/local/bin/cwltoil", line 9, in <module>
    load_entry_point('toil==3.2.0a2', 'console_scripts', 'cwltoil')()
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 580, in main
    adjustFiles(builder.job, functools.partial(writeFile, toil.importFile, {}))
  File "/usr/local/lib/python2.7/dist-packages/cwltool/process.py", line 128, in adjustFiles
    adjustFiles(rec[d], op)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/process.py", line 131, in adjustFiles
    adjustFiles(d, op)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/process.py", line 131, in adjustFiles
    adjustFiles(d, op)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/process.py", line 126, in adjustFiles
    rec["path"] = op(rec["path"])
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 162, in writeFile
    index[x] = (writeFunc(rp), os.path.basename(x))
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 570, in importFile
    return self.jobStore.importFile(srcUrl)
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/abstractJobStore.py", line 260, in importFile
    return self._importFile(findJobStoreForUrl(url), url)
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/aws/jobStore.py", line 279, in _importFile
    info.copyFrom(srcKey)
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/aws/jobStore.py", line 826, in copyFrom
    headers=self._s3EncryptionHeaders()).version_id
  File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/aws/jobStore.py", line 865, in _copyKey
    headers=headers)
  File "/usr/local/lib/python2.7/dist-packages/boto/s3/bucket.py", line 888, in copy_key
    response.reason, body)
boto.exception.S3ResponseError: S3ResponseError: 400 Bad Request
<Error><Code>InvalidRequest</Code><Message>The specified copy source is larger than the maximum allowable size for a copy source: 5368709120</Message><RequestId>CCB50A66B881B563</RequestId><HostId>1dAVKee/hXEaLCvisKyxUHlWtLbZwqcedEXOsKVSzAoE5ISrsuykzg9teW9vzidQaIqImJxbNgI=</HostId></Error>
```

Is it really necessary to copy the file into the jobStore bucket?

### LINKED ISSUE #829: AWS multipart copy does not pass encryption headers
Minor bug that causes encrypted copies to fail

JakeNarkizian:
boto.new_key was used instead of boto.get_key

### DIFF
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -274,7 +274,7 @@ class AWSJobStore(AbstractJobStore):
 
     def _importFile(self, otherCls, url):
         if issubclass(otherCls, AWSJobStore):
-            srcBucket, srcKey = self._extractKeyInfoFromUrl(url)
+            srcBucket, srcKey = self._extractKeyInfoFromUrl(url, existing=True)
             info = self.FileInfo.create(srcKey.name)
             info.copyFrom(srcKey)
             info.save()
@@ -292,7 +292,7 @@ class AWSJobStore(AbstractJobStore):
 
     @classmethod
     def _readFromUrl(cls, url, writable):
-        srcBucket, srcKey = cls._extractKeyInfoFromUrl(url)
+        srcBucket, srcKey = cls._extractKeyInfoFromUrl(url, existing=True)
         srcKey.get_contents_to_file(writable)
 
     @classmethod
@@ -301,13 +301,33 @@ class AWSJobStore(AbstractJobStore):
         dstKey.set_contents_from_string(readable.read())
 
     @staticmethod
-    def _extractKeyInfoFromUrl(url):
+    def _extractKeyInfoFromUrl(url, existing=None):
         """
+        Extracts bucket and key from URL. The existing parameter determines if a
+        particular state of existence should be enforced. Note also that if existing
+        is not True and the key does not exist.
+
+        :param existing: determines what the state
         :return: (bucket, key)
         """
         s3 = boto.connect_s3()
         bucket = s3.get_bucket(url.netloc)
-        key = bucket.new_key(url.path[1:])
+        key = bucket.get_key(url.path[1:])
+
+        if existing is True:
+            if key is None:
+                raise RuntimeError('Key does not exist.')
+        elif existing is False:
+            if key is not None:
+                raise RuntimeError('Key exists.')
+        elif existing is None:
+            pass
+        else:
+            assert False
+
+        if key is None:
+            key = bucket.new_key(url.path[1:])
+
         return bucket, key
 
     @classmethod
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -71,7 +71,7 @@ def copyKeyMultipart(srcKey, dstBucketName, dstKeyName, headers=None):
     totalSize = srcKey.size
 
     # initiate copy
-    upload = s3.get_bucket(dstBucketName).initiate_multipart_upload(dstKeyName)
+    upload = s3.get_bucket(dstBucketName).initiate_multipart_upload(dstKeyName, headers=headers)
     try:
         start = 0
         partIndex = itertools.count()
--- a/src/toil/test/jobStores/jobStoreTest.py
+++ b/src/toil/test/jobStores/jobStoreTest.py
@@ -723,7 +723,7 @@ class AWSJobStoreTest(hidden.AbstractJobStoreTest):
     @staticmethod
     def _hashUrl(url):
         from toil.jobStores.aws.jobStore import AWSJobStore
-        bucket, key = AWSJobStore._extractKeyInfoFromUrl(urlparse.urlparse(url))
+        bucket, key = AWSJobStore._extractKeyInfoFromUrl(urlparse.urlparse(url), existing=True)
         return hashlib.md5(key.get_contents_as_string()).hexdigest()
 
     @staticmethod


hannes-ucsc:
Can you change the commit title to `Fix imports of files > 5GiB from S3 (resolves #816)`?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -307,7 +307,8 @@ def _extractKeyInfoFromUrl(url):
         """
         s3 = boto.connect_s3()
         bucket = s3.get_bucket(url.netloc)
-        key = bucket.new_key(url.path[1:])
+        key = bucket.get_key(url.path[1:])
+        key = key if key is not None else bucket.new_key(url.path[1:])
```
hannes-ucsc: Dubious about this line. If you left it at None, how much other code would have to change to accommodate that?

Also don't like `key = key if key …`. It's actually more code than a conditional assignment.

Alternatively, just do `key = bucket.get_key(…) or bucket.new_key(…)`

JakeNarkizian: If it remains None the following methods would need to be changed: ```AWSJobStoreTest._hashUrl, AWS.JobStore._writeToUrl, AWS.JobStore._readFromUrl, AWS.JobStore._importFile, AWS.JobStore._exportFile  ```

hannes-ucsc:
Don't forget to change the commit title.

src/toil/jobStores/aws/jobStore.py
hannes-ucsc: I'm surprised that there are that many. Let's try it. This method is when we first learn whether bucket and key exist. If bucket doesn't exist we can fail right away. If the key doesn't exists we should return (bucket, None). Methods that rely on the key to exist should test for None. Methods that write to the key, and hence need a Key object, should use new_key explicitly. I can only think of one place where that is necessary.

JakeNarkizian: Doing this creates an issue of having to mess around with the url again after ```_extractKeyInfoFromUrl()``` is called. 
e.g. 
```
dstBucket, newKey = self._extractKeyInfoFromUrl(url)
assert newKey is None
newKey = self.s3.new_key(url.path[1:])      
```
How about instead adding a parameter that explicitly specifies what should happen if the key does not already exist
```
_extractKeyInfoFromUrl(url, createKey)
```
This way the consideration for either case must be made with one boolean value rather than testing for None.

JakeNarkizian: _extractKeyInfoFromUrl is also not an implementation of an abstract method since every jobstore has a different method for extracting data from URLs, so changing the signature does not have many repercussions

hannes-ucsc: I don't like `createKey` because you're not really creating the key when you call `new_key`. You're just creating an instance of the Key class, not the remote key it represents. Bool args should always be keyword args and passed as such. How about

`_extractKeyInfoFromUrl(url, existing=None)`

If `existing` is `True` the key must exist, if existing is `False` it must not exist, otherwise it doesn't matter. 

The returned Key instance is never None but it may be unpopulated if the key is absent and `existing` is not True.

JakeNarkizian: sounds good to me

JakeNarkizian:
Jenkins, retest this please

hannes-ucsc:
Commit title still wrong.

JakeNarkizian:
Jenkins, retest this please

JakeNarkizian:
Jenkins, retest this please

JakeNarkizian:
@hannes-ucsc

src/toil/jobStores/aws/jobStore.py
```diff
@@ -301,13 +301,25 @@ def _writeToUrl(cls, readable, url):
...
         :return: (bucket, key)
         """
         s3 = boto.connect_s3()
         bucket = s3.get_bucket(url.netloc)
-        key = bucket.new_key(url.path[1:])
+        key = bucket.get_key(url.path[1:])
+        if existing is True:
+            assert key is not None
+        elif existing is False:
+            assert key is None
```
hannes-ucsc: We can't use `assert` to check for user errors. It would be a user error to pass a URL to an absent key for import.

Add branch for `is None`. The `else` branch should just do assert False.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -71,7 +71,7 @@ def copyKeyMultipart(srcKey, dstBucketName, dstKeyName, headers=None):
     totalSize = srcKey.size
 
     # initiate copy
-    upload = s3.get_bucket(dstBucketName).initiate_multipart_upload(dstKeyName)
+    upload = s3.get_bucket(dstBucketName).initiate_multipart_upload(dstKeyName, headers=headers)
```
hannes-ucsc: I feel like this is a different bug. Separate commit?

JakeNarkizian:
@hannes-ucsc

PR #822, opened 2016-04-27
Updated Azure to latest stable build (v1.0.3)

JakeNarkizian:
Minor changes to the codebase

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -27,10 +27,12 @@ from datetime import datetime, timedelta
 
 from ConfigParser import RawConfigParser, NoOptionError
 
-from azure import WindowsAzureMissingResourceError, WindowsAzureError
+from azure.common import AzureMissingResourceHttpError, AzureException
+from azure.storage import SharedAccessPolicy, AccessPolicy
+from azure.storage.table import TableService, EntityProperty
+from azure.storage.blob import BlobService, BlobSharedAccessPermissions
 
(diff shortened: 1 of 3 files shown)


JakeNarkizian:
@hannes-ucsc ready

src/toil/jobStores/azureJobStore.py
```diff
@@ -713,12 +715,12 @@ def toItem(self, chunkSize=maxAzureTablePropertySize):
 def retryOnAzureTimeout(exception):
     timeoutMsg = "could not be completed within the specified time"
     busyMsg = "Service Unavailable"
-    return isinstance(exception, WindowsAzureError) and (timeoutMsg in str(exception)
-        or busyMsg in str(exception))
+    return isinstance(exception, AzureException) and (timeoutMsg in str(exception)
+                                                      or busyMsg in str(exception))
```
hannes-ucsc: wrap **before** `and` and keep the disjunction on a line. When wrapping expressions (which are trees), prefer keeping the leafs together on a line and instead wrap at higher levels in the tree.

hannes-ucsc:
Is there a ticket for this? If so, it should be referenced here. If not, the branch name should have been `features/…` instead of `issues/…` but it's ok to leave it as is for this PR. I never mentioned this explicitly as I made that convention up a few weeks ago and never told anyone.

Branch needs to be rebased and see my line notes.

JakeNarkizian:
@hannes-ucsc

Issue #760, opened 2016-04-07
Sporadic NoSuchJobException in sort tests


hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/1017/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsSingle/

hannes-ucsc:
Another one: http://jenkins.cgcloud.info/job/toil-pull-requests/1164/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsSingle/

cket:
AWS with mesos: http://jenkins.cgcloud.info/job/toil-pull-requests/1205/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsMesos/

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/lastCompletedBuild/testReport/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceWithCheckpoints/search/?q=NoSuchJobException

hannes-ucsc:
I've added a Lucene indexer to Jenkins so we can do full text search on build results:

http://jenkins.cgcloud.info/search/?q=NoSuchJobException

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-branches/branch/master/167/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsMesos/

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/1445/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsMesos/

joelarmstrong:
Hmm, I was sort of hoping the fix to #808 would fix this too, but after looking at it, it looks like it's either a race condition (exists check followed right after by a load) or stale reads from SDB. I tried to aggravate the race locally by sleeping in between but no dice.

hannes-ucsc:
I think this is a stale read after delete. We've been in touch with AWS support about this. They didn't outright reject the idea of SDB not being fully consistent and recommended that we upgrade to DynamoDB. 

I wonder is why this happens in the same code location every time.

hannes-ucsc:
… meaning that if it is a stale read, the time window should be somewhat flexible and therefore cause different problems in different places.

hannes-ucsc:
[...]

PR #823, opened 2016-04-27
Fix stale promise job store cache (resolves #817)

### LINKED ISSUE #817: Consecutive workflows in same process use wrong cached promised job store

hannes-ucsc:
Resolves #817, continued from #818.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -388,7 +388,7 @@ line::
     
 The value of *j1.rv()* is a *promise*, rather than the actual return value of the function, 
 because j1 for the given input has at that point not been evaluated. A promise
-(:class:`toil.job.PromisedJobReturnValue`) is essentially a pointer to the return value
+(:class:`toil.job.Promise`) is essentially a pointer to the return value
 that is replaced by the actual return value once it has been evaluated. Therefore when j2
 is run the promise becomes 2.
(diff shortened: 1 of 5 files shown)


hannes-ucsc:
Test failed for unrelated reasons: https://github.com/BD2KGenomics/toil/issues/760

Jenkins, test this please.

hannes-ucsc:
@benedictpaten you asked to be pinged for review: ping

PR #826, opened 2016-04-28
All values in generated function names are lowercase (resolves #825)

### LINKED ISSUE #825: Generated test method names lack intuitively cased value names
The make_tests method generates test method names by calling title() on value names  which returns a string of all lowercase letters with the exception of the first character which is capitalized e.g. ```test_generalMethodName__param1_Valuename()```. 

These generated tests cannot be run individually via clicking on their method signatures (since such methods are _generated_). Thus if one wishes to run only a single test from the set of generated tests, they must explicitly type the name of the desired method in PyCharm's configuration. This change makes this process more smooth as all value name characters are now consistently  lowercased.

JakeNarkizian:
Resolves #825 

Previously the first letter of the value name was uppercase and every subsequent letter was lowercase. e.g. ```test_generalMethodName__param1_Valuename()```. This is a bit arbitrary since value names are fairly discernible given that a single underscore separates parameter names and their respective value names. Now the same method would look like: ```test_generalMethodName__param1_valuename()```.

### DIFF
--- a/src/toil/test/__init__.py
+++ b/src/toil/test/__init__.py
@@ -317,37 +317,37 @@ def make_tests(generalMethod, targetClass=None, **kwargs):
 
     >>> b = Bar()
 
-    >>> assert b.test_has__num_One__letter_A() == b.has(1, 'a')
+    >>> assert b.test_has__num_one__letter_a() == b.has(1, 'a')
 
-    >>> assert b.test_has__num_One__letter_B() == b.has(1, 'b')
+    >>> assert b.test_has__num_one__letter_b() == b.has(1, 'b')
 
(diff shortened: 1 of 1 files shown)


JakeNarkizian:
@hannes-ucsc

hannes-ucsc:
Commit title lacks issue reference. Otherwise ready to merge.

JakeNarkizian:
jenkins, retest this please.

JakeNarkizian:
@hannes-ucsc (there were no changes to the code between this test and the preceding one)

hannes-ucsc:
Commit title should be "resolves" not "Resolves".

And rebase on latest master.

JakeNarkizian:
@hannes-ucsc

PR #830, opened 2016-04-29
Implement Google job store (resolves #605, resolves #832)

### LINKED ISSUE #605: Implement GCS job store
due end of January 2016) as a storage medium (see https://github.com/BD2KGenomics/toil/blob/master/src/toil/jobStores/abstractJobStore.py for API that needs to be implemented)

### LINKED ISSUE #832: Updating FileInfo attributes can delete file
When a FileInfo object is modified with non-inlined content and without rewriting the content itself, the file is mistakenly assumed to be outdated and deleted.

cket:
Encryption not yet supported

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,8 @@ import sys
 from version import version
 from setuptools import find_packages, setup
 
+botoVersionRequired = 'boto==2.38.0'
+
 kwargs = dict(
     name='toil',
     version=version,
@@ -33,12 +35,15 @@ kwargs = dict(
         'mesos': [
             'psutil==3.0.1'],
         'aws': [
-            'boto==2.38.0',
+            botoVersionRequired,
             'cgcloud-lib==1.4a1.dev195' ],
         'azure': [
             'azure==1.0.3'],
         'encryption': [
             'pynacl==0.3.0'],
+        'google': [
+            'gcs_oauth2_boto_plugin==1.9',
+            botoVersionRequired],
         'cwl': [
             'cwltool==1.0.20160425140546']},
     package_dir={'': 'src'},
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -384,13 +384,19 @@ class AbstractJobStore(object):
 
         def getJob(jobId):
             if jobCache is not None:
-                return jobCache[jobId]
+                try:
+                    return jobCache[jobId]
+                except KeyError:
+                    self.load(jobId)
             else:
                 return self.load(jobId)
 
         def haveJob(jobId):
             if jobCache is not None:
-                return jobCache.has_key(jobId)
+                if jobCache.has_key(jobId):
+                    return True
+                else:
+                    return self.exists(jobId)
             else:
                 return self.exists(jobId)
 
@@ -419,11 +425,23 @@ class AbstractJobStore(object):
         getConnectedJobs(self.loadRootJob())
         logger.info("%d jobs reachable from root." % len(reachableFromRoot))
 
-        # Cleanup the state of each jobWrapper
-        for jobWrapper in getJobs():
-            changed = [False]  # Flag to indicate if we need to update the jobWrapper
-            # on disk
+        # Cleanup invalid jobs
+        for jobWrapper in (x for x in getJobs() if x not in reachableFromRoot):
+            # clean up any associated files before deletion
+            for fileID in jobWrapper.filesToDelete:
+                # Delete any files that should already be deleted
+                logger.critical(
+                    "Removing file in job store: %s that was marked for deletion but not previously removed" % fileID)
+                self.deleteFile(fileID)
+            jobWrapper.filesToDelete = []
 
+        # clean up valid jobs
+        for jobWrapper in (getJob(x) for x in reachableFromRoot):
+            changed = [False]
+            # jobWrappers here are necessarily in reachable from root.
+            # While jobs at the end of the stack are already deleted remove
+            # those jobs from the stack (this cleans up the case that the jobWrapper
+            # had successors to run, but had not been updated to reflect this)
             if len(jobWrapper.filesToDelete) != 0:
                 # Delete any files that should already be deleted
                 for fileID in jobWrapper.filesToDelete:
@@ -433,14 +451,7 @@ class AbstractJobStore(object):
                 jobWrapper.filesToDelete = []
                 changed[0] = True
 
-            # Delete a jobWrapper if it is not reachable from the rootJob
-            if jobWrapper.jobStoreID not in reachableFromRoot:
-                logger.critical(
-                    "Removing job: %s that is not a successor of the root job in cleanup" % jobWrapper.jobStoreID)
-                self.delete(jobWrapper.jobStoreID)
-                continue
-
-            # For a job whose command is already execute, remove jobs from the 
+            # For a job whose command is already execute, remove jobs from the
             # stack that are already deleted. 
             # This cleans up the case that the jobWrapper
             # had successors to run, but had not been updated to reflect this
@@ -631,11 +642,13 @@ class AbstractJobStore(object):
 
     def jobs(self):
         """
-        Return an iterator with all jobs in this job store as an iterator. All valid jobs will be
-        returned, but not all returned jobs are valid. Invalid jobs are jobs that have already
-        finished running and should not be rerun. To guarantee you only get jobs that can be run,
-        construct a ToilState object instead.
+        Best effort attempt to return iterator on all jobs in the store. The iterator may not
+        return all jobs and may also contain orphaned jobs that have already finished succesfully
+        and should not be rerun. To guarantee you get any and all jobs that can be run instead
+        construct a more expensive ToilState object
 
+        :return: Returns iterator on jobs in the store. The iterator may or may not contain all jobs and may contain
+                 invalid jobs
         :rtype: Iterator[toil.jobWrapper.JobWrapper]
         """
         raise NotImplementedError()
@@ -700,14 +713,14 @@ class AbstractJobStore(object):
     def getEmptyFileStoreID(self, jobStoreID=None):
         """
         Creates an empty file in the job store and returns its ID.
-
-        :param str jobStoreID: the ID of a job, or None. If specified, the file will be associated
-               with that job and when when jobStore.delete(job) is called all files associated with
-               the given job will be removed from the job store.
+        Call to fileExists(getEmptyFileStoreID(jobStoreID)) will return True.
+        
+        :param str jobStoreID: the id of a job, or None. If specified, the file will be associated with
+               that job and when jobStore.delete(job) is called a best effort attempt is made to delete
+               all files written with the given job.jobStoreID
 
         :return: a jobStoreFileID that references the newly created file and can be used to reference the
                  file in the future.
-
         :rtype: str
         """
         raise NotImplementedError()
--- a/src/toil/jobStores/googleJobStore.py
+++ b/src/toil/jobStores/googleJobStore.py
@@ -2,7 +2,6 @@ import base64
 from contextlib import contextmanager
 import hashlib
 import os
-import random
 import uuid
 from StringIO import StringIO
 from bd2k.util.threading import ExceptionalThread
@@ -79,14 +78,12 @@ class GoogleJobStore(AbstractJobStore):
 
         super(GoogleJobStore, self).__init__(config=config)
         self.sseKeyPath = self.config.sseKey
-        # functionally equivalent to dictionary1.update(dictionary2) but preserves immutability
+        # functionally equivalent to dictionary1.update(dictionary2) but works with our immutable dicts
         self.encryptedHeaders = dict(self.encryptedHeaders, **self._resolveEncryptionHeaders())
 
         self.statsBaseID = 'f16eef0c-b597-4b8b-9b0c-4d605b4f506c'
         self.statsReadPrefix = '_'
         self.readStatsBaseID = self.statsReadPrefix+self.statsBaseID
-        self.start = 0  # tracks the starting index of available statsFileIDs
-        self.statsIDRange = 30
 
     def deleteJobStore(self):
         # no upper time limit on this call keep trying delete calls until we succeed - we can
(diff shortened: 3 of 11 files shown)


cket:
@hannes-ucsc I think google tests will be skipped until the .boto file on jenkins master is distributed to the slaves

hannes-ucsc:
I'll do an in-depth review ASAP.

src/toil/test/__init__.py
```diff
@@ -140,20 +140,41 @@ def needs_aws(test_item):
         dot_boto_path = os.path.expanduser('~/.boto')
         dot_aws_credentials_path = os.path.expanduser('~/.aws/credentials')
         hv_uuid_path = '/sys/hypervisor/uuid'
-        if (os.path.exists(dot_boto_path)
-            or os.path.exists(dot_aws_credentials_path)
-            # Assume that EC2 machines like the Jenkins slave that we run CI on will have IAM roles
```
hannes-ucsc: Why delete the comment?

src/toil/test/__init__.py
```diff
@@ -140,20 +140,41 @@ def needs_aws(test_item):
...
-            or os.path.exists(hv_uuid_path) and file_begins_with(hv_uuid_path,'ec2')):
+        try:
+            with open(dot_boto_path) as f:
+                if 'aws_access_key_id' in f.read().decode('utf8'):
+                    return test_item
+        except IOError:
```
hannes-ucsc: Should be more specific. Only errno.ENOENT is ok everything else should be reraised.

src/toil/test/__init__.py
```diff
@@ -140,20 +140,41 @@ def needs_aws(test_item):
         dot_boto_path = os.path.expanduser('~/.boto')
         dot_aws_credentials_path = os.path.expanduser('~/.aws/credentials')
         hv_uuid_path = '/sys/hypervisor/uuid'
-        if (os.path.exists(dot_boto_path)
-            or os.path.exists(dot_aws_credentials_path)
-            # Assume that EC2 machines like the Jenkins slave that we run CI on will have IAM roles
-            or os.path.exists(hv_uuid_path) and file_begins_with(hv_uuid_path,'ec2')):
+        try:
+            with open(dot_boto_path) as f:
+                if 'aws_access_key_id' in f.read().decode('utf8'):
```
hannes-ucsc: This is a bit wonky. What if 

```
# aws_access_key_id = 
```

occurs in the file? Spend like 10 min to see if you can use boto.config instead. If not, this is acceptable.

src/toil/test/__init__.py
```diff
@@ -140,20 +140,41 @@ def needs_aws(test_item):
...
+    try:
+        import boto
+    except ImportError:
+        return unittest.skip("Skipping test. Install Toil with the 'google' extra to include this "
+                             "test.")(test_item)
+    else:
+        dot_boto_path = os.path.expanduser('~/.boto')
+        try:
+            with open(dot_boto_path) as w:
+                if 'gs_access_key_id' in w.read().decode('utf-8'):
```
hannes-ucsc: See above.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -278,29 +277,28 @@ def callback(f2):
...
 
-            # Delete parent and its associated files
+            # test the readAll parameter
+            self.assertEqual(4, master.readStatsAndLogging(callback, readAll=True))
+
+            # Delete parent
             #
             master.delete(jobOnMaster.jobStoreID)
             self.assertFalse(master.exists(jobOnMaster.jobStoreID))
-            # Files should be gone as well. NB: the fooStream() methods return context managers
```
hannes-ucsc: Why delete the comment?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -377,13 +377,19 @@ def clean(self, jobCache=None):
 
         def getJob(jobId):
             if jobCache is not None:
-                return jobCache[jobId]
+                try:
+                    return jobCache[jobId]
+                except ValueError:
```
hannes-ucsc: Shouldn't this be KeyError?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -412,11 +418,24 @@ def getConnectedJobs(jobWrapper):
...
         logger.info("%d jobs reachable from root." % len(reachableFromRoot))
 
-        # Cleanup the state of each jobWrapper
-        for jobWrapper in getJobs():
-            changed = [False]  # Flag to indicate if we need to update the jobWrapper
-            # on disk
+        # Cleanup invalid jobs
+        for jobWrapper in [x for x in getJobs() if x not in reachableFromRoot]:
+            # clean up any associated files before deletion
+            if len(jobWrapper.filesToDelete) != 0:
```
hannes-ucsc: `if jobWrapper.filesToDelete:`

src/toil/jobStores/abstractJobStore.py
```diff
@@ -412,11 +418,24 @@ def getConnectedJobs(jobWrapper):
...
+                # Delete any files that should already be deleted
+                for fileID in jobWrapper.filesToDelete:
+                    logger.critical(
+                        "Removing file in job store: %s that was marked for deletion but not previously removed" % fileID)
+                    self.deleteFile(fileID)
+                jobWrapper.filesToDelete = []
 
+        # clean up valid jobs
+        for jobWrapper in [getJob(x) for x in reachableFromRoot]:
```
hannes-ucsc: Change list comprehension to generator expression.

hannes-ucsc:
Change all inline comments to begin with upper case. Change all log messages to begin with upper case and end in period.

setup.py
```diff
@@ -39,6 +39,9 @@
             'azure==1.0.3'],
         'encryption': [
             'pynacl==0.3.0'],
+        'google': [
+            'gcs_oauth2_boto_plugin==1.9',
+            'boto==2.38.0'],
```
hannes-ucsc: Can you factor this out into a variable to ensure that the `aws` and `google` extras don't use conflicting versions?

src/toil/jobStores/googleJobStore.py
```diff
@@ -0,0 +1,536 @@
...
+
+    @headerValues.setter
+    def headerValues(self, value):
+        self._headerValues = value
+
+    def __init__(self, namePrefix, projectID=None, config=None):
+        #  create 2 buckets
+        self.projectID = projectID
+
+        self.bucketName = namePrefix+"--toil"
```
hannes-ucsc: We could just remove the suffix if you are reasonably sure that there will only ever be one bucket in this implementation.

src/toil/jobStores/googleJobStore.py
```diff
@@ -0,0 +1,536 @@
...
+        create = config is not None
+        self._checkJobStoreCreation(create, exists, projectID+':'+namePrefix)
+
+        if not exists:
+            self.files = self.uri.create_bucket(headers=self.headerValues)
+
+        super(GoogleJobStore, self).__init__(config=config)
+        self.sseKeyPath = self.config.sseKey
+        # functionally equivalent to dictionary1.update(dictionary2) but preserves immutability
+        self.encryptedHeaders = dict(self.encryptedHeaders, **self._resolveEncryptionHeaders())
```
hannes-ucsc: The property takes care of the copying, right?

src/toil/jobStores/googleJobStore.py
```diff
@@ -0,0 +1,536 @@
...
+        try:
+            key.set_contents_from_file(fileObj, headers=headers)
+        except boto.exception.GSDataError:
+            # per https://cloud.google.com/storage/docs/encryption#customer
+            # -supplied_encryption_keys the etag and md5 will not match with customer supplied
+            # keys. however boto didn't get the memo apparently, and will raise this error if
+            # they dont match reported: https://github.com/boto/boto/issues/3518
+            pass
+
+    def _writeString(self, jobStoreID, stringToUpload, update=False, encrypt=True):
```
hannes-ucsc: With wrapper methods like this, I often use **kwargs so I don't have two covariant signatures to maintain.

src/toil/jobStores/googleJobStore.py
```diff
@@ -0,0 +1,536 @@
...
+            except NoSuchFileException:
+                pass
+
+    def _getKey(self, jobStoreID=None, headers=None):
+        # gets remote key, in contrast to self._newKey
+        key = None
+        try:
+            key = self.files.get_key(jobStoreID, headers=headers)
+        except boto.exception.GSDataError:
+            pass  # see self._writeFile for more
```
hannes-ucsc: Should probably elaborate here, even if it is redundant.

src/toil/jobStores/googleJobStore.py
```diff
@@ -0,0 +1,536 @@
...
+        if key is None:
+            raise NoSuchFileException(jobStoreID)
+        else:
+            return key
+
+    def _newKey(self, jobStoreID):
+        # does not create a key remotely. Careful- does not ensure key name is unique
```
hannes-ucsc: This is why I like comments to always start in upper-case. As soon as you have to sentences, it look awkward. Nit: `--` is commonly used in place of em dash.

hannes-ucsc:
Two main questions: 

Firstly, is there a limit on the size of a single-request upload? I've seen recommendations to use multipart or resumable uploads for files > 5MiB but I don't see a hard limit documented anywhere. Even if there isn't a hard limit we might still want to use multipart upload because it would make retries feasible. You can't really retry a 1GB upload if you have to start from scratch every time.

Second, how do you feel about the test coverage of the stats and logging code? That seems reasonably complex to me and might warrant a special, more aggressive test than what we currently have.

Last but not least, you should document the Google job store in the Sphinx doc. In yesterday's sprint planning we deemed Google and Azure support "experimental" and the documentation should clearly state that.

cket:
@hannes-ucsc when you get a chance can you move the google .boto file from slave-secrets in jenkin master to the slaves? I think thats why the google tests didn't run on this PR. 

As we discussed I created #871 to track multipart uploading.

I think that if we decide to stick with the current incarnation of the stats and logging we should include a more aggressive test but following @benedictpaten's comments at the sprint demo meeting it might make more sense to lose the complexity and just live with google's eventually consistent listing. This would mean that we wouldn't be able to guarantee that we will receive the stats or worker's logs but would be much simpler and efficient. 

I'm leaning towards simplifying the code - this current approach works but is noticeably slower and will likely be a pain to maintain.

hannes-ucsc:
+1 on the simplification. Maybe add a grace period of 10s at the end?

hannes-ucsc:
Do you still need .boto.prelim?

hannes-ucsc:
Added .boto to the slave deployment.

Jenkins, test this please.

src/toil/jobStores/googleJobStore.py
cket: I think we are only going to need one bucket but the suffix has been helpful to me in identifying toil buckets in job stores so I'd prefer to keep it, if possible.

hannes-ucsc: Makes sense.

src/toil/test/jobStores/jobStoreTest.py
cket: The comment originally provided context for the next assertions that showed that the job's files were deleted when the job was. Since we no longer guarantee or test that behavior the comment seemed outdated. 

The note about the streams seemed unneeded but it doesn't hurt to leave it in if it could be helpful.

src/toil/test/__init__.py
cket: I don't see a way to use boto.config to solve this issue so I'll keep the code as is.

src/toil/test/jobStores/jobStoreTest.py
hannes-ucsc: Yes, just leave the NB: because that helps explain the use of __enter__ below.

src/toil/test/__init__.py
hannes-ucsc: Can you show me what you tried? It should really be as easy as

```
from boto import config
config.get('Credentials', 'aws_access_key_id')
```

cket: My mistake, I think we are talking about different things. I was looking here 
http://boto.cloudhackers.com/en/latest/boto_config_tut.html
My understanding is that the config file is just another location to specify credentials which could also be commented out. Ill look at the config module and get back to you!

src/toil/jobStores/googleJobStore.py
cket: Yes, which is why I can't use the dict update method here -- I would just be updating a copy of the dictionary that would be thrown away. I can't use headers = dict.update(dict2) either because the update operation returns None

cket:
I don't need .boto.prelim, it was a draft version of the boto file I was testing out. Sorry, I thought I deleted it.

src/toil/test/jobStores/jobStoreTest.py
cket: Makes sense. I moved the note further down where the __enter__ is actually used -- the original usages were part of the job file deletion assertions.

cket:
Failed on unrelated Azure issues

cket:
Jenkins, test this please.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -277,21 +277,24 @@ def callback(f2):
...
             self.assertEquals({'1', '2'}, stats)
             largeLogEntry = os.urandom(self._largeLogEntrySize())
             stats = set()
-            master.writeStatsAndLogging(largeLogEntry)
+            worker.writeStatsAndLogging(largeLogEntry)
             self.assertEquals(1, master.readStatsAndLogging(callback))
             self.assertEquals({largeLogEntry}, stats)
 
+            # test the readAll parameter
+            self.assertEqual(4, master.readStatsAndLogging(callback, readAll=True))
```
cket: Due to stats simplifications, I can no longer guarantee we will always get the specified number of stats files in any of these assertions. So far I haven't seen any failures and when/if I do I can increase the grace period at the end of readStatsAndLogging to a higher value. Another alternative would be to just remove these assertions altogether but I think we should keep them in for now

benedictpaten:
I've reviewed this and am happy to merge. @cket , are you happy now?

cket:
Yes, I think it's ready to merge

Issue #547, opened 2015-11-13
Parasol, GridEngine: Ability to communicate environment variables from leader to worker


hannes-ucsc:
Follow-up on #441 and implement for GridEngine and Parasol what we did for Mesos and SingleMachine.

PR #847, opened 2016-05-05
Remove out of date doc comment

jtratner:
Was reading through docs and noticed that #547 appears to be resolved now, so the doc comment in abstractBatchSystem is wrong.

Commit 67c7cf7 appears to have fixed the underlying issue.


hannes-ucsc:
Continued in #848

PR #855, opened 2016-05-10
Fix off by factor 1024 in Mesos disk allocation (resolves #854)

### LINKED ISSUE #854: Fix disk requirement conversion for Mesos

jpfeil:
resolves #854

### DIFF
--- a/src/toil/test/batchSystems/batchSystemTest.py
+++ b/src/toil/test/batchSystems/batchSystemTest.py
@@ -19,6 +19,7 @@ from fractions import Fraction
 from inspect import getsource
 import logging
 import os
+import fcntl
 import tempfile
 from textwrap import dedent
 import time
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Great job. This is a very valuable test and I'm embarassed that we didn't think of it earlier. 

I think I want this test in toil/test/batchSystems/batchSystemTest.py. Either in a separate class or preferably in AbstractBatchSystemTest. I think at least SingleMachine and Mesos should support this test. GridEngine, Slurm and Parasol are probably too sluggish for the current sleep time. But you could factor the sleep time out into a class attribute that those three override to like 30-60 seconds.

hannes-ucsc:
I see. You already have this for Single Machine and Mesos. I still think integrating it with the other batch system tests is worth a shot.

jpfeil:
@hannes-ucsc

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -217,6 +276,20 @@ def tearDown(self):
...
 
+@needs_mesos
+class MesosBatchSystemJobTest(hidden.AbstractBatchSystemJobTest, MesosTestSupport):
+    """
+    Tests against the Mesos batch system
+    """
+
+    def getBatchSystemName(self):
+        self._startMesos(self.cpu_count)
+        return "mesos", 5
```
hannes-ucsc: Use class attribute.

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -195,6 +196,64 @@ def _waitForJobsToStart(self, numJobs):
                 runningIDs = self.batchSystem.getRunningBatchJobIDs().keys()
             return runningIDs
 
+    class AbstractBatchSystemJobTest(ToilTest):
+        """
+        A base test case with generic tests that every batch system should pass
```
hannes-ucsc: `An abstract base class for batch system tests that use a full Toil workflow rather than using the batch system directly.`

jpfeil:
Jenkins, test this please

jpfeil:
@hannes-ucsc

hannes-ucsc:
See #861 for follow up.

PR #833, opened 2016-04-29
Various Azure fixes (resolves #593, resolves #714, resolves #478, resolves #543, resolves #860)

JakeNarkizian:
Implemented retry policies and added exponential back off. Added validation check for both job store name and shared file names. Added server busy error to retry-able  exceptions.

PR #866, opened 2016-05-13
Various Azure fixes (resolves #593, resolves #543, resolves #714, resolves #478, resolves #860)

### LINKED ISSUE #478: Azure job store should validate names of shared files
See readSharedFileStream() and writeSharedFileStream() in the file and AWS implementations.

### LINKED ISSUE #543: Azure job store name isn't being validated
See #542 and the associated PR.

### LINKED ISSUE #593: Azure job store retries forever and without delay
Number of retries needs to be limited and there should be an exponential back-off in between. Just copy the logic from the AWS job store.

### LINKED ISSUE #714: Azure job store should retry on server busy errors
We got ServerBusy error during a test run. Probably in the stats aggregator thread/process.

http://jenkins.cgcloud.info/job/toil-pull-requests/885/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAzureSingle/

### LINKED ISSUE #860: Azure test doesn't use factory construction method
The Azure factory method was not added in #835 thus the tests also could not be added

JakeNarkizian:
continues #833 
resolves #593, resolves #543, resolves #714, resolves #478, resolves #860

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -84,6 +84,44 @@ class AzureJobStore(AbstractJobStore):
     A job store that uses Azure's blob store for file storage and
     Table Service to store job info with strong consistency."""
 
+    @classmethod
+    def loadOrCreateJobStore(cls, jobStoreString, config=None, **kwargs):
+        account, namePrefix = jobStoreString.split(':', 1)
+        if '--' in namePrefix:
+            raise ValueError("Invalid name prefix '%s'. Name prefixes may not contain "
+                             "%s." % (namePrefix, cls.nameSeparator))
+
+        if not cls.containerNameRe.match(namePrefix):
+            raise ValueError("Invalid name prefix '%s'. Name prefixes must contain only digits, "
+                             "hyphens or lower-case letters and must not start or end in a "
+                             "hyphen." % namePrefix)
+
+        # Reserve 13 for separator and suffix
+        if len(namePrefix) > cls.maxContainerNameLen - cls.maxNameLen - len(cls.nameSeparator):
+            raise ValueError(("Invalid name prefix '%s'. Name prefixes may not be longer than 50 "
+                              "characters." % namePrefix))
+
+        if '--' in namePrefix:
+            raise ValueError("Invalid name prefix '%s'. Name prefixes may not contain "
+                             "%s." % (namePrefix, cls.nameSeparator))
+
+        return cls(account, namePrefix, config=config, **kwargs)
+
+    # Dots in container names should be avoided because container names are used in HTTPS bucket
+    # URLs where the may interfere with the certificate common name. We use a double
+    # underscore as a separator instead.
+    #
+    containerNameRe = re.compile(r'^[a-z0-9](-?[a-z0-9]+)+[a-z0-9]$')
+
+    # See https://msdn.microsoft.com/en-us/library/azure/dd135715.aspx
+    #
+    minContainerNameLen = 3
+    maxContainerNameLen = 63
+    maxNameLen = 10
+    nameSeparator = 'xx'  # Table names must be alphanumeric
+
+    # Do not invoke the constructor, use the factory method above.
+
     def __init__(self, accountName, namePrefix, config=None, jobChunkSize=maxAzureTablePropertySize):
         self.jobChunkSize = jobChunkSize
         self.keyPath = None
@@ -122,9 +160,6 @@ class AzureJobStore(AbstractJobStore):
 
         if self.config.cseKey is not None:
             self.keyPath = self.config.cseKey
-
-    # Table names must be alphanumeric
-    nameSeparator = 'xx'
 
     # Length of a jobID - used to test if a stats file has been read already or not
     jobIDLength = len(str(uuid.uuid4()))
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -343,12 +343,14 @@ class AzureJobStore(AbstractJobStore):
 
     @contextmanager
     def writeSharedFileStream(self, sharedFileName, isProtected=None):
+        assert self._validateSharedFileName(sharedFileName)
         sharedFileID = self._newFileID(sharedFileName)
         with self._uploadStream(sharedFileID, self.files, encrypted=isProtected) as fd:
             yield fd
 
     @contextmanager
     def readSharedFileStream(self, sharedFileName):
+        assert self._validateSharedFileName(sharedFileName)
         sharedFileID = self._newFileID(sharedFileName)
         if not self.fileExists(sharedFileID):
             raise NoSuchFileException(sharedFileID)
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -23,7 +23,8 @@ import cPickle
 import socket
 import httplib
 import time
-
+from collections import Iterable, Iterator
+from typing import Callable
 from datetime import datetime, timedelta
 
 from ConfigParser import RawConfigParser, NoOptionError
@@ -122,7 +123,8 @@ class AzureJobStore(AbstractJobStore):
 
     # Do not invoke the constructor, use the factory method above.
 
-    def __init__(self, accountName, namePrefix, config=None, jobChunkSize=maxAzureTablePropertySize):
+    def __init__(self, accountName, namePrefix, config=None,
+                 jobChunkSize=maxAzureTablePropertySize):
         self.jobChunkSize = jobChunkSize
         self.keyPath = None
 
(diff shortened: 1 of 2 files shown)


JakeNarkizian:
Jenkins, retest this please.

JakeNarkizian:
@hannes-ucsc please review this first

src/toil/jobStores/azureJobStore.py
```diff
@@ -784,26 +825,48 @@ def attempt(last=False):
...
-        if num_tries == 1:
-            yield attempt(last=True)
-        else:
-            yield attempt()
+        try:
+            delay = delays[totalTries - numTries]
+        except IndexError:
+            delay = delays[-1]
+        finally:
+            if delay < 0:
```
hannes-ucsc: You should assert this earlier.

hannes-ucsc: And by assert I mean just check. Don't use the `assert` statement. And the correct exceptions type to raise would be ValueError.

src/toil/jobStores/azureJobStore.py
```diff
@@ -784,26 +825,48 @@ def attempt(last=False):
...
+                raise RuntimeError('Delay cannot be negative.')
+            totalDelay += delay
+
+        expiration = time.time() + timeout
```
hannes-ucsc: Shouldn't this be outside the loop?

src/toil/jobStores/azureJobStore.py
```diff
@@ -784,26 +825,48 @@ def attempt(last=False):
         try:
             yield
         except retriable_exceptions as e:
-            # Any instance of these exceptions is automatically retriable.
-            if last:
-                raise
+            if time.time() < expiration:
```
hannes-ucsc: It would be desirable to include the next delay in the condition. If the next delay puts us past the expiration we don't need to delay or try again.

hannes-ucsc: Please model this after https://github.com/BD2KGenomics/cgcloud/blob/9cfe6b894a7609244452e14e6f3ff90e0db8611d/lib/src/cgcloud/lib/ec2.py#L35

src/toil/jobStores/azureJobStore.py
JakeNarkizian: Where do you propose it be moved?

 Line 860 of the finally branch is the first point after which ```delay``` represents the proper value from ```delays```. Assigning ```delay``` earlier wouldn't work. 

The fastest average case I can think of ```O(n)``` if we call ```min()``` on the list see [wiki.python.org/moin/TimeComplexity](https://wiki.python.org/moin/TimeComplexity) or iterate through the list manually. If a negative number is passed into ```delays``` I think it'd be a pretty rare occurrence, not worth the extra computation.

JakeNarkizian: Actually I propose that this check be omitted entirely. ```time.sleep(n)``` will raise an error if ```n < 0``` that I think will be fairly obvious in a stack trace.

JakeNarkizian:
Jenkins, test this please.

JakeNarkizian:
Jenkins, test this please.

JakeNarkizian:
@benedictpaten if you are still reviewing any of my PR's please look at this guy first.

src/toil/jobStores/azureJobStore.py
```diff
@@ -777,33 +818,45 @@ def retry_on_error(num_tries=5, retriable_exceptions=(socket.error, socket.gaier
...
-    while go:
-        if num_tries == 1:
-            yield attempt(last=True)
-        else:
-            yield attempt()
-        # It's safe to do this, even with Python's weird default
-        # arguments behavior, since we are assigning to num_tries
-        # rather than mutating it.
-        num_tries -= 1
+    if timeout > 0:
```
benedictpaten: Woah, I find the following code really inscrutable. 

I am particularly perplexed by the while loop, which, as far as I can tell, will never break. I also dislike the use of the direct and inelegant call to the __iter__() function.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -846,6 +845,20 @@ def _cleanUpExternalStore(url):
         blobService, containerName, _ = AzureJobStore._extractBlobInfoFromUrl(urlparse.urlparse(url))
         blobService.delete_container(containerName)
 
+@needs_azure
```
benedictpaten: Are the addition of these tests sufficient for testing the issues being resolved?

benedictpaten:
@JakeNarkizian Left you some comments.

src/toil/test/jobStores/jobStoreTest.py
JakeNarkizian: The exponential back off already has some pretty comprehensive doctests. I don't think there's any functionality that is not accounted for by existing tests other than the added test here.

JakeNarkizian:
@benedictpaten I modeled the exponential back off commit after [this code](https://github.com/BD2KGenomics/cgcloud/blob/9cfe6b894a7609244452e14e6f3ff90e0db8611d/lib/src/cgcloud/lib/ec2.py#L35) per Hannes's request. The while loop terminates when ```pop()``` is called on ```go``` in the nested attempt function. However I do agree that it is difficult to understand. I've added a comment block that hopefully clears up the role of the ```go``` variable.

JakeNarkizian:
Jenkins, test this please.

src/toil/jobStores/azureJobStore.py
hannes-ucsc: @benedictpaten It is complex code, no doubt. It uses a generator that yields context managers. The complexity is localized and well-tested (via doctests). In return you get very concise usage. I recommend that you look at it not in a diff but in the actual code.

I don't understand what you mean by "direct and inelegant call to the iter() function". Can you elaborate?

src/toil/jobStores/azureJobStore.py
```diff
@@ -777,33 +818,48 @@ def retry_on_error(num_tries=5, retriable_exceptions=(socket.error, socket.gaier
...
+        delays = (i for i in delays)
```
hannes-ucsc: Just use iter(). That's what it is for.

src/toil/jobStores/azureJobStore.py
```diff
@@ -777,33 +818,48 @@ def retry_on_error(num_tries=5, retriable_exceptions=(socket.error, socket.gaier
...
+        delay = delays.next()
+        go = [None]
+        expiration = time.time() + timeout
+        while go:
+            # Recall that an empty list is False, and [None] is True.
+            # go is used so that it can be popped the attempt method.
```
hannes-ucsc: There is a word missing in this sentence.

JakeNarkizian:
@hannes-ucsc

src/toil/jobStores/azureJobStore.py
benedictpaten: I don't see why it doesn''t read something like:

try:
      for delay in delays:
              time.sleep(delay)
             rValue, stop = attempt()
             yield rValue
             if stop:
                 break
   else:
          while 1:
               time.sleep(delay)
             rValue, stop = attempt()
             yield rValue
             if stop:
                 break

Longer for sure (but could be cut with nested function), and without:

(1) The horrible side effect in attempt to pop go.
(2) The (I think), fugly use of the iterator. 

But fair enough - I can still read the code, I just don't love it ;)

benedictpaten: Ugh, it broke all my formatting.

hannes-ucsc: `attempt()` is a context manager and can't return anything.

go.pop() is needed because Python 2.7 does not have `nonlocal` and can't assign to closed over variables.

src/toil/jobStores/azureJobStore.py
```diff
@@ -777,33 +818,49 @@ def retry_on_error(num_tries=5, retriable_exceptions=(socket.error, socket.gaier
...
-        if num_tries == 1:
-            yield attempt(last=True)
-        else:
-            yield attempt()
-        # It's safe to do this, even with Python's weird default
-        # arguments behavior, since we are assigning to num_tries
-        # rather than mutating it.
-        num_tries -= 1
+    if timeout > 0:
+        delays = delays.__iter__()
```
hannes-ucsc: iter() not .__iter__

hannes-ucsc:
Just that one change. Otherwise +1 from me.

src/toil/jobStores/azureJobStore.py
JakeNarkizian: I tried that earlier and got an error, something like this 
```
>>> a = (1,2,3)
>>> a.iter()
AttributeError: 'tuple' object has no attribute 'iter'
```

hannes-ucsc: Not the `.iter()` method. The `iter()` built-in function.

JakeNarkizian:
@hannes-ucsc The build failed because of another ```GSResponseError: 429 Too Many Requests``` error. It should be good to merge anyway since this PR has nothing to do with that.

JakeNarkizian:
Jenkins, test this please.

hannes-ucsc:
I'm still not happy with the retry logic. I asked repeatedly to make it look like the one in cgcloud but this one looks still very convoluted to me. I will take it from here.

JakeNarkizian:
@hannes-ucsc I'm disappointed to hear that you are unhappy with it. What about it is convoluted? I want to make sure I do a better job next time, but I'm a bit confused.

PR #876, opened 2016-05-17
Azure and boto are now imported lazily (resolves #874)

### LINKED ISSUE #874: Job store tests import boto and Azure unconditionally
They should be imported lazily only such that they are only imported if Toil was installed with the corresponding extra.

This is high priority as it blocks other developers from running the tests.

JakeNarkizian:


hannes-ucsc:
Commit title could have given more context, e.g.

```
Import Azure and boto lazily in job store tests (resolves #874)
```

Also note how changing passive to active voice and removing the redundant `now` shortened the title.

PR #505, opened 2015-10-20
Allow job wrapping of more types of functions (resolves #264)

hannes-ucsc:

PR #879, opened 2016-05-23
Use dill to pickle job functions (resolves #264)

### LINKED ISSUE #264: Allow jobs to wrap class functions, nested functions and lambda functions

jpfeil:
continued from #505 
resolves #264

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -26,6 +26,8 @@ import logging
 import shutil
 import stat
 import inspect
+import dill
+import types
 from collections import defaultdict
 from threading import Thread, Semaphore, Event
 from Queue import Queue, Empty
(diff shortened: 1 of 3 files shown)


jpfeil:
@benedictpaten

hannes-ucsc:
I don't like how the function is always pickled but then only unpickled as a fall-back when the old mechanism fails. The pickle string may be huge and we need to defend against that. We need to thoroughly understand if we really need both mechanisms and how they interact with each other.

src/toil/job.py
```diff
@@ -1457,17 +1459,19 @@ def __init__(self, userFunction, *args, **kwargs):
...
-        #TODO: Add dill support
-        #else use indirect method
+
         self.userFunctionModule = ModuleDescriptor.forModule(userFunction.__module__).globalize()
-        self.userFunctionName = str(userFunction.__name__)
-        self._args=args
-        self._kwargs=kwargs
+        self.pickledUserFunction = dill.dumps(userFunction)
+        pickledLen = len(self.pickledUserFunction)
+        assert len(self.pickledUserFunction) < 16000, \
```
hannes-ucsc: Don't use assert for this. Assertion are for catching errors you made.

src/toil/job.py
```diff
@@ -1457,17 +1459,19 @@ def __init__(self, userFunction, *args, **kwargs):
...
+            'User function {} is too large to pickle: {}'.format(userFunction.__name__, pickledLen)
+        logger.info('{} pickled size: {}'.format(userFunction.__name__,
+                                                 pickledLen))
+        self._args = args
+        self._kwargs = kwargs
 
     def _getUserFunction(self):
-        userFunctionModule = self._loadUserModule(self.userFunctionModule)
```
hannes-ucsc: Now you've completely disabled hot deployment. This will not work.

hannes-ucsc:
We are dealing with this the wrong way around. All of the tests test the special case of running on a single machine where hot deployment isn't necessary. This causes you to draw the wrong conclusions. Like today when you completely disabled hot-deployment and the tests still magically pass. However, in the real-world that change would break Toil for everyone using Mesos, GridEngine etc.

hannes-ucsc:
@benedictpaten and I decided to post-pone this until after the 3.2 release.

hannes-ucsc:
We now have integration tests that could help validate changes to—or affecting the—hot-deployment subsystem. I can give direction wrt what a good integration test for this would look like.

DailyDreaming:
Closing this stale PR.  Especially since dill is currently implemented in Toil.  @benedictpaten @hannes-ucsc

PR #888, opened 2016-05-26
Make bad worker a non-daemon thread (resolves #661)

### LINKED ISSUE #661: Nonetype object has no attribute "terminate"
An exception generated on interpreter shutdown while running SortTest. Seems benign but the exception should still be either avoided, caught, or documented. See http://stackoverflow.com/questions/1745232/solving-thread-cleanup-on-paramiko for more on this exception

cket:


hannes-ucsc:
+1

cket:
Jenkins, test this please.

PR #910, opened 2016-05-31
CWL: new cwltool pin; raise errors in batch setup

chapmanb:
These are two small fixes from testing running cwltoil on bcbio CWL
runs:

- Does not error out on batch system shutdown if _batchSystem setup
  failed. This avoids a non-helpful assertion error and allows
  propagation of the initial error for debugging.
- Pins against a version of cwltool compatible with arvados-cwl-runner
  so users can have both in a single python environment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ kwargs = dict(
             'gcs_oauth2_boto_plugin==1.9',
             botoVersionRequired],
         'cwl': [
-            'cwltool==1.0.20160425140546']},
+            'cwltool==1.0.20160427142240']},
     package_dir={'': 'src'},
     packages=find_packages('src', exclude=['*.test']),
     entry_points={
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -762,12 +762,11 @@ class Toil(object):
         """
         Shuts down current batch system if it has been created.
         """
-        assert self._batchSystem is not None
-
-        startTime = time.time()
-        logger.debug('Shutting down batch system')
-        self._batchSystem.shutdown()
-        logger.debug('Finished shutting down the batch system in %s seconds.' % (time.time() - startTime))
+        if self._batchSystem is not None:
+            startTime = time.time()
+            logger.debug('Shutting down batch system')
+            self._batchSystem.shutdown()
+            logger.debug('Finished shutting down the batch system in %s seconds.' % (time.time() - startTime))
 
     def _assertContextManagerUsed(self):
         if not self._inContextManager:


hannes-ucsc:
Need this as two separate commits. The common.py fix needs an associated issue and more motivation. Please take a look at the other PRs for a reference on how to format commit title with ticket reference etc.

What does bumping cwltool buy us?

chapmanb:
Hannes;
Thanks for the response. To answer your questions:

- The common.py code as written masks any failures in creating a `batchSystem` (https://github.com/BD2KGenomics/toil/blob/467e5985adc087eb8abc033609f9d075828ab158/src/toil/common.py#L526). On failure, this try/finally block (https://github.com/BD2KGenomics/toil/blob/467e5985adc087eb8abc033609f9d075828ab158/src/toil/common.py#L548) executes and raises an assertion, masking the original error.The code avoids trying to shut it down if it doesn't exist, so the original exception is propagated and you can see what the actual problem was.

- With the aggressive cwltool version pinning here and in arvados-cwl-runner it's not currently possible to have both installed in the same python environment, since they pin two slightly different version. After testing against the newer version that arvados-cwl-runner pins, this bumps it to resolve that.

Hope this additional information helps. Also, trying to get a handle on all the process for contributing to Toil. Do you need separate issues plus separate PRs for these? That's two issues and two PRs for a two line fix -- just trying to help with Toil and hopefully streamline the process some.

hannes-ucsc:
@chapmanb, thank you for the additional info. There clearly is sth wrong with that code. It would help us most if you could put your reasoning in a separate issue. While we agree that there is a problem, I might disagree on your particular solution offered in this PR. With a separate issue, the community can discuss the problem and evaluate alternative solutions.

Re the pin: I'm afraid simply moving the pin is not a sustainable solution. I want to ask @tetron for his opinion, so another ticket would help here, too.

And BTW: It is possible to address more than one issue in a single PR. Just not in the same commit. That way I can revert the changes related to one issue without affecting the changes that address another issue.

hannes-ucsc:
@chapmanb, did we lose you? :-)

hannes-ucsc:
I need two tickets, one about the pin and another one about the assertion masking a primary exception. Sorry. We have so many PRs and tickets, if we don't go by the book, stuff will get dropped.

chapmanb:
Hannes -- thanks so much for getting in the fix for the batch setup problem. I followed up on the cwltool pinning issue with a new PR #961. Hope this works. Thanks again.

PR #912, opened 2016-05-31
Handle EOFError while unpicking promises (resolves #517)

### LINKED ISSUE #517: Better error needed if job is not encapsulated by the user
If the statically defined child of a job F (that should be encapsulated but isn't) is passed a return value that was obtained from a dynamically defined child of F, then and EOFError is thrown. @hannes believes the error should rather be an unfulfilled promise error.

Toy code below reproduces error. 

```
#!/usr/bin/env python2.7
'''
Author : Arjun Arkal Rao
Affiliation : UCSC BME, UCSC Genomics Institute
File : toy_code_for_unfulfilled_promise_error.py
'''
from toil.job import Job

import argparse
import time
import os

def sbe_job(job):
    '''
    This is a job that Should Be Encapsulated (SBE) for the logic to work.
    It will spawn 3 children, A, B and C that will each sleep for 5 seconds
    after which they will return their jobname.  The promises are store in dicts
    with the key=jobname and value=PromisedJobReturnValue
    '''
    return_dict = {}
    for jobname in ['A', 'B', 'C']:
        return_dict[jobname] = job.addChildJobFn(sleepy, jobname).rv()
    return return_dict


def sleepy(job, return_val):
    '''
    This function will sleep for 5 seconds and then return the value it was
    passed.
    '''
    time.sleep(5)
    return return_val


def dopey(job, test_val):
    '''
    This is a completely useless function.  It's only task is to be a child job
    of SBE_JOB.
    '''
    return None


def test_unfulfilled_promise_error_fail():
    '''
    This will run the code that fails. F should be encapsulated but it isn't.
    The DAG looks like
             F
        _____|_____ 
       /    /  \   \
      A     B   C   G
    since A, B and C are all sibling jobs and share the same parent as G.

    G is being passed the return values of A, B and C through F.rv() and since
    A, B and C sleep for 5 seconds, G starts running before they start and gets
    an EOFError when it whould really be returning an unfulfilled promise error.
    '''
    parser.add_argument('--dummy', dest='dummy', default=None, required=False)
    params = parser.parse_args()
    F = Job.wrapJobFn(sbe_job)
    G = Job.wrapJobFn(dopey, F.rv())
    F.addChild(G)


def test_unfulfilled_promise_error_pass():
    '''
    This will run the code that passes. F should be encapsulated and is.
    The DAG looks like
          _____________
         |      A      |
         |   ___|___   |
         |  /   |   \  |
         | B    C    D |
         |_____________|
                |
                X

    G is being passed the return values of A, B and C through F.rv().  This time
    A, B and C sleep finish running before G starts running since they are
    encapsulated within A -- Hence no error is thrown.
    '''
    parser.add_argument('--dummy', dest='dummy', default=None, required=False)
    params = parser.parse_args()
    F = Job.wrapJobFn(sbe_job).encapsulate()
    G = Job.wrapJobFn(dopey, F.rv())
    F.addChild(G)

if __name__ == '__main__':
    test_unfulfilled_promise_error_fail()  # This will FAIL
    test_unfulfilled_promise_error_pass()  # This will PASS
```

arkal:
resolves #517

The EOFError is caught in a  try-except and raised as a RuntimeError with an informative message.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -2044,7 +2044,13 @@ class Job(object):
                 return getattr(importlib.import_module(module_name), class_name)
 
         unpickler.find_global = filter_main
-        return unpickler.load()
+        try:
+            return unpickler.load()
+        except EOFError:
+            raise RuntimeError('EOFError encountered while unpickling a promise. Is the job '
+                               'attempting to receive a promise from a non-encapsulated parent?')
+
+
 
     def getUserScript(self):
         return self.userModule


hannes-ucsc:
I've seen these EOFErrors in more situations than the one that's suggested by the new error message. So we are exchanging a generic but nondescript error message with another one that might mislead the user  into just one particular scenario.

arkal:
This was the only scenario I've seen it in... I'll try and look through old tickets for other mentions of such an error

hannes-ucsc:
I'm still rather unclear as to why an unresolved promise results in EOF during unpickling.

Can you establish whether unresolved promises are the only reason for EOF during unpickling?

Can we prevent the EOF altogether?

arkal:
When a promise is set up, an empty file is created in the job store that is filled in when the promise is fulfilled. Attempting to unpickle an empty file gives an EOFError

```
>>> import cPickle
>>> open('/tmp/poo', 'w').close()
>>> with open('/tmp/poo') as poo:
...   cPickle.load(poo)
... 
  File "<stdin>", line 2, in <module>
EOFError
```

@jpfeil is this correct?

hannes-ucsc:
If that is the case, we should not store an empty file but a file with a pickled dummy value that we can test for after we deserialize it. Then we won't have to guess what went wrong.

jvivian:
The other case I've seen this in was when the root job failed, but I believe @cket fixed that a while ago.

hannes-ucsc:
@jpfeil, could you reply to @arkal's question?

I'm strongly leaning to pickling a special sentinel value for unresolved promises rather than an empty file.

@benedictpaten, do you have an opinion on this?

Issue #908, opened 2016-05-31
NoSuchJobException in service tests


hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-branches/branch/master/159/testReport/junit/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceWithCheckpoints/

```
NoSuchJobException: The job 'J/X/jobWhAf5O' does not exist
```

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/1367/testReport/junit/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceParallelRecursive/

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-branches/branch/master/162/testReport/junit/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceWithCheckpoints/

hannes-ucsc:
Also see #760.

PR #915, opened 2016-05-31
Document __name__ == '__main__' boilerplate in user scripts (resolves #617)

### LINKED ISSUE #617: Documentation examples don't explain need for if name=="__main__" idiom
very low priority

The toil examples waste mental bandwidth and continues to brainwash young minds into thinking programming means cut-n-paste instead of actually understanding what you are doing.

This is done by repeating the pointless python idiom
   if __name__=="__main__":

Which is only use when a single file is both a python module and a python command line program.  This is almost never a useful thing to do in well structured code, yet people just copy and paste this idiom without thinking.

Please, the future of humanity, or at least the part that programs, depends on leading by example.

hannes-ucsc:
Resolves #617

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ __pycache__
 /test-report.xml
 /venv/
 /src/toil/test/cwl/spec
+/docs/generated_rst/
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Test failure tracked as #908.

PR #961, opened 2016-06-09
Synchronize cwltool version with arvados-cwl-runner

chapmanb:
Moves cwltool version pin to match that used in arvados-cwl-runner so
both can exist in the same python environment. This has been tested with
full bcbio generated CWL runs on arvados-cwl-runner and cwltoil.
Discussed with @tetron who recommended synchronizing until we move to
semantic versioning on cwltool. Follows up on #910.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
             'gcs_oauth2_boto_plugin==1.9',
             botoRequirement],
         'cwl': [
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
+1 from me as long as build succeeds.

benedictpaten:
Jenkins, test this please.

PR #962, opened 2016-06-09
Make successorCounts map of jobStoreIds to successors (resolves #958).

### LINKED ISSUE #958: Jobs with totally-failed successors will not wait for other successors to finish as intended
This line in leader.py is broken and I'm pretty sure it will never evaluate to true:
https://github.com/BD2KGenomics/toil/blob/1195effda167eea9d1d623f2de0794c138de75b9/src/toil/leader.py#L756

The `toilState.successorCounts` dict is keyed by JobWrappers, not jobStoreIDs.

(The `__hash__` method of JobWrapper gives the same hash as its jobStoreID, but that only determines its bucket in the dict. `__eq__` is still used to compare items with the same hash value.)

I think the only side effect is that jobs with totally failed successors will terminate early as totally failed, rather than waiting for the successors to finish.

The fix should be trivial, let me know if you want it as a PR or to fix on your own.

benedictpaten:


hannes-ucsc:
+1

Issue #960, opened 2016-06-09
AssertionError in testMultipleJobsReadSameCacheHitGlobalFile


hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/1410/

hannes-ucsc:
It just occurred to me that toil-jenkins-slave slaves are shared by toil-scripts and toil. It is possible that toil-scripts pollutes the slave.

arkal:
To the order of 70G? I find that a little hard to believe

hannes-ucsc:
This is caused by running on a slave were for some reason ephemeral wasn't mounted. I'm leaving this open because I don't know why ephemeral is only mounted sometimes. The AMI has a suitable block device mapping. The useEphemeralDevices for the slave was set to false in config.xml. I changed that to true for all slaves. We'll have to keep an eye out for it.

See also https://github.com/BD2KGenomics/cgcloud/issues/184

hannes-ucsc:
We should also talk about detecting lack of temporary space in jenkins.sh, the Makefile, or in a special unit test.

arkal:
Wow! How did you find this?! Awesome. Can we have a line in jenkins.sh that checks for (say) 20G of space on ephemeral before starting any tests?

hannes-ucsc:
Yes, we should.

PR #965, opened 2016-06-11
Add support for hot-deploying virtualenvs (resolves #937)

### LINKED ISSUE #937: Add ability to hot-deploy entire virtualenvs
This is needed to support user distributions with dependencies. A user distribution is a distribution  with user modules. A distribution is a collection of modules, created with distutils/setuptools. The most prominent user distribution is toil-scripts. It may soon have a dependency (toil-lib, #922) and that will require this feature.

hannes-ucsc:
Resolves #937

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -32,12 +32,16 @@ def toilPackageDirPath():
     return result
 
 
+def inVirtualEnv():
+    return hasattr(sys, 'real_prefix')
+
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
See https://github.com/BD2KGenomics/toil/issues/960#issuecomment-225341663 about build failures.

Jenkins, test this please.

PR #971, opened 2016-06-14
Improved getJobStoreClasses (resolves #969, resolves #970)

### LINKED ISSUE #969: Google job store missing from getJobStoreClasses
```getJobStoreClasses``` returns a list of job store classes for which the dependancies have been installed. The names of each job store class are hard coded into the method [see](https://github.com/BD2KGenomics/toil/blob/master/src/toil/jobStores/abstractJobStore.py#L94). The ```GoogleJobStore``` class will never be included in this list because it is missing from the aforementioned hard coding. 

As a result calling ```importFile``` on a google-only supported url, ```<url> ``` will result in a runtime error with the following message:
```No existing job store supports importing for URL '<url>'```

### LINKED ISSUE #970: getJobStoreClasses log message is not descriptive enough
If a job store class cannot be imported the following messaged is logged: 
```toil.jobStores.abstractJobStore: Unable to import 'toil.jobStores.azureJobStore'```

This message should be more descriptive.

### PR DESCRIPTION
**Pull request by JakeNarkizian**

### CODE REVIEW

File: src/toil/jobStores/abstractJobStore.py [hunk 1]
```diff
@@ -94,18 +94,18 @@ def getJobStoreClasses():
...
-
     jobStoreClasses = []
     for className in jobStoreClassNames:
         moduleName, className = className.rsplit('.', 1)
         from importlib import import_module
         try:
             module = import_module(moduleName)
         except ImportError:
-            logger.info("Unable to import '%s'" % moduleName)
+            logger.info("Unable to import '%s'. This may be because its dependencies are not installed." % moduleName)
```
hannes-ucsc: You probably want to mention the `extra` mechanism here. And please wrap your code at column 100.

hannes-ucsc:
Commit title is too long. You may want to split commits per issue.

hannes-ucsc:
Here's what I would have used for the commit message:

```
Added Google to list of job stores to import from (resolves #969)
Improve log message on job store module import failure (resolves #970)
```

PR #972, opened 2016-06-14
Attempting to fail UtilsTest after job completion

arkal:
but before job cleanup.


hannes-ucsc:
This was used to confirm the cause for #973. This was never intended to be merged.

Issue #978, opened 2016-06-15
SSLError during SDB requests


hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-pull-requests/1462/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsMesos/

Needs retry.

arkal:
This has exactly the same stacktrace
http://jenkins.cgcloud.info/job/toil-pull-requests/1465/testReport/junit/src.toil.test.sort.sortTest/SortTest/testAwsMesos/

and this one is similar, but crashes with the same SSLError
http://jenkins.cgcloud.info/job/toil-pull-requests/1465/testReport/junit/src.toil.test.src.jobCacheTest/AwsJobStoreCacheTest/testWriteNonLocalFileToJobStore/

hannes-ucsc:
I fear SDB is going bonkers.

PR #980, opened 2016-06-15
Added return statement to retry_azure (resolves #979)

### LINKED ISSUE #979: retry_azure is broken
There is a missing return statement. [See](https://github.com/BD2KGenomics/toil/blob/master/src/toil/jobStores/azureJobStore.py#L789)

JakeNarkizian:


hannes-ucsc:
Build failure tracked as #978

PR #981, opened 2016-06-15
Retry certain SSL errors (fixes #978)

### LINKED ISSUE #978: SSLError during SDB requests

hannes-ucsc:
This is based on the assumption that those errors are intermittent and caused by the server side. Alternatively, there is the possibility that they are caused by a threading issue or a connection leak. In the latter case, retrying would aggravate the situation.

### DIFF
--- a/src/toil/jobStores/aws/utils.py
+++ b/src/toil/jobStores/aws/utils.py
@@ -20,6 +20,8 @@ import logging
 import types
 
 import errno
+from ssl import SSLError
+
 from boto.exception import (SDBResponseError,
                             BotoServerError,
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Had to abort build because we were running out of domains.

hannes-ucsc:
Jenkins, test this please.

arkal:
+1,000,000 if this fixes it!

PR #983, opened 2016-06-16
Added google to getJobStoreClasses (resolves #969)

### LINKED ISSUE #969: Google job store missing from getJobStoreClasses

### LINKED ISSUE #970: getJobStoreClasses log message is not descriptive enough

JakeNarkizian:


JakeNarkizian:
continues in #971

hannes-ucsc:
This was opened in error. It does not continue on #971.

PR #990, opened 2016-06-21
Modified copyright in license header (resolves #989)

### LINKED ISSUE #989: Copyright header should refer to `Regents of the University of California`
… because we have multiple UCs contributingt to Toil.

See also https://github.com/BD2KGenomics/toil-scripts/pull/311

fnothaft:


hannes-ucsc:
Ran out of domains/buckets.

Jenkins, test this please.

version.py
```diff
@@ -1 +0,0 @@
-src/toil/version.py
```
hannes-ucsc: This must be a symlink. Looks like you replaced the symlink with a copy?

fnothaft: Yeah, not sure how that happened. Let me patch this up tomorrow.

fnothaft: I think a similar thing may have happened on https://github.com/BD2KGenomics/toil-scripts/pull/311 as well.

hannes-ucsc:
Continues in #999

PR #1013, opened 2016-06-25
Removed registry from Azure job store (resolves #975)

### LINKED ISSUE #975: Remove registry from Azure job store
The AWS job store maintains a registry due to eventual consistency issues. @joelarmstrong do you recall why Azure also maintains a registry? If there does not exist a compelling reason it may be worth forgoing the cost of writing to a reading from an extra table in Azure. 

Instead the existence of a job store can be determined by probing for some component of it that is eventually consistent.

JakeNarkizian:
A new jobStoreExists method is used to test for existence. A unit test has been added for the new method.

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -139,12 +139,8 @@ class AzureJobStore(AbstractJobStore):
         self.tableService = TableService(account_key=self.account_key, account_name=accountName)
         self.blobService = BlobService(account_key=self.account_key, account_name=accountName)
 
-        # Register our job-store in the global table for this storage account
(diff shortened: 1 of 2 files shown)


src/toil/jobStores/azureJobStore.py
```diff
@@ -225,13 +221,28 @@ def delete(self, jobStoreID):
...
 
     def deleteJobStore(self):
-        self.registryTable.delete_entity(row_key=self.namePrefix)
         self.jobItems.delete_table()
         self.jobFileIDs.delete_table()
         self.files.delete_container()
         self.statsFiles.delete_container()
         self.statsFileIDs.delete_table()
 
+    def jobStoreExists(self):
```
hannes-ucsc: This should be a protected method.

PR #1015, opened 2016-06-25
Add attributes to return value of writeGlobalFile (resolves #1014)

### LINKED ISSUE #1014: Give attributes to writeGlobalFile's return value
Start with size. So we can use it in promise lambdas

arkal:
resolves #1014

the return value of writeGlobalFile is now a class  jobStoreFile that has an
__repr__ magic method returning the jobStoreFIleID string, and has a class
variable size that describes the size of the file in bytes. Naive implementation
at the moment.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -660,6 +660,7 @@ class Job(object):
             If not the global file must be deleted manually.
 
             :returns: an ID that can be used to retrieve the file.
+            :rtype: FileID
             """
             #Put the file into the cache if it is a path within localTempDir
             absLocalFileName = os.path.abspath(localFileName)
@@ -679,7 +680,7 @@ class Job(object):
             else:
                 #Write the file directly to the file store
                 jobStoreFileID = self.jobStore.writeFile(localFileName, cleanupID)
-            return jobStoreFileID
+            return FileID.forPath(jobStoreFileID, absLocalFileName)
 
         def writeGlobalFileStream(self, cleanup=False):
             """
@@ -692,12 +693,14 @@ class Job(object):
             The yielded file handle does not need to and should not be closed explicitly.
             """
             #TODO: Make this work with the caching??
+            #TODO: Make this work with FileID
             return self.jobStore.writeFileStream(None if not cleanup else self.jobWrapper.jobStoreID)
 
         def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=None):
             """
             Get a copy of a file in the job store.
 
+            :param FileID fileStoreID: The unique job store ID for the file
             :param string userPath: a path to the name of file to which the global \
             file will be copied or hard-linked (see below).
 
@@ -1086,6 +1089,7 @@ class Job(object):
             If not the global file must be deleted manually.
 
             :returns: an ID that can be used to retrieve the file.
+            :rtype: FileID
             """
             absLocalFileName = self._abspath(localFileName)
             # What does this do?
@@ -1145,7 +1149,7 @@ class Job(object):
                 # Non local files are NOT cached by default, but they are tracked as local files.
                 self._JobState.updateJobSpecificFiles(self, jobStoreFileID, None,
                                                       0.0, False)
-            return jobStoreFileID
+            return FileID.forPath(jobStoreFileID, absLocalFileName)
 
         def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=None):
             """
@@ -1159,7 +1163,7 @@ class Job(object):
             The cache parameter will be used only if the file isn't already in the cache, and
             provided user path (if specified) is in the scope of local temp dir.
 
-            :param fileStoreID: file store id for the file
+            :param FileID fileStoreID: job store id for the file
 
             :param string userPath: a path to the name of file to which the global \
             file will be copied or hard-linked (see below).
@@ -1851,6 +1855,7 @@ class Job(object):
                     # Based on answer by user "Mark" at:
                     # http://stackoverflow.com/questions/2709800/how-to-pickle-yourself
                     # We can't pickle nested classes. So we have to pickle the variables of the class
+                    # If we ever change this, we need to ensure it doesn't break FileID
                     cPickle.dump(self.__dict__, fH, cPickle.HIGHEST_PROTOCOL)
                 os.rename(fileName + '.tmp', fileName)
 
@@ -3026,3 +3031,24 @@ class PromisedRequirement(object):
                 foundPromisedRequirement = True
         return foundPromisedRequirement
 
+
+class FileID(str):
+    """
+    A class to wrap the job store file id returned by writeGlobalFile and any attributes we may want
+    to add to it.
+    """
+    def __new__(cls, fileStoreID, *args):
+        return super(FileID, cls).__new__(cls, fileStoreID)
+
+    def __init__(self, fileStoreID, size):
+        super(FileID, self).__init__(fileStoreID)
+        self.size = size
+
+    @classmethod
+    def forPath(cls, fileStoreID, filePath):
+        size = cls._resolveFileSize(filePath)
+        return cls(fileStoreID, size)
+
+    @staticmethod
+    def _resolveFileSize(filePath):
+        return os.stat(filePath).st_size
--- a/src/toil/test/src/promisedRequirementTest.py
+++ b/src/toil/test/src/promisedRequirementTest.py
@@ -103,6 +103,54 @@ class hidden:
         def testJobConcurrency(self):
             pass
 
+        def testPromisesWithJobStoreFileObjects(self):
+            """
+            Check whether FileID objects are being pickled properly when used as return
+            values of functions.  Then ensure that lambdas of promised FileID objects can be
+            used to describe the requirements of a subsequent job.  This type of operation will be
+            used commonly in Toil scripts.
+            :return: None
+            """
+            file1 = 1024
+            file2 = 512
+            F1 = Job.wrapJobFn(_writer, file1)
+            F2 = Job.wrapJobFn(_writer, file2)
+            G = Job.wrapJobFn(_follower, file1+file2,
+                              disk=PromisedRequirement(lambda x, y: x.size + y.size,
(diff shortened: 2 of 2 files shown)


hannes-ucsc:
I don't fully understand why we are doing this. Anyways, we never really told anyone what those IDs are. We kinda a did when we wrote the docstrings, because we typed them as `str`. But I think it its still ok to assume that no one is relying on the fact that IDs are strings. We should definitely change the docstrings (see #1018). With that in mind, if we are doing this–and that's a big if, it'd be better if we didn't emulate or extend `str` and just refactored the codebase to always use `str()` before using and ID as a string.

arkal:
I was initially inheriting object and trying the typecasting to str before doing an operation on it. There's way too much code to touch for that to be viable. That's why I rolled back to inheriting string and making it work. Really, it's working in every case except 2 of the 61 CWL conformance tests and I really have no idea what those tests are doing because the output is not very verbose.

I'm doing this so we don't bake file size into the output of every tool in our toil scripts/protect libraries

arkal:
Jenkins, test this please.

arkal:
@hannes-ucsc this is ready for a review

src/toil/job.py
```diff
@@ -1522,7 +1526,7 @@ def encodedFileID(self, JobStoreFileID):
             :return: str outCachedFile: A path to the hashed file in localCacheDir
             '''
             outCachedFile = os.path.join(self.localCacheDir,
-                                         base64.urlsafe_b64encode(JobStoreFileID))
+                                         base64.urlsafe_b64encode(str(JobStoreFileID)))
```
arkal: redundant

src/toil/job.py
```diff
@@ -3026,3 +3030,32 @@ def convertPromises(kwargs):
                 foundPromisedRequirement = True
         return foundPromisedRequirement
 
+
+class JobStoreFile(str):
+    """
+    A class to wrap the job store file id returned by writeGlobalFile and any attributes we may want
+    to add to it.
+
+    It is easier to implement a startswith method here, than to inherit string because that makes
```
arkal: stale comment

src/toil/job.py
```diff
@@ -3026,3 +3030,32 @@ def convertPromises(kwargs):
...
+    pickling and unpickling really hard.
+    """
+    # str isa  new-style class and we should modify __new__ instead of __init__
+    @classmethod
+    def __new__(cls, *args, **kwargs):
```
hannes-ucsc: Checkout the Promise class. I think this can be simplified by overriding __reduce__ and __new__ instead of __new__ and __getnewargs__. I don't like that class is passed twice. Also, omit **kwargs and use super() instead of hard-coding `str`.

src/toil/job.py
```diff
@@ -3026,3 +3030,32 @@ def convertPromises(kwargs):
...
+        # args becomes (<class>, fileStoreID, filePath)
+        obj = str.__new__(cls, args[1])
```
hannes-ucsc: s/obj/self

src/toil/job.py
```diff
@@ -3026,3 +3030,32 @@ def convertPromises(kwargs):
...
+        _, obj.fileStoreID, obj.filePath = args
+        assert isinstance(obj.fileStoreID, (str, unicode))
+        if obj.filePath is not None:
+            # This is the case when we are unpickling.  At this point, size is already in __dict__,
+            # moreover it probably doesn't even exist anymore.
+            assert isinstance(obj.filePath, (str, unicode))
+            assert os.path.exists(obj.filePath)
```
hannes-ucsc: Why? Stat will raise IOError which is more informative.

hannes-ucsc: Also, pull this out into factory method `JobStoreFile.forPath(path)` such that `__new__` doesn't mix all these concerns. `__new__` takes care of construction and unpickling, `__reduce__` takes care of pickling.

src/toil/job.py
```diff
@@ -3026,3 +3030,32 @@ def convertPromises(kwargs):
                 foundPromisedRequirement = True
         return foundPromisedRequirement
 
+
+class JobStoreFile(str):
```
hannes-ucsc: Rename to JobStoreFileID, don't you think?

hannes-ucsc:
Some scary sh*t going on here, but ok. We probably want @benedictpaten's opinion, too.

src/toil/job.py
arkal: i was thinking of that and settled on this. I could change to that if you'd like

arkal:
Haha! I'll work on your comments till @benedictpaten takes a look at what it's all about.
I really want this in asap so i can get protects CI working

arkal:
@hannes-ucsc i implemented the suggested changes. I think this is a much nicer implementation. @benedictpaten this is ready for you to look at!

src/toil/job.py
hannes-ucsc: To qualify for the name JobStoreFile, I should be able to do more things with it. This still is an ID—a glorified one, sure—but the name will conflict if we actually want create an full-blown class that wholly represents a file in the job store.

I also dislike that the "jobStore" part leaks to the user. The job store should be hidden. In the context of the file store I would have preferred to refer to file IDs as just fileID. This naming choice solidifies that mistake.

How about FileID or FileRef or FileInfo?

src/toil/job.py
```diff
@@ -3026,3 +3031,22 @@ def convertPromises(kwargs):
...
+    """
+    A class to wrap the job store file id returned by writeGlobalFile and any attributes we may want
+    to add to it.
+    """
+    # Need to modify this so this subclass of the mutable str can accept two arguments.
+    @classmethod
+    def __new__(cls, *args):
+        # args is (<class>, fileStoreID, filePath) during instantiation, and (<class>, fileStoreID)
+        # during unpickling.  We care about the fileStoreID.
+        return str.__new__(cls, args[1])
```
hannes-ucsc: You really should use super() here.

Also I don't think this is needed anymore since you don't have custom serialization anymore.

src/toil/job.py
```diff
@@ -3026,3 +3031,22 @@ def convertPromises(kwargs):
...
+
+    def __init__(self, fileStoreID, filePath):
```
hannes-ucsc: The constructor should have no intelligence, i.e. be dumb. That will give you more freedom to construct objects of this class in different ways. You can tell that there is sth wrong with this approach because it smells: the bad smell is that _resolveFileSize sets an instance attribute. PyCharm actually warns you about that.

The signature of the constructor should be `__init__(self, fileID, size)`. Then add a class method that has this signature: `forPath(cls, fileID, path)`. Construction then reads like FileInfo.forPath().

hannes-ucsc:
Concern: Files in the job store are mutable, if we persist their size through an independent channel (pickling these new magic strings) there is a chance that they become out of sync. How do we deal with that?

src/toil/job.py
arkal: I like FileID

src/toil/job.py
arkal: It is! I added a comment explaining why. If we don't do this, str can't take 2 arguments

src/toil/job.py
arkal: ah now i see what you did there

arkal:
Are user-written files mutable to the user? I don't this the user is exposed to updateFile, etc. Files written by Toil bypass writeGlobalFile and directly use writeFile, readFile, updateFile, etc so it's not relevant

arkal:

src/toil/job.py
hannes-ucsc: You are right. But can you tweak this a liitle bit and add doctests:

```
class FancyString(str):
    """
    >>> f = FancyString(1,'x')
    >>> f
    '1'
    >>> f.extra
    'x'
    """

    def __new__(cls, obj, *args):
        return super(FancyString, cls).__new__(cls, obj)

    def __init__(self, obj, extra):
        super(FancyString, self).__init__(obj)
        self.extra = extra
```

hannes-ucsc: The @classmethod/@staticmethod decorator is unnecessary and I prefer calling out the initial argument to super's __new__. Your use of mutable makes no sense in this context. `str` instances aren't mutable.

hannes-ucsc:
We overlapped, my comments were clobbered:

https://github.com/BD2KGenomics/toil/pull/1015#discussion_r68808973
https://github.com/BD2KGenomics/toil/pull/1015#discussion_r68809316

src/toil/job.py
arkal: i like the tweak. I vaguely remember reading that the reason we had to do that __new__ hack had something to do with mutability. Then again I was running on 3 hours of sleep and was still awake working at 7:00am

arkal:

src/toil/job.py
```diff
@@ -3026,3 +3031,24 @@ def convertPromises(kwargs):
...
+        self.size = size
+
+    @classmethod
+    def forPath(cls, fileStoreID, filePath):
+        size = cls._resolveFileSize(filePath)
+        return cls(fileStoreID, size)
+
+    @staticmethod
+    def _resolveFileSize(filePath):
+        return os.stat(filePath).st_size
```
hannes-ucsc: newline missing at EOL

hannes-ucsc:
Just the NL missing at EOL. Otherwise, great.

arkal:

src/toil/job.py
```diff
@@ -3026,3 +3031,24 @@ def convertPromises(kwargs):
...
+    def __init__(self, fileStoreID, size):
+        super(FileID, self).__init__(fileStoreID)
+        self.size = size
+
+    @classmethod
+    def forPath(cls, fileStoreID, filePath):
+        size = cls._resolveFileSize(filePath)
+        return cls(fileStoreID, size)
+
+    @staticmethod
```
hannes-ucsc: I didn't catch this. Separating this out into a method is overkill. Can you fix this in your next PR?

arkal:
Haha! I'll do it right now

arkal:
#1024

PR #1025, opened 2016-06-29
Cwltoil restart (resolves #1241)

### LINKED ISSUE #1241: `cwltoil --restart` does not work

tetron:
Resolves #1241 

Support --restart in cwltoil.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -506,8 +506,8 @@ cwltool.process.supportedProcessRequirements = ("DockerRequirement",
 def main(args=None, stdout=sys.stdout):
     parser = ArgumentParser()
     Job.Runner.addToilOptions(parser)
-    parser.add_argument("cwltool", type=str)
-    parser.add_argument("cwljob", type=str)
+    parser.add_argument("cwltool", type=str, nargs='?', default='')
+    parser.add_argument("cwljob", type=str, nargs='?', default='')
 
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
You might want to add a test for `--restart`. Is it possible to force a task to fail once in CWL?

benedictpaten:
@tetron Can you revise this please?

hannes-ucsc:
@chapmanb do you feel like taking this over? It looks like a straightforward change and we're open to merging it. It just needs a test.

chapmanb:
Hannes -- I could definitely use this functionality and saw the PR but am stuck at thinking of a good way to get in a test for this. Would it work to run a pipeline twice and the second `--restart` run should skip all operations? If so, I can work with Peter to get this up to date with all of the latest changes.

hannes-ucsc:
Brad, yes, sort of. A minimal test should invoke a workflow three times. During the first invocation, one workflow step should consistently fail. It will be retried during the invocation but since it always fails it should cause the invocation to fail. The second invocation should be with `--restart`, and during this invocation the workflow step should not fail. The second invocation will consequently succeed. The a third invocation, again with `--restart`, should fail since the second one already succeeded and the job store should have been clean up.

This is pretty straight-forward I think, I just don't know how to make a CWL workflow step fail controllably.

hannes-ucsc:
Continued in #1107

PR #1032, opened 2016-06-30
Remove legacy FileStore (resolves #1029)

### LINKED ISSUE #1029: k thx bye Job.FileStore
remove job.FileStore and retain just CachedFileStore

arkal:
resolves #1029

Removed legacy Job.FileStore.
Removed all mentions of the legacy cache requirement.
Removed --useSharedCache flag.
Modified worker.py to remove all legacy FileStore operations.
Renamed Job.CachedFileStore to Job.FileStore

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -82,9 +82,7 @@ class Config(object):
         self.defaultMemory = 2147483648
         self.defaultCores = 1
         self.defaultDisk = 2147483648
-        self.disableSharedCache = False
         self.readGlobalFileMutableByDefault = False
-        self.defaultCache = self.defaultDisk
         self.defaultPreemptable = False
         self.maxCores = sys.maxint
         self.maxMemory = sys.maxint
@@ -191,9 +189,7 @@ class Config(object):
         setOption("defaultMemory", h2b, iC(1))
         setOption("defaultCores", float, fC(1.0))
         setOption("defaultDisk", h2b, iC(1))
-        setOption("disableSharedCache")
         setOption("readGlobalFileMutableByDefault")
-        setOption("defaultCache", h2b, iC(0))
         setOption("maxCores", int, iC(1))
         setOption("maxMemory", h2b, iC(1))
         setOption("maxDisk", h2b, iC(1))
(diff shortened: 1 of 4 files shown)


arkal:
@hannes-ucsc this is ready!

arkal:
@hannes-ucsc this is ready!

src/toil/common.py
```diff
@@ -661,10 +650,10 @@ def createBatchSystem(config, jobStore=None, userScript=None):
         else:
             raise RuntimeError('Unrecognised batch system: %s' % config.batchSystem)
 
-        if not (config.disableSharedCache or batchSystemClass.supportsWorkerCleanup()):
-            raise RuntimeError('%s currently does not support shared caching.  Use the '
-                               '--disableSharedCache option when running Toil with this batch '
-                               'system.' % config.batchSystem)
+        if not batchSystemClass.supportsWorkerCleanup():
+            raise RuntimeError('%s currently does not support shared caching.  Use Toil versions'
+                               '< 3.2.0 (Legacy Toil) if you want to use this batch system.' %
```
hannes-ucsc: `Use Toil version 3.1.6 along with the --disableSharedCache if you want to …`

The notion of what is "legacy" will change over time.

src/toil/job.py
```diff
@@ -1076,7 +597,45 @@ def open(self,job):
...
+            :rtype: string
+            """
+            handle, tmpFile = tempfile.mkstemp(prefix="tmp",
+                                               suffix=".tmp", dir=self.localTempDir)
+            os.close(handle)
+            return os.path.abspath(tmpFile)
+
+        def getLocalTempFileName(self):
+            '''
+            Get a valid name for a new local file. Do it in a really stupid way by creating and then
```
hannes-ucsc: Move the caveat out of the docstring to a `#` comment.

hannes-ucsc: Also, I don't think there is anything stupid about the approach.

hannes-ucsc:
Only two nitpicks. Otherwise ok.

src/toil/job.py
```diff
@@ -1234,7 +807,7 @@ def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=None):
                     # either the file was successfully downloaded and added to cache, or something
                     # failed. To prevent code duplication, we recursively call readGlobalFile.
                     flock(lockFileHandle, LOCK_UN)
-                    return self.readGlobalFile(fileStoreID, userPath, cache)
+                    return self.readGlobalFile(fileStoreID, userPath, cache, mutable)
```
hannes-ucsc: Optional arguments must always be called by name. I hope you don't do this all over the place!

The reason is that optional arguments should be order-insensitive. When modifying a function, I should be free to reorder the optional arguments with affecting existing call sites of that function. PyCharm doesn't warn you about this?

hannes-ucsc:
And another one.

src/toil/job.py
arkal: it didn't. maybe i've turned off that inspection?

hannes-ucsc: PEP-8 doesn't mention it. I think I still have a valid point.

arkal: i agree. Made the fix

arkal:

PR #1034, opened 2016-06-30
Various smallish bug fixes

### LINKED ISSUE #631: Toil script as a program generates assertion

### LINKED ISSUE #996: AWS job store's `.exist()` is slower than needed
… because it fetches all attributes of the SDB item. Passing attribute_name=[] turns out to be ineffective. Discovered by @cket.

Every item should have the '000' attribute, so we should limit the set of requested attributes to that.

### LINKED ISSUE #1021: Cannot index promise if job returns another promise
If I have 4 chained functions:

A -> B -> C -> D

Where A returns `job.AddChildJobFn(B).rv()` and so forth until D returns a tuple ('foo', 'bar'),

```
a = job.wrapJobFn(A)
a.rv(0)
```
will return an error as it attempts to return the index of A, which is a promise, instead of the tuple in D.  @hannes-ucsc knows where in the code this bug originates.

### LINKED ISSUE #1023: Cannot index promise using dictionary keys
If a job function returns a dictionary, the promise cannot be indexed using a dictionary key: i.e.   `bqsr.rv('bam')`, but indexing a tuple or list works fine: `bqsr.rv(1)`.

I'll try and get a code example and stack trace up when I have time.

hannes-ucsc:
Resolves #631
Resolves #1021
Resolves #996 
Resolves #1023

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     author_email='benedict@soe.usc.edu',
     url="https://github.com/BD2KGenomics/toil",
     install_requires=[
-        'bd2k-python-lib==1.13.dev21',
+        'bd2k-python-lib==1.14a1.dev28',
         'dill==0.2.5'],
     extras_require={
         'mesos': [
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -527,7 +527,7 @@ class Toil(object):
             rootJobReturnValueID = self._jobStore.getEmptyFileStoreID()
 
             # Add the root job return value as a promise
-            rootJob._rvs[None].append(rootJobReturnValueID)
+            rootJob._rvs[()].append(rootJobReturnValueID)
 
             # Write the name of the promise file in a shared file
             with self._jobStore.writeSharedFileStream("rootJobReturnValue") as fH:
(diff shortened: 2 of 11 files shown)


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Ran out of buckets and domains.

hannes-ucsc:
@benedictpaten, would you have time to give this a sanity check? Only need to look at 38f7088 and 3124052.

hannes-ucsc:
@benedictpaten, turns out that slices are supported out of the box and I addressed that in the docstring to rv(). Also rewrote the docstring on rv() and added examples as suggested.

PR #1036, opened 2016-07-01
Prevents exception masking in Toil context manager (resolves #921)

### LINKED ISSUE #921: NoSuchBucket error during Toil context manager exit
I have the feeling this happens while `__exit__` is invoked with a primary exception. The primary exception is shadowed by the secondary exception that results from not being able to delete the job store. The robust deletion PR might improve this but we should also log the primary exception eagerly in `__exit__`.

JakeNarkizian:
Resolves #921

Secondary exceptions raised in  the ```__exit__``` method of the Toil context manger can mask a primary exception. This has been fixed with a simple try except block.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -490,15 +490,21 @@ class Toil(object):
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """
-        Clean up after a workflow invocaton. Depending on the configuration, delete the job store.
+        Clean up after a workflow invocation. Depending on the configuration, delete the job store.
         """
+        try:
+            if (exc_type is not None and self.config.clean == "onError" or
+                            exc_type is None and self.config.clean == "onSuccess" or
+                        self.config.clean == "always"):
+                logger.info("Attempting to delete the job store")
+                self._jobStore.deleteJobStore()
+                logger.info("Successfully deleted the job store")
+        except Exception as e:
+            if exc_type is None:
+                raise
+            else:
+                logger.exception('The following error was raised during clean up:')
         self._inContextManager = False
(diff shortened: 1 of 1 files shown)


src/toil/common.py
```diff
@@ -490,15 +490,21 @@ def __enter__(self):
...
+                            exc_type is None and self.config.clean == "onSuccess" or
+                        self.config.clean == "always"):
+                logger.info("Attempting to delete the job store")
+                self._jobStore.deleteJobStore()
+                logger.info("Successfully deleted the job store")
+        except Exception as e:
+            if exc_type is None:
+                raise
+            else:
+                logger.exception(e)
```
hannes-ucsc: Please read the documentation on this method. You are using it wrong. The log message should give some context.

src/toil/common.py
```diff
@@ -490,15 +490,21 @@ def __enter__(self):
...
+                            exc_type is None and self.config.clean == "onSuccess" or
+                        self.config.clean == "always"):
+                logger.info("Attempting to delete the job store")
+                self._jobStore.deleteJobStore()
+                logger.info("Successfully deleted the job store")
+        except Exception as e:
+            if exc_type is None:
+                raise
+            else:
+                logger.exception('The following error was raised during clean up:')
```
hannes-ucsc: clean-up or cleanup

hannes-ucsc:
Fix the spelling error "on the side" in one of your next PRs. Separate commit but no issue reference or anything.

src/toil/common.py
JakeNarkizian: Addressed in #1038

Issue #918, opened 2016-06-01
SLURM integration follow up


guillermo-carrasco:
Hi there!

This is more a question/discussion than an issue, but I though this was the right place. Please forgive me if its not :)

On the past few weeks we all have been working in some way or another mainly in two things:
- Integrating the SLURM batch system into Toil
- Running CWL jobs and workflows using `cwltoil`

In our team, so far we've managed to run [the bcbio CWL workflow example](http://bcbio-nextgen.readthedocs.io/en/latest/contents/cwl.html#running) both using `cwltool` and `cwltoil` in single machines after the issues with `--no-container` and `--preserve-environment` were solved (see #863 and #882 ).

Yesterday @chapmanb managed to run also the complete workflow using SLURM, which is awesome! We were getting some strange errors about caching, but adding the `--disableSharedCache` option was enough to solve them.

We would love to keep the work on this SLURM and CWL integration with Toil, so we decided to open this discussion to decide how to proceed. What we would need is mainly:
1. Maybe a bit more documentation on how the caching and I/O works in Toil. Couldn't find anything relevant in the documentation, did I skip something @hannes-ucsc ?
2. Progress on the SLURM integration: In our group we would like to at least be able to select walltime and queues other than the default ones. @dleehr do you have any plans on working on this? Can we help?

Hopefully this will be helpful to all of us, cheers!

Ping @dleehr @tetron @chapmanb @brainstorm @ohofmann

hannes-ucsc:
Adding @benedictpaten (the original author of Toil) and @arkal (the author of the shared caching).

> We were getting some strange errors about caching

Please file issues. Caching should be transparent to the user, with one exception: files read from the FileStore are immutable. This is a backwards-incompatible API change but it can be disabled with a command line option. @arkal, can you elaborate?

arkal:
Hi @guillermo-carrasco.  Do you have a stack trace for the issues you saw with caching? As @hannes-ucsc mentioned, teh biggest change to how Toil works with the caching integrations is that read files (with readGlobalFile) are now immutable by default, and files associated with a job store ID cannot be explicitly deleted, moved or renamed by the user (use deleteGlobalFile or deleteLocalFile instead)

chapmanb:
[...]

Issue #677, opened 2016-02-10
Introduce concept of job name for improved diagnostic


hannes-ucsc:
The task name should be logged alongside the ID whenever possible. It could default to the function name or class name but should be overridable via a keyword argument to the job factories.

fnothaft:
+1

hannes-ucsc:
Remove the changes introduced by #1048.

hannes-ucsc:
Proposal: Add the ability to supply a job name with each job. This could be as a pseudo-requirement (`disk='1G', name='foo', …`) and/or as parameter to the `addChild…` and `addFollowOn…` family of methods. The job name should default to the function name. Users would typically want to add the sample UUID. At the end of a workflow invocation, Toil should print the list of failed jobs along with their name and optionally their stdout and stderr. The printout should only contain the jobs that failed during that invocation. There should also be a command line utility to produce the same listing for all failed jobs' output.

jvivian:
@arkal and I +1 as well.   

@Jeltje — Once this is integrated, the final output of Toil will list the name of the job that failed and what UUID, which means you can quickly ascertain what subset of samples failed.

Issue #652, opened 2016-01-22
User tutorial has no documentation of debugging pipelines


diekhans:
Users moving to Toil will often have no experience debugging in a cloud environment.  There needs to be a chapter on how a user approaches debugging.  print and pydb don't exactly work.

hannes-ucsc:
See also #677 and #662.

PR #1048, opened 2016-07-07
Improve logging message for filestore CWL runs

chapmanb:
The output logging messages when running cwltoil pipelines were generic
references to jobstore directories, making it hard for users to
determine which stage of the pipeline was running. This fixes in two
ways:

- Moves references to internal file deletion to debug, instead of info.
- Handles the case where we're running CWL-based runs with filesystem
  storage, reading the underlying CWL and providing a useful logging
  message from the baseCommand.

This is an attempt to improve the logging output during cwltoil runs to make it easier for users to identify what stage of the pipeline they were at. In the previous iteration, users could tell jobs were running but had little insight into the current step since all the outputs are references to `_toil_worker` commands, internal jobstore IDs and cleanup of files:
```
Executing command: '_toil_worker /home/chapmanb/drive/work/cwl/test_bcbio_cwl/cwltoil_work/cwltoil_jobstore l/1/jobt5ZNIZ'.
Got message from job at time 07-05-2016 10:34:15: Successfully deleted local copies of file with ID 'k/K/tmpABzeZz.tmp'.
Got message from job at time 07-05-2016 10:34:15: Successfully deleted local copies of file with ID 'T/u/tmpC8PsBX.tmp'.
Executing command: '_toil_worker /home/chapmanb/drive/work/cwl/test_bcbio_cwl/cwltoil_work/cwltoil_jobstore v/i/jobxQaXXO'.
```
After the change, the output looks like:
```
Executing command: 'bcbio_nextgen.py runfn prep_samples cwl'.
Executing command: 'bcbio_nextgen.py runfn prep_align_inputs cwl'.
Jobstore directory is: /home/chapmanb/drive/work/cwl/test_bcbio_cwl/cwltoil_work/cwltoil_jobstore
Executing command: '_toil_worker file:/home/chapmanb/drive/work/cwl/test_bcbio_cwl/cwltoil_work/cwltoil_jobstore K/2/jobVgjucx'.
Executing command: '_toil_worker file:/home/chapmanb/drive/work/cwl/test_bcbio_cwl/cwltoil_work/cwltoil_jobstore 6/n/job2ND5wI'.
Executing command: 'bcbio_nextgen.py runfn process_alignment cwl'.
Executing command: 'bcbio_nextgen.py runfn process_alignment cwl'.
```
I'm happy for suggestions to make this better and more generalized. Right now it would only work for filesystem-based JobStores and CWL runs. I was not able to understand the abstractions to know the best way to get these commands for all cases so it will fall back to the previous method where it is not these cases. You can see that behavior in the new example above where it falls back to the old method when executing non CWLTool commands (like workflows and scatter operations). Thanks much for looking at this.

### DIFF
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -120,6 +120,33 @@ class SingleMachineBatchSystem(BatchSystemSupport):
             self.workerThreads.append(worker)
             worker.start()
 
+    def _getDebugCmd(self, jobCommand):
+        """Calculate useful debugging command line, handling CWL runs.
+
+        Tries to print out underlying CWL base command being run if possible,
+        otherwise defaulting to the toil jobCommand.
+        """
+        debug_cmd = jobCommand
+        cmd_args = jobCommand.split()
+        if len(cmd_args) == 3 and cmd_args[0] == "_toil_worker":
+            _, job_store_locator, job_store_id = cmd_args
+            if job_store_locator.startswith("file:") or os.path.exists(job_store_locator):
+                import cPickle
+                import marshal as pickler
+                job_store_locator = job_store_locator.replace("file:", "")
+                job_cmd_file = os.path.join(job_store_locator, "tmp", job_store_id, "job")
+                with open(job_cmd_file) as in_handle:
+                    job = pickler.load(in_handle)
+                if job.get("command"):
+                    command_parts = job["command"].split()
+                    input_pickle_file = os.path.join(job_store_locator, "tmp", command_parts[1])
+                    with open(input_pickle_file) as in_handle:
+                        input = cPickle.load(in_handle)
+                    if (hasattr(input, "cwltool") and hasattr(input.cwltool, "tool")
+                            and "baseCommand" in input.cwltool.tool):
+                        debug_cmd = " ".join(input.cwltool.tool["baseCommand"])
+        return debug_cmd
+
     # Note: The input queue is passed as an argument because the corresponding attribute is reset
     # to None in shutdown()
 
@@ -141,7 +168,7 @@ class SingleMachineBatchSystem(BatchSystemSupport):
                                   jobCores)
                         with self.coreFractions.acquisitionOf(coreFractions):
                             with self.disk.acquisitionOf(jobDisk):
-                                log.info("Executing command: '%s'.", jobCommand)
+                                log.info("Executing command: '%s'.", self._getDebugCmd(jobCommand))
                                 startTime = time.time() #Time job is started
                                 with self.popenLock:
                                     popen = subprocess.Popen(jobCommand,


benedictpaten:
Jenkins, test this please.

hannes-ucsc:
We shouldn't have to parse the worker command line to get at that information. That's just is brittle and circuitous. This will be solved permanently and allencompassingly by #652 and #677. However, I am torn as to whether we should include this anyways as a temporary patch for a real paint point for our users.

Would like to get @benedictpaten and @fnothaft position on this.

fnothaft:
> However, I am torn as to whether we should include this anyways as a temporary patch for a real paint point for our users.

What's the timeline for a fix to #652 and #677? My thought is that since the pain point is real and big, unless the all-encompassing fix is expected to hit very soon (e.g. within the next release), I would merge this in, and then revert it when we've got an all-encompassing fix.

hannes-ucsc:
> What's the timeline for a fix to #652 and #677?

I guessing, three months.

hannes-ucsc:
@arkal there are some subtle changes to how caching is logged. Could you take a look? That one indention change worries me most.

hannes-ucsc:
@chapmanb, we use camelCase in Toil. Could you rename _get_debug_cmd accordingly? And, pending @arkal's feedback, please rebase with conflict resolution.

arkal:
Uhm... deleteGlobal and deleteLocal are actually user callable functions too. In that case we definitely want to see the logs in INFO and not in DEBUG. An option is to change the signature of `deleteGlobalFile` and `deleteLocalFile` to become `(self, fileStoreID, verbose=False)` and then modify [this line](https://github.com/chapmanb/toil/blob/967fc1f79792592164a9ec39d1ad223fa8571c22/src/toil/job.py#L2291) to `fileStore.deleteGlobalFile(jobStoreFileID, quiet=True)`
quiet = True will put those 2 logs into DEBUG, quiet=False will keep then in INFO

src/toil/job.py
```diff
@@ -982,9 +982,9 @@ def deleteLocalFile(self, fileStoreID):
                             os.remove(cachedFile)
                             cacheInfo.cached -= fileSize
                     self.logToMaster('Successfully deleted cached copy of file with ID '
-                                     '\'%s\'.' % fileStoreID)
+                                     '\'%s\'.' % fileStoreID, logging.DEBUG)
                 self.logToMaster('Successfully deleted local copies of file with ID '
-                                 '\'%s\'.' % fileStoreID)
+                                 '\'%s\'.' % fileStoreID, logging.DEBUG)
```
benedictpaten: I agree with this change, I would go further and turn these into standard log messages. @arkal - in the case that the job succeeds do you really want to see this message? Seems like a lot of low level information. The current design for "global" logging messages is to convey them all back to the leader. If you want to persist this behavior we should filter log to masters on the worker by the log level to prevent excessive log passing during a production run.

src/toil/job.py
```diff
@@ -1013,7 +1013,7 @@ def deleteGlobalFile(self, fileStoreID):
             # Add the file to the list of files to be deleted once the run method completes.
             self.filesToDelete.add(fileStoreID)
             self.logToMaster('Added file with ID \'%s\' to the list of files to be' % fileStoreID +
-                             ' globally deleted.')
+                             ' globally deleted.', logging.DEBUG)
```
benedictpaten: Ditto comment above on logging. @arkal

benedictpaten:
I left comments on the log messages. I agree that sorting out the log messages is important. What we have now is the accumulated cruft of our development cycles. Totally agree with Hannes about this being a pain point.

I really like the attempt to get at the "job name", but I agree with Hannes, I don't like that this is a single machine only fix. I had forgotten that single machine had those log messages. For only a modest amount of storage cost we could afford to serialise the job name in the job wrapper and pass it around, and then all log messages could access it. We could make the job name a class variable is users wanted a nicer string (not the limitations of the class name), or even just a variable of the class if a user wanted to identify a particular instance of a class. The biggest cost of this approach would be an increase of memory usage on the leader.

arkal:
Yes I agree they should be standard logging. I think i made those logToMaster because at the time I thought that was replacing standard logging.

@benedictpaten i think if we see "Downloaded file from file store " and "wrote file to file store" for all files read and written by the user, we should also see "deleted file from file store" messages. In the interest of symmetricity

brainstorm:
That pull request would be really useful since I'm now trying to implement support for SLURM flags in CWLToil (https://github.com/BD2KGenomics/toil/issues/918) but running into some issues with some code in `leader.py` and `cwltoil.py` when it tries to unpickle the `rootJobReturnValue`.

Will pull it in my branch, but happy to see it in `master`. Good job @chapmanb!

chapmanb:
Hannes, Arjun, Benedict and Roman;
Thanks for the thoughts. I'm happy to try some of the suggestions and improvements in this patch if there is consensus we should move forward. I'd love to have the less brittle fix command serialization suggested by Benedict but don't have enough Toil code understanding to implement it myself.

In defense of this patch I can give you the CWL user perspective. The current set of log messsages are completely meaningless to me. I can tell that something is running but have no idea about where in the process it is. I don't care at all about Toil staging and deleting files unless it's somehow erroring out, and the jobStore outputs have no useful information to act on. I feel most strongly that we need something to indicate status that is meaningful to the user who has no concept of the underlying CWL or code, so happy to get there with whatever path you think is best.

src/toil/job.py
hannes-ucsc: +1 on changing these to standard logging.

@chapmanb, why did you change this? Depending on your response, we might ask you take undo these two changes out and we might address them in a second PR.

hannes-ucsc:
We will merge this once the sticking points are resolved. The code will be removed when we have a more complete solution.

Also see my line comments.

src/toil/job.py
hannes-ucsc: Just fixed a bunch of typos in my previous line note.

chapmanb: I switched these from the default (logging.INFO) to DEBUG as they fill up the logging input and don't offer much helpful information from the user perspective. For my CWL runs I get hundreds of messages like:
```
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'h/y/tmpMJXEcB.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'j/W/tmp2cwAzR.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'G/2/tmpGT1jJU.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'a/W/tmpUWRh4C.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'A/l/tmpwS225a.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'Q/f/tmpMt3I3j.tmp'.
```
In my mind these seemed better suited for debugging than running information.

chapmanb:
Thanks, I left a note. My logging changes are meant to adjust messages from INFO to DEBUG to avoid some noise in the output. Let me know if I can provide more information.

src/toil/job.py
arkal: are you using job.fileStore.deleteLocalFile?

hannes-ucsc: CWL might be triggering the calls to deleteLocalFile.

chapmanb: That's right on. We're calling this through CWL so are a layer separated from using the Toil API.

hannes-ucsc: It seem like there is a theme developing: we want invocations of file store methods made by Toil code to be logged at DEBUG, while user script invocations should be logged at INFO. We have @arkal's `quiet=` kwarg proposal. We could also use `inspect` to figure out whether the calling stack frame is for a function belonging to Toil. Other ideas?

arkal: Make it a command line flag?

arkal: Also i realised, the messages you're seeing are from the job cleanup calls to deleteLocalFile. I've made a comment on how you can fix this inline

src/toil/job.py
```diff
@@ -982,9 +982,9 @@ def deleteLocalFile(self, fileStoreID):
                             os.remove(cachedFile)
                             cacheInfo.cached -= fileSize
                     self.logToMaster('Successfully deleted cached copy of file with ID '
-                                     '\'%s\'.' % fileStoreID)
+                                     '\'%s\'.' % fileStoreID, logging.DEBUG)
                 self.logToMaster('Successfully deleted local copies of file with ID '
```
arkal: If this statement is moved into the if block, it will be printed only when the user explicitly calls job.fileStore.deleteLocalFile.

arkal: I can address this in a future PR

src/toil/job.py
hannes-ucsc: Ok, it's clear that we need a different ticket to track this. @chapmanb can you remove that change from this PR and create a GH issue for it?

chapmanb:
Arjun and Hannes -- thanks for the comments and discussion. I removed the logging messages from this PR and opened up an issue to track that (#1061). You both have a much better clue than I do about what has to happen so I'll stay away from a PR on that.

Hope this works for getting the debugging merged. Thanks again.

src/toil/batchSystems/singleMachine.py
```diff
@@ -120,6 +120,33 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
             self.workerThreads.append(worker)
             worker.start()
 
+    def _get_debug_cmd(self, jobCommand):
```
hannes-ucsc: Toil uses camel case. Please rename to _getDebugCmd.

chapmanb: Thanks Hannes, fixed. Apologies, it's quite hard to get used to camelCase when you normally use the pep8 styles.

hannes-ucsc:
Please squash.

src/toil/batchSystems/singleMachine.py
hannes-ucsc: I dislike camelCase in Python. But I dislike inconsistency even more.

On Thu, Jul 14, 2016 at 10:18 AM, Brad Chapman <notifications@github.com>
wrote:

> In src/toil/batchSystems/singleMachine.py
> <https://github.com/BD2KGenomics/toil/pull/1048#discussion_r70845448>:
>
> > @@ -120,6 +120,33 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
> >              self.workerThreads.append(worker)
> >              worker.start()
> >
> > +    def _get_debug_cmd(self, jobCommand):
>
> Thanks Hannes, fixed. Apologies, it's quite hard to get used to camelCase
> when you normally use the pep8 styles.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/BD2KGenomics/toil/pull/1048/files/d71a123c4a03b56ef322cdeb0270a847bf83f6bc#r70845448>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AE562MZN5BXY41uAREp6dfnJeLRzShJoks5qVm9QgaJpZM4JG9ej>
> .
>



-- 
Hannes Schmidt
Software Engineer
Center for Biomolecular Science and Engineering
University of California, Santa Cruz

(206) 696-2316 (cell)
hannes@ucsc.edu

chapmanb:
Also squashed

hannes-ucsc:
Will merge once build passes.

hannes-ucsc:
Jenkins, test this please.

PR #1052, opened 2016-07-08
Add singleMachine disk tracking (resolves #770)

### LINKED ISSUE #770: Add tracking of disk to single machine
Should be straight-forward to implement but it will increase the likelihood of deadlock due to the  nested, blocking nature of resource allocation in SingleMachine.

arkal:
resolves #770

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -352,5 +352,6 @@ class InsufficientSystemResources(Exception):
         self.resource = resource
 
     def __str__(self):
-        return 'Requesting more {} than available. Requested: {}, Available: {}' \
-               ''.format(self.resource, self.requested, self.available)
+        return 'Requesting more {} than either physically available, or enforced by --max{}. ' \
+               'Requested: {}, Available: {}'.format(self.resource, self.resource.capitalize(),
+                                                     self.requested, self.available)
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -25,7 +25,7 @@ from threading import Lock, Condition
 from Queue import Queue, Empty
 
 import toil
-from toil.batchSystems.abstractBatchSystem import BatchSystemSupport, AbstractBatchSystem
+from toil.batchSystems.abstractBatchSystem import BatchSystemSupport, InsufficientSystemResources
 
 log = logging.getLogger(__name__)
 
@@ -62,6 +62,10 @@ class SingleMachineBatchSystem(BatchSystemSupport):
         if maxMemory > self.physicalMemory:
             log.warn('Limiting maxMemory to physically available memory (%i).', self.physicalMemory)
             maxMemory = self.physicalMemory
+        self.physicalDisk = toil.physicalDisk(config)
+        if maxDisk > self.physicalDisk:
+            log.warn('Limiting maxDisk to physically available disk (%i).', self.physicalDisk)
+            maxDisk = self.physicalDisk
         super(SingleMachineBatchSystem, self).__init__(config, maxCores, maxMemory, maxDisk)
         assert self.maxCores >= self.minCores
         assert self.maxMemory >= 1
@@ -94,12 +98,20 @@ class SingleMachineBatchSystem(BatchSystemSupport):
         """
         :type list[Thread]
         """
+        # Variables involved with non-blocking resource acquisition
+        self.acquisitionTimeout = 5
+        self.acquisitionRetryDelay = 10
+        self.aquisitionCondition = Condition()
+
         # A pool representing available CPU in units of minCores
-        self.coreFractions = ResourcePool(self.numWorkers)
+        self.coreFractions = ResourcePool(self.numWorkers, 'cores', self.acquisitionTimeout)
         # A lock to work around the lack of thread-safety in Python's subprocess module
         self.popenLock = Lock()
         # A pool representing available memory in bytes
-        self.memory = ResourcePool(self.maxMemory)
+        self.memory = ResourcePool(self.maxMemory, 'memory', self.acquisitionTimeout)
+        # A pool representing the available space in bytes
+        self.disk = ResourcePool(self.maxDisk, 'disk', self.acquisitionTimeout)
+
         log.info('Setting up the thread pool with %i workers, '
                  'given a minimum CPU fraction of %f '
                  'and a maximum CPU value of %i.', self.numWorkers, self.minCores, maxCores)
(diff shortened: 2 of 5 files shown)


arkal:
Jenkins, test this please.

hannes-ucsc:
I haven't done a review yet, but write the commit message for 025adaf as if you didn't know that it was going  to fail. Just the describe what the test would prove, assuming they pass. The current commit message is going to confuse people in a few months. Also, the title to should refer to the ticket as `(connected to #…)`

arkal:
okay :)

arkal:

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -352,5 +352,6 @@ def __init__(self, resource, requested, available):
         self.resource = resource
 
     def __str__(self):
-        return 'Requesting more {} than available. Requested: {}, Available: {}' \
-               ''.format(self.resource, self.requested, self.available)
+        return 'Requesting more {} than either physically available, or enforced by --max{}. ' \
+               'Requested: {}, Available: {}'.format(self.resource.capitalize(), self.resource,
```
hannes-ucsc: shouldn't the 2nd one be capitlized?

arkal: whoops. I'll fix that.

src/toil/batchSystems/singleMachine.py
```diff
@@ -118,37 +125,51 @@ def worker(self, inputQueue):
...
+                                                          "from command '%s'.", statusCode,
+                                                          jobCommand)
+                                    finally:
+                                        self.runningJobs.pop(jobID)
+                                finally:
+                                    if statusCode is not None and not info.killIntended:
+                                        self.outputQueue.put((jobID, statusCode,
+                                                              time.time() - startTime))
+                except InsufficientSystemResources as err:
+                    # Suppressing the error message and printing this to debug instead.
```
hannes-ucsc: This is obvious.

src/toil/batchSystems/singleMachine.py
```diff
@@ -118,37 +125,51 @@ def worker(self, inputQueue):
...
+                                                log.error("Got exit code %i (indicating failure) "
+                                                          "from command '%s'.", statusCode,
+                                                          jobCommand)
+                                    finally:
+                                        self.runningJobs.pop(jobID)
+                                finally:
+                                    if statusCode is not None and not info.killIntended:
+                                        self.outputQueue.put((jobID, statusCode,
+                                                              time.time() - startTime))
+                except InsufficientSystemResources as err:
```
hannes-ucsc: Exceptions always `e`

src/toil/batchSystems/singleMachine.py
```diff
@@ -241,15 +262,27 @@ def __init__(self, startTime, popen, killIntended):
...
     def acquire(self, amount):
         with self.condition:
+            startTime = time.time()
             while amount > self.value:
-                self.condition.wait()
+                if time.time() - startTime >= 5:
+                    # This means the thread timed out waiting for the resource.  We exit the nested
+                    # context managers in worker to prevent blocking of a resource due to
+                    # unavailability of a nested resource request.
+                    raise InsufficientSystemResources(resource=self.resourceType, requested=amount,
```
hannes-ucsc: This is not the right exception type. Create new one, nest it in ResourcePool and call it `AcquisitionTimeoutException`.

arkal: Haha I had created an `InsufficientResourceAvailable` Exception and realized it was almost exactly the same as `InsufficientSystemResources` (in terms of what variables it holds) barring the error message so I thought i'd appropriate this instead.

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -577,6 +577,118 @@ class SingleMachineBatchSystemJobTest(hidden.AbstractBatchSystemJobTest):
...
+        But the order of execution should be
+        root > bl, fJ, sJ > sJC > sJCC > fJC
+        since fJC cannot run till bl finishes but sJC and sJCC can(fJC blocked by disk). If the
+        resource acquisition is written properly, then fJC which is scheduled before sJC and sJCC
+        should not block them, and should only run after they finish.
+        """
+        Job.Runner.startToil(root, options)
+        with open(outFile) as oFH:
+            outString = oFH.read()
+        self.assertEqual(outString, 'AAACDB')
```
hannes-ucsc: Can you make the writeVal values consistent with the job name? That way I won't have to hunt down each job's writeVal to verify your assertion against what you write in the comment above.

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -577,6 +577,118 @@ class SingleMachineBatchSystemJobTest(hidden.AbstractBatchSystemJobTest):
...
+                                         disk='1M')
+
+        root.addChild(blocker)
+        root.addChild(firstJob)
+        root.addChild(secondJob)
+
+        firstJob.addChild(firstJobChild)
+        secondJob.addChild(secondJobChild)
+
+        secondJobChild.addChild(secondJobChildChild)
```
hannes-ucsc: "Grand child"

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -577,6 +577,118 @@ class SingleMachineBatchSystemJobTest(hidden.AbstractBatchSystemJobTest):
...
+
+        root = Job()
+        # writes A
+        blocker = Job.wrapFn(_resourceBlockTestAuxFn, outFile=outFile, sleepTime=30, writeVal='A',
+                             cores=2, memory='1M', disk='1M')
+        firstJob = Job.wrapFn(_resourceBlockTestAuxFn, outFile=outFile, sleepTime=5, writeVal='A',
+                              cores=1, memory='1M', disk='1M')
+        secondJob = Job.wrapFn(_resourceBlockTestAuxFn, outFile=outFile, sleepTime=10, writeVal='A',
+                               cores=1, memory='1M', disk='1M')
+        # Writes B
```
hannes-ucsc: Obvious, but see below.

src/toil/batchSystems/singleMachine.py
```diff
@@ -118,37 +125,51 @@ def worker(self, inputQueue):
...
-                                                  "command '%s'.", statusCode, jobCommand)
-                            finally:
-                                self.runningJobs.pop(jobID)
-                        finally:
-                            if statusCode is not None and not info.killIntended:
-                                self.outputQueue.put((jobID, statusCode, time.time() - startTime))
-            finally:
-                log.debug('Finished job. self.coreFractions ~ %s and self.memory ~ %s',
-                          self.coreFractions.value, self.memory.value)
+            while True:
```
hannes-ucsc: Here's what I don't understand: this loops get tries to satisfy the same job with every iteration. How can this be effective in increasing job throughput?

arkal: From what i've understood, each job gets one worker. This loop keeps the current worker alive and doesn't return a value saying the job passed or failed. Another concurrent worker can then attempt to use the resources that this one would have potentially blocked.

src/toil/test/batchSystems/batchSystemTest.py
arkal: Done!

src/toil/batchSystems/singleMachine.py
hannes-ucsc: The likelihood is incidental.

src/toil/batchSystems/singleMachine.py
hannes-ucsc: Ahh, yes. Now I remember.

src/toil/batchSystems/singleMachine.py
hannes-ucsc: Using a specific exception will increase the resusability of ResourcePool.

src/toil/batchSystems/singleMachine.py
```diff
@@ -241,15 +262,27 @@ def __init__(self, startTime, popen, killIntended):
...
+                    # This means the thread timed out waiting for the resource.  We exit the nested
+                    # context managers in worker to prevent blocking of a resource due to
+                    # unavailability of a nested resource request.
+                    raise InsufficientSystemResources(resource=self.resourceType, requested=amount,
+                                                      available=self.value)
+                # Allow 5 seconds to get the resource, else quit through the above if condition.
+                # This wait + timeout is the last thing in the loop such that a request that takes
+                # longer than 5s due to multiple wakes under the 5 second threshold are still
+                # honored.
+                self.condition.wait(timeout=5)
```
hannes-ucsc: This `5` is related to the other `5` above. It and the sleep of 10 are important parameters that should be promoted to class attributes.

src/toil/test/batchSystems/batchSystemTest.py
```diff
@@ -577,6 +577,118 @@ class SingleMachineBatchSystemJobTest(hidden.AbstractBatchSystemJobTest):
...
+        resource acquisition is written properly, then fJC which is scheduled before sJC and sJCC
+        should not block them, and should only run after they finish.
+        """
+        Job.Runner.startToil(root, options)
+        with open(outFile) as oFH:
+            outString = oFH.read()
+        self.assertEqual(outString, 'AAACDB')
+
+    @staticmethod
+    def _resourceBlockTestAuxFn(outFile, sleepTime, writeVal):
```
hannes-ucsc: Considering that this is the end the testcase, can we forgo the "export static method as global" hack?

src/toil/batchSystems/singleMachine.py
arkal:
```
self.resourceAcquisitionTimeout = 5
self.resourceAcquisitionRetryDelay = 10
```
sound good?

hannes-ucsc: Considering that this is a *Resource*Pool, I would drop the `resource` prefix.

arkal: okai

src/toil/test/batchSystems/batchSystemTest.py
arkal: so just make it a global? that's fine by me!

src/toil/batchSystems/singleMachine.py
arkal: sounds good

src/toil/test/batchSystems/batchSystemTest.py
hannes-ucsc: yes, as long as it is in close proximity with the test method.

arkal: done

arkal:
Tests were timing out due to insufficient disk during setup of single machine batch system for tests. Wasn't a problem before because we didn't track disk haha.

arkal:
@hannes-ucsc this is ready!

src/toil/batchSystems/singleMachine.py
```diff
@@ -275,3 +319,26 @@ def acquisitionOf(self, amount):
             yield
         finally:
             self.release(amount)
+
+
+class AcquisitionTimeoutException(Exception):
```
hannes-ucsc: This should be nested in ResourcePool or renamed to ResourceAcquisitionTimeoutException. I'd prefer the former.

arkal:

PR #1050, opened 2016-07-07
cwltoil: bring up to date with CWL v1.0 release

chapmanb:
Synchronizes Toil CWL support with the latest cwltool supporting v1.0 of
the CWL specification:

- Handles back compatibility of path/location specifications for files
  (path is old, location is new).
- Adjusts imports to match cwltool reorganization changes
- Some functions use kwargs instead of default arguments
  (single_job_executor, _init_job).
- Changes from output to outputSource in job outputs.

cc @tetron for sanity checks

PR #1056, opened 2016-07-12
CWL: test update and fixes for v1.0 support

chapmanb:
Update of #1050 which I messed up trying to rebase.

In progress work towards v1.0 support. Current status, failing 19 out of
72 unit tests.

- Apply path/location fixes globally, required for v1.0 support
  (which uses location) and back compatibility with CWL (which uses path).
- Adjusts imports to match cwltool reorganization changes
- Some functions use kwargs instead of default arguments
  (single_job_executor, _init_job).
- Changes from output to outputSource in job outputs.
- Start of work to improve scatter support for valueFrom by applying
  after scatter process. Still in progress.
- Update internal CWL revsort test to v1.0 syntax
- Update conformance test to v1.0

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,8 +12,8 @@ Toil is a workflow engine entirely written in Python. It features:
 
 * `Common Workflow Language`_ (`CWL`_) support
 
-  Complete support for the draft-3 CWL specification, allowing it to execute
-  CWL workflows.
+  Nearly complete support for the stable CWL v1.0 specification,
+  allowing it to execute CWL workflows.
 
 * Cross platform support
 
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -44,11 +44,24 @@ workflows that are portable across multiple workflow engines and platforms.  To
 run workflows written using CWL, first ensure that Toil is installed with the
 "cwl" extra as described in :ref:`installation-ref`.  This will install the
 executables ``cwl-runner`` and ``cwltoil`` (these are identical, where
-``cwl-runner`` is the portable name for the default system CWL runner).  To
-learn more about CWL, see the `CWL User Guide`_.
+``cwl-runner`` is the portable name for the default system CWL runner).
 
-To run in local batch mode, simply provide the CWL file and the input object
-file::
+To learn more about CWL, see the `CWL User Guide`_. Toil has nearly full support
+for the stable v1.0 specification, only lacking the following features:
+
+- `Directory <http://www.commonwl.org/v1.0/CommandLineTool.html#Directory>`_
+  inputs and outputs in pipelines. Currently you need to enumerate directory
+  inputs as Files.
+- `InitialWorkDirRequirement
+  <http://www.commonwl.org/v1.0/CommandLineTool.html#InitialWorkDirRequirement>`_
+  to create files together within a specific work directory. Collecting
+  associated files using `secondaryFiles
+  <http://www.commonwl.org/v1.0/CommandLineTool.html#CommandInputParameter>`_ is
+  a good workaround.
+- `File literals <http://www.commonwl.org/v1.0/CommandLineTool.html#File>`_ that
+  specify only ``contents`` to a File without an explicit file name.
+
+To run in local batch mode, provide the CWL file and the input object file::
 
     cwltoil example.cwl example-job.yml
 
@@ -57,7 +70,7 @@ command line parameters to select and configure the batch system to use.
 Consult the appropriate sections.
 
 .. _Common Workflow Language: http://commonwl.org
-.. _CWL User Guide: http://commonwl.org/draft-3/UserGuide.html
+.. _CWL User Guide: http://www.commonwl.org/v1.0/UserGuide.html
 
 .. _runningDetail:
 
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
             'gcs_oauth2_boto_plugin==1.9',
             botoRequirement],
         'cwl': [
-            'cwltool==1.0.20160712154127']},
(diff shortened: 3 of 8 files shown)


chapmanb:
Hannes -- sorry, I don't have enough git skills to rebase against master and ended up pulling in all of the upstream changes as separate commits in #1050, so I closed that and started a clean fix. I need to find some documentation to learn how to rebase without messing everything up.

On the CWL side, I updated the tests and added additional fixes. This is not fully v1.0 compliant but runs complicated bcbio workflows so is a good step in the right direction. I'm currently stuck and some of the v1.0 migration might need help from @tetron. What are you thoughts about moving forward with this? Could we merge this as a starting point for draft-3 and partial v1.0 support and then continue to add fixes? Happy for suggestions.

hannes-ucsc:
@chapmanb, no worries. In the interest of keeping the discussion in one place for future PRs, let me mention that you can always delete the messed up branch in your local clone, create a new branch but name it the same and finally just force push. That will leave the old PR open. We explicitly allow force pushes on PR branches as long as that PR isn't tagged `multi-author`.

hannes-ucsc:
Jenkins, test this please.

chapmanb:
Hannes -- I talked with Peter and managed to push Toil CWL support forward so we support everything in CWL v1.0 except for these features:

- Support for directory inputs
  (http://www.commonwl.org/v1.0/CommandLineTool.html#Directory)
- InitialWorkDirRequirement
  http://www.commonwl.org/v1.0/CommandLineTool.html#InitialWorkDirRequirement

Implementing those new features is beyond my skill level with CWL and Toil and Peter estimated that he wouldn't have availability to get to them for about a month. Would it be possible to review this current set of changes and incorporate so we have better v1.0 support in the meantime? I will plan to revisit the additional features once Peter has time to coordinate.

The latest fixes also includes a fileStore assertion change which caused lots of intermittent failures when running the tests. I kept this as a separate commit in case you want to omit or treat separately. Thanks much for looking at this.

tetron:
Incidentally, if something is failing due to an unimplemented feature, it should quit with exit code 33 to indicate to cwltest that it's a known feature gap rather than an actual test failure.

src/toil/fileStore.py
```diff
@@ -851,7 +851,8 @@ def cleanCache(self, newJobReqs):
                 #                  (self.decodedFileID(cachedFile), cachedFileSize))
                 logger.debug('CACHE: Evicted  file with ID \'%s\' (%s bytes)' %
                              (self.decodedFileID(cachedFile), cachedFileSize))
-            assert cacheInfo.isBalanced(), 'Unable to free up enough space for caching.'
```
hannes-ucsc: This assertion is rather important. It usually happens because the job requirements are not accurate. We might need to add some explanation to the assertion message.

Pinging @arkal. We might need to change the assertion message to include a few hints as to possible reasons. If user error can lead to the assertion failure, it shouldn't be an assertion. Use require() from BPL.

hannes-ucsc:
I would merge this if the tests passed and the assertion removal was reverted. If the assertion is causing you trouble, I would open a ticket.

src/toil/fileStore.py
arkal: That's a very important error. We could make it more verbose but really, we should never see this issue.

hannes-ucsc: > we should never see this issue.

But we are. Whose problem is this? Toil's or the user's?

arkal: I'm going to take a guess and say you're using singleMachine? This won't happen after the disk tracking we just put into Toil

chapmanb: Hannes and Arjun -- thanks for the additional information on this and the pointer to the latest update. I pulled in #1052 and it does fix the problem running the tests. Brilliant. I reverted the change and have a local version passing all of the CWL v1.0 tests. As soon as common-workflow-language/common-workflow-language#268 gets merged I can update to that and will push the final CWL fixes needed. Thanks again for the help.

hannes-ucsc:
Please remove commits 6430548 and 1dbaab0 using interactive rebase.

hannes-ucsc:
Jenkins, test this please.

chapmanb:
Hannes, sorry I was too slow getting in the final fixes for v1.0 support for your most recent test run. I just checked in fixes to allow CWL v1.0 compatibility minus a few features (Directories, InitialWorkDir and File literals) which Toil will sensibly error out with a useful message. This now passes the CWL v1.0 test suite and cwltool is synchronized with Arvados CWL runner.

I hope this will now pass tests and be good to go. I'd rather not rebase my steps here if we can avoid it so it's easier to trace back logic and changes in the futures via the commit details.  Once we have this in I will prepare a separate PR updating the documentation about CWL support.

Thanks for all the useful feedback and fingers crossed this passes.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
The last commit message has a typo, Toil 1.0 is long gone. You mean CWL, correct? In fact, I like the prefix you used for the first two commits. I'm torn about not squashing because the commits and am willing to make an exception but the documentation needs to be added in this PR. I want to reduce the PR churn and avoid procrastinating on documentation which we do too frequently already. You are not likely to run into conflicts with this PR because no one is likely to touch the same files as you, so there is no real advantage in doing a separate PR.

hannes-ucsc:
I just updated my previous above comment to finish a sentence I had left open.

chapmanb:
Hannes -- fair enough, I added documentation to this PR to describe that we support CWL v1.0 and the specific features we currently lack. Thanks again.

hannes-ucsc:
Jenkins, ok to test.

Issue #928, opened 2016-06-03
Slurm currently does not support shared caching


chapmanb:
Hi all -- this extracts the caching discussion from #918 (cc'ing @arkal) and provides more details so that PR can focus on Slurm.

Running cwltoil with Slurm using:
```
cwltoil --batchSystem slurm --jobStore $JOBSTORE --logFile $LOGFILE --workDir $WORKDIR --preserve-environment PATH HOME --no-container run_info-cwl-workflow/main-run_info-cwl.cwl run_info-cwl-workflow/main-run_info-cwl-samples.json
```
(The CWL is from https://s3.amazonaws.com/bcbio/cwl/test_bcbio_cwl.tar.gz but not sure this has anything to do with CWL versus toil cache usage).

After applying the fix from #910 (which masks the real error), we see:
```
node04: 2016-06-03 10:43:16,323 INFO: toil.lib.bioio: Logging set at level: INFO
node04: 2016-06-03 10:43:16,323 INFO: toil.lib.bioio: Logging to file: cwltoil.log
node04: 2016-06-03 10:43:57,411 INFO: toil.lib.bioio: Logging set at level: INFO
node04: 2016-06-03 10:43:57,412 INFO: toil.lib.bioio: Logging to file: cwltoil.log
node04: 2016-06-03 10:43:57,476 INFO: toil.jobStores.fileJobStore: Jobstore directory is: /scratch/bchapman/cwl_testing/test_bcbio_cwl/cwltoil_jobstore
node04: 2016-06-03 10:43:57,477 INFO: toil.jobStores.abstractJobStore: The workflow ID is: 'e93d192a-a3ab-4027-9902-44db0f0837ad'
node04: 2016-06-03 10:43:57,685 INFO: toil.jobStores.abstractJobStore: Unable to import 'toil.jobStores.azureJobStore'
Traceback (most recent call last):
  File "/scratch/bchapman/cwl_testing/test_bcbio_cwl/../anaconda/bin/cwltoil", line 9, in <module>
    load_entry_point('toil==3.2.0a2', 'console_scripts', 'cwltoil')()
  File "/scratch/bchapman/cwl_testing/anaconda/lib/python2.7/site-packages/toil-3.2.0a2-py2.7.egg/toil/cwl/cwltoil.py", line 595, in main
    outobj = toil.start(wf1)
[...]

Issue #987, opened 2016-06-20
Docker containers can become zombies


fnothaft:
If a job terminates but that job was running docker containers, the docker containers can persist in the background and not get killed off.

Depends on #1189 (PR #1199)

fnothaft:
As suggested by @hannes-ucsc, look at https://docs.docker.com/swarm/ before doing anything.

hannes-ucsc:
@benedictpaten to write up his proposal here.

hannes-ucsc:
Renaming to zombie as that is established Unix terminology for processes and Docker containers are processes.

arkal:
@benedictpaten and I had a long chat yesterday and we built on the chat @hannes-ucsc and I had last week. We will use the inherent caching mechanism to track pids of jobs and then use the filestore context manager to allow jobs to clean up behind themselves on completion in a finally clause. On entering the filestore, jobs will check for other jobs' integrity and purge any dead jobs. Users will have the ability to register cleanup functions with toil that will be run in these situations. We will add a `docker_call` and a corresponding `docker_cleanup` function in toil-lib that can be used by any user.

hannes-ucsc:
Try to design this mechanism in a way that let's us do arbitrary per-job cleanup: Let's say we have a job function `foo`:

```py
def foo(job):
    job.defer(myCleanUp) # register a function to be called when this function returns or raises
    job.recover(myRecovery) # register a function to be invoked before this job is to be retried
    …
```

The particulars would have to be discussed further. For example, we might want Toil to guarantee that either the deferred function or the clean-up function are going to be invoked, but never both. Not sure.

`docker_call()` would then have to be extended to exploit this mechanism. The deferred function would `docker rm` one specific container—the one that will be created by this invocation of `docker_call`, the recovery function would list all containers and kill the ones that could have been created by any number of docker_call invocations.
```

arkal:
can we discuss this tomorrow after lunch?

fnothaft:
@hannes-ucsc one thing to keep in mind is that the clean up function would need to run on the same node as the original job, right? This is hard to do on some batch systems, without "going around" the batch system (i.e., ssh'ing directly to the node).

arkal:
yes which is why we need to chat. We were thinking of a generic cleanup that could be registered with toil. Then any job on a node can check for failed jobs and handle them

hannes-ucsc:
[...]

PR #1067, opened 2016-07-21
Shared caching: re-enable SLURM, SGE, LSF

chapmanb:
The current version of Toil does not run CWL with batch schedulers like
SLURM, SGE, and LSF since the disableSharedCache flag has been removed
(https://github.com/BD2KGenomics/toil/pull/1032/files#diff-ec7c0fff4aad88175b4920cd3d1e5f8fR653).
Following the previous discussion (https://github.com/BD2KGenomics/toil/issues/928#issuecomment-223639111),
this enables these schedulers to work provided files are staged into a
shared filesystem directory, allowing Toil to work for these cases.

### DIFF
--- a/src/toil/batchSystems/gridengine.py
+++ b/src/toil/batchSystems/gridengine.py
@@ -206,7 +206,7 @@ class GridengineBatchSystem(BatchSystemSupport):
 
     @classmethod
     def supportsWorkerCleanup(cls):
-        return False
+        return True
 
     @classmethod
     def supportsHotDeployment(cls):
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -129,7 +129,7 @@ class LSFBatchSystem(BatchSystemSupport):
     """
     @classmethod
     def supportsWorkerCleanup(cls):
-        return False
+        return True
 
     @classmethod
     def supportsHotDeployment(cls):
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -243,7 +243,7 @@ class SlurmBatchSystem(BatchSystemSupport):
 
     @classmethod
     def supportsWorkerCleanup(cls):
-        return False
+        return True
 
     @classmethod
     def supportsHotDeployment(cls):


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
-1

If a batch system doesn't support worker clean-up, it shouldn't pretend that it does. This needs to be a little smarter. Even if the cache is on a shared FS, who cleans up those directories? Who enforces that the FS is indeed shared? Last but not least, why are we caching files on a network file system if those files originate in a job store that's most likely hosted on a network file system?

chapmanb:
Hannes -- thanks for the feedback. My preferred option would be to not cache files at all, as I agree that I don't see a benefit for batch systems. Is it possible to do this? As far as I understood it #1032 removed the ability to not use shared caches. Is there another way to do this?

If you don't like the approach in this patch, I could use some practical suggestions for how to proceed. I don't have the Toil expertise to implement this suggestion (https://github.com/BD2KGenomics/toil/issues/928#issuecomment-223680488) and don't think it provides a ton of benefit for the costs of tracking and extra job scheduling.

Right now, batch scheduling doesn't work at all with Toil > 3.1.6 (and thus CWL v1.0 support) so I'm trying to work on this to unblock our efforts to expand SLURM support.

hannes-ucsc:
There are two things we can do. 

1) Implement true worker clean-up for those batch systems. We would need a way to determine all nodes in a cluster and then submitting a final clean-up job when the batch system is being shut-down.

2) Add back a non-caching FileStore implementation. We had one but it was too complicated and the caching file store duplicated some logic from it (asynchronous writing).

I think we need 2) anyways as a quick and easy way to rule out caching when diagnosing bugs, so we should start with that. It should be fairly easy: At its conception, FileStore used to be a mere proxy in front of the job store that only exposed its file handling methods.

@arkal, who would be the best person to do this, is pretty busy with his day time job (working on ProTECT) so I can't promise when we would get to this. If anyone wants to play with this, go way back in the history when FileStore was initially created and resurrect that implementation.

chapmanb:
Hannes -- thanks for the details. I agree #2 makes the most sense and happy to work on that. Arjun or Hannes, if you can help with specific pointers I will try and give it a go.

I'd still like to try to get this PR in as a temporary fix. Right now batch schedulers don't work at all which is bad for my efforts to promote Toil as a CWL runner for local clusters.

arkal:
So i could do 2 relatively easily but I can't do it till Monday Night/Tuesday morning.... my only laptop is going to be in the shop for the weekend.

@hannes-ucsc fora  test i was thinkging of writing a simple pipeline that runs 6 jobs.... one writes, one writes to a stream, one reads either of the written files normally, one reads the other written file as a stream (both assert the contents of the file), one deletes a file locally and one deletes it globally. This should be enough to test the simple filestore we have planned. I'll work on this asa i get my mac back.

hannes-ucsc:
2) is tracked in #1072 
1) is not tracked yet. Once 2) is in we can see if the community still really craves caching on the legacy BSs. Note, however, that worker cleanup intersects with the zombie container problem that @benedictpaten is currently pondering over (#987)

PR #1079, opened 2016-07-27
Add ability to disable caching (resolves #1072)

### LINKED ISSUE #1072: Add back ability to disable caching
Create a second FileStore implementation that can fill in for the current caching one. Add `--disableCaching` and mention it in the error message that is raised when a BS is used that doesn't support the caching FileStore.

cket:


src/toil/fileStore.py
```diff
@@ -92,6 +92,13 @@ def __init__(self, jobStore, jobWrapper, localTempDir, inputBlockFn):
         # job if required.
         self._setupCache()
 
+    @staticmethod
+    def createFileStore(jobStore, jobWrapper, localTempDir, inputBlockFn, caching):
```
hannes-ucsc: This could be rewritten as

```python
fileStoreCls = FileStore if caching else NonCachingFileStore
return fileStoreCls(…)
```

src/toil/fileStore.py
```diff
@@ -1326,6 +1336,102 @@ def __del__(self):
...
+        self.filesToDelete = set()
+
+    @contextmanager
+    def open(self, job):
+        startingDir = os.getcwd()
+        self.localTempDir = makePublicDir(os.path.join(self.localTempDir, str(uuid.uuid4())))
+        try:
+            os.chdir(self.localTempDir)
+            yield
+
```
hannes-ucsc: no empty line before finally

src/toil/fileStore.py
```diff
@@ -1326,6 +1336,102 @@ def __del__(self):
...
+    def _blockFn(self):
+        # there is no asynchronicity in this file store so no need to block at all
+        return True
+
+    def _updateJobWhenDone(self):
+        try:
+            # Indicate any files that should be deleted once the update of
+            # the job wrapper is completed.
+            self.jobWrapper.filesToDelete = list(self.filesToDelete)
+
```
hannes-ucsc: I'd rather not have these empty lines.

src/toil/worker.py
```diff
@@ -319,7 +320,7 @@ def recursiveDelete(jobWrapper2):
                     jobWrapper.checkpoint = jobWrapper.command
 
                 # Create a fileStore object for the job
-                fileStore = FileStore(jobStore, jobWrapper, localWorkerTempDir, blockFn)
+                fileStore = FileStore.createFileStore(jobStore, jobWrapper, localWorkerTempDir, blockFn, caching)
```
hannes-ucsc: If a Config instance is in scope, it's probably clearer to no extract `caching` and call as `…, caching=not config.disableCaching`.

hannes-ucsc:
Fantastic! I only have a few nit picks.

PR #1090, opened 2016-07-29
disableCaching: avoid batch error message when set

chapmanb:
@cket @arkal @hannes-ucsc -- thank you for all the work on implementing disableCaching for the batch schedulers. I tested on SLURM and all works great with the exception of this small problem with the initial sanity checks. Thank you again for tackling this.

During testing of #1072 #1079, batch schedulers would error out with a
warning to add `--disableCaching` even when set. This corrects the check
to include the status of that flag.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -681,7 +681,7 @@ class Toil(object):
         else:
             raise RuntimeError('Unrecognised batch system: %s' % config.batchSystem)
 
-        if not batchSystemClass.supportsWorkerCleanup():
+        if not config.disableCaching and not batchSystemClass.supportsWorkerCleanup():
             raise RuntimeError('%s currently does not support shared caching.  Set the '
                                '--disableCaching flag if you want to '
                                'use this batch system.' % config.batchSystem)


benedictpaten:
jenkins, test this please.

src/toil/common.py
```diff
@@ -681,7 +681,7 @@ def createBatchSystem(config, jobStore=None, userScript=None):
         else:
             raise RuntimeError('Unrecognised batch system: %s' % config.batchSystem)
 
-        if not batchSystemClass.supportsWorkerCleanup():
+        if not batchSystemClass.supportsWorkerCleanup() and not config.disableCaching:
```
hannes-ucsc: Lets exploit boolean shortcut here and test for config.disableCaching first by swapping the sides of the conjunction. This is not just an optimization, I wouldn't ask for it if it were just that, I don' want the BS to be needlessly interrogated because there could be side-effects.

hannes-ucsc:
Ignore the failing jenkins-it check. This is because the branch isn't based on the latest master, but that's ok for this particular set of changes.

chapmanb:
Thanks for reviewing. I swapped it the other way, hopefully this works.

hannes-ucsc:
Jenkins, ok to test.

hannes-ucsc:
These two tests failures are expected and were fixed in #1092 by e684f5a.

PR #1095, opened 2016-08-01
Include AWS region in ToilTest class (resolves #998)

### LINKED ISSUE #998: Consolidate mentions of 'us-west-2' in AWS-related tests
Ideally, tests should use the region they are bing run in, if they are run on EC2.

cket:


src/toil/test/__init__.py
```diff
@@ -48,6 +54,10 @@ class ToilTest(unittest.TestCase):
     _tempBaseDir = None
     _tempDirs = None
 
+    # Use us-west-2 unless running on EC2, in which case use the region in which
+    # the instance is located
+    AWSRegion = _region() if runningOnEC2() else 'us-west-2'
```
cket: The property decorator can only be used with an instance function. Since the ToilClass is not always instantiated when the region is referred to, I made AWSRegion a class variable instead of a property.

cket:
jenkins, test this please.

src/toil/test/__init__.py
cket: I just found a downside to this approach. This code will throw NameErrors if the functions are not defined above their usage (_region, runningOnEC2, and functions used within those functions). I think this is because it's run at import time since it's class level and not protected by __init__...

cket:
The current approach has the desired interface but its a little hacky. I think it would be better to change the AWSRegion variable into a regular function and live with the slightly uglier invocation rather than risk breaking things if the code is reorganized.

src/toil/test/__init__.py
hannes-ucsc: Attributes (class or instance) should be lower-case.

hannes-ucsc: Let's just make awsRegion an `@classmethod`, and the support functions too. That would keep all the related code in a single place. We'd be trading in the omission of `()` but that's an acceptable trade-off.

hannes-ucsc:
> regular function

Agreed.

cket:
jenkins, test this please

cket:
@hannes-ucsc test failures are unrelated

PR #1101, opened 2016-08-02
Add native flags support for SLURM, via TOIL_SLURM_ARGS, (resolves issue #918)

brainstorm:
This supports SLURM native flags in a similar way [SGE does](https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/gridengine.py#L177).

This is hopefully a continuation and closer to closing issue #918 and siblings.

@chapmanb

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -193,6 +193,15 @@ class Worker(Thread):
         if cpu is not None:
             sbatch_line.append('--cpus-per-task={}'.format(int(math.ceil(cpu))))
 
+        # "Native extensions" for SLURM (see DRMAA or SAGA)
+        nativeConfig = os.getenv('TOIL_SLURM_ARGS')
+        if nativeConfig is not None:
+            logger.debug("Native SLURM options appended to sbatch from TOIL_SLURM_RESOURCES env. variable: {}".format(nativeConfig))
+            if "--mem" or "--cpus-per-task" in nativeConfig:
+                raise ValueError("Some resource arguments are incompatible: {}".format(nativeConfig))
+
+            sbatch_line.extend([nativeConfig])
+
         return sbatch_line
 
     def sbatch(self, sbatch_line):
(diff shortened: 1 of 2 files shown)


benedictpaten:
Jenkins, test  this please

cket:
@brainstorm thanks for the PR! Would this fix be enough to close #918? If so, would you mind adding (resolves #918) to the end of the PR so the ticket and the PR are linked? 

Besides that, this looks good to me but I don't have any expertise with the SLURM batch system so pinging @benedictpaten for final review/ merge.

brainstorm:
Thanks @cket, @benedictpaten for the attention to this PR!

benedictpaten:
LGTM! I'm merging this. I personally dislike these global environment variables, and would prefer a better hierarchical system for config that would allow these options to be specified, but I'll leave that as a future option.

src/toil/batchSystems/slurm.py
```diff
@@ -193,6 +193,15 @@ def prepareSbatch(self, cpu, mem, jobID):
         if cpu is not None:
             sbatch_line.append('--cpus-per-task={}'.format(int(math.ceil(cpu))))
 
+        # "Native extensions" for SLURM (see DRMAA or SAGA)
+        nativeConfig = os.getenv('TOIL_SLURM_ARGS')
+        if nativeConfig is not None:
+            logger.debug("Native SLURM options appended to sbatch from TOIL_SLURM_RESOURCES env. variable: {}".format(nativeConfig))
```
hannes-ucsc: For future reference, this should have been `… variable: %s", nativeConfig)`

src/toil/batchSystems/slurm.py
```diff
@@ -193,6 +193,15 @@ def prepareSbatch(self, cpu, mem, jobID):
...
+            if "--mem" or "--cpus-per-task" in nativeConfig:
+                raise ValueError("Some resource arguments are incompatible: {}".format(nativeConfig))
+
+            sbatch_line.extend([nativeConfig])
```
hannes-ucsc: why not `.append(nativeConfig)`?

src/toil/batchSystems/slurm.py
brainstorm: Was just trying to future-proof it:

https://pyformat.info/

`%s` is old format.

hannes-ucsc: % formatting isn't going to go away anytime soon. Letting the logger do the interpolation is  more concise and more efficient because the interpolation is performed only if the logging threshold is at debug.

PR #1105, opened 2016-08-03
GridEngine: adjust qsub with TOIL_GRIDENGINE_ARGS

chapmanb:
- Provides support for specifying TOIL_GRIDENGINE_ARGS to pass
  queue and other information to qsub. Environmental variable
  approach synchronized with #1101
- Documentation on environmental variables
- Fix and allows debugging on qsub command line. Previous version
  was missing -V so would not pass through environmental variables, and
  also swallowed all qsub error output making it hard to debug problems.

SGE with cwltoil works cleanly with these improvements.

### DIFF
--- a/docs/batchSystem.rst
+++ b/docs/batchSystem.rst
@@ -4,9 +4,22 @@ The batch system interface
 ==========================
 
 The batch system interface is used by Toil to abstract over different ways of running
-batches of jobs, for example GridEngine, Mesos, Parasol and a single node. The 
+batches of jobs, for example Slurm, GridEngine, Mesos, Parasol and a single node. The 
 :class:`toil.batchSystems.abstractBatchSystem.AbstractBatchSystem` API is implemented to
 run jobs using a given job management system, e.g. Mesos.
+
+Environmental variables allow passing of scheduler specific parameters. 
+
+For SLURM::
+
+    export TOIL_SLURM_ARGS="-t 1:00:00 -q fatq"
+
+For GridEngine (SGE, UGE), there is an additional environmental variable to define the
+`parallel environment <https://blogs.oracle.com/templedf/entry/configuring_a_new_parallel_environment>`_
+for running multicore jobs::
+
+    export TOIL_GRIDENGINE_PE='smp'
+    export TOIL_GRIDENGINE_ARGS='-q batch.q'
 
 .. autoclass:: toil.batchSystems.abstractBatchSystem::AbstractBatchSystem
    :members:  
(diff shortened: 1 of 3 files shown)


benedictpaten:
jenkins, test this please.

cket:
@chapmanb thanks for the PR! The Jenkins build failed because there were some files left behind by one of the tests. 
toil_job_0.o1
toil_job_0.o4
toil_job_1.o2
toil_job_1.o5
toil_job_2.o3
were left behind, are those produced by Gridengine? Is there any way to insure those are cleaned up?

chapmanb:
Thank so much for looking at the PR. This does change the gridengine so the logs are maintained, since I found them useful for debugging. Previously they got sent do /dev/null, so there were no extra files but it was also hard to figure out why things failed. I guess I'm trying to error on the side of providing more logs here. Is it required to not leave these behind? If so, is there a target location to do cleanup on success so we could remove these if the job finished successfully and leave them behind for debugging if not?

cket:
The logs don't have to be cleaned up after each workflow invocation but they should be deleted after each unittest to minimize side effects and keep logs from building up on disk (the Mesos batch system does this in the MesosTestSupport class in Toil, for reference).

Where are the logs being written? I'm not familiar with Grid Engine but Mesos' logging directory is outside of Toil's control. If we _can_ set Grid Engine's work directory from within Toil then it probably makes sense to have Grid Engine respect Toil's --workDir flag. What do you think?

chapmanb:
Christopher -- thanks for the test cleanup pointer, that's a great idea to cleanup the tests without removing the ability to have these. Gridengine should write these to the current working directory (with the `-cwd` flag to qsub) so have the same behavior as SLURM. I pushed a fix to the test scripts to clean this up, hopefully that makes the tests happy.

hannes-ucsc:
Jenkins, ok to test.

hannes-ucsc:
@cket, @benedictpaten: The comment I just made enables a PR to be tested automatically on every push.

hannes-ucsc:
I don't know why Jenkins claims that no test results were found but http://jenkins.cgcloud.info/job/toil-pull-requests/1721/consoleText contains some pretty obvious type errors at the end.

chapmanb:
Hannes -- is this something I need to fix? I don't have access to jenkins test outputs so not sure what the next steps are. Thanks for the help getting this integrated.

hannes-ucsc:
Brad, I added you to our "Jenkins Guests" group. So you should now have permissions to view that log. Let me know if that's not the case. And yes, I think the error from the logs is something in your code that you can fix.

chapmanb:
Thanks for the group permissions, that was a big help for catching my copy/paste typo. Sorry about the mistake, it's updated now and looks like the tests restarted. Fingers crossed this passes cleanly.

hannes-ucsc:
Test failures will be addressed by #1118. I'll merge this once #1118 has been merged.

PR #1112, opened 2016-08-19
Eliminate double autoscaling by fixing issued job code (resolves #1086)

### LINKED ISSUE #1086: Auto-scaling creates twice as many nodes as needed
… because non-preemptable nodes are scaled up even though all jobs are preemptable.

fnothaft:
Currently, the autoscaler launches both a preemptable and a non-preemptable instance when growing the cluster. This occurs because both cluster scaler threads looks at the total number of issued jobs, instead of just looking at the specific number of (non-)preemptable jobs issued. This commit changes the `toil.leader.JobBatcher` `getNumberOfJobsIssued` method to add a preemptable flag. This flag is then used by the two cluster scalers so that each thread looks only at their specific type of job.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -355,6 +355,12 @@ def _addOptions(addGroupFn, config):
     addOptionFn("--scaleInterval", dest="scaleInterval", default=None,
                 help=("The interval (seconds) between assessing if the scale of"
                       " the cluster needs to change. default=%s" % config.scaleInterval))
+    addOptionFn("--slackPreemptablePreference", dest="slackPreemptablePreference",
+                default=None,
+                help=("The preference for the autoscaler to replace non-preemptable nodes"
+                      " with preemptable nodes, when preemptable nodes cannot be started."
+                      " Defaults to %s. This value must be between 0 and 1,"
+                      " inclusive." % config.slackPreemptablePreference))
 
     #
     #Resource requirements
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -77,6 +77,7 @@ class Config(object):
         self.alphaPacking = 0.8
         self.betaInertia = 1.2
         self.scaleInterval = 10
+        self.slackPreemptablePreference = 0.0
 
         #Resource requirements
         self.defaultMemory = 2147483648
@@ -195,6 +196,7 @@ class Config(object):
         setOption("alphaPacking", float)
         setOption("betaInertia", float)
         setOption("scaleInterval", float)
+        setOption("slackPreemptablePreference", float)
 
         #Resource requirements
         setOption("defaultMemory", h2b, iC(1))
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -28,6 +28,14 @@ from toil.provisioners.abstractProvisioner import AbstractProvisioner, Shape
 
 logger = logging.getLogger(__name__)
 
+# slack exists when we have more jobs that can run on preemptable nodes than
+# we have preemptable nodes. in order to not block these jobs, we want to scale
+# up the number of non-preemptable nodes that we have. however, we may still
+# prefer waiting for preemptable instances to come available.
+#
+# to accomodate this, we set the slack to the delta between requested and
+# provisioned preemptable nodes times a preemptible node preference factor.
+_preemptableQueueSlack = 0
 
 class RecentJobShapes(object):
     """
@@ -282,6 +290,10 @@ class ScalerThread(ExceptionalThread):
         self.maxNodes = scaler.config.maxPreemptableNodes if preemptable else scaler.config.maxNodes
 
     def tryRun(self):
+
+        # we will make reference to the global preemptable slack variable
+        global _preemptableQueueSlack
+
         if isinstance(self.scaler.jobBatcher.batchSystem, AbstractScalableBatchSystem):
             totalNodes = len(self.scaler.jobBatcher.batchSystem.getNodes(self.preemptable))
         else:
@@ -292,6 +304,12 @@ class ScalerThread(ExceptionalThread):
                 # Calculate the approx. number nodes needed
                 # TODO: Correct for jobs already running which can be considered fractions of a job
                 queueSize = self.scaler.jobBatcher.getNumberOfJobsIssued(preemptable=self.preemptable)
+
+                # if we're in the non-preemptable queue, we need to see if we have any slack
+                # coming over from the preemptable queue
+                if not self.preemptable:
+                    queueSize += _preemptableQueueSlack
+
                 recentJobShapes = self.jobShapes.get()
                 assert len(recentJobShapes) > 0
                 nodesToRunRecentJobs = binPacking(recentJobShapes, self.nodeShape)
@@ -331,6 +349,26 @@ class ScalerThread(ExceptionalThread):
                                 estimatedNodes)
                     totalNodes = self.scaler.provisioner.setNodeCount(numNodes=estimatedNodes,
                                                                       preemptable=self.preemptable)
+                    
+                    # if we were scaling up the number of preemptable nodes and failed to
+                    # meet our target, we need to update the slack so that non-preemptable
+                    # nodes will be allocated and we won't block. if we _did_ meet our target,
+                    # we need to reset the slack to 0
+                    if totalNodes < estimatedNodes and self.preemptable:
+                        
+                        # slack is derived from the delta (the number of nodes we did _not_ allocate)
+                        # times a preference for preemptable nodes
+                        require(self.scaler.config.slackPreemptablePreference >= 0.0 and
+                                self.scaler.config.slackPreemptablePreference <= 1.0,
+                                "Slack preference for preemptable nodes (%f) must be >= 0.0 and <= 1.0" % self.scaler.config.slackPreemptablePreference)
+                        nonPreemptablePreference = 1.0 - self.scaler.config.slackPreemptablePreference
+                        delta = estimatedNodes - totalNodes
+                        _preemptableQueueSlack = int(round(delta * nonPreemptablePreference))
+                        logger.debug('Had delta of %d, setting slack to %d.',
+                                     delta, _preemptableQueueSlack)
+                    else:
+                        _preemptableQueueSlack = 0
+                    
         logger.info('Forcing provisioner to reduce cluster size to zero.')
         totalNodes = self.scaler.provisioner.setNodeCount(numNodes=0,
                                                           preemptable=self.preemptable,
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -291,7 +291,7 @@ class ScalerThread(ExceptionalThread):
             with throttle(self.scaler.config.scaleInterval):
                 # Calculate the approx. number nodes needed
                 # TODO: Correct for jobs already running which can be considered fractions of a job
-                queueSize = self.scaler.jobBatcher.getNumberOfJobsIssued()
+                queueSize = self.scaler.jobBatcher.getNumberOfJobsIssued(preemptable=self.preemptable)
                 recentJobShapes = self.jobShapes.get()
                 assert len(recentJobShapes) > 0
                 nodesToRunRecentJobs = binPacking(recentJobShapes, self.nodeShape)
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -290,8 +290,6 @@ class ScalerThread(ExceptionalThread):
         self.maxNodes = scaler.config.maxPreemptableNodes if preemptable else scaler.config.maxNodes
 
     def tryRun(self):
-
-        # we will make reference to the global preemptable slack variable
         global _preemptableQueueSlack
 
         if isinstance(self.scaler.jobBatcher.batchSystem, AbstractScalableBatchSystem):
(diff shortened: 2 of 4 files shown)


hannes-ucsc:
IIRC, preemptable jobs will be scheduled on non-preemptable nodes. Does this change still work if the user has only preemptable jobs but has disabled preemptable nodes from being provisioned? In that case no nodes would be created, right?

fnothaft:
@hannes-ucsc ah yes... Good point. A lame (but simple) solution would be to allow the preemptable thread to launch non-preemptable images in that case. Arguably, a better solution might be to just launch one scaler thread if you've disabled preemptable machines, and to have this look at the whole job queue (simple flag change in this PR).

How exactly does the user disable preemptable nodes from being launched? I.e., what is the command line option? I'll look through that code path, and see what else makes sense.

hannes-ucsc:
> How exactly does the user disable preemptable nodes from being launched

`--maxPreemptableNodes=0` IIRC

A too low spot bid may also have that effect, inadvertently.

> Arguably, a better solution might be to just launch one scaler thread if you've disabled preemptable machines, and to have this look at the whole job queue (simple flag change in this PR).

Yes. What do you think about implementing some kind of overflow from the P scaler to the NP scaler? There is a point in the code path in which the P scaler knows how many nodes should ideally be provisioned, call it i<sub>p</sub>. i<sub>p</sub> then undergoes the application of the maximum (and minimum) limits before being given to the provisioner. The provisioner returns how many nodes where actually provisioned, call it a<sub>p</sub>. What if we saved d<sub>p</sub> = i<sub>p</sub> - a<sub>p</sub> in a global variable such that the NP scaler can i<sub>np</sub> += d<sub>p</sub>.

fnothaft:
@hannes-ucsc I've added an implementation of the slack concept that we discussed yesterday in a second commit (https://github.com/BD2KGenomics/toil/pull/1112/commits/0bb89cf7868312dfd2d02bf6ef289a9cd66a7705). Does this implementation look faithful to what we discussed? Let me know your thoughts. I'm OOO at some meetings tomorrow, so I may not be on scrum, but I will definitely be on Friday morning's scrum.

src/toil/provisioners/clusterScaler.py
```diff
@@ -331,6 +349,26 @@ def tryRun(self):
...
+                    
+                    # if we were scaling up the number of preemptable nodes and failed to
+                    # meet our target, we need to update the slack so that non-preemptable
+                    # nodes will be allocated and we won't block. if we _did_ meet our target,
+                    # we need to reset the slack to 0
+                    if totalNodes < estimatedNodes and self.preemptable:
+                        
+                        # slack is derived from the delta (the number of nodes we did _not_ allocate)
+                        # times a preference for preemptable nodes
+                        require(self.scaler.config.slackPreemptablePreference >= 0.0 and
```
hannes-ucsc: In Python, you can do this: `0 <= self.scaler.config.slackPreemptablePreference <= 1`

src/toil/provisioners/clusterScaler.py
```diff
@@ -331,6 +349,26 @@ def tryRun(self):
...
+                                self.scaler.config.slackPreemptablePreference <= 1.0,
+                                "Slack preference for preemptable nodes (%f) must be >= 0.0 and <= 1.0" % self.scaler.config.slackPreemptablePreference)
```
hannes-ucsc: Replace percent with comma and let require do the interpolation lazily.

src/toil/provisioners/clusterScaler.py
```diff
@@ -282,6 +290,10 @@ def __init__(self, scaler, preemptable):
         self.maxNodes = scaler.config.maxPreemptableNodes if preemptable else scaler.config.maxNodes
 
     def tryRun(self):
+
+        # we will make reference to the global preemptable slack variable
```
hannes-ucsc: This is obvious.

src/toil/provisioners/clusterScaler.py
```diff
@@ -291,7 +303,13 @@ def tryRun(self):
             with throttle(self.scaler.config.scaleInterval):
                 # Calculate the approx. number nodes needed
                 # TODO: Correct for jobs already running which can be considered fractions of a job
-                queueSize = self.scaler.jobBatcher.getNumberOfJobsIssued()
+                queueSize = self.scaler.jobBatcher.getNumberOfJobsIssued(preemptable=self.preemptable)
+
+                # if we're in the non-preemptable queue, we need to see if we have any slack
+                # coming over from the preemptable queue
+                if not self.preemptable:
+                    queueSize += _preemptableQueueSlack
```
hannes-ucsc: I think we should add slack to estimatedNodes further down. The unit of queueSize is # of jobs, the unit of estimatedNodes is # of nodes.

hannes-ucsc:
Done with review. A few nits and one substantial issue (the one about whether we should add to queueSize or estimatedNodes.

src/toil/provisioners/clusterScaler.py
fnothaft: Agreed, that SGTM.

fnothaft:
Rebased and pushed a commit https://github.com/BD2KGenomics/toil/pull/1112/commits/195d4ceb77a9ff49552509169c1f50ba2cf80604 to address review comments.

hannes-ucsc:
Looks like clusterScalingTest and testAutoScaledCluster timed out:

http://jenkins.cgcloud.info/job/toil-pull-requests/1757/console
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/68/console

Jenkins, test this please.

fnothaft:
I probably messed up something. I'll look on my end.

fnothaft:
@hannes-ucsc just pushed a fix for the unit test errors, but it looks like the integration tests fail. I'll take another look tomorrow.

fnothaft:
@hannes-ucsc can I get you to look at the integration test failure. It looks like this failure is inside of the test that ran, not in the autoscaling code. Was this a spurious test failure, or was the `UserError` that the integration test threw expected?

hannes-ucsc:
Thanks, @fnothaft. The UserError is a bug in toil-scripts:

https://github.com/BD2KGenomics/toil-scripts/issues/436

Jenkins, test this please.

fnothaft:
Thanks for catching that @hannes-ucsc! I was looking through other `jenkins-it` runs that were passing and was not seeing a trend!

fnothaft:
Also, if the tests pass and the implementation looks good to you, would you like me to squash on down?

src/toil/provisioners/clusterScaler.py
```diff
@@ -331,6 +347,26 @@ def tryRun(self):
...
+                    if totalNodes < estimatedNodes and self.preemptable:
+                        
+                        # slack is derived from the delta (the number of nodes we did _not_ allocate)
+                        # times a preference for preemptable nodes
+                        require(1.0 >= self.scaler.config.slackPreemptablePreference >= 0.0,
+                                "Slack preference for preemptable nodes (%f) must be >= 0.0 and <= 1.0",
+                                self.scaler.config.slackPreemptablePreference)
+                        nonPreemptablePreference = 1.0 - self.scaler.config.slackPreemptablePreference
+                        delta = estimatedNodes - totalNodes
+                        _preemptableQueueSlack = int(round(delta * nonPreemptablePreference))
```
hannes-ucsc: Can we have the producer (the preemptable scaler thread) write the raw value to the global variable and apply the preference factor in the consumer? The global variable should reflect how many preemptable nodes are currently missing. This information can be used in various ways. Only one of them is to replace them with preemptable notes. But there may be others. Publishing the raw value would enable those other ways in the future. Last but not least, you wouldn't have to mention the preference factor when you document the variable.

src/toil/common.py
```diff
@@ -187,6 +188,7 @@ def fC(minValue, maxValue=None):
         setOption("alphaPacking", float)
         setOption("betaInertia", float)
         setOption("scaleInterval", float)
+        setOption("slackPreemptablePreference", float)
```
hannes-ucsc: Looks like the actual argparse option definition is missing?

hannes-ucsc:
Wait with the squashing, I would say.

src/toil/provisioners/clusterScaler.py
fnothaft: Yeah, that sounds reasonable. Shouldn't be a hard change. I'll do that tonight/tomorrow.

hannes-ucsc:
One more thing we didn't think about: The node types may be different, so one NP node may not replace one P node. But I guess that's what the preference factor is for.

hannes-ucsc:
… one P node may not be an appropriate substitute for one P node …

fnothaft:
> One more thing we didn't think about: The node types may be different, so one NP node may not replace one P node. But I guess that's what the preference factor is for.

I don't think there's a straightforward way to handle that anyways.

fnothaft:
Er, not "straightforward", but "rigorous".

fnothaft:
@hannes-ucsc just addressed your two comments. Let's see if the tests passed/what I messed up this time. ;)

src/toil/common.py
```diff
@@ -347,6 +347,12 @@ def _addOptionFn(*name, **kwargs):
     addOptionFn("--scaleInterval", dest="scaleInterval", default=None,
                 help=("The interval (seconds) between assessing if the scale of"
                       " the cluster needs to change. default=%s" % config.scaleInterval))
+    addOptionFn("--slackPreemptablePreference", dest="slackPreemptablePreference",
+                default=None,
+                help=("The preference for the autoscaler to replace non-preemptable nodes"
+                      " with preemptable nodes, when preemptable nodes cannot be started."
+                      " Defaults to %s. This value must be between 0 and 1,"
+                      " inclusive." % config.slackPreemptablePreference)
```
hannes-ucsc: `)` missing at the end.

src/toil/common.py
hannes-ucsc: Fixed by 5931c5f.

fnothaft:
This last run of the integration tests threw an error inside of Boto:

```
boto.exception.S3ResponseError: S3ResponseError: 409 Conflict
<Error><Code>OperationAborted</Code><Message>A conflicting conditional operation is currently in progress against this resource. Please try again.</Message><RequestId>5F7C1887D8B8D26A</RequestId><HostId>xBKJbG+mCJsEFzevDconu9QBYf7+P9LxTZIwFyUFuMnu2qv5YEPzwWq8bG7hTfntt/QD0xusq4E=</HostId></Error>
```

@hannes-ucsc have you seen this before?

cket:
@fnothaft IIRC that error occurs when multiple processes are messing with the same S3 file at the same time. Each process asserts that it's modifying a specific version but as soon as 1 operation succeeds the version is updated and subsequent operations fail

fnothaft:
Jenkins, retest this please.

src/toil/provisioners/clusterScaler.py
```diff
@@ -331,6 +356,19 @@ def tryRun(self):
...
+                    # meet our target, we need to update the slack so that non-preemptable
+                    # nodes will be allocated and we won't block. if we _did_ meet our target,
+                    # we need to reset the slack to 0
+                    if totalNodes < estimatedNodes and self.preemptable:
+                        
+                        # slack is derived from the delta (the number of nodes we did _not_ allocate)
+                        _delta = estimatedNodes - totalNodes
+                        logger.debug('Preemptable thread had delta of %d.', _delta)
+                    else:
+                        _delta = 0
```
hannes-ucsc: This causes the NP thread to continually reset _delta.

src/toil/provisioners/clusterScaler.py
```diff
@@ -331,6 +356,19 @@ def tryRun(self):
                                 estimatedNodes)
                     totalNodes = self.scaler.provisioner.setNodeCount(numNodes=estimatedNodes,
                                                                       preemptable=self.preemptable)
+                    
+                    # if we were scaling up the number of preemptable nodes and failed to
+                    # meet our target, we need to update the slack so that non-preemptable
+                    # nodes will be allocated and we won't block. if we _did_ meet our target,
+                    # we need to reset the slack to 0
+                    if totalNodes < estimatedNodes and self.preemptable:
+
```
hannes-ucsc: Empty line not necessary here.

hannes-ucsc:
Do we need the test for this? The test could specify a spot bid that is too low to be ever fulfilled, enable the slack factor and only schedule preemptable jobs by setting defaultPreemptable to True.

fnothaft:
I won't be on scrum today, but I'll make a pass to address these comments.

hannes-ucsc:
@cket is in the process of adding another integration test for his fix to #1148. He refactored the CgcloudProvisionerTest  class to make it easier to share more test setup and have other workflows to be tested (currently only the rnaseq pipeline is being run as part of the integration tests). His refactorings should make it easier for you, @fnothaft, to add a small integration test that exercises the slack stuff.

fnothaft:
Sounds good @hannes-ucsc. I'll wait for @cket's #1148 patch to go ahead first.

src/toil/provisioners/clusterScaler.py
hannes-ucsc: IOW, I think this should read

```py
if self.preemptable:
    if totalNodes < estimatedNodes:
        # slack is derived from the delta (the number of nodes we did _not_ allocate)
        _delta = estimatedNodes - totalNodes
        logger.debug('Preemptable thread had delta of %d.', _delta)
    else:
        _delta = 0
```

fnothaft:
CC @almussel

hannes-ucsc:
I'll be adding tests.

hannes-ucsc:
Continued in #1176

PR #1123, opened 2016-08-26
Insure default resource requirement values are set correctly (resolves #1119)

### LINKED ISSUE #1119: Fix default resource attributes for JobFunctionWrappingJob
We are now using JobFunctionWrappingJob resource attributes to configure function calls, but currently the resource attributes default to None when the job function is called. 

@cket

cket:
First commit modifies sort test to cover this bug.


hannes-ucsc:
@cket and I discussed this. He wants to try out a property-based approach that relies on config being injected during unpickling.

cket:
Closing, continuing in #1133

PR #1133, opened 2016-08-30
Use config to discover proper resource values (resolves #1119)

### LINKED ISSUE #1119: Fix default resource attributes for JobFunctionWrappingJob

cket:
Fixes this bug in both Job and Service resource requirements

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -41,55 +41,76 @@ from toil.resource import ModuleDescriptor
 logger = logging.getLogger( __name__ )
 
 
-class Job(object):
+class ResourceRequirementMixin(object):
     """
-    Class represents a unit of work in toil.
+    Inherit from this class to add requirement properties to a job (or job-like) object.
+    If the object doesn't specify explicit requirements, these properties will fall back
+    to the configured defaults. If the value cannot be determined, an AttributeError is raised.
     """
-    def __init__(self, memory=None, cores=None, disk=None, preemptable=None, checkpoint=False):
-        """
-        This method must be called by any overriding constructor.
-        
-        :param memory: the maximum number of bytes of memory the job will require to run.
-        :param cores: the number of CPU cores required.
-        :param disk: the amount of local disk space required by the job, expressed in bytes.
-        :param preemptable: if the job can be run on a preemptable node.
-        :param checkpoint: if any of this job's successor jobs completely fails,
-            exhausting all their retries, remove any successor jobs and rerun this job to restart the
-            subtree. Job must be a leaf vertex in the job graph when initially defined, see
-            :func:`toil.job.Job.checkNewCheckpointsAreCutVertices`.
-        :type cores: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type disk: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type preemptable: boolean
-        :type cache: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        :type memory: int or string convertable by bd2k.util.humanize.human2bytes to an int
-        """
-        self.cores = self._parseResource('cores', cores)
-        self.memory = self._parseResource('memory', memory)
-        self.disk = self._parseResource('disk', disk)
-        self.checkpoint = checkpoint
-        self.preemptable = preemptable
-        #Private class variables
+    def __init__(self, memory=None, cores=None, disk=None, preemptable=None):
+        self._cores = self._parseResource('cores', cores)
+        self._memory = self._parseResource('memory', memory)
+        self._disk = self._parseResource('disk', disk)
+        self._preemptable = preemptable
+        self._config = None
 
-        #See Job.addChild
-        self._children = []
-        #See Job.addFollowOn
-        self._followOns = []
-        #See Job.addService
-        self._services = []
-        #A follow-on, service or child of a job A, is a "direct successor" of A, if B
-        #is a direct successor of A, then A is a "direct predecessor" of B.
-        self._directPredecessors = set()
-        # Note that self.__module__ is not necessarily this module, i.e. job.py. It is the module
-        # defining the class self is an instance of, which may be a subclass of Job that may be
-        # defined in a different module.
-        self.userModule = ModuleDescriptor.forModule(self.__module__)
-        # Maps index paths into composite return values to lists of IDs of files containing
-        # promised values for those return value items. An index path is a tuple of indices that
-        # traverses a nested data structure of lists, dicts, tuples or any other type supporting
-        # the __getitem__() protocol.. The special key `()` (the empty tuple) represents the
-        # entire return value.
-        self._rvs = collections.defaultdict(list)
-        self._promiseJobStore = None
+    @property
+    def disk(self):
+        """
+        The maximum number of bytes of disk the job will require to run.
+        """
+        if self._disk is not None:
+            return self._disk
+        elif self._config is not None:
+            return self._config.defaultDisk
+        else:
+            raise AttributeError("Default value for 'disk' cannot be determined")
+
+    @property
+    def memory(self):
+        """
+        The maximum number of bytes of memory the job will require to run.
+        """
+        if self._memory is not None:
+            return self._memory
+        elif self._config is not None:
+            return self._config.defaultMemory
+        else:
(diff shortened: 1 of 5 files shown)


cket:
@hannes-ucsc

src/toil/job.py
```diff
@@ -30,6 +30,7 @@
 from contextlib import contextmanager
 from io import BytesIO
 
+from bd2k.util import memoize
```
hannes-ucsc: Unused import?

src/toil/job.py
```diff
@@ -41,55 +42,54 @@
...
 
 
-class Job(object):
+class ResourceRequirementMixin(object):
     """
-    Class represents a unit of work in toil.
+    Inherit from this class to add resource requirements for an object
     """
-    def __init__(self, memory=None, cores=None, disk=None, preemptable=None, checkpoint=False):
+    def __init__(self, memory=None, cores=None, disk=None):
```
hannes-ucsc: So IIRC, preemptability is also a requirement. I think there is even a defaultPreemptable. We should add that to the mixin.

src/toil/job.py
```diff
@@ -41,55 +42,54 @@
...
+        self._cores = self._parseResource('cores', cores)
+        self._memory = self._parseResource('memory', memory)
+        self._disk = self._parseResource('disk', disk)
+        self._config = None
+
+    @property
+    def disk(self):
+        """
+        This property insures that even if the user does not pass a resource requirment, the
```
hannes-ucsc: This paragraph is an implementation note. It should be a # comment in the body of the method. The docstring should describe the semantics of the property, e.g. "The amount of free local disk required on the machine this job is run on" or so.

hannes-ucsc: Also no need to duplicate that # comment in every method. Maybe it should go above the first property.

src/toil/job.py
```diff
@@ -41,55 +42,54 @@
...
+        default value will be available when the user's code is run in Job.run or in
+        Job.Service.start. If a value is not set and the Job is not yet ready to be run, this method
+        will return None.
+        """
+        if self._disk or self._config:
```
hannes-ucsc: I forgot the specifics, but doesn't calling this method without _config set constitute an error?

src/toil/job.py
```diff
@@ -798,10 +848,12 @@ def effectiveRequirements(self, config):
         :rtype: Expando
         :return: a dictionary/object hybrid with one entry/attribute for each requirement
         """
+        if self._config is None:
```
hannes-ucsc: Is there any way we can eliminate this method? Maybe _config is already set in all call sites of this method?

cket: I thought about eliminating this but effectiveRequirements is still useful as a way to wrap up resource requirements into a dictionary/expando. What do you think? IMO, it's pretty convenient especially for argument unpacking so I'm in favor of keeping it.

src/toil/job.py
cket: You're right, it was going to be an error but there is a case in ServiceJob where we refer to the resources without _config set, in which case we want to return None and pass the burden of determining the actual default values into Job. Alternatively, I could make this an error and try-except it in ServiceJob.

hannes-ucsc: I'd prefer the latter as it doesn't sacrifice strictness for one special case. If the scenario of _config being unset were more common, raising an exception would place a burden on many call sites but based on what you say, that doesn't seem to be the case here.

src/toil/job.py
hannes-ucsc: Cool!

cket:
@hannes-ucsc

src/toil/job.py
```diff
@@ -1282,7 +1326,11 @@ def __init__(self, job):
         :param toil.job.Job job: the job to encapsulate.
         """
         # Giving the root of the subgraph the same resources as the first job in the subgraph.
-        Job.__init__(self, disk=job.disk, memory=job.memory, cores=job.cores)
+        # The service resource requirments are not guaranteed to be set yet so default to None
+        Job.__init__(self, memory=getattr(job, 'memory', None),
```
hannes-ucsc: This makes me curious. Under what circumstance can `job` not have the `memory` attribute?

hannes-ucsc: And now I read your comment above. Ignore the question but lets discuss after stand-up.

hannes-ucsc:
@cket to try out the `**job._resources()` approach.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
IT passed. Any idea why some unit tests failed?

cket:
jenkins, test this please

src/toil/job.py
```diff
@@ -1282,7 +1331,7 @@ def __init__(self, job):
         :param toil.job.Job job: the job to encapsulate.
         """
         # Giving the root of the subgraph the same resources as the first job in the subgraph.
-        Job.__init__(self, disk=job.disk, memory=job.memory, cores=job.cores)
+        Job.__init__(self, **job._resources)
```
cket: The tests were failing because this was previously **self._resources which made the new job always have the default requirements no matter what was passed to the constructor.

hannes-ucsc: Sounds like we need `_resources` and `_effectiveResources`.

BTW, can we call those properties `_requirements` and `_effectiveRequirements` instead? That would avoid the conflict with the use of the term `resource` for hot deployment. It would also be consistent with PromisedRequirement etc. …

cket: These changes are now outdated- they are **job._requirements, as requested

cket:
@hannes-ucsc I addressed your comments

src/toil/job.py
```diff
@@ -41,55 +41,69 @@
...
 
 
-class Job(object):
+class ResourceRequirementMixin(object):
     """
-    Class represents a unit of work in toil.
+    Inherit from this class to add resource requirements for an object. These
+    properties insure that even if the user does not pass a resource requirement the
+    default value will be available when the user's code is run in Job.run or in
+    Job.Service.start. If the value cannot be determined an AttributeError is raised.
```
hannes-ucsc: comma after determined

hannes-ucsc:
```
Inherit from this class to add requirement properties to a job (or job-like) object. If the object doesn't specify explicit requirements, these properties will fall back to the configured defaults. If the value cannot …
```

src/toil/job.py
```diff
@@ -816,23 +878,10 @@ def _createEmptyJobWrapperForJob(self, jobStore, command=None, predecessorNumber
...
-        """
-        requirements = Expando(
-            memory=float(config.defaultMemory) if self.memory is None else self.memory,
-            cores=float(config.defaultCores) if self.cores is None else self.cores,
-            disk=float(config.defaultDisk) if self.disk is None else self.disk,
-            preemptable=config.defaultPreemptable if self.preemptable is None else self.preemptable)
-        return requirements
+        # set _config to determine user determined default values for resource requirements
+        self._config = jobStore.config
+        return jobStore.create(command=command, predecessorNumber=predecessorNumber,
```
hannes-ucsc: PEP-8 says wrap all kwargs or none.

src/toil/job.py
```diff
@@ -41,55 +41,69 @@
...
-        self._promiseJobStore = None
+    @property
+    def memory(self):
+        """
+        The maximum number of bytes of memory the job will require to run.
+        """
+        if self._memory or self._config:
+            return self._memory if self._memory is not None else self._config.defaultMemory
+        else:
+            raise AttributeError('Default value for memory cannot be determined')
```
hannes-ucsc: `Default value for 'memory' requirement cannot be determined`

Same everywhere else.

hannes-ucsc: There are unnecessary comparisons here. 

```
if self._memory is not None:
   return …
elif self._config is not None:
    return …
else:
    raise
```

hannes-ucsc:
Test failure could be a eventual consistency issue with S3.

Jenkins, test this please.

cket:
@hannes-ucsc this is ready to be reviewed

PR #1137, opened 2016-08-31
Use dedicated logger for worker output logs (resolves #1136)

### LINKED ISSUE #1136: Toil logs kinda suck
Debugging Toil pipelines is harder than it needs to be because the stacktrace returned by Toil Worker Output Log has each line prepended by unnecessary log information.

cket:
Resolves #1136

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -29,10 +29,12 @@ from bd2k.util.expando import Expando
 
 from toil import resolveEntryPoint
 from toil.jobStores.abstractJobStore import NoSuchJobException
-from toil.lib.bioio import getTotalCpuTime, logStream
+from toil.lib.bioio import getTotalCpuTime, logStream, getOutputLogger
 from toil.provisioners.clusterScaler import ClusterScaler
 
 logger = logging.getLogger( __name__ )
+
+
 
 ####################################################
 ##Stats/logging aggregation
@@ -56,6 +58,7 @@ class StatsAndLogging( object ):
         The following function is used for collating stats/reporting log messages from the workers.
         Works inside of a thread, collates as long as the stop flag is not True.
         """
+        outputLogger = getOutputLogger(__name__)
         #  Overall timing
         startTime = time.time()
         startClock = getTotalCpuTime()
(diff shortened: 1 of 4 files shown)


cket:
@jvivian can you test this out when you get a chance?

src/toil/leader.py
```diff
@@ -34,6 +34,15 @@
 
 logger = logging.getLogger( __name__ )
 
+# create new logger for worker output logs so we can format them appropriately
+outputLogger = logging.getLogger('workerOutput')
```
hannes-ucsc: I think this should be `__name__ + ".workerOutput" such that it is a child of the "toil" logger.

src/toil/leader.py
```diff
@@ -34,6 +34,15 @@
...
+
+logger_handler = logging.StreamHandler()
```
hannes-ucsc: Camel case.

hannes-ucsc:
Generally +1. But see comments.

hannes-ucsc:
Remember to add a ticket reference into the description or Waffle won't stack PR and issue.

cket:
Don't you just have to reference it in the title? I think they were stacked before

hannes-ucsc:
It wasn't stacked before and after adding the reference, it was.

hannes-ucsc:
But this suggests that a reference in the title should be enough:

https://github.com/waffleio/waffle.io/wiki/FAQs#prs-closing-keywords

cket:
Yeah, #1119 is stacked without a reference. Maybe waffle just needed to be refreshed

hannes-ucsc:
Ahh, but a reference in the title doesn't create a backlink from the issue to the PR.

cket:
Jenkins, test this please

cket:
The test is failing because it redirects logs by adding a handler to the root logger. It then parses the files looking for worker output, which aren't there because they are logged by the new outputLogger and specifically not propagated to the root logger.

The solution would be to propagate the logs back to the root logger, but that would also cause all the outputLogger's messages to be double logged.

cket:
closing, continued in #1177

Issue #1127, opened 2016-08-29
Restarted job forgets about user script


adamnovak:
My script runs fine in a new job store (apart from some failing jobs due to errors from the tools they run), but when I go to restart the job, I get this:

```
WARNING:toil.leader:The jobWrapper seems to have left a log file, indicating failure: 52d0d400_4080_480d_9c53_e7a7624b20e9
WARNING:toil.leader:Reporting file: 52d0d400_4080_480d_9c53_e7a7624b20e9
WARNING:toil.leader:52d0d400_4080_480d_9c53_e7a7624b20e9:       ---TOIL WORKER OUTPUT LOG---
WARNING:toil.leader:52d0d400_4080_480d_9c53_e7a7624b20e9:       WARNING:toil.resource:Can't find resource for leader path '/home/anovak/hgvm-graph-bakeoff-evaluations/scripts'
WARNING:toil.leader:52d0d400_4080_480d_9c53_e7a7624b20e9:       WARNING:toil.resource:Can't localize module ModuleDescriptor(dirPath='/home/anovak/hgvm-graph-bakeoff-evaluations/scripts', name='parallelMappingEvaluation')
WARNING:toil.leader:52d0d400_4080_480d_9c53_e7a7624b20e9:       ERROR:toil.job:Failed to import user module ModuleDescriptor(dirPath='/home/anovak/hgvm-graph-bakeoff-evaluations/scripts', name='parallelMappingEvaluation') from sys.path=['/usr/local/bin', '/usr/local/lib/python2.7/dist-packages/pip-8.1.2-py2.7.egg', '/usr/local/lib/python2.7/dist-packages/mesos.interface-0.22.0-py2.7.egg', '/usr/local/lib/python2.7/dist-packages/mesos-0.22.0-py2.7-linux-x86_64.egg', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/usr/local/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages', '/home/anovak/hgvm-graph-bakeoff-evaluations/scripts']
WARNING:toil.leader:52d0d400_4080_480d_9c53_e7a7624b20e9:       Traceback (most recent call last):
[...]

PR #1151, opened 2016-09-12
Autoscale now works with --restart (resolves #1148)

### LINKED ISSUE #1148: Autoscaling does not work with restarted pipeline
After a pipeline is stopped & its nodes are terminated, rerunning the script with --restart will not spin up any more workers.

cket:


hannes-ucsc:
The fix for #1127 was just merged. Please rebase.

cket:
Jenkins, test this please.

cket:
Jenkins, ok to test.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, ok to test.

hannes-ucsc:
Jenkins, test this please.

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,41 @@ def testAutoScaledCluster(self):
     @integrative
     def testAutoScaledSpotCluster(self):
         self._test(autoScaled=True, spotInstances=True)
+
+restartScript = textwrap.dedent("""\"
```
hannes-ucsc: Why is the escaped double quote necessary?

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,41 @@ def testAutoScaledCluster(self):
...
+                                    from toil.job import Job
```
hannes-ucsc: This should be indented by four spaces exactly.

src/toil/test/__init__.py
```diff
@@ -444,6 +445,29 @@ def integrative(test_item):
 methodNamePartRegex = re.compile('^[a-zA-Z_0-9]+$')
 
 
+@contextmanager
+def timeLimit(seconds):
+    """
+    http://stackoverflow.com/a/601168
+    Use to limit the execution of a function. Run the function wrapped in the context manager:
```
hannes-ucsc: `execution time`

src/toil/test/__init__.py
```diff
@@ -444,6 +445,29 @@ def integrative(test_item):
...
+    ```
```
hannes-ucsc: Remove triple backquotes and convert example to a doctest.

hannes-ucsc:
Looks like one of the `run_tests.py` child processes (the one that runs the added `--restart` integration tests) basically disappeared without creating a test report. My hunch is that this is caused by the time_limit context manager. OS signals can have behave counter-intuitively. Can you push a version that doesn't apply the time limit.

cket:
@hannes-ucsc this is ready to be reviewed

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,44 @@ def testAutoScaledCluster(self):
     @integrative
     def testAutoScaledSpotCluster(self):
         self._test(autoScaled=True, spotInstances=True)
+
+
+restartScript = """\"from toil.job import Job
```
hannes-ucsc: Can you use this pattern instead. It's a bit unconventional but I like it because it applies syntax checking to embedded scripts:

https://github.com/BD2KGenomics/toil/blob/master/src/toil/test/src/hotDeploymentTest.py#L44
https://github.com/BD2KGenomics/toil/blob/master/src/toil/test/__init__.py#L718

hannes-ucsc: And the embedded quotes should be added programmatically.

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,44 @@ def testAutoScaledCluster(self):
...
+    options = parser.parse_args()
+    i = Job.Runner.startToil(Job.wrapJobFn(f0), options)
+\""""
+
+
+class CGCloudRestartTest(AbstractCGCloudProvisionerTest):
+
+    def _getScript(self):
+        self.scriptName= "restartScript.py"
+        self._leader('echo %s > %s' % (restartScript, 'restartScript.py'), shell=True)
```
hannes-ucsc: `self._leader( "tee restartScript.py", input=…)` works for me.

hannes-ucsc: If not lets debug that together. The `echo` technique can break, if, for example, you had double quotes in your script.

hannes-ucsc: If, for some reason, you had to use `echo`, and to add another tool to your belt, look at the `quote` function from the `pipes` module.

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,44 @@ def testAutoScaledCluster(self):
...
+
+    def _runScript(self, toilOptions):
+        # clean = onSuccess
+        restartCompatible = lambda x : ('clean' not in x and 'retryCount' not in x)
+        newOptions = [option for option in toilOptions if restartCompatible(option)]
```
hannes-ucsc: This can be generator expression. And there is no need to extract restartCompatible. And if there were, I'd prefer a `def x():…` instead of a `x = lambda …`.

src/toil/test/provisioners/cgcloud/provisionerTest.py
```diff
@@ -265,3 +266,44 @@ def testAutoScaledCluster(self):
...
+
+class CGCloudRestartTest(AbstractCGCloudProvisionerTest):
+
+    def _getScript(self):
+        self.scriptName= "restartScript.py"
+        self._leader('echo %s > %s' % (restartScript, 'restartScript.py'), shell=True)
+
+    def _runScript(self, toilOptions):
+        # clean = onSuccess
+        restartCompatible = lambda x : ('clean' not in x and 'retryCount' not in x)
```
hannes-ucsc: I think I'd prefer an explicit list of disallowed options rather than a substring match.

cket:
Jenkins, test this please

cket:
Jenkins, test this please

cket:
Jenkins, test this please

cket:
Jenkins, test this please

cket:
Jenkins, test this please

src/toil/test/provisioners/cgcloud/provisionerTest.py
cket: In this case isn't the generator expression just as expensive as the list comprehension? The result has to be evaluated before being unpacked, right? 

If they're equivalent I would rather stick with the list comprehension since it's a little more conventional and you can inspect the resulting 'newOptions' variable when debugging.

cket:
@hannes-ucsc I addressed all of your comments, minus the minor question I have about the generator expression above

src/toil/test/provisioners/cgcloud/provisionerTest.py
hannes-ucsc: It's probably more expensive because to allocate a tuple, Python needs the size, and to know the size, it needs a list and to get a list it has to build one from the generator. I'm not sure. Anyhow,  I think a list comprehension is better.

PR #1156, opened 2016-09-14
Fix `make docker` to not pass -f to `docker tag` (resolves #1155)

### LINKED ISSUE #1155: `make docker` fails on docker >= 1.10
… because their `docker tag` doesn't support `-f`.

hannes-ucsc:
resolves #1155


hannes-ucsc:
Merging this despite IT failures because IT failures are unrelated (caused by Ubuntu keyserver outage) and I can see my changes in the Makefile logic working.

PR #1164, opened 2016-09-20
Fix hot-deployment of venv with jobs from distinct root packages (resolves #1152)

### LINKED ISSUE #1152: Hot-deployment fails with job functions spread over two top-level packages
The following structure in a venv on the leader /path/to/virtualenv/lib/python/site-packages/toil_scripts
/path/to/virtualenv/lib/python/site-packages/toil_lib

causes hot-deployment to fail localization in a job function if parent job function is from toil_scripts.foo.bla while the other job function is from toil_lib.bar.fasel.

hannes-ucsc:
resolves #1152

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -690,7 +690,7 @@ class Job(object):
         """
         commandTokens = command.split()
         assert "_toil" == commandTokens[0]
-        userModule = ModuleDescriptor(*(commandTokens[2:]))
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
IT failure is definitely unrelated and @cket is blocked on this, so I'll be merging this despite the IT failure.

PR #1172, opened 2016-09-23
Add test for promised requirements with non-caching file store (resolves #1172)

cket:


hannes-ucsc:
… in *the* subclass.

cket:
Wrong issues # in the branch name & title, closing.

PR #797, opened 2016-04-15
Add YARNBatchSystem to Toil (resolves #796)

fnothaft:
Adds a YARNBatchSystem that enables Toil to run jobs on a cluster that is managed with the Apache Hadoop YARN resource manager. A job maps directly to a YARN application. Specifically, we run a batch command line under YARN using the YARN DistributedShell application. This way, we do not need to implement a YARN ApplicationManager, etc.

This is completely untested and doesn't have docs. I'll be adding tests in the next few days. @hannes-ucsc: IIRC, we'll need to add a YARN setup on Jenkins. I will ask you about this once I've got the tests done. I think it should be straightforward to do.

PR #1195, opened 2016-10-06
Initial PBS/Torque support and AbstractBatchSystem refactoring

### LINKED ISSUE #1163: Add support for PBS/Torque
It looks like we would need to make a torque version of this module: https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/slurm.py 
This module includes 5 calls to slurm commands: sbatch, scancel, squeue, sinfo, sacct
which would need to be replaced with the equivalents: qsub, qdel, qstat, qstat -a, q?
A cheat sheet like this one might help: https://slurm.schedmd.com/rosetta.pdf
Also we would need to : 
- convert the command options 
- modify the processing of the outputs.... not sure how much they differ... 

There might be more ramifications... 
Comments, Thoughts ? Any one tried before ?
Thanks

cjfields:
This is a preliminary run at adding PBS/Torque support, designed off the SLURM and SGE classes.  Tests have also been added.

Resolves #1163

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -17,13 +17,21 @@ from __future__ import absolute_import
 
 import os
 import shutil
+import logging
+import time
 from abc import ABCMeta, abstractmethod
 from collections import namedtuple
+from Queue import Queue, Empty
 
 from bd2k.util.objects import abstractclassmethod
 
 from toil.common import Toil, cacheDirName
 from toil.fileStore import shutdownFileStore
+
+logger = logging.getLogger(__name__)
+
+# TODO: should this be an attribute?  Used in the worker and the batch system
+sleepSeconds = 1
 
 # A class containing the information required for worker cleanup on shutdown of the batch system.
 WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
@@ -33,7 +41,6 @@ WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
     'workflowID',
     # The value of the cleanWorkDir flag
     'cleanWorkDir'))
-
 
 class AbstractBatchSystem(object):
     """
@@ -311,7 +318,6 @@ class BatchSystemSupport(AbstractBatchSystem):
             and workflowDirContents in ([], [cacheDirName(info.workflowID)])):
             shutil.rmtree(workflowDir)
 
-
 class NodeInfo(namedtuple("_NodeInfo", "cores memory workers")):
     """
     The cores attribute  is a floating point value between 0 (all cores idle) and 1 (all cores
@@ -344,7 +350,6 @@ class AbstractScalableBatchSystem(AbstractBatchSystem):
         :rtype: dict[str,NodeInfo]
         """
         raise NotImplementedError()
-
 
 class InsufficientSystemResources(Exception):
     """
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -1,0 +1,168 @@
+# Copyright (C) 2015 UCSC Computational Genomics Lab
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+import logging
+import os
+from pipes import quote
+import subprocess
+import time
+import math
+import sys
+import xml.etree.ElementTree as ET
+import tempfile
+
+from toil.batchSystems import MemoryString
+from toil.batchSystems.abstractGridEngineBatchSystem import AbstractGridEngineBatchSystem
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+class TorqueBatchSystem(AbstractGridEngineBatchSystem):
+
+    # class-specific Worker
+    class Worker(AbstractGridEngineBatchSystem.Worker):
+
+        """
+        Torque-specific AbstractGridEngineWorker methods
+        """
+        def getRunningJobIDs(self):
+            times = {}
+            currentjobs = dict((str(self.batchJobIDs[x][0]), x) for x in self.runningJobs)
+            process = subprocess.Popen(["qstat"], stdout=subprocess.PIPE)
+            stdout, stderr = process.communicate()
+
+            # TODO: qstat supports XML output which is more comprehensive, so maybe use that?
+            for currline in stdout.split('\n'):
+                items = currline.strip().split()
+                if items:
+                    jobid = items[0].strip().split('.')[0]
+                    if jobid in currentjobs and items[4] == 'R':
+                        walltime = items[3]
+                        # normal qstat has a quirk with job time where it reports '0'
+                        # when initially running; this catches this case
+                        if walltime == '0':
+                            walltime = time.mktime(time.strptime(walltime, "%S"))
+                        else:
+                            walltime = time.mktime(time.strptime(walltime, "%H:%M:%S"))
+                        times[currentjobs[jobid]] = walltime
+
+            return times
+
+        def killJob(self, jobID):
+            subprocess.check_call(['qdel', self.getBatchSystemID(jobID)])
+
+        def prepareSubmission(self, cpu, memory, jobID, command):
+            return self.prepareQsub(cpu, memory, jobID) + [self.generateTorqueWrapper(command)]
+
+        def submitJob(self, subLine):
+            process = subprocess.Popen(subLine, stdout=subprocess.PIPE)
+            so, se = process.communicate()
+            # TODO: the full URI here may be needed on complex setups, stripping
+            # down to integer job ID only may be bad long-term
+            result = int(so.strip().split('.')[0])
+            return result
+
+        def getJobExitCode(self, torqueJobID):
+            args = ["qstat", "-f", str(torqueJobID)]
+
+            process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+            for line in process.stdout:
+                line = line.strip()
+                if line.startswith("failed") and int(line.split()[1]) == 1:
+                    return 1
+                if line.startswith("exit_status"):
+                    status = line.split(' = ')[1]
+                    logger.debug('Exit Status: ' + status)
+                    return int(status)
+            return None
+
+        """
+        Implementation-specific helper methods
+        """
+        def prepareQsub(self, cpu, mem, jobID):
+
+            # TODO: passing $PWD on command line not working for -d, resorting to
+            # $PBS_O_WORKDIR but maybe should fix this here instead of in script?
+
+            # TODO: we previosuly trashed the stderr/stdout, as in the commented
+            # code, but these may be retained by others, particularly for debugging.
+            # Maybe an option or attribute w/ a location for storing the logs?
+
+            # qsubline = ['qsub', '-V', '-j', 'oe', '-o', '/dev/null',
+            #             '-e', '/dev/null', '-N', 'toil_job_{}'.format(jobID)]
+
+            qsubline = ['qsub', '-V', '-N', 'toil_job_{}'.format(jobID)]
+
+            if self.boss.environment:
+                qsubline.append('-v')
+                qsubline.append(','.join(k + '=' + quote(os.environ[k] if v is None else v)
+                                         for k, v in self.boss.environment.iteritems()))
+
+            reqline = list()
+            if mem is not None:
+                memStr = str(mem / 1024) + 'K'
+                reqline += ['-l mem=' + memStr]
+
+            if cpu is not None and math.ceil(cpu) > 1:
+                qsubline.extend(['-l ncpus=' + str(int(math.ceil(cpu)))])
+
+            return qsubline
+
+        def generateTorqueWrapper(self, command):
+            """
+            A very simple script generator that just wraps the command given; for
+            now this goes to default tempdir
+            """
+            _, tmpFile = tempfile.mkstemp(suffix='.sh', prefix='torque_wrapper')
+            fh = open(tmpFile , 'w')
+            fh.write("$!/bin/bash\n\n")
+            fh.write("cd $PBS_O_WORKDIR\n\n")
+            fh.write(command + "\n")
+            fh.close
+            return tmpFile
+
+    """
+    The interface for the PBS/Torque batch system
+    """
+
+    @classmethod
+    def obtainSystemConstants(cls):
+
+        maxCPU = 0
+        maxMEM = MemoryString("0K")
+
+        # parse XML output from pbsnodes
+        root = ET.fromstring(subprocess.check_output(["pbsnodes","-x"]))
+
+        # for each node, grab status line
+        for node in root.findall('./Node/status'):
+            # then split up the status line by comma and iterate
+            status = {}
+            for state in node.text.split(","):
+                statusType, statusState = state.split("=")
+                status[statusType] = statusState
+            if status['ncpus'] is None or status['totmem'] is None:
+                RuntimeError("pbsnodes command does not return ncpus or totmem columns")
+            if status['ncpus'] > maxCPU:
+                maxCPU = status['ncpus']
+            if MemoryString(status['totmem']) > maxMEM:
+                maxMEM = MemoryString(status['totmem'])
+
+        if maxCPU is 0 or maxMEM is MemoryString("0K"):
+            RuntimeError('pbsnodes returned null ncpus or totmem info')
+        else:
+            logger.info("Got maxCPU: %s and maxMEM: %s" % (maxCPU, maxMEM, ))
+
+        return maxCPU, maxMEM
--- a/src/toil/batchSystems/__init__.py
+++ b/src/toil/batchSystems/__init__.py
@@ -23,9 +23,12 @@ if sys.version_info >= (3, 0):
 
 class MemoryString:
     def __init__(self, string):
-        if string[-1] == 'K' or string[-1] == 'M' or string[-1] == 'G' or string[-1] == 'T':
+        if string[-1] == 'K' or string[-1] == 'M' or string[-1] == 'G' or string[-1] == 'T': #10K
             self.unit = string[-1]
             self.val = float(string[:-1])
+        elif len(string) >= 3 and (string[-2] == 'k' or string[-2] == 'M' or string[-2] == 'G' or string[-2] == 'T'):
+            self.unit = string[-2]
+            self.val = float(string[:-2])
         else:
             self.unit = 'B'
             self.val = float(string)
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -1,0 +1,393 @@
+# Copyright (C) 2015-2016 Regents of the University of California
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import os
+import shutil
+import logging
+import time
+from threading import Thread
+from abc import ABCMeta, abstractmethod
+
+# Python 3 compatibility imports
+from six.moves.queue import Empty, Queue
+
+from bd2k.util.objects import abstractclassmethod
+
+from toil.batchSystems.abstractBatchSystem import BatchSystemSupport
+
+logger = logging.getLogger(__name__)
+
+# TODO: should this be an attribute?  Used in the worker and the batch system
+
+class AbstractGridEngineBatchSystem(BatchSystemSupport):
+    """
+    A partial implementation of BatchSystemSupport for batch systems run on a
+    standard HPC cluster. By default worker cleanup and hot deployment are not
+    implemented.
+    """
+
+    class Worker(Thread):
+
+        __metaclass__ = ABCMeta
+
+        def __init__(self, newJobsQueue, updatedJobsQueue, killQueue, killedJobsQueue, boss):
+            """
+            Abstract worker interface class. All instances are created with five
+            initial arguments (below). Note the Queue instances passed are empty.
+
+            :param newJobsQueue: a Queue of new (unsubmitted) jobs
+            :param updatedJobsQueue: a Queue of jobs that have been updated
+            :param killQueue: a Queue of active jobs that need to be killed
+            :param killedJobsQueue: Queue of killed jobs for this worker
+            :param boss: the AbstractGridEngineBatchSystem instance that controls this AbstractGridEngineWorker
+
+            """
+            Thread.__init__(self)
+            self.newJobsQueue = newJobsQueue
+            self.updatedJobsQueue = updatedJobsQueue
+            self.killQueue = killQueue
+            self.killedJobsQueue = killedJobsQueue
+            self.waitingJobs = list()
+            self.runningJobs = set()
+            self.boss = boss
+            self.allocatedCpus = dict()
+            self.batchJobIDs = dict()
+
+        def getBatchSystemID(self, jobID):
+            """
+            Get batch system-specific job ID
+
+            Note: for the moment this is the only consistent way to cleanly get
+            the batch system job ID
+
+            :param: string jobID: toil job ID
+            """
+            if not jobID in self.batchJobIDs:
+                RuntimeError("Unknown jobID, could not be converted")
+
+            (job, task) = self.batchJobIDs[jobID]
+            if task is None:
+                return str(job)
+            else:
+                return str(job) + "." + str(task)
+
+        def forgetJob(self, jobID):
+            """
+            Remove jobID passed
+
+            :param: string jobID: toil job ID
+            """
+            self.runningJobs.remove(jobID)
+            del self.allocatedCpus[jobID]
+            del self.batchJobIDs[jobID]
+
+        def createJobs(self, newJob):
+            """
+            Create a new job with the Toil job ID. Implementation-specific; called
+            by AbstractGridEngineWorker.run()
+
+            :param string newJob: Toil job ID
+            """
+            activity = False
+            # Load new job id if present:
+            if newJob is not None:
+                self.waitingJobs.append(newJob)
+            # Launch jobs as necessary:
+            while (len(self.waitingJobs) > 0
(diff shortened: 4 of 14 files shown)


hannes-ucsc:
This is yet another almost identical copy of the SGE batch system. As much as I understand the desire to get stuff done, this path is unsustainable because, once I merge it, it puts the onus on the Toil team to maintain three nearly identical copies until eternity.

Also, the tests will fail since TORQUE isn't installed on the CI slaves. This would have to be addressed by a change to CGCloud's ToilJenkinsSlave.

cjfields:
@hannes-ucsc what's the best course forward?  Focus on a single DRMAA-based approach (not sure where that stands these days; it's been a little flaky in the past)?  Or move the common code into a parent class or module?

hannes-ucsc:
I think extracting a common abstract base class using the template [method pattern](https://en.wikipedia.org/wiki/Template_method_pattern) is the obvious solution.

cjfields:
At the moment there is the `AbstractBatchSystem` abstract base class, and the `BatchSystemSupport` base class for SLURM, Torque, Gridengine, etc implementing common methods (which seems like the obvious place).  If that seems fine I can start on pulling out the common code from these into `BatchSystemSupport`.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
I think this should be the hierarchy:

```
AbstractBatchSystem
    BatchSystemSupport
        AbstractGridEngineBatchSystem
             GridEngineBatchSystem
             SlurmBatchSystem
             TorqueBatchSystem
```

hannes-ucsc:
@cjfields, I sent an invite to our GH org. If you accept, you will be able to see the Jenkins test results.

cjfields:
@hannes-ucsc thanks!

cjfields:
@hannes-ucsc should we merge this in and have a separate issue tracking re-implementation of the grid schedulers.  It seems SLURM is set up w/ Jenkins so it can act as a check, and we can test Torque locally.  The only wild card is SGE/OGE, which we don't have installed locally.

hannes-ucsc:
I'd prefer not merging until the testing and code duplication issues are squared away. The changes in this PR are mostly additions so keeping the branch rebased on mainline should be easy, meaning there is no rush to merge it.

cjfields:
@hannes-ucsc okay, that works for me.

cjfields:
@hannes-ucsc I added an initial `AbstractGridEngineBatchSystem` implementation that works for Torque, will work on SLURM next (which Jenkins should catch).  Anywhere to test this w/ SGE?

cket:
@cjfields just FYI every time you push to this branch you're triggering an integration test build. I cancelled all but the most recent push so the test results are showing up as failed in the history although they may not have failed if I had let them finish running. 

Since the tests are run on only the most recent commit in a push and not every commit, you can avoid this by committing locally and waiting to push the commits until you want Jenkins to run the test suite :)

cjfields:
@cket Thanks!  I was using `[ci-skip]` in the commit msg to try skipping CI, and it seemed to skip a couple of those intermittent ones I pushed, e.g. https://github.com/BD2KGenomics/toil/pull/1195/commits/4afc3b669c593aa856114da0a764a84028fe0969 but it appears to work inconsistently.  I'll stick to the local branch for now.

cket:
@cjfields makes sense! We use the directive 'Jenkins, skip this please' in a PRs description to skip testing. However, I'm not sure how Jenkins directives work with outside contributors so you're probably better off just committing locally

cjfields:
@cket @hannes-ucsc seems the failures in the last jenkins run is AWS-related, [not due to `batchSystemTests` issues](http://jenkins.cgcloud.info/job/toil-pull-requests/1911/testReport/src.toil.test.batchSystems.batchSystemTest/)

```
boto.exception.S3ResponseError: S3ResponseError: 400 Bad Request
<Error>
<Code>TooManyBuckets</Code>
<Message>You have attempted to create more buckets than allowed</Message>
<CurrentNumberOfBuckets>101</CurrentNumberOfBuckets>
<AllowedNumberOfBuckets>101</AllowedNumberOfBuckets>
</Error>
ERROR
```

hannes-ucsc:
Yes, only push if you want Jenkins to test. I think Jenkins builds on every push because you are now a member of our organization.

> Anywhere to test this w/ SGE

The Jenkins slave running the build has SGE and Slurm installed. Here are the important bits for SGE:

https://github.com/BD2KGenomics/cgcloud/blob/master/jenkins/src/cgcloud/jenkins/toil_jenkins_slave.py#L45

https://github.com/BD2KGenomics/cgcloud/blob/master/jenkins/src/cgcloud/jenkins/toil_jenkins_slave.py#L59

I hope the TORQUE installation is less involved.

cjfields:
Okay, it appears to skip SGE tests then, likely due to trying to find `qmake`.  I'll work locally on the SGE module and see if there is a better alternative (`qsub` is used by both SGE and PBS/Torque).

cjfields:
@hannes-ucsc small but important thing: the current name of the SGE-specific class is `GridengineBatchSystem`, but the abstract class is `AbstractGridEngineBatchSystem` (note the cap for `...Engine...` in the latter).  Switching the former to be consistent with the naming scheme (e.g. to `GridEngineBatchSystem`) is breaking the test and my guess would break any SGE-dependent code on case-sensitive file systems.  Also note I renamed the various `Worker` classes to be batch-system specific (eg. SGE's `Worker` is now `GridEngineWorker`).  

I am reverting this to the old name for now, but any thoughts on a reasonable workaround, or should we just go ahead and make the change in the test?

hannes-ucsc:
Feel free touch all code necessary to refactor the name. No code external to Toil is dependent on the names of these classes, so you should be good to change the names to whatever you like, as long as you change all occurrences of that name in the Toil source. Since the official name is/was "Sun Grid Engine"/"Oracle Grid Engine" I would prefer `GridEngine…`, not `Gridengine…`.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -17,14 +17,22 @@
...
+from Queue import Queue, Empty
 
 from bd2k.util.objects import abstractclassmethod
 
 from toil.common import Toil, cacheDirName
 from toil.fileStore import shutdownCache
 
+logger = logging.getLogger(__name__)
+
+# TODO: should this be an attribute?  Used in the worker and the batch system
```
hannes-ucsc: yes

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -331,6 +337,136 @@ def getNodes(self, preemptable=None):
         """
         raise NotImplementedError()
 
+class AbstractGridEngineBatchSystem(BatchSystemSupport):
```
hannes-ucsc: This should be in a separate module. I propose the following module structure

```
toil
      batchSystems
          gridEngine
               abstractGridEngineBatchSystem.py
               gridEngineBatchSystem.py
               torqueBatchSystem.py
               slurmBatchSystem.py
```

hannes-ucsc:
Did a preliminary review. The history is polluted with commits from the master branch. 

The code duplication concern applies to the workers too, obviously, so you might need to add a parallel worker class hierarchy as well, starting in AbstractGridEngineWorker.

Since you will be touching a lot of code, I would nest the definition of the workers in the batch system class. So GridEngineWorker becomes, GridEngineBatchSystem.Worker.

cjfields:
Thanks @hannes-ucsc .  The commits from master were rebased in, I'll just keep this branch clean from future commits from master from now on.  I have the updates running through Jenkins now, including running with `AbstractGridEngineBatchSystem` in a separate module.  

Re: the separate `Worker`, I agree though I'm wondering whether it might be cleaner to do this in a separate pull request.

cjfields:
@hannes-ucsc regarding abstracting `Worker` I'm currently testing this in a separate branch off the current PBS `issues/1163-pbs-support`, specifically [this one](https://github.com/HPCBio/toil/tree/issues/abstract-worker).  So far embedding the class definition is working for Torque.

hannes-ucsc:
This should be boiled down to two commits: one that eliminates the duplication in the existing GridEngine-based batch systems and another one that adds TORQUE support, in that order. If you would like to do each in a separate PR, that's fine with me.

cjfields:
@hannes-ucsc the current run of commits was developed inversely.  The initial pull request added simple PBS/Torque based on the old SGE code but had code duplication.  Based on your feedback additional downstream commits implemented a base batch system class and removed code duplication in the Torque, SGE, and SLURM batch modules (at least, to this point, relegated primarily in the batch system part).  

I could basically start from a clean branch off master and implement the changes as stated as two pull requests, though it would remove much of the commi history and render this pull request obsolete.

hannes-ucsc:
Clean slate sounds good to me.

But you don't need to start a new PR. You can soft-reset the PR branch to master without changing your working copy. Then commit the changes again. Then force push. But new PR works, too, as long as both are linked.

cjfields:
@hannes-ucsc went ahead and squashed these down to just one commit, and will renaming pull accordingly.

hannes-ucsc:
Thank you, @cjfields. To summarize what I already mentioned in one way or the other:

- worker classes still duplicate too much functionality
- TORQUE tests aren't being run on CI

And more thing: there should be a sort test that uses TORQUE. It will yield better state coverage compared to the more unit-testy batch system tests. But don't worry, adding a sort test for a particular batch system is very simple. See sortTest.py.

cjfields:
> worker classes still duplicate too much functionality

Working on this now, though (bad timing) our cluster is down with file system issues until Monday

> TORQUE tests aren't being run on CI

That will be the next step

> And more thing: there should be a sort test that uses TORQUE. It will yield better state coverage compared to the more unit-testy batch system tests. But don't worry, adding a sort test for a particular batch system is very simple. See sortTest.py.

I actually had prelim tests on this as well.

hannes-ucsc:
Another collaborator had mentioned to me that he was working on a batch system implementation for DRMAA which would allow us to use one AbstractBatchSystem implementation with GridEngine, Slurm and Torque. I know virtually nothing about DRMAA and I don't can't tell yet if or when his efforts will be merged.

I feel this shouldn't stop you from pursuing this PR. I will merge this regardless of the DRMAA effort if it reduces the code duplication and has tests that pass on our CI. I think the DRMAA implementation and your refactored implementations can and should coexist.

hannes-ucsc:
> EDIT: forgot to ask, should we pull this into a separate class file? abstractBatchSystem.py is getting a little long.

Yes. abstractBatchSystem.py should only contain generic code, not code specific to any particular concrete implementation.

cjfields:
@hannes-ucsc agreed, DRMAA should probably be implemented as a separate co-existing module.  We actually have DRMAA on our cluster so this is great, but IIRC some schedulers had (as least in the past) required having DRMAA support compiled in prior to deployment (at least PBS/Torque was this way, and it wasn't on by default).

benedictpaten:
@cjfields @hannes-ucsc This looks like great, great work. Hope we can get it merged soon and under CI.

cjfields:
Thx @benedictpaten .  Slight delay, our cluster had a major disk issue that took it offline for the week, so we were busy triaging analyses (it's back to normal now).

cjfields:
I have a squashed commit from a branch with the AbstractGridEngineWorker extracted out which passes Torque tests.  I am going ahead with the merge to test w/ jenkins to see if SGE and SLURM modifications work (I may test on our new cluster as well, which is using SLURM).

cjfields:
Only have a bit of minor cleanup and to implement `sort` tests.

@hannes-ucsc @benedictpaten on getting Torque tests running on jenkins: are all tests run within one cloud instance?  If so, a significant hurdle when implementing testing of Torque and GridEngine from the same instance will be that many of the exe names are the same in both job schedulers (e.g. `qsub`, `qstat`, etc).  The current tests on this branch only run GridEngine tests when SGE is installed (and Torque similarly) via looking for scheduler-specific applications, but internally they both use `qsub` and `qstat` calls.

benedictpaten:
This one is for @hannes-ucsc to comment on.

benedictpaten:
@hannes-ucsc Probably good to move on this, I think this is a really useful refactor.

hannes-ucsc:
I will do a detailed review shortly.

> Torque tests running on jenkins: are all tests run within one cloud instance?

Yes. We call that instance the *Jenkins slave*.

> If so, a significant hurdle when implementing testing of Torque and GridEngine from the same instance will be that many of the exe names are the same in both job schedulers (e.g. qsub, qstat, etc). The current tests on this branch only run GridEngine tests when SGE is installed (and Torque similarly) via looking for scheduler-specific applications, but internally they both use qsub and qstat calls.

Good point.

I would use https://hub.docker.com/r/agaveapi/torque/ to start the TORQUE daemons in a container on the Jenkins slave. The slave that runs the tests has docker installed and the user that the tests are running is a member of the `docker` group so that it can invoke the `docker` command.

You could start the container during test setup and stop it during test teardown. You then need a mechanism to override the various TORQUE command names used by your batch system implementation. Let's say the default for `qsub` is just `qsub`. Your tests could override that with `docker exec CONTAINER_ID qsub`, thereby running `qsub` in the container started during test setup that's hosting the TORQUE daemons. IIRC, there already is a way to override commands but I don't think it is complete. This doesn't have to be a fancy mechanism, simply extracting global variables for the command names and monkey-patching those during test setup would be sufficient.

The beauty of this approach is that you don't have to submit a PR to CGCloud that performs TORQUE installation (which looks complicated) and that it doesn't collide with the existing SGE installation on the slave.

Toil is already using that approach to run the Toil appliance container for certain tests. Two containers actually. Take a look at `toil.test.ApplianceTestSupport` and `toil.test.src.hotDeploymentTest.HotDeploymentTest`.

src/toil/batchSystems/abstractGridEngineWorker.py
```diff
@@ -0,0 +1,183 @@
...
+import os
+from abc import ABCMeta, abstractmethod
+from threading import Thread
+from Queue import Queue, Empty
+import time
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+class AbstractGridEngineWorker(Thread):
```
hannes-ucsc: This class should be nested in AbstractGridEngineBatchSystem and called Worker. You can then refer to it as `AbstractGridEngineBatchSystem.Worker`.  It makes the code organization is consistent between abstract and concrete classes.

src/toil/batchSystems/abstractGridEngineWorker.py
```diff
@@ -0,0 +1,183 @@
...
+
+    __metaclass__ = ABCMeta
+
+    def __init__(self, newJobsQueue, updatedJobsQueue, killQueue, killedJobsQueue, boss):
+        '''
```
hannes-ucsc: We use `"""` for docstrings as in PEP-8. I've been informally asking people to prefer `'` for regular strings if possible but have not made this a "rule".

src/toil/batchSystems/torque.py
```diff
@@ -0,0 +1,189 @@
...
+            fh.write("cd $PBS_O_WORKDIR\n\n")
+            fh.write(command + "\n")
+            fh.close
+            return tmpFile
+
+    """
+    The interface for the PBS/Torque batch system
+    """
+
+    @staticmethod
```
hannes-ucsc: We discourage use of @staticmethod and prefer @classmethod. It's incidental that a method doesn't access  class attributes and therefore has no use for the `cls` parameter. And because its incidental, always using @classmethod avoids unnecessary refactorings in the future when changes to that method do suddenly require the `cls` argument.

cjfields:
Hi @hannes-ucsc I'll start having a look at it this week.

src/toil/batchSystems/abstractGridEngineWorker.py
cjfields: Moved into `AbstractGridEngineBatchSystem` in a6bd60f

src/toil/batchSystems/abstractGridEngineWorker.py
cjfields: Addressed in  	8bd8e0d

src/toil/batchSystems/torque.py
cjfields: Addressed in c40c329

cjfields:
@cket this request is reporting possible configuration error issues with Jenkins.  Any ideas?  Maybe it needs a refresh from master (merge conflicts for `gridengine.py` and `slurm.py` need to be addressed anyway, could do this on a local branch and merge in...)

cket:
@cjfields merging/rebasing on master should fix everything!

cjfields:
Looks like I just missed the python3 compat pull; I'll see how this round of jenkins tests goes and try merging those once I have tests running cleanly again

cjfields:
**Remaining issues**

- [ ] Move `createJobs` into base class
- [ ] Refactor `getRunningJobIDs` ?
- [ ] Consolidation of process logging 

From @hannes-ucsc review:

> This is looking much improved and I'm happy about the path this is taking. I recommend working on refactoring more duplication out of the existing methods in each concrete implementation. For example, createJobs looks virtually identical across all of them.

Agreed.

> getJobExitCode looks very similar between TORQUE and SGE, so maybe you want to pull that up into the abstract base class and override it in SLURM only.

This one is tricky, as the various implementations for this method are output-specific.  Though there are similarities enough differences exist between the SGE and Torque methods, the output parsing differs enough that I could see refactoring these creating more code than currently exists and possibly causing more problems than it's worth.

> getRunningJobIDs could be refactored by pulling up it up and factoring out an abstract method called parseQstatLine that parses an individual line and returns a namedtuple with the information needed. This would be similar to what you did with killJob vs killJobs. The name parseQstatLine glosses over the fact that SLURM doesn't have qstat but I think that's ok. After all these are all SGE clones so I think people will get it.

Will check into that

> Logging every subprocess call is very helpful in diagnosing problems, so I propose adding that to TORQUE, too. In fact, I would write a wrapper around subprocess.check_call that logs implicitly and, again, eliminates a lot of duplication.

Okay

> Lastly, do you know why SLURM doesn't seem to have setEnv?

It seemed to fall back to the default implementation in class `BatchSystemSupport`.  This now is commonly set within the `AbstractGridEngineBatchSystem` base class, as all current implementations using this can't use commas in the variable values

cjfields:
@cket latest tests seem to be passing, but I am seeing a Java issue on Jenkins that is causing a problem:

http://jenkins.cgcloud.info/job/toil-pull-requests/2447/consoleText

Any idea?

cket:
Jenkins, test this please

cket:
@cjfields I haven't seen that before, I retesting might fix the issue

cket:
@cjfields looks like that fixed it. Thanks again for the PR, it's looking good. Did you see this related slurm PR, #1391? Does this PR address any of those issues?

cjfields:
> Did you see this related slurm PR, #1391? Does this PR address any of those issues?

@cket I can have a look, though the changes for SLURM/SGE are quite significant in this branch (it's really a misnamed PR as it adds Torque support and refactors all three to have a common base). We would likely need to hand-merge the SLURM PR changes in.

cket:
@cjfields just a heads up, it looks like there is still a little work to be done in this PR so I’ll likely merge #1391 first - sorry to have you do another merge from master, hopefully its not too troublesome.

cjfields:
@cket not a problem, I can work on resolving conflicts once it's in.

cjfields:
@cket I'm seeing some API differences when keeping batch system IDs that seems to break SLURM.  The primary issues is that Torque, SGE (and LSF, which needs to be moved over) all seem to store the mapping of toil job IDs to batch system job IDs as a tuple of `(job ID, Task name)`, but by default passes in `None` for the task.  It seems like this is behavior inherited from the original gridengine (SGE) code, but I can't see anywhere that the task from the tuple is actively used.

**Master branch**

SGE: https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/gridengine.py#L131

LSF: https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/lsf.py#L115

On this pull request I adopted this behavior when refactoring (this occurs in `createJobs()`, in abstract base class `AbstractGridEngineBatchSystem`): 

https://github.com/HPCBio/toil/blob/issues/1163-pbs-support/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L123

However, SLURM doesn't use this approach and simply sets the toil ID -> batch system ID mapping as a simple value (not tuple):

**Master branch**

https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/slurm.py#L150

I attempted to modify SLURM to adopt the other tuple-based behavior, but tests are hanging and are likely due to some inconsistency at this level.  However, I can't easily test this until we have our local SLURM cluster ready to go, and testing using Jenkins is touch-and-go at best until this is ready.

My suggestion is to back out the SLURM changes that are breaking tests, but it's definitely an inconsistency that needs to be addressed.  We can file a related issue tagging this.  

**The key question:** which behavior should we end up adopting: tuple-based, like SGE, or simple value, like SLURM?  A cursory look through the code doesn't indicate where the task value from the tuple is used, but maybe I'm missing something there (e.g. it's settable at the API level).

cjfields:
@cket actually, as a quick followup, it looks like our local SLURM setup is working, just did a checkout from our fork and tests on master pass.  I'll see if I can get these at least passing on this.

cjfields:
Digging in, it looks like the 'Task' stuff mentioned [previously](https://github.com/BD2KGenomics/toil/pull/1195#issuecomment-272638744) is from SGE and probably came from a few years back.  Seems like it really pertains to allowing for job array support, where the task is the job array ID.  I have gone ahead and modified SLURM to allow for this, but the question is how much overall support toil has for job arrays currently.

cket:
@cjfields I'm a little unsure of the exact question, but Toil doesn't have the concept of job arrays. If a batch system class wants to implement them to improve performance or behavior when interacting with the batch system they can, as long as it's transparent to the rest of Toil. 

It sounds like you've already made this behavior consistent across the relevant batch systems, which is great!

ohofmann:
Can I follow up on this issue? We are on the fence tackling a PBS Pro integration with the National Computational Infrastructure in Australia, but it would make a lot of sense for us to build on @cjfields 's work. 

What are the biggest stumbling blocks to getting this pull request merged?

cket:
@ohofmann I think https://github.com/BD2KGenomics/toil/pull/1195#issuecomment-270690866 summarizes it best. IMO this is nearing completion and the remaining issues are (probably) simple compared to the work already done, but I would bet @cjfields has a better feel for the work to be done.

cjfields:
@cket I'm primarily concerned this fork has diverged enough that this will cause problems with additional pull requests in the same batch system code; we will likely see this with #797, #1385, and #1392.  

To be honest, IMHO I think it would make better sense to pull in the changes once the SLURM conflict is dealt with (which I'll focus on now) and the commits are squashed.  We can reconcile the additional changes as (much smaller, atomic) pull requests, adding issues that track these.  The remaining two items in that comment are pretty minor and I think can be addressed separately, and if there are any tweaks for PBS-Pro that need to be made these can be addressed as a separate pull request.   Does that make sense?

cket:
@cjfields that makes sense to me - this refactoring is huge improvement and I'd like to get it into master so we can just track the last couple issues in their own tickets. 

Once you deal with the slurm conflicts and create tickets for the last issues, I'll do a final review and hopefully merge it ASAP!

cjfields:
@cket Just resolved that conflict manually; I'll add the tickets and refer to this pull request.

cjfields:
@cket seeing integration test fails but this seems linked to the same issues with #1385 (AWS resource limits).

cket:
@cjfields yeah, sorry about that. I've cleared out the outdated IAM roles. 

Jenkins, test this please

cjfields:
Okay, I'll work on squashing the commits for the branch

cjfields:
@cket commits have been squashed; I did a test merge locally to master which looks good.

cket:
@cjfields thanks again, great work!

cjfields:
Awesome, thanks @cket!

src/toil/batchSystems/torque.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    @classmethod
+    def obtainSystemConstants(cls):
+
+        maxCPU = 0
+        maxMEM = MemoryString("0K")
+
+        # parse XML output from pbsnodes
+        root = ET.fromstring(subprocess.check_output(["pbsnodes","-x"]))
```
brainstorm: For the record/future reference: https://twitter.com/braincode/status/847721319328137216

Will take it from here, I've started looking at both formats (XML and -a), that way I also have a way to write a test for it and not break things, thanks @cjfields!

cjfields: @brainstorm sure not a problem, let me know if you have a pull with the modifications.We've moved to SLURM now, where Toil works just fine, but we still have two other Torque-based clusters still running for the time being that I can test on.

brainstorm: Fantastic, thanks @cjfields! Could you please confirm that:

`pbsnodes -a -F json`

Is not an option in your system? `pbsnodes` in `PBSPro` spits json so I'm adding logic to differentiate between them, but if your PBS supports it already, it would be awesome and less code/effort ;)

cjfields: @brainstorm unfortunately not though it would be much better than XML (guess they are saving JSON for their paying customers):

```
-system-specific-4.1$ pbsnodes -a -F json
pbsnodes: invalid option -- 'F'
usage:	pbsnodes [-{c|d|l|o|p|r}] [-s server] [-n] [-N "note"] [-q] node ...
	pbsnodes [-{a|x}] [-s server] [-q] [node]
-system-specific-4.1$ pbsnodes --version
Version: 4.2.9
Commit: a78379d06ed3c309a4a38cdaafb146fba60608dc
```

brainstorm: Work in progress, @cjfields. I'll have to wait a few hours since our cluster is going for a short downtime today: https://github.com/BD2KGenomics/toil/compare/master...brainstorm:pbspro?expand=1

brainstorm: @cjfields Sorry to bother you again with this, but could you paste here an example `qstat -u $USER` from your cluster with running/queued jobs? I'm chasing a bug with the torque/PBSPro code and need to compare outputs there as well and not break your implementation along the way.

"Fun" fact: while `pbsnodes` supports XML and PBSPro version supports JSON, `qstat` supports neither, just "plain" commandline format O_o''

/cc @ohofmann

cjfields: @brainstorm `qstat` on ours has a `-x` switch for XML output.

https://gist.github.com/cjfields/7b652048f9a17cfee0ece0de30c87c5b

brainstorm: @cjfields I know, I didn't explain myself right: PBSPro does not have neither XML nor JSON (I would expect that a paid solution would, but oh well...). 

The -x flag is reserved for sth else in PBSPro and your command does not output XML but just plain qstat outputs:

```
-x        Displays status information for finished and moved jobs in addition 
to running and queued jobs.
```

But anyway, I'm trying to fix the `qstat` info gathering function(s) from `torque.py` *without* breaking your output parsing in the process, that's why I enquired back to you. For instance, my `qstat` output looks like this:

```
Job id            Name             User              Time Use S Queue
----------------  ---------------- ----------------  -------- - -----

687694.r-man2     name1      user1                   0 H normal-def
4589526.r-man2    name2      user2            584:51:0 R gpu-def
4595282.r-man2    name3          user3                   0 W copyq
```

When adding debugging loggers in `getRunningJobIDs` I can see that the times are not gathered properly by Toil's leader which according to @cket might be the reason my master is waiting indefinitely since it's not getting timing feedback from the queue system: 

```
getRunningJobIDs current jobs are: {'4780506.r-man2': 0}
Job times from qstat are: {}
No activity, sleeping for 1s
Job times from qstat are: {}
No activity, sleeping for 1s
Job times from qstat are: {}
No activity, sleeping for 1s
(...)
```

So ideally I would like to have both vanilla qstat outputs and make it work for me *and* you ;)

PR #1197, opened 2016-10-07
Add job name (resolves #677, resolves #1219)

### LINKED ISSUE #677: Introduce concept of job name for improved diagnostic

### LINKED ISSUE #1219: testCacheEvictionFailCase is too brittle
The test relies on specific log messages being in the output to determine if an exception was raised. If those log messages are modified, the test breaks.

cket:


cket:
Jenkins, test this please

cket:
Jenkins, test this please

src/toil/test/src/jobCacheTest.py
```diff
@@ -252,9 +252,9 @@ def _testCacheEviction(self, file1MB, file2MB, diskRequestMB):
                 Job.Runner.startToil(A, self.options)
             except FailedJobsException as err:
                 self.assertEqual(err.numberOfFailedJobs, 1)
-                errType, errMsg = self._parseAssertionError(self.options.logFile)
-                if (errType == 'AssertionError' and
-                        errMsg == 'Unable to free up enough space for caching.'):
+                with open(self.options.logFile) as f:
+                    logContents = f.read()
+                if 'AssertionError: Unable to free up enough space for caching.' in logContents:
```
hannes-ucsc: This is just as brittle, no? It's still duplicating a hard-coded error message. There must have been a reason for the intelligent parsing done by `_parseAssertionError`.

hannes-ucsc: ping @arkal for opinion

cket: This is a step in the right direction IMO, but you're right it's still brittle. The current approach's weak point is 
https://github.com/BD2KGenomics/toil/blob/master/src/toil/test/src/jobCacheTest.py#L353
where is tests for a particular log message from leader.py, and this is mainly what I wanted to remove. The parser, when successful, returned the exception message which was then compared against another hardcoded string. I'm still relying on a hardcoded string but the test now would only break if you touch caching code, specifically that particular assertion.

I could also define a caching specific exception to raise and test for it in the output, that would probably be more flexible. Ideally though my approach wouldn't rely on output parsing at all, I'll keep thinking about it.

cket:
@hannes-ucsc ping for review

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
-class Job(ResourceRequirementMixin):
+class JobLikeObject(ResourceRequirementMixin):
+    def __init__(self, name, job=None, **requirements):
+        super(JobLikeObject, self).__init__(**requirements)
+        self.name = name
+        self.job = job or self.__class__.__name__
+
+    def __str__(self):
+        printedName = '\''+self.job + '\''
+        if self.name:
```
hannes-ucsc: With the same reasoning for disliking how the constructor treats `None` and `""` the same, I think if someone passes and empty string for `name`, that should be reflected in the output.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+            printedName += ' '+self.name
```
hannes-ucsc: space around operators

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        return str(printedName)
```
hannes-ucsc: str seem redundant

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+
+class JobNode(JobLikeObject):
+    """
+    This object bridges the job graph, job, and batchsystem classes
+    """
+    def __init__(self, memory, cores, disk, preemptable, job, name, jobStoreID,
+                 command, predecessorNumber=1):
```
hannes-ucsc: Don't like how this constructor signature duplicates the signatures of the superclass constructors.

Interestingly, `name` is required here but optional in the superclass constructor.

I may want to attempt a refactoring against your branch once we resolve the other issues.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        super(JobNode, self).__init__(memory=memory, cores=cores, disk=disk,
+                                      preemptable=preemptable, name=name, job=job)
+        self.jobStoreID = jobStoreID
+        self.predecessorNumber = predecessorNumber
+        self.command = command
+
+    def __hash__(self):
+        return hash(self.jobStoreID)
+
+    def __ne__(self, other):
```
hannes-ucsc: You need to override `__eq__` to compare based on ID. This overrides `__ne__` with the default implementation.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        return not self.__eq__(other)
+
+    def __repr__(self):
```
hannes-ucsc: JobLikeObject should also override this. This implementation should also override `__str__`. Both methods here should forward to the inherited ones from JobLikeObject. Might also want to do the same thing for `ResourceRequirementMixin`.

src/toil/job.py
hannes-ucsc: I think the key to a successful refactoring would be capture all requirements in a dictionary and pass that around, rather than always keeping all requirements as separate kwargs. If that makes sense. I can explain in person tomorrow.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        return '%s( **%r )' % (self.__class__.__name__, self.__dict__)
+
+    @classmethod
+    def fromJobGraph(cls, jobGraph):
+        return cls(jobStoreID=jobGraph.jobStoreID, memory=jobGraph.memory,
```
hannes-ucsc: In general the rule is to wrap all or none. I'd be ok with grouping all requirements on one line, though, as that would increase readability.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        return hash(self.jobStoreID)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return '%s( **%r )' % (self.__class__.__name__, self.__dict__)
+
+    @classmethod
+    def fromJobGraph(cls, jobGraph):
```
hannes-ucsc: Docstring with type annotations.

src/toil/job.py
hannes-ucsc: These copy constructors would also benefit from capturing requirements into a dict.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+        return cls(jobStoreID=jobGraph.jobStoreID, memory=jobGraph.memory,
+                   cores=jobGraph.cores, disk=jobGraph.disk,
+                   preemptable=jobGraph.preemptable,
+                   command=jobGraph.command,
+                   job=jobGraph.job,
+                   name=jobGraph.name,
+                   predecessorNumber=jobGraph.predecessorNumber)
+
+    @classmethod
+    def fromJob(cls, job, command, predecessorNumber):
```
hannes-ucsc: Same comments as for the other copy constructor.

src/toil/job.py
```diff
@@ -157,11 +159,65 @@ def _parseResource(name, value):
...
+                   name=job.name,
+                   predecessorNumber=predecessorNumber)
+
+
+class Job(JobLikeObject):
     """
     Class represents a unit of work in toil.
     """
-    def __init__(self, memory=None, cores=None, disk=None, preemptable=None, checkpoint=False):
+    def __init__(self, memory=None, cores=None, disk=None, preemptable=None, name='', checkpoint=False):
```
hannes-ucsc: Again, empty string vs. None ugliness. Basically, don't treat empty string like None, but rather like any other string and make sure that `__str__` and `__repr__` return values that make it obvious that an empty string was used for either `name` or `job`.

src/toil/jobGraph.py
```diff
@@ -14,43 +14,38 @@
...
-        # Max number of bytes on disk space used by the job
-        self.disk = disk
-        # Can the job be run on a node that can be preempted
-        self.preemptable = preemptable
-        # The jobStoreID of the job. JobStore.load(jobStoreID) will return the job
-        self.jobStoreID = jobStoreID
-        
+                  checkpointFilesToDelete=None ):
+
+        super(JobGraph, self).__init__(command=command, memory=memory, cores=cores, disk=disk,
```
hannes-ucsc: Wrap all or none. Alternatively group semantically.

cket:
Jenkins, test this please

cket:
jenkins, test this please

cket:
jenkins, test this please

cket:
jenkins, test this please

cket:
@hannes-ucsc ready for review

cket:
jenkins, test this please

cket:
jenkins, test this please

cket:
@hannes-ucsc I think we should go ahead with merging this, if you approve. I'm still thinking about the inspection magic but if it works out I'd rather do that in a new branch next week rather than keep this PR open longer.

cket:
@hannes-ucsc ready for review- I believe this PR had conflicts earlier but merging the multi-region PR seems to have resolved them.

hannes-ucsc:
Will merge once queued-up master builds succeed.

arkal:
Congrats @cket !

cket:
Thanks!

PR #1198, opened 2016-10-07
Allow Job.Service to access the underlying Toil Job (resolves #1191)

### LINKED ISSUE #1191: Job.Service needs access to the underlying job instance
We're running into this in https://github.com/BD2KGenomics/toil-lib/pull/46. Specifically, if the `Job.Service.start` method invokes a Docker container via [docker_call](https://github.com/BD2KGenomics/toil-lib/blob/master/src/toil_lib/programs.py#L20) that cleans up the ghost container via a [`defer`red method](https://github.com/BD2KGenomics/toil/blob/master/src/toil/job.py#L549), we need the wrapping `ServiceJob`. To do this, we can swap out the `fileStore` parameter that is passed in the `start`/`stop` methods with the underlying job (since the fileStore can be accessed through the job).

fnothaft:
Changes the `Job.Service` `start` and `stop` methods to pass `job` instead of `fileStore`, which allows `Job.Service` instances to register defer methods, etc. To do this, we also modify `ServiceJob.run` to set `self.fileStore` within the service job, which was previously not exposed.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -656,11 +656,12 @@ class Job(ResourceRequirementMixin):
             self._hasParent = False
 
         @abstractmethod
-        def start(self, fileStore):
+        def start(self, job):
             """
             Start the service.
             
-            :param toil.job.Job.FileStore fileStore: A fileStore object to create temporary files with.
+            :param toil.job.Job job: The underlying job that is being run. Can be used to register
+            deferred functions, or to access the fileStore for creating temporary files.
 
             :returns: An object describing how to access the service. The object must be pickleable \
             and will be used by jobs to access the service (see :func:`toil.job.Job.addService`).
@@ -668,12 +669,12 @@ class Job(ResourceRequirementMixin):
             pass
 
         @abstractmethod
-        def stop(self, fileStore):
+        def stop(self, job):
             """
-            Stops the service. 
+            Stops the service. Function can block until complete.
             
-            :param toil.job.Job.FileStore fileStore: A fileStore object to create temporary files with.
-                Function can block until complete.
+            :param toil.job.Job job: The underlying job that is being run. Can be used to register
+            deferred functions, or to access the fileStore for creating temporary files.
             """
             pass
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -92,9 +92,8 @@ clean_sdist:
 	- rm -rf dist
 
 
-test: check_venv check_build_reqs docker
```
hannes-ucsc: This shouldn't be committed.

src/toil/job.py
```diff
@@ -1384,12 +1385,16 @@ def __init__(self, service):
         self.jobWrapper = None
 
     def run(self, fileStore):
+        
+        # we need access to the filestore from underneath the service job
+        self.fileStore = fileStore
```
hannes-ucsc: The fact that you have to do this is a bad smell, coming from Toil. JobFunctionWrappingJob does the same assignment but other Job subclasses don't, most likely because they don't need it. We should decide if we pass the fileStore down or if we set it on the job *before* `run()` is being invoked.

Makefile
fnothaft: Yes, sorry, did this for local testing. Fixed.

fnothaft:
@hannes-ucsc tests are passing, ping for a review. If it looks good to you, I can squash. We need this for https://github.com/BD2KGenomics/toil-lib/pull/46 downstream.

src/toil/job.py
hannes-ucsc: I think it should be pretty easy to make sure that every job has `.fileStore` but if you don't want to do this as part of this PR, could you create a ticket?

fnothaft:
> I think it should be pretty easy to make sure that every job has .fileStore but if you don't want to do this as part of this PR, could you create a ticket?

Agreed. It is more of a bandwidth issue, unfortunately. I have a deadline next Thursday and thus don't have time for non-critical path issues. Since this is blocking @almussel via the upstream PR in toil-lib, I want to get this in. I've just squashed and rebased.

fnothaft:
Opened https://github.com/BD2KGenomics/toil/issues/1224, will follow up later.

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Can I get a merge?

hannes-ucsc:
The prospects of a linear history are starting to grow on me. If only I could get Jenkins to build a rebase instead of a merge. I do like the fact that the commit that's built by the PR is identical to the one that lands on master.

PR #1190, opened 2016-10-05
Add disk-usage logging to filestores (resolves #1188)

arkal:
resolves #1188

Added a simple check at the end of each job of how much disk space is used by the
job at the end of its lifespan. Log the results to INFO and then add a message to
WARNING if the job used more space than it requested.
Added a function to recursively get the size of a directory (including subdirectories).
Added a test for the function under a new test file `src/miscTests.py`

PR #1199, opened 2016-10-07
Add defer to uncached filestore (resolves #1189)

### LINKED ISSUE #1189: Add defer to inferior filestore
Depended on by #987

The uncached filestore does not support defer and it should. Proposing to add a jobstate file to each job's working directory such that it can be used to store defer information

arkal:
Resolves #1189, depended on by #987.

Add defer functionality to the uncached file store.


arkal:
This is incomplete and currently only addresses the abstraction of the file store.

arkal:
Jenkins, test this please.

arkal:

hannes-ucsc:
I had started a review a long time ago, didn't finish it and never clicked submit. I submitted it just now but still need to finish the review. Doing that right now.

src/toil/fileStore.py
```diff
@@ -47,11 +51,13 @@
     'kwargs'))
 
 
-class FileStore(object):
-    '''
-    Class used to manage temporary files, read and write files from the job store\
-    and log messages, passed as argument to the :func:`toil.job.Job.run` method.
-    '''
+class AbstractFileStore(object):
```
hannes-ucsc: Set meta class to abc.ABCMeta

src/toil/fileStore.py
```diff
@@ -47,11 +51,13 @@
...
+    """
+    An abstract (as far as Python currently allows) base class to represent the interface between a
```
hannes-ucsc: What does the "as far as Python currently allows" part mean?

src/toil/fileStore.py
```diff
@@ -64,10 +70,337 @@ def __init__(self, jobStore, jobWrapper, localTempDir, inputBlockFn):
...
+    @abstractmethod
+    def _blockFn(self):
+        """
+        Blocks while _updateJobWhenDone is running. This function is called by this job's
+        successor to ensure that it does not begin modifying the job store until after this job has
+        finished doing so.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
```
hannes-ucsc: I wouldn't explicitly require this method. It's an implementation detail that subclasses 

The method that you'll introduce for #1240 should be abstract, though.

src/toil/fileStore.py
```diff
@@ -64,10 +70,337 @@ def __init__(self, jobStore, jobWrapper, localTempDir, inputBlockFn):
...
+        :returns: Absolute path to key
+        :rtype: str
+        """
+        if key.startswith('/'):
+            return os.path.abspath(key)
+        else:
+            return os.path.join(self.localTempDir, key)
+
+    # A class to read and write dill-ed state dictionaries from/to a file into a namespace.
+    class _FileState(object):
```
hannes-ucsc: Rename to _StateFile.

src/toil/common.py
```diff
@@ -904,3 +908,50 @@ def cacheDirName(workflowID):
     :return: Name of the cache directory.
     """
     return 'cache-' + workflowID
+
+
+def getSizeOfDirectory(dirToSizeUp):
```
hannes-ucsc: I'd prefer `path` or `dirPath` for argument name

src/toil/common.py
```diff
@@ -904,3 +908,50 @@ def cacheDirName(workflowID):
...
+    '''
+    Returns the size of the directory including all linked files provided their inode was not in
+    linkedFileInodes.
+
+    :param str dirPath: Dir path
+    :param list fileNames: Files in the directory
+    :param set seenInodes: set of inode numbers of files that have multiple links, and have already
+           been acccounted for.
+    :return: Size of the directory in bytes
+    :rtype: float
```
hannes-ucsc: Why float? You can't have half a byte. And folderSize is initialized to an int and only ints are added so I don't see where that value gets converted to a float.

src/toil/fileStore.py
```diff
@@ -47,11 +51,13 @@
...
-    '''
-    Class used to manage temporary files, read and write files from the job store\
-    and log messages, passed as argument to the :func:`toil.job.Job.run` method.
-    '''
+class AbstractFileStore(object):
+    """
+    An abstract (as far as Python currently allows) base class to represent the interface between a
+    worker and the job store.  Classes inheriting this abstract class will be used to manage
+    temporary files, read and write files from the job store and log messages, passed as argument
+    to the :func:`toil.job.Job.run` method.
```
hannes-ucsc: `:meth:` ?

hannes-ucsc: See if link actually works (`make docs`).

src/toil/test/src/miscTests.py
```diff
@@ -0,0 +1,87 @@
...
+                os.link(linkSrc, fileName)
+                files[fileName] = 'Link to %s' % linkSrc
+
+        computedDirectorySize = getSizeOfDirectory(self.testDir)
+        totalExpectedSize = sum([x for x in files.values() if isinstance(x, int)])
+        self.assertEqual(computedDirectorySize, totalExpectedSize)
+
+    @staticmethod
+    def _getRandomName():
+        return uuid4().hex
```
hannes-ucsc: Consider using the base64 encoded random binary string. I like that better because it gives shorter, easier to distinguish random string.

hannes-ucsc: But its also a only a test method, so no biggy.

hannes-ucsc: BTW, I stopped declaring methods that don't access `self` as `@staticmethod`. In fact, I don't use `@staticmethod` at all. I use `@classmethod`, and only for methods that are truly used in a static context. Often it's sort of incidental that a method doesn't use `self`. And due to the incidental nature of that fact, it can also change because its not inherent to the way the method is used. I just grew tired of constantly removing and adding that decorator. I disabled the PyCharm warning "this method can be made static".

src/toil/fileStore.py
```diff
@@ -64,10 +70,337 @@ def __init__(self, jobStore, jobWrapper, localTempDir, inputBlockFn):
...
+
+        :param str key: The absolute or relative path to the file. If relative, it must be
+        relative to the local temp working dir
+        :returns: Absolute path to key
+        :rtype: str
+        """
+        if key.startswith('/'):
+            return os.path.abspath(key)
+        else:
+            return os.path.join(self.localTempDir, key)
```
arkal: Do you mean "what if key (renamed to filePath) is a symlink"?

src/toil/fileStore.py
arkal: I actually copied that first line from abstractBatchSystem (which I used as a template for an abstract class)

src/toil/common.py
arkal: This comes from #1190 (the branch i used as the base for this PR).

src/toil/common.py
arkal: This comes from #1190 (the branch i used as the base for this PR).

src/toil/test/src/miscTests.py
arkal: This comes from #1190 (the branch i used as the base for this PR).

src/toil/common.py
hannes-ucsc: Can you fix it there and rebase this branch?

src/toil/fileStore.py
hannes-ucsc: If the point of this method is to resolve symlinks, then this branch in the control flow doesn't achieve that. 

In /a/b/c, any component thereof could be a symlink. So /a/b could be a symlink, while c isn't.

src/toil/test/src/miscTests.py
hannes-ucsc: Can you fix it there and rebase?

arkal:
@hannes-ucsc this is ready!

src/toil/fileStore.py
hannes-ucsc: resolved.

src/toil/fileStore.py
```diff
@@ -1542,8 +1638,95 @@ def _updateJobWhenDone(self):
...
+                if filename == '.jobState':
+                    jobStateFiles.append(os.path.join(root, file))
+        for filename in jobStateFiles:
+            with open(filename) as fH:
+                yield dill.load(fH)
+
+    def _registerDeferredFunction(self, callable, *args, **kwargs):
+        """
+        This is the backend function to register a function that will be run after the completion of
+        the current job on the same node
```
hannes-ucsc: period at end of sentence and empty line thereafter

src/toil/fileStore.py
```diff
@@ -1542,8 +1638,95 @@ def _updateJobWhenDone(self):
...
+        :param bool batchSystemShutdown: Is the batch system in the process of shutting down?
+        :return:
+        """
+        # A list of tuples of (job name, pid or process running job, registered defer functions)
+        for jobState in cls._getAllJobStates(workflowDir):
+            if not FileStore.HarbingerFile._pidExists(jobState['jobPID']):
+                logger.warning('Detected that job (%s) prematurely terminated.  Fixing the state '
+                               'of the cache.', jobState['jobName'])
+                if not batchSystemShutdown:
+                    logger.debug("Deleting the stale working direcotry.")
```
hannes-ucsc: directory

src/toil/fileStore.py
```diff
@@ -1542,8 +1638,95 @@ def _updateJobWhenDone(self):
...
+                    # Delete the old work directory if it still exists.  Do this only during the
+                    # life of the program and dont' do it during the batch system cleanup.  Leave
+                    # that to the batch system cleanup code.
+                    shutil.rmtree(jobState['jobDir'])  # Ignore_errors?
```
hannes-ucsc: No

hannes-ucsc:
Once those nits are fixed, I will merge.

src/toil/fileStore.py
```diff
@@ -52,10 +52,9 @@
 
 class AbstractFileStore(object):
     """
-    An abstract (as far as Python currently allows) base class to represent the interface between a
-    worker and the job store.  Classes inheriting this abstract class will be used to manage
-    temporary files, read and write files from the job store and log messages, passed as argument
-    to the :meth:`toil.job.Job.run` method.
+    An abstract base class to represent the interface between a worker and the job store.  Classes
+    inheriting this abstract class will be used to manage temporary files, read and write files from
+    the job store and log messages, passed as argument to the :meth:`toil.job.Job.run` method.
```
arkal: the signature fro toil.job.Job.run is `(self, fileStore)`

src/toil/fileStore.py
```diff
@@ -1642,12 +1644,87 @@ def __del__(self):
...
+                cls._runDeferredFunctions(jobState['deferredFunctions'])
+
+    @staticmethod
+    def _getAllJobStates(workflowDir):
+        """
+        Generator function that undills and yields the jobState for every job on the node, one at a
+        time.
+
+        :param str workflowDir: The location of the workflow directory on the node.
+        :return: dict with keys (jobName,  pid, workDir, [deferredFunctions])
```
arkal: I'll get on it

arkal:
Jenkins, test this please.

arkal:
Failure in ba620bc is not a problem. The test is being discarded in the next commit anyway

arkal:
Jenkins, test this please.

arkal:
@hannes-ucsc could you please look at this? I will squash after I implement all your feedback since your feedback will be binned into one of the 4 final commits
```
1. Fix bug in NonCachingFileStore regarding local files
2. Add an abstract file store and refactor accordingly
3. Refactor filestore tests
4. Implement Defer in NonCachingFileStore (resolves #1189)
```

src/toil/test/src/fileStoreTest.py
```diff
@@ -1126,42 +1253,58 @@ def _deleteFileClassMethod(cls, nonLocalFile, nlf=None):
...
+class NonCachingFileStoreTestWithAzureJobStore(hidden.AbstractNonCachingFileStoreTest):
+    jobStoreType = 'azure'
+
+
+@needs_azure
+@experimental
+class CachingFileStoreTestWithAzureJobStore(hidden.AbstractCachingFileStoreTest):
+    jobStoreType = 'azure'
 
     @unittest.skipIf(testingIsAutomatic, "To save time")
```
hannes-ucsc: @unittest.skipIf(testingIsAutomatic and jobStoreType != 'file', "To save time")

src/toil/fileStore.py
```diff
@@ -49,26 +51,352 @@
...
+        raise NotImplementedError()
+
+
+class CachingFileStore(FileStore):
+    """
+    A cache-enabled file store that attempts to use hard-links and asynchronous job store writes to
+    reduce I/O between, and during jobs.
+    """
+
+    def __init__(self, jobStore, jobWrapper, localTempDir, inputBlockFn):
```
arkal: wow i completely overlooked that.

src/toil/fileStore.py
```diff
@@ -103,19 +429,13 @@ def __init__(self, jobStore, jobGraph, localTempDir, inputBlockFn):
...
     @contextmanager
     def open(self, job):
-        '''
+        """
         This context manager decorated method allows cache-specific operations to be conducted
         before and after the execution of a job in worker.py
-        :param job:
-        :return:
-        '''
+        :param job: job
```
arkal: Haha.... consistency?

src/toil/fileStore.py
```diff
@@ -821,11 +1040,13 @@ def returnFileSize(self, fileStoreID, cachedFileSource, lockFileHandle,
 
     @staticmethod
     def _isHidden(filePath):
-        '''
+        """
```
arkal: resolved

arkal:
@hannes-ucsc @cket this is ready

hannes-ucsc:
We should probably squash this down on merge.

Leaving final decision to @cket.

PR #1203, opened 2016-10-08
Delete SDB domains on failed jobstore bucket creation (resolves #1202)

### LINKED ISSUE #1202: Delete sdb domains if unable to create job store bucket
refer [This log](http://jenkins.cgcloud.info/job/toil-pull-requests/1902/consoleFull). Take for example, the test with uuid `6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9`. The logs say

```
INFO:toil.jobStores.aws.jobStore:Binding to job store domain 'toil-registry'.
INFO:toil.jobStores.aws.jobStore:Binding to job store domain 'toil-registry'.
INFO:toil.jobStores.aws.jobStore:Binding to job store domain 'jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--jobs'.
INFO:toil.jobStores.aws.jobStore:Binding to job store domain 'jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--files'.
INFO:toil.jobStores.aws.jobStore:Binding to job store bucket 'jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--files'.
DEBUG:toil.jobStores.aws.jobStore:Bucket 'jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--files' does not exist.
DEBUG:toil.jobStores.aws.jobStore:Creating bucket 'jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--files'.
 AWSJobStoreTest.test_testImportExportFile__otherCls_awsjobstoretest__size_partsize 
src/toil/test/jobStores/jobStoreTest.py:104: in setUp
    self.master.initialize(self.config)
src/toil/jobStores/aws/jobStore.py:204: in initialize
    self._bind(create=True)
src/toil/jobStores/aws/jobStore.py:232: in _bind
    versioning=True)
src/toil/jobStores/aws/jobStore.py:674: in _bindBucket
    bucket = self.s3.create_bucket(bucket_name, location=location)
...
...
E           S3ResponseError: S3ResponseError: 400 Bad Request
E           <?xml version="1.0" encoding="UTF-8"?>
E           <Error><Code>TooManyBuckets</Code><Message>You have attempted to create more buckets than allowed</Message><CurrentNumberOfBuckets>101</CurrentNumberOfBuckets><AllowedNumberOfBuckets>101</AllowedNumberOfBuckets><RequestId>CA9B35247D82A7B9</RequestId><HostId>atOhiIDkPFkyEOdnBGfPkEzJbswJpWZpz1hLC4ofc7eSF6jQX0Zg/SyMNYhIfeaXlOaV1RAmWJ0=</HostId></Error>
```

We've apparently reached the limit on number of buckets, so Toil fails..... without deleting the bound jobstore domains -- as evident by

```
>>> [z for z in x.get_all_domains(max_domains=100) if '6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9' in z.name]
[Domain:jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--files,
 Domain:jobstore-test-6cdc03f3-e1c1-4fab-97ea-19cc74fb1ea9--jobs]
```

arkal:
resolves #1202

We see a lot of leaked `jobstore-test-UUID` and `cache-test-UUID` domains in SDB
and those are due to the domains being bound before creating the s3 bucket for the
jobstore. If the bucket creation fails, Toil exited without removing the domains. This
PR adds a flag to ensure the jobstore is properly initialized, and adds logic to handle
the deletion of a failed fresh jobstore.

Added a test for this feature.

### DIFF
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -100,6 +100,10 @@ class JobStoreExistsException(Exception):
             "the job store with 'toil clean' to start the workflow from scratch" % locator)
 
 
+class JobStoreBucketExistsException(Exception):
+    def __init__(self, message):
+        super(JobStoreBucketExistsException, self).__init__(message)
+
 class AbstractJobStore(object):
     """ 
     Represents the physical storage for the jobs and files in a Toil workflow.
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -100,9 +100,12 @@ class JobStoreExistsException(Exception):
             "the job store with 'toil clean' to start the workflow from scratch" % locator)
 
 
-class JobStoreBucketExistsException(Exception):
-    def __init__(self, message):
-        super(JobStoreBucketExistsException, self).__init__(message)
+class BucketLocationConflictException(Exception):
+    def __init__(self, bucketRegion):
+        message = ('A bucket with the same name as the jobstore was found in another region (%s). '
+                   'Cannot proceed as the unique bucket name is already in use.' % bucketRegion)
+        super(BucketLocationConflictException, self).__init__(message)
+
 
 class AbstractJobStore(object):
     """ 
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -48,6 +48,7 @@ from toil.jobStores.abstractJobStore import (AbstractJobStore,
                                              ConcurrentFileModificationException,
                                              NoSuchFileException,
                                              NoSuchJobStoreException,
+                                             JobStoreBucketExistsException,
                                              JobStoreExistsException)
 from toil.jobStores.aws.utils import (SDBHelper,
                                       retry_sdb,
@@ -681,6 +682,21 @@ class AWSJobStore(AbstractJobStore):
                             raise
                         else:
                             return None
+                    elif e.status == 301:
+                        # This is raised if the user attempts to get a bucket in a region outside
+                        # the specified one, if the specified one is not `us-east-1`.  The us-east-1
+                        # server allows a user to use buckets from any region.
+                        if self.region == 'us-east-1':
+                            message = ('Region is us-east-1 but 301 error was raised when '
+                                       'attempting to create the bucket. 301 Errors are usually '
+                                       'raised when a non-us-east-1 connection attempts to get a '
+                                       'bucket from another region. Cannot proceed further.')
+                        else:
+                            b = self.s3.get_bucket(bucket_name, validate=False)
+                            message = ('A bucket with the same name as the jobstore was found in '
+                                       'another region (%s). Cannot proceed further as the unique '
+                                       'jobstore name is already claimed.' % b.get_location())
+                        raise JobStoreBucketExistsException(message)
                     else:
                         raise
                 else:
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -48,7 +48,7 @@ from toil.jobStores.abstractJobStore import (AbstractJobStore,
                                              ConcurrentFileModificationException,
                                              NoSuchFileException,
                                              NoSuchJobStoreException,
-                                             JobStoreBucketExistsException,
+                                             BucketLocationConflictException,
                                              JobStoreExistsException)
 from toil.jobStores.aws.utils import (SDBHelper,
                                       retry_sdb,
@@ -202,10 +202,15 @@ class AWSJobStore(AbstractJobStore):
         if self._registered:
             raise JobStoreExistsException(self.locator)
         self._registered = None
-        self._bind(create=True)
-        super(AWSJobStore, self).initialize(config)
-        # Only register after job store has been full initialized
-        self._registered = True
+        try:
+            self._bind(create=True)
+        except:
+            with panic(log):
+                self.destroy()
+        else:
+            super(AWSJobStore, self).initialize(config)
+            # Only register after job store has been full initialized
+            self._registered = True
 
     @property
     def sseKeyPath(self):
@@ -222,6 +227,9 @@ class AWSJobStore(AbstractJobStore):
             assert len(name) <= self.maxNameLen
             return self.namePrefix + self.nameSeparator + name
 
+        # The order in which this sequence of events happens is important.  We can easily handle the
+        # inability to bind a domain, but it is a little harder to handle some cases of binding the
+        # jobstore bucket.  Maintaining this order allows for an easier `destroy` method.
         if self.jobsDomain is None:
             self.jobsDomain = self._bindDomain(qualify('jobs'), create=create, block=block)
         if self.filesDomain is None:
@@ -686,21 +694,13 @@ class AWSJobStore(AbstractJobStore):
                         # This is raised if the user attempts to get a bucket in a region outside
                         # the specified one, if the specified one is not `us-east-1`.  The us-east-1
                         # server allows a user to use buckets from any region.
-                        if self.region == 'us-east-1':
-                            message = ('Region is us-east-1 but 301 error was raised when '
-                                       'attempting to create the bucket. 301 Errors are usually '
-                                       'raised when a non-us-east-1 connection attempts to get a '
-                                       'bucket from another region. Cannot proceed further.')
-                        else:
-                            b = self.s3.get_bucket(bucket_name, validate=False)
-                            message = ('A bucket with the same name as the jobstore was found in '
-                                       'another region (%s). Cannot proceed further as the unique '
-                                       'jobstore name is already claimed.' % b.get_location())
-                        raise JobStoreBucketExistsException(message)
+                        bucket = self.s3.get_bucket(bucket_name, validate=False)
+                        raise BucketLocationConflictException(self.__getBucketRegion(bucket))
                     else:
                         raise
                 else:
-                    assert self.__getBucketRegion(bucket) == self.region
+                    if self.__getBucketRegion(bucket) != self.region:
+                        raise BucketLocationConflictException(self.__getBucketRegion(bucket))
                     assert versioning is self.__getBucketVersioning(bucket)
                     log.debug("Using existing job store bucket '%s'.", bucket_name)
                     return bucket
@@ -1271,7 +1271,14 @@ class AWSJobStore(AbstractJobStore):
     def destroy(self):
         # FIXME: Destruction of encrypted stores only works after initialize() or .resume()
         # See https://github.com/BD2KGenomics/toil/issues/1041
-        self._bind(create=False, block=False)
+        try:
+            self._bind(create=False, block=False)
+        except BucketLocationConflictException:
+            # If the unique jobstore bucket name existed, _bind would have raised a
+            # BucketLocationConflictException before calling destroy.  Calling _bind here again
+            # would reraise the same exception so we need to catch and ignore that exception.
+            pass
+        # TODO: Add other failure cases to be ignored here.
         self._registered = None
         if self.filesBucket is not None:
             self._delete_bucket(self.filesBucket)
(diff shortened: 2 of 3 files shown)


arkal:
@hannes-ucsc I can add a test to ensure domains are not deleted in
restart case but that will either leak domains, or require automated domain removal
and I can implement it if you're ok with that!

arkal:
Jenkins, test this please.

arkal:
Booyeah! @hannes-ucsc this is ready!

src/toil/jobStores/aws/jobStore.py
```diff
@@ -201,10 +202,16 @@ def initialize(self, config):
...
-        self._bind(create=True)
-        super(AWSJobStore, self).initialize(config)
-        # Only register after job store has been full initialized
-        self._registered = True
+        try:
+            self._bind(create=True)
+        except:
+            with panic(log):
+                self.destroy()
+            raise
```
hannes-ucsc: I believe panic() raises already. But double check.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -681,6 +688,21 @@ def bucket_creation_pending(e):
...
+                        if self.region == 'us-east-1':
+                            message = ('Region is us-east-1 but 301 error was raised when '
+                                       'attempting to create the bucket. 301 Errors are usually '
+                                       'raised when a non-us-east-1 connection attempts to get a '
+                                       'bucket from another region. Cannot proceed further.')
+                        else:
+                            b = self.s3.get_bucket(bucket_name, validate=False)
+                            message = ('A bucket with the same name as the jobstore was found in '
+                                       'another region (%s). Cannot proceed further as the unique '
+                                       'jobstore name is already claimed.' % b.get_location())
```
hannes-ucsc: I think you can only call get_location() if the connection is to us-east-1. if I am right, then this branch of the control flow is not covered by tests.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -681,6 +688,21 @@ def bucket_creation_pending(e):
                             raise
                         else:
                             return None
+                    elif e.status == 301:
+                        # This is raised if the user attempts to get a bucket in a region outside
+                        # the specified one, if the specified one is not `us-east-1`.  The us-east-1
+                        # server allows a user to use buckets from any region.
+                        if self.region == 'us-east-1':
+                            message = ('Region is us-east-1 but 301 error was raised when '
```
hannes-ucsc: I think that we don't need to make the message that detailed. The message to the user should be that a bucket of the given name exists but is not the the specified reason. The us-east-1 thing is a mere implementation detail.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -36,10 +36,11 @@
 # (installed by `make prepare`)
 from mock import patch
 
-from toil.common import Config
+from toil.common import Config, Toil
+from toil.job import Job
 from toil.jobStores.abstractJobStore import (AbstractJobStore,
                                              NoSuchJobException,
-                                             NoSuchFileException)
+                                             NoSuchFileException, JobStoreBucketExistsException)
```
hannes-ucsc: Wrap all or none.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -100,6 +100,10 @@ def __init__(self, locator):
             "the job store with 'toil clean' to start the workflow from scratch" % locator)
 
 
+class JobStoreBucketExistsException(Exception):
+    def __init__(self, message):
+        super(JobStoreBucketExistsException, self).__init__(message)
```
hannes-ucsc: Consider hardcoding the message here and just passing the actual location:

```
def __init__(self, bucketLocation):
    …
```

src/toil/jobStores/abstractJobStore.py
```diff
@@ -100,6 +100,10 @@ def __init__(self, locator):
             "the job store with 'toil clean' to start the workflow from scratch" % locator)
 
 
+class JobStoreBucketExistsException(Exception):
```
hannes-ucsc: Shouldn't the name be `BucketLocationConflictException`? We do handle existing buckets, just not in the wrong region.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -839,6 +840,49 @@ def _corruptJobStore(self):
         assert isinstance(self.master, AWSJobStore)  # type hinting
         self.master.filesBucket.delete()
 
+    def testSDBDomainsDeletedOnFailedJobstoreBucketCreation(self):
+        """
+        This test ensures that SDB domains bound to a jobstore are deleted if the jobstore bucket
+        failed to be created.  We simulate a failed jobstore bucket creation by using a bucket in a
+        different region with the same name.
+
+        :returns: None
```
hannes-ucsc: Probably redundant.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -839,6 +840,49 @@ def _corruptJobStore(self):
...
+            sdb = connect_to_region(self.awsRegion())
+            next_token = None
+            allDomainNames = []
+            while True:
+                domains = sdb.get_all_domains(max_domains=100, next_token=next_token)
+                allDomainNames.extend([x.name for x in domains])
+                next_token = domains.next_token
+                if next_token is None:
+                    break
+            # Will the user ever have 0 domains?
```
hannes-ucsc: They might. The registry should exist but domain creation is eventually consistent, so you might not see it if this was the first time a job store was created in the region.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -839,6 +840,49 @@ def _corruptJobStore(self):
...
+        failed to be created.  We simulate a failed jobstore bucket creation by using a bucket in a
+        different region with the same name.
+
+        :returns: None
+        """
+        from boto.sdb import connect_to_region
+        from boto.s3.connection import Location, S3Connection
+        aws_west_regions = {'us-west-1': Location.USWest,
+                            'us-west-2': Location.USWest2}
+        externalAWSRegion = [R for r, R in aws_west_regions.items() if r != self.awsRegion()][0]
```
hannes-ucsc: `region, location` instead of `r, R` would probably be more readable. If line gets too long wrap before `if`

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -839,6 +840,49 @@ def _corruptJobStore(self):
...
+        """
+        This test ensures that SDB domains bound to a jobstore are deleted if the jobstore bucket
+        failed to be created.  We simulate a failed jobstore bucket creation by using a bucket in a
+        different region with the same name.
+
+        :returns: None
+        """
+        from boto.sdb import connect_to_region
+        from boto.s3.connection import Location, S3Connection
+        aws_west_regions = {'us-west-1': Location.USWest,
```
hannes-ucsc: Camel case

src/toil/jobStores/aws/jobStore.py
arkal: yup it does

src/toil/test/jobStores/jobStoreTest.py
arkal: huh weird. Must have hit backspace twice. This was unintentional.

src/toil/jobStores/abstractJobStore.py
arkal: true. I like that name better

src/toil/jobStores/abstractJobStore.py
arkal: I was thinking of it but there are two possible messages that make it through hence the format i used.

arkal: However, given you rejected my other log message, i can implement this.

src/toil/test/jobStores/jobStoreTest.py
arkal: On further inspection, I learnt that this happened because i used `alt+enter` to import the exception into the test.

arkal:

src/toil/jobStores/aws/jobStore.py
hannes-ucsc: Tested this manually and seems to work.

```
>>> from boto.s3 import connect_to_region
>>> s3=connect_to_region('us-west-1')
>>> b=s3.get_bucket('bd2k-artifacts',validate=False)
>>> b.get_location()
u'us-west-2'
```

The other branch is going away right?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -681,6 +687,13 @@ def bucket_creation_pending(e):
                             raise
                         else:
                             return None
+                    elif e.status == 301:
+                        # This is raised if the user attempts to get a bucket in a region outside
+                        # the specified one, if the specified one is not `us-east-1`.  The us-east-1
+                        # server allows a user to use buckets from any region.
+                        assert self.region != 'us-east-1'
```
hannes-ucsc: I wouldn't assert this. Assertions are for things that you could break by changing your code. This assertion depends on Amazon. If Amazon decides to send redirects from us-east-1 this would break.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1255,7 +1268,14 @@ def __getBucketRegion(self, bucket):
     def destroy(self):
         # FIXME: Destruction of encrypted stores only works after initialize() or .resume()
         # See https://github.com/BD2KGenomics/toil/issues/1041
-        self._bind(create=False, block=False)
+        try:
+            self._bind(create=False, block=False)
+        except BucketLocationConflictException:
+            # If the jobstore bucket existed, _bind would have raised BucketLocationConflictException
```
hannes-ucsc: I dislike that this is dependent on the fact that domains are bound before the bucket. But I don't have any other idea. Except adding another mode to bind, like `ignore_errors=False` which I also dislike.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -839,6 +841,45 @@ def _corruptJobStore(self):
...
+            next_token = None
+            allDomainNames = []
+            while True:
+                domains = sdb.get_all_domains(max_domains=100, next_token=next_token)
+                allDomainNames.extend([x.name for x in domains])
+                next_token = domains.next_token
+                if next_token is None:
+                    break
+            self.assertFalse([d for d in allDomainNames if testJobStoreUUID in d])
+        finally:
```
hannes-ucsc:
```
else:
    self.fail()
```

Right?

src/toil/jobStores/aws/jobStore.py
arkal: Hmmm... we still want to fail in this case though, right?

hannes-ucsc: Yes, I would just remove the assertion. The assumption we operate on is: 

If a `HEAD bucket` request on a connection to region A results in a 301, the bucket is in a region other than A. If it succeeds (200) but a subsequent `GET location` request over the same connection returns a region other than A, the bucket is also not in A.

Important thing to keep in mind: location is the same as region, except for region 'us-east-1' where location is the empty string. You should use toil.jobStores.aws.utils.region_to_bucket_location and toil.jobStores.aws.utils.bucket_location_to_region for translating between the two if you need to.

If we want to have 100% coverage we should also test this with a connection to `us-east-1` and the bucket in one of the other two US regions.

src/toil/jobStores/aws/jobStore.py
hannes-ucsc: Maybe add a comment in _bind() to alert the developer of this ordering sensitivity.

arkal: I'm actually not sure what you'd want here and I guarantee that you will edit whatever i put in so do you have anything in particular you wanted?

arkal:
Jenkins, test this please.

arkal:

PR #1213, opened 2016-10-18
Periodically check on scaler thread continuance (resolves #1208)

### LINKED ISSUE #1208: Failures in the cluster scaler thread aren't propagated
If an exception is raised that kills the scaler thread Toil will try to continue running, potentially hanging until it is killed.

cket:
Resolves #1208

### DIFF
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -244,7 +244,7 @@ class ClusterScaler(object):
         for scalerThread in [self.preemptableScaler, self.scaler]:
             if scalerThread is not None:
                 try:
-                    scalerThread.join(0)
+                    scalerThread.join(timeout=0)
                 except Exception as e:
                     logger.exception(e)
                     exception = True
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -235,6 +235,23 @@ class ClusterScaler(object):
         else:
             self.scaler = None
 
+    def check(self):
+        """
+        Attempt to join any existing scaler threads that may have died or finished. This insures
+        any exceptions raised in the threads are propagated in a timely fashion.
+        """
+        exception = False
+        for scalerThread in [self.preemptableScaler, self.scaler]:
+            if scalerThread is not None:
+                try:
+                    scalerThread.join(0)
+                except Exception as e:
+                    logger.exception(e)
+                    exception = True
+        if exception:
+            raise RuntimeError('The cluster scaler has exited due to an exception')
+
+
     def shutdown(self):
         """
         Shutdown the cluster.
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -905,9 +905,9 @@ def innerLoop(jobStore, config, batchSystem, toilState, jobBatcher, serviceManag
 
                 elif jobWrapper.jobStoreID in toilState.servicesIssued:
                     logger.debug("Telling job: %s to terminate its due to the successful completion of its successor jobs",
-                                jobWrapper.jobStoreID)
+                                 jobWrapper.jobStoreID)
                     serviceManager.killServices(toilState.servicesIssued[jobWrapper.jobStoreID],
-                                                    error=False)
+                                                error=False)
 
                 #There are no remaining tasks to schedule within the jobWrapper, but
                 #we schedule it anyway to allow it to be deleted.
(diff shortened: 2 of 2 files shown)


hannes-ucsc:
A few thoughts: This mildly different to what I proposed in https://github.com/BD2KGenomics/toil/issues/1208#issuecomment-253812788

With this approach, can you guarantee that both Threads are joined? If both fail, will we be able to detect both failures and print stack traces for them?

cket:
Jenkins, test this please

cket:
@hannes-ucsc ahh, joining with a timeout would be better - I missed that comment, thanks for catching that!

src/toil/provisioners/clusterScaler.py
```diff
@@ -235,6 +235,30 @@ def __init__(self, provisioner, jobBatcher, config):
...
+        Attempt to join any existing scaler threads with a short timeout. This insures an exception
+        raised in the threads is propagated in a timely fashion. If there are exceptions in both threads,
+        one exception is logged before the other is raised to insure that both tracebacks are visible
+        """
+        exceptions = []
+        if self.preemptableScaler:
+            try:
+                self.preemptableScaler.join(1)
+            except Exception as e:
+                exceptions.append(e)
```
hannes-ucsc: I think to get an accurate backtrace, you need to log the exception right away with log.error(…,excinfo=True). That is because the backtrace is not part of the exception, but stored in a global variable called [sys.exc_info](https://docs.python.org/2/library/sys.html#sys.exc_info) (one of Python's design flaws).

src/toil/provisioners/clusterScaler.py
```diff
@@ -235,6 +235,30 @@ def __init__(self, provisioner, jobBatcher, config):
...
+        if self.scaler:
+            try:
+                self.scaler.join(1)
+            except Exception as e:
+                exceptions.append(e)
+        if len(exceptions) > 1:
+            # we want stacktraces for both so we have to log one exception before raising the other
```
hannes-ucsc: Addressing my comment above would make the list and this special casing obsolete.

I'd be ok with raising a secondary exception if the primary ones are being logged with stack trace.

Independently, I'd feel strange about the asymmetric handling of those two exceptions. But probably unfounded.

src/toil/provisioners/clusterScaler.py
```diff
@@ -235,6 +235,30 @@ def __init__(self, provisioner, jobBatcher, config):
         else:
             self.scaler = None
 
+    def check(self):
+        """
+        Attempt to join any existing scaler threads with a short timeout. This insures an exception
+        raised in the threads is propagated in a timely fashion. If there are exceptions in both threads,
```
cket: I kept the original wording here since my new approach doesn't actually raise the errors, just logs them

cket:
My new approach is slightly different from my old one but I think it solves all of the problems you brought up. Now both exceptions are handled symmetrically and their tracebacks are visible and I removed the special case and the exception aggregation. 

The tradeoff is that neither exception is actually raised, just logged. Instead, a runtime error is raised at the end of the method if there was an exception in either thread.

cket:
Jenkins, test this please

cket:
@hannes-ucsc ready for review

src/toil/provisioners/clusterScaler.py
hannes-ucsc: Makes sense. I didn't read the sentence correctly.

hannes-ucsc:
https://github.com/BD2KGenomics/toil/pull/1213#discussion_r84047755

cket:
I missed another comment, I need to be more careful with this PR.. Sorry about that!

cket:
Jenkins, test this please

cket:
Jenkins, test this please

cket:
Jenkins, test this please

cket:
@hannes-ucsc ping for review

src/toil/provisioners/clusterScaler.py
```diff
@@ -235,6 +235,23 @@ def __init__(self, provisioner, jobBatcher, config):
...
+    def check(self):
+        """
+        Attempt to join any existing scaler threads that may have died or finished. This insures
+        any exceptions raised in the threads are propagated in a timely fashion.
+        """
+        exception = False
+        for scalerThread in [self.preemptableScaler, self.scaler]:
+            if scalerThread is not None:
+                try:
+                    scalerThread.join(0)
```
hannes-ucsc: `timeout=0` to make the intent more obvious

src/toil/provisioners/clusterScaler.py
```diff
@@ -235,6 +235,23 @@ def __init__(self, provisioner, jobBatcher, config):
...
+                except Exception as e:
+                    logger.exception(e)
+                    exception = True
+        if exception:
+            raise RuntimeError('The cluster scaler has exited due to an exception')
```
hannes-ucsc: `A scaler thread …` or `One of the …` and mention that the exception was logged.

hannes-ucsc: Never mind. Leave as is.

PR #1215, opened 2016-10-18
DOC: fix syntax error in dynamic example

cswarth:
Dynamic example in toil/docs/developing.rst has a syntax error:

```
$ python dynamic.py
  File "dynamic.py", line 3
    def binaryStringFn(job, message="", depth):
SyntaxError: non-default argument follows default argument
```

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -335,7 +335,7 @@ For example::
 
     from toil.job import Job
     
-    def binaryStringFn(job, message="", depth):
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Thank you for taking the time to fix this. Apparently, this example was never actually tested ;-(

Would it be possible to swap the `depth` and `message` arguments such that we don't need to provide a default value for `depth`?

PR #1217, opened 2016-10-18
Add ftp support for file import

### LINKED ISSUE #1209: Import files from FTP
It looks like there's a [fake JobStore that reads from http and https for file import](https://github.com/BD2KGenomics/toil/blob/master/src/toil/jobStores/abstractJobStore.py#L963) using urllib2. Since urllib2 can also do FTP, it would be easy to just add "ftp" to the list of protocols that this JobStore supports, to enable importing files from FTP.

Since many bioinformatics data sets live on FTP (like 1000 Genomes and the GRC reference assembly), it would be nice to just be able to point Toil at those URLs and have those resources available.

adamnovak:
Closes #1209.

There are no tests for the other schemes (http and https), so I'm not adding a test for ftp.

### DIFF
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -960,7 +960,7 @@ class JobStoreSupport(AbstractJobStore):
 
     @classmethod
     def _supportsUrl(cls, url, export=False):
-        return url.scheme.lower() in ('http', 'https') and not export
+        return url.scheme.lower() in ('http', 'https', 'ftp') and not export
 
     @classmethod
     def _readFromUrl(cls, url, writable):


JakeNarkizian:
@adamnovak There are [tests for http imports](https://github.com/BD2KGenomics/toil/blob/master/src/toil/test/jobStores/jobStoreTest.py#L456). 

However since you proposed the ftp feature I've been thinking about the way we are testing these. What we have currently is not actually a unit test as it uses some github url. This could theoretically fail because of problems outside of the code we are trying to test.

 I propose rewriting the test to use a [docker image with apache web server](https://hub.docker.com/_/httpd/)  which can be used for testing both ftp and http imports. We then do not run the risk of the test failing if for whatever reason [the github web server is down](http://arstechnica.com/security/2015/03/massive-denial-of-service-attack-on-github-tied-to-chinese-government/) @adamnovak I am willing to write this in a separate PR if you don't have the time.  

@hannes-ucsc What do you think?

hannes-ucsc:
> There are no tests for the other schemes (http and https), so I'm not adding a test for ftp.

This rationale made me smile.

hannes-ucsc:
@JakeNarkizian I like the idea of running a server locally. However, Python actually has an HTTP server [built-in](https://docs.python.org/2/library/simplehttpserver.html). For FTP, I've had good experience with https://github.com/giampaolo/pyftpdlib. We'd have to add it as a test dependency, see `make prepare`.

adamnovak:
I definitely don't want to redesign the test system, and while I know there
should be FTP tests, it is harder to find an FTP server to borrow for
testing than an HTTP one, so the redesign does sound necessary.

hannes-ucsc:
> I definitely don't want to redesign the test system

That's totally fine. I on the other hand can't accept new features without tests. At the very least we should have a test against a public FTP server URL. I can't imagine that finding a stable FTP URL on the internet is as hard as you claim it is.

@JakeNarkizian, could you pick this one up in a alternative PR?

PR #1223, opened 2016-10-20
Add ftp and test

### LINKED ISSUE #1209: Import files from FTP

adamnovak:
Here's my FTP support with a test using a random Internet FTP server, while we wait for @JakeNarkizian to redesign the tests that download things.

Resolves #1209

### DIFF
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -960,7 +960,7 @@ class JobStoreSupport(AbstractJobStore):
 
     @classmethod
     def _supportsUrl(cls, url, export=False):
-        return url.scheme.lower() in ('http', 'https') and not export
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
For future reference, commits should refer to the ticket they resolve. And there's no need to split new features and test in separate commits. That's why I squashed this down.

hannes-ucsc:
https://github.com/BD2KGenomics/toil/issues/1226

PR #1228, opened 2016-10-20
Enable TTY mode when SSH-ing into Toil Appliance (resolves #1227)

### LINKED ISSUE #1227: Missing -t flag for AWS provisioner ssh-cluster util
Without the flag the util is functional but very ugly

cket:
resolves #1227

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -19,6 +19,7 @@ import logging
 import time
 from contextlib import contextmanager
 
+import sys
 from boto.ec2.blockdevicemapping import BlockDeviceMapping, BlockDeviceType
 from boto.exception import BotoServerError, EC2ResponseError
 from cgcloud.lib.ec2 import (ec2_instance_types, retry_ec2, wait_spot_requests_active, a_short_time,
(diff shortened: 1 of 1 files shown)


cket:
The tests are failing with 
 ```Pseudo-terminal will not be allocated because stdin is not a terminal.```
This seems to be a jenkins specific problem but thats probably why I disabled it in the first place. I'll fix it so tty is enabled only when the method is invoked via the ssh util.

cket:
@hannes-ucsc ready

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -82,11 +83,13 @@ def getNodeShape(self, preemptable=False):
...
-        cls._sshAppliance(leader.ip_address, 'bash')
+        tty = sys.stdin.isatty()
+        cls._sshAppliance(leader.ip_address, 'bash', tty=tty)
 
     @classmethod
-    def _sshAppliance(cls, leaderIP, command):
-        command = "ssh -o \"StrictHostKeyChecking=no\" -t core@%s \"docker exec -i leader %s\"" % (leaderIP, command)
+    def _sshAppliance(cls, leaderIP, command, tty=False):
+        ttyFlag = 't' if tty else ''
+        command = "ssh -o \"StrictHostKeyChecking=no\" -t core@%s \"docker exec -i%s leader %s\"" % (leaderIP, ttyFlag, command)
```
hannes-ucsc: Change outer string to single quotes and eliminate backslashes before inner double quotes.

Think of a backslash as a character to avoid. Sorry for pushing back this PR for such a small thing but I have been asking people to avoid backslashes for about a year now. This should not be news.

hannes-ucsc:
I will squash this down when I merge.

PR #1233, opened 2016-10-21
Use stub servers for import tests (resolves #1226)

### LINKED ISSUE #1226: Eliminate reliance on external servers for import tests
HTTP and FTP. See also #1209 and #1217 and, in particular,

https://github.com/BD2KGenomics/toil/pull/1217#issuecomment-254801319

JakeNarkizian:
@hannes-ucsc I ended up going with [stubserver](https://github.com/tarttelin/Python-Stub-Server) for the FTP test. I think it is better because it does not rely on any particular file system and would not require writing a test file to said file system.

With that same reasoning I subclassed the built-in http mock server into a stub server. Please let me know what you think!

### DIFF
--- a/src/toil/test/jobStores/jobStoreTest.py
+++ b/src/toil/test/jobStores/jobStoreTest.py
@@ -14,9 +14,11 @@
 
 from __future__ import absolute_import
 
+import SocketServer
 import hashlib
 import logging
-
+import threading
+import SimpleHTTPServer
 import os
 import shutil
 import tempfile
@@ -24,6 +26,7 @@ import time
 import urllib2
 import urlparse
 import uuid
+from stubserver import FTPStubServer
 from Queue import Queue
 from abc import abstractmethod, ABCMeta
 from itertools import chain, islice, count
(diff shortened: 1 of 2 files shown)


src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1060,4 +1078,14 @@ class EncryptedAzureJobStoreTest(AzureJobStoreTest, AbstractEncryptedJobStoreTes
     pass
 
 
+class StubHttpServer(SimpleHTTPServer.SimpleHTTPRequestHandler):
```
hannes-ucsc: This isn't a server, it's a handler, correct? Shouldn't the name reflect that?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -454,17 +457,32 @@ def testImportSharedFile(self, otherCls):
                        otherCls=activeTestClassesByName)
 
         def testImportHttpFile(self):
-            srcUrl, srcHash = ('https://raw.githubusercontent.com/BD2KGenomics/toil'
-                               '/12c711ff91caa5d8fb37a072bad376f447faa797/Makefile',
-                               '16a38fd7ae233616083d43d81159d4e8')
-            with self.master.readFileStream(self.master.importFile(srcUrl)) as f:
-                self.assertEqual(hashlib.md5(f.read()).hexdigest(), srcHash)
+            httpd = SocketServer.TCPServer(('', 0), StubHttpServer)
```
hannes-ucsc: Not a daemon, so why call this httpd?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -454,17 +457,32 @@ def testImportSharedFile(self, otherCls):
...
+            httpdThread = threading.Thread(target=httpd.handle_request)
```
hannes-ucsc: If this line raises, httpd.server_close() isn't called.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -454,17 +457,32 @@ def testImportSharedFile(self, otherCls):
...
+            httpdThread.start()
+            try:
+                assignedPort = httpd.server_address[1]
```
hannes-ucsc: cool

src/toil/test/jobStores/jobStoreTest.py
hannes-ucsc: It might be desirable to not handle just one request, but be prepared to handle multiple requests, even though right now the import code does exactly one request. In the future, we could then support HEAD requests, multi-part downloads etc. for HTTP imports. I think this wouldn't involve more code, just using `serve_forever` instead of handle_request and `shutdown` at the end. If that turns out to be complicated, leave it the way it is.

JakeNarkizian: I originally had it that way but changed it to handle one request for simplicity. But I see your point, and you are correct it only requires two changes.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -456,17 +459,35 @@ def testImportSharedFile(self, otherCls):
...
+                httpThread = threading.Thread(target=http.serve_forever)
+                httpThread.start()
+                try:
+                    assignedPort = http.server_address[1]
+                    url = 'http://localhost:%d' % assignedPort
+                    with self.master.readFileStream(self.master.importFile(url)) as readable:
+                        self.assertEqual(readable.read(), StubHttpRequestHandler.fileContents)
+                finally:
+                    try:
+                        http.shutdown()
```
JakeNarkizian: @hannes-ucsc shutdown is idempotent and must be called before the thread is joined.

JakeNarkizian:
@hannes-ucsc

PR #1235, opened 2016-10-24
Run volume-mounting.service before docker.service (resolves #1230)

### LINKED ISSUE #1230: Toil appliance must start docker after mounting volumes
@cket and I ran into this bug this afternoon when testing out the autoscaler. On machines with multiple ephemeral volumes, the Toil appliance JBODs the volumes together and mounts them to `/var/lib/docker`, `/var/lib/toil`, etc. If this happens after the Docker daemon launches, docker fails sporadically when writing to `/var/lib/docker`, and the Toil appliance does an infinite failure loop.

cket:
resolves #1230


hannes-ucsc:
Would you like me to merge this?

cket:
Yes please

PR #1243, opened 2016-10-25
Handle orphaned jobstore writes on job failure (resolves #1239)

### LINKED ISSUE #1239: Ensure that job store files created by a failed job attempt are deleted
Write test first.

arkal:
resolves #1239

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -1580,8 +1580,8 @@ class NonCachingFileStore(FileStore):
                                  "script to avoid the chance of failure due to incorrectly "
                                  "requested resources. " + logString, level=logging.WARNING)
             os.chdir(startingDir)
-            jobState = self._readJobState(self.jobStateFile)
-            deferredFunctions = jobState['deferredFunctions']
+            jobState = self._JobState.read(self.jobStateFile)
+            deferredFunctions = jobState.deferredFunctions
             failures = self._runDeferredFunctions(deferredFunctions)
             for failure in failures:
                 self.logToMaster('Deferred function "%s" failed.' % failure, logging.WARN)
@@ -1676,11 +1676,11 @@ class NonCachingFileStore(FileStore):
         """
         # A list of tuples of (job name, pid or process running job, registered defer functions)
         for jobState in cls._getAllJobStates(nodeInfo):
-            if not cls._pidExists(jobState['jobPID']):
+            if not cls._pidExists(jobState.jobPID):
                 # using same logic to prevent races as CachingFileStore._setupCache
                 myPID = str(os.getpid())
-                cleanupFile = os.path.join(jobState['jobDir'], '.cleanup')
-                with open(os.path.join(jobState['jobDir'], '.' + myPID), 'w') as f:
+                cleanupFile = os.path.join(jobState.jobDir, '.cleanup')
+                with open(os.path.join(jobState.jobDir, '.' + myPID), 'w') as f:
                     f.write(myPID)
                 while True:
                     try:
@@ -1700,16 +1700,16 @@ class NonCachingFileStore(FileStore):
                             raise
                     else:
                         logger.warning('Detected that job (%s) prematurely terminated.  Fixing the '
-                                       'state of the job on disk.', jobState['jobName'])
+                                       'state of the job on disk.', jobState.jobName)
                         if not batchSystemShutdown:
                             logger.debug("Deleting the stale working directory.")
                             # Delete the old work directory if it still exists.  Do this only during
                             # the life of the program and dont' do it during the batch system
                             # cleanup.  Leave that to the batch system cleanup code.
-                            shutil.rmtree(jobState['jobDir'])
+                            shutil.rmtree(jobState.jobDir)
                         # Run any deferred functions associated with the job
                         logger.debug('Running user-defined deferred functions.')
-                        cls._runDeferredFunctions(jobState['deferredFunctions'])
+                        cls._runDeferredFunctions(jobState.deferredFunctions)
                         break
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


arkal:
@hannes-ucsc I've implemented the test i believe shows that the orphaned file exists If you approve this, I'll start on the actual code to fix it.

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            """
+            jsID = job.fileStore.writeGlobalFile(job.fileStore.getLocalTempFile())
+            with open(nonLocalFile, 'w') as fH:
+                dill.dump({'jsID': jsID, 'pid': os.getpid()}, fH)
+            os.kill(os.getpid(), signal.SIGKILL)
+
+        @staticmethod
+        def _testOrphanedFilesAreDeletedFromJobStore_B(job, nonLocalFile):
+            """
+            Wait for nonLocalFile to get populated. Ensure that the file exists in teh jobstore.
```
hannes-ucsc: the

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            Then spawn the child that will check for non-existence of the jobstore file after the
+            scheduled cleanup.
+
+            :param job: Job
+            :param nonLocalFile: A file outside the scope of the working directory that can persist
+                   across jobs.
+            :returns: None
```
hannes-ucsc: redundant

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            with open(nonLocalFile, 'w') as fH:
+                dill.dump({'jsID': jsID, 'pid': os.getpid()}, fH)
+            os.kill(os.getpid(), signal.SIGKILL)
+
+        @staticmethod
+        def _testOrphanedFilesAreDeletedFromJobStore_B(job, nonLocalFile):
+            """
+            Wait for nonLocalFile to get populated. Ensure that the file exists in teh jobstore.
+            Then spawn the child that will check for non-existence of the jobstore file after the
+            scheduled cleanup.
```
hannes-ucsc:
```
After this job function returns, the sole child of this job will then …
```

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            while attempt <= 10:
+                try:
+                    with open(nonLocalFile, 'r') as fH:
+                        jobAInfo = dill.load(fH)
+                except EOFError:
+                    time.sleep(3)
+                    attempt += 1
+                else:
+                    while job.fileStore.HarbingerFile._pidExists(jobAInfo['pid']):
+                        # This should never infinite loop.
```
hannes-ucsc: You might still want to protect against a hang here just in case something in Toil breaks in such a way that this loops forever. A breakage in the code under test should cause a test to fail, not cause it to hang. I think you can continue continue to maintain the `attempt` counter in this inner loop.

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+                        time.sleep(1)
+                    break
+            else:
+                raise RuntimeError('Timed out waiting for nonLocalFile.')
+            # The file should exist at this point.
+            job.fileStore.readGlobalFile(jobAInfo['jsID'])
```
hannes-ucsc: Isn't this racy? The file could be deleted before you get to this. If you argue that its not racy due to some implementation detail of the deferred mechanism, I would say that the test shouldn't rely on that. 

I would actually be satisfied if the existence of the file is asserted in A and the non-existence asserted in C.

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+
+        @staticmethod
+        def _testOrphanedFilesAreDeletedFromJobStore_C(job, nonLocalFile):
+            """
+            Ensure that the test jsID in nonLocalFile does not exist in the jobstore.
+
+            :param job: Job
+            :param nonLocalFile: A file outside the scope of the working directory that can persist
+                   across jobs.
+            :returns: None
```
hannes-ucsc: redundant

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            with open(nonLocalFile, 'r') as fH:
+                jobAInfo = dill.load(fH)
+            # The file should not exist at this point.
+            try:
+                job.fileStore.readGlobalFile(jobAInfo['jsID'])
+            except NoSuchFileException as e:
+                assert "does not exist" in e.message
+            else:
+                raise RuntimeError('The file still exists in the job store.')
+            return None
```
hannes-ucsc: redundant

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,106 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            root.addChild(A)
+            root.addChild(B)
+            B.addChild(C)
+            try:
+                Job.Runner.startToil(root, self.options)
+            except FailedJobsException as err:
+                self.assertEqual(err.numberOfFailedJobs, 2) # root and A
+            assert not os.path.exists(nonLocalFile)
+
+        @staticmethod
```
hannes-ucsc: We can use static methods as job functions? I seem to remember that we used to have to alias those as global variable at the module level. Or did we automate the aliasing?

arkal: Yeah check out line 1305

src/toil/test/src/jobCacheTest.py
```diff
@@ -1125,6 +1126,154 @@ def testBatchSystemCleanupCanHandleWorkerDeaths(self):
...
+            # There can be no retries
+            self.options.retryCount = 0
+            self.options.logLevel = 'DEBUG'
+            workdir = self._createTempDir(purpose='nonLocalDir')
+            nonLocalFile = os.path.join(workdir, str(uuid4()))
+            # nonLocalFile has to exist and be empty
+            open(nonLocalFile, 'w').close()
+            assert os.path.exists(nonLocalFile)
+            root = Job()
+            A = Job.wrapJobFn(self._testOrphanedFilesAreDeletedFromJobStore_A,
```
hannes-ucsc: Not necessarily for this PR but I thought of a naming convention for jobs: letters for sibliings (left to right in the tree) numbers for descendants. So here it would be A, B1 and B2. If B1 had two children they would be B2A and B2B.

arkal: I can use that from now on

cket:
@arkal what happened to the tests?

cket:
Continued in #1351

PR #1260, opened 2016-11-01
Appliance improvements

### LINKED ISSUE #1157: Tag appliance images with Jenkins build number
… and consider not pip-installing the sdist but the official release from PyPI. @cmarkello noticed that inside the image only `3.5.0a1` is reported by `pip list`. This may be confusing as there is no easy way quickly find the matching release on PyPI.

### LINKED ISSUE #1162: Combine appliance images for leader and worker
The base images for master and slave from mesosphere are only trivially different: only the entry point differs. We can easily eliminate that dichotomy. This would avoid having to alias the `toil-leader` images as `toil`, cutting three repos down to one. We already need one set of repos for the integration tests and the unit tests, respectively, so this would reduce the maintenance overhead considerably. It also makes for a simpler overall system, as we don't need to explain what those three images are for, and that one is just an alias of another one.

@cket: no intelligence required here, your code already parameterizes the `docker run` command in the instance user data, this would just change how you parameterize it. Instead of varying the image name, you vary the run command.

### LINKED ISSUE #1182: Docker now required for tests?
I'm working on a fork that is adding PBS/Torque job manager support, and am attempting to run tests as I had in the past (and as described in the docs).  I noticed the following is now occurring:

```
(venv) -system-specific-4.1$ make test tests=src/toil/test/sort/sortTest.py::SortTest::testSort
Please set docker_registry, e.g. to quay.io/USER.
make: *** [check_docker_registry] Error 1
```

Is there any way to run tests that do not require Docker, such as standard batch system tests like SLURM, etc?  Our local HPC doesn't implement Docker for security reasons (they are looking into alternatives), so further development is kind of a non-starter for me at the moment.

### LINKED ISSUE #1244: Improve tagging of the appliance image
Supply the following tags:
- `latest` (but only if building on master and only if version is a release (no .dev suffix)
- semantic version `3.5.0a1.dev123` if building on jenkins
- git commit sha1 (abbreviated)
- git commit sha1 (full)

### LINKED ISSUE #1245: Make bash the entry point of the appliance
… so users can get an interactive shell environment with Toil installed:

```
docker run toil
toil create-cluster …
```

Obviously, they should also be able to 

```
docker run toil toil create-cluster
```

### LINKED ISSUE #1246: Use reasonable default for TOIL_APPLIANCE_SELF
… such that we don't need to set it before invoking auto-scaling commands, e.g. toil create-cluster.

### LINKED ISSUE #1295: Eliminate need to set docker registry during build
See also #1182

hannes-ucsc:
- fixes #1245 
- fixes #1162 
- fixes #1244 
- fixes #1246 
- fixes #1295 
- fixes #1182
- fixes #1157


hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

PR #1277, opened 2016-11-03
Force removal of files left over by test containers (fixes #1276)

### LINKED ISSUE #1276: CI builds fail if previous build left over files owned by root
The hot deployment tests now use the appliance. The appliance runs as root. The tests do clean up after themselves but if the build times out, that clean-up is never run and subsequent builds will fail early when jenkins.sh attempts to remove the files.

hannes-ucsc:
fixes #1276


hannes-ucsc:
I cancelled the IT build to save money. The changes affect both the IT and the regular build equally so testing just one seemed sufficient.

PR #1279, opened 2016-11-04
Fixes related to encapsulated jobs (fixes #1259, fixes #1278)

### LINKED ISSUE #1259: Worker fails to unpickle job if job arguments refer to user-defined types
A JobFunctionWrappingJob/FunctionWrappingJob instance refers to 

1) the user module
2) the user function 
3) the args and kwargs for the function invocation

While 1) and 2) are symbolic references, 3) is not. The reason they are symbolic is to allow for the instance to be unpickled without having to load the user module. Once the job instance exists, we can use 1) to load the user module and 2) to extract the user function. If, however, the object graph from 3) references any types defined in the the user script, the unpickling will fail.

Update: I can't reproduce this without encapsulated jobs, so this seems to be specific to encapsulation. Moreover, what I describe above is partially wrong: the actual user script has already been loaded when the job instance is unpickled. This is because JobFunctionWrappingJob/FunctionWrappingJob override getUserScript() to return the module that the function is defined in. This module also contains the user-defined types, allowing them to be loaded during unpickling.

### LINKED ISSUE #1278: Encapsulating root jobs always returns None
… independently of whether the encapsulated job returns a value.

hannes-ucsc:
fixes #1259, fixes #1278

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -595,15 +595,14 @@ class Toil(object):
             self._serialiseEnv()
             self._cacheAllJobs()
 
-            # Make a file to store the root job's return value in
-            rootJobReturnValueID = self._jobStore.getEmptyFileStoreID()
-
-            # Add the root job return value as a promise
-            rootJob._rvs[()].append(rootJobReturnValueID)
-
-            # Write the name of the promise file in a shared file
-            with self._jobStore.writeSharedFileStream("rootJobReturnValue") as fH:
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
Test failure tracked in #1280.

hannes-ucsc:
All UT and some of the IT passed (the failures are unrelated) which sufficiently convinced me of the   stability of the changes.

Issue #1289, opened 2016-11-08
Fix job scheduling on restart after partially failed DAGs


arkal:
If a job in a DAG fails and there is another path to its child, the child is run in the restart irrespective of the status of the previously failed parent in the new run.

@arkal will write a test to prove this behaviour and wither @cket or @benedictpaten will probably write the fix.

cket:
Likely related to #1100

PR #1264, opened 2016-11-02
Add tests for stranded successors in DAGs (resolves #1122, resolves #1289, resolves #1294)

cket:
@benedictpaten this adds a couple simple test cases for #1122

Issue #1122, opened 2016-08-26
cwltoil non-clean job failures


chapmanb:
While debugging #1107 and responding to comments from @hannes-ucsc I realized the underlying issues with CWL restarts might be that cwltoil does not exit cleanly. Here is a small example with the test directory, forcing revtool to fail by messing up the `rev` command:

```
cd src/toil/test/cwl
sed -i 's/baseCommand: rev/baseCommand: rev-fail/' revtool.cwl
cwltoil --no-container revsort.cwl revsort-job.json
```

This fails with an assertion error during consistency checks at the end of `innerLoop` (https://github.com/BD2KGenomics/toil/blob/92f64a5d94bd16514979bc5c784f72863f2ddf5d/src/toil/leader.py#L728) which I suspect is not the right outcome. That function is pretty complex and deeply nested so it's not clear to me how to start debugging this. I hope that there's something we can include from the CWL side to create cleaner job exits here.

Can anyone offer any advice/directions?

Here is the full traceback from the run:

```
kunkel 2016-08-26 11:18:31,313 MainThread WARNING toil.leader: The jobWrapper seems to have left a log file, indicating failure: B/U/jobWBMw6I
kunkel 2016-08-26 11:18:31,313 MainThread WARNING toil.leader: Reporting file: B/U/jobWBMw6I
kunkel 2016-08-26 11:18:31,313 MainThread WARNING toil.leader: B/U/jobWBMw6I:   ---TOIL WORKER OUTPUT LOG---
kunkel 2016-08-26 11:18:31,313 MainThread WARNING toil.leader: B/U/jobWBMw6I:   INFO:rdflib:RDFLib Version: 4.2.1
kunkel 2016-08-26 11:18:31,314 MainThread WARNING toil.leader: B/U/jobWBMw6I:   INFO:toil.fileStore:CACHE: Read file with ID 'a/2/tmp2RYy7z.tmp' from the cache.
kunkel 2016-08-26 11:18:31,314 MainThread WARNING toil.leader: B/U/jobWBMw6I:   [job revtool.cwl] /tmp/user/1001/tmpZ7hLZ6$ rev-fail \
kunkel 2016-08-26 11:18:31,314 MainThread WARNING toil.leader: B/U/jobWBMw6I:       /tmp/user/1001/tmppbMf7F/stg5c0b758e-c348-411b-b260-877b9491c6da/whale.txt > /tmp/user/1001/tmpZ7hLZ6/output.txt
[...]

PR #1292, opened 2016-11-08
Add test to show children of failed jobs in a DAG are run on restart

arkal:
connected to #1289

Also, Toil raises an `AssertionError` on workflow failure instead of `FailedJobsException`.
This is also a bug that is caught by this test.

### DIFF
--- a/src/toil/test/src/restartJobTest.py
+++ b/src/toil/test/src/restartJobTest.py
@@ -1,0 +1,175 @@
+# Copyright (C) 2015-2016 Regents of the University of California
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
@benedictpaten, I consider this a show-stopper. It may be related to the issue you are currently working on (#1122).

arkal:
It most certainly is a showstopper with the ProTECT computes given the DAG-gyness of the pipeline.

cket:
I think found the problem, I'll push my solution soon

benedictpaten:
@cket do not push your solution. My solution I just pushed fixes the problem (at least it will when the code is straight). We should merge #1264 #1294 and #1292 and we should write a generative unit test that properly test restart. I thought I had one in jobTest, but clearly I am wrong.

benedictpaten:
Superseded by #1122 fix.

benedictpaten:
Superseded by #1122 fix

PR #1294, opened 2016-11-08
Fix DAG scheduling on restart (resolves #1289)

### LINKED ISSUE #1289: Fix job scheduling on restart after partially failed DAGs

cket:


benedictpaten:
This test should work with the changes made by PR #1264 . @cket let's chat, I think my fix is more general but we should compare notes.

benedictpaten:
This is superseded by PR #1122 fix.

hannes-ucsc:
> This is superseded by PR #1122 fix.

That is not the case. This was superseded by #1264.

PR #1304, opened 2016-11-14
Propagate worker logs to master (resolves #1298, resolves #697)

### LINKED ISSUE #697: Better, more persistent handling of logs
As discussed on #687, there is good reason to treat logfiles as akin to data generated by an analysis pipeline. Unfortunately, right now, the only options are to steal the logfiles from the clutches of the temporary job folders on the toil workers, or to grab truncated files that are stashed away in the jobstore. We should solve this scourge and help those betrayed by the log files that have dashed off into the sunset.

### LINKED ISSUE #1298: Make log truncation configurable
Currently hard-coded to 50k, should be configurable. There is a command line option but it is ineffective.

@cket has a fix for this in one of his open (but stale) PRs. @cket which one? It might make sense to pull that change out.

cket:


cket:
@hannes-ucsc this encompasses the fix for #1298 and a naive implementation for #697 that works but doesn't include the housekeeping measures you proposed in the ticket

src/toil/common.py
```diff
@@ -101,7 +101,8 @@ def __init__(self):
 
         #Misc
         self.disableCaching = False
-        self.maxLogFileSize=50120
+        self.maxLogFileSize = 50000
```
hannes-ucsc: Prefer powers of two. 64KiB

src/toil/common.py
```diff
@@ -446,8 +448,13 @@ def _addOptionFn(*name, **kwargs):
                      'LSF, or Slurm')
     addOptionFn("--maxLogFileSize", dest="maxLogFileSize", default=None,
                       help=("The maximum size of a job log file to keep (in bytes), log files larger "
-                            "than this will be truncated to the last X bytes. Default is 50 "
-                            "kilobytes, default=%s" % config.maxLogFileSize))
+                            "than this will be truncated to the last X bytes. Setting this option to a negative "
+                            "value will prevent any truncation. Default=%s" % config.maxLogFileSize))
```
hannes-ucsc: Apply bytes2human to default value? Should print nicely with a default that is a power of two.

src/toil/leader.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, jobStore):
         self._worker.start()
 
     @staticmethod
+    def writeLogs(jobStoreID, jobLogList):
+        def logFileName(jobStoreID):
```
hannes-ucsc: Inline. No need to extract this.

src/toil/leader.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, jobStore):
...
+            logName = jobStoreID.replace('/', '-') + '.log'
+            return str(logName)
+        fileName = logFileName(jobStoreID)
+        with open(fileName, 'a+') as f:
+            f.writelines("%s\n" % l for l in jobLogList)
+            f.write('\nEnd of the log for this invocation of the job\n')
+
+
+    @staticmethod
```
hannes-ucsc: @classmethod

src/toil/leader.py
```diff
@@ -93,10 +106,14 @@ def logWithFormatting(jobStoreID, jobLogs):
...
                         # we have reached the next job, output the aggregated logs and continue
                         logWithFormatting(currentJobStoreID, jobLogs)
+                        if config.writeLogs:
+                            StatsAndLogging.writeLogs(currentJobStoreID, jobLogs)
                         jobLogs = []
                         currentJobStoreID = jobStoreID
                 # output the last job's logs
                 logWithFormatting(currentJobStoreID, jobLogs)
+                if config.writeLogs:
+                    StatsAndLogging.writeLogs(currentJobStoreID, jobLogs)
```
hannes-ucsc: `cls.`

src/toil/leader.py
```diff
@@ -93,10 +106,14 @@ def logWithFormatting(jobStoreID, jobLogs):
                     else:
                         # we have reached the next job, output the aggregated logs and continue
                         logWithFormatting(currentJobStoreID, jobLogs)
+                        if config.writeLogs:
+                            StatsAndLogging.writeLogs(currentJobStoreID, jobLogs)
```
hannes-ucsc: cls.

src/toil/leader.py
hannes-ucsc: Basically avoid @staticmethod. I can explain in person.

src/toil/leader.py
```diff
@@ -331,6 +348,8 @@ def processRemovedJob(jobStoreID):
                     logFormat = '\n%s    ' % jobStoreID
                     logger.warn('The job seems to have left a log file, indicating failure: %s\n%s',
                                 jobStoreID, logFormat.join(messages))
+                    if self.config.writeLogs:
+                        StatsAndLogging.writeLogs(jobStoreID, messages)
```
hannes-ucsc: cls.

src/toil/leader.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, jobStore):
         self._worker.start()
 
     @staticmethod
+    def writeLogs(jobStoreID, jobLogList):
+        def logFileName(jobStoreID):
+            logName = jobStoreID.replace('/', '-') + '.log'
+            return str(logName)
+        fileName = logFileName(jobStoreID)
+        with open(fileName, 'a+') as f:
+            f.writelines("%s\n" % l for l in jobLogList)
```
hannes-ucsc: This can be made faster by replacing string interpolation with concatenation.

src/toil/leader.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, jobStore):
         self._worker.start()
 
     @staticmethod
+    def writeLogs(jobStoreID, jobLogList):
+        def logFileName(jobStoreID):
+            logName = jobStoreID.replace('/', '-') + '.log'
```
hannes-ucsc: replace '-' with '--' for bijective transformation.

src/toil/leader.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, jobStore):
...
+            return str(logName)
```
hannes-ucsc: `str` makes no sense here

src/toil/leader.py
cket: The method became more complex so I kept it in for now

hannes-ucsc: If it's only called once there is no need to extract a function. It just adds confusion.

src/toil/leader.py
hannes-ucsc: Got it.

cket:
jenkins, test this please

hannes-ucsc:
@cket decided to add symlinks for chained jobs and put job name in file. That would mean that this is dependent on #677 and PR #1197.

Job name might have to be transliterated and truncated to be used in a file name.

cket:
jenkins, test this please

cket:
@hannes-ucsc ready for review

src/toil/leader.py
```diff
@@ -50,15 +52,54 @@ def __init__(self, jobStore):
...
 
-    @staticmethod
-    def statsAndLoggingAggregator(jobStore, stop):
+    @classmethod
+    def writeLogFiles(cls, jobNames, jobLogList, config):
+        def createName(logPath, jobName, logExtension):
+            logName = jobName.replace('-', '--')
+            logName = logName.replace('/', '-')
+            logName = logName.replace(' ', '_')
+            logName = logName.replace("'", '')
```
hannes-ucsc: double quotes, too?

hannes-ucsc: Spaces, single and double quotes are actually allowed in filenames but require escaping when you want to refer to such filenames in a shell. So I understand the motivation for this.

src/toil/leader.py
```diff
@@ -50,15 +52,54 @@ def __init__(self, jobStore):
...
+                counter += 1
+
+        mainFileName = jobNames[0]
+        path = None
+        writeFn = None
+        extension = '.log'
+        if config.writeLogs:
+            path = config.writeLogs
+            writeFn = open
+        if config.writeLogsGzip:
```
hannes-ucsc: These two options are are mutually exclusive, right? If so, I would make second `if` an `elif` and `assert not (config.writeLogs and config.writeLogsGzip)` above.

src/toil/leader.py
```diff
@@ -50,15 +52,54 @@ def __init__(self, jobStore):
...
+            path = config.writeLogsGzip
+            writeFn = gzip.open
+            extension += '.gz'
+        if path is None:
```
hannes-ucsc: This would then become an `else`

hannes-ucsc: And the initializations above the branching could be eliminated.

src/toil/test/src/regularLogTest.py
```diff
@@ -13,12 +13,32 @@
...
+    def _getFiles(self, dir):
+        return [f for f in os.listdir(dir) if os.path.isfile(os.path.join(dir, f))]
+
+    def _assertFileTypeExists(self, dir, extension, encoding=None):
+        # an encoding of None implies no compression
+        onlyFiles = self._getFiles(dir)
+        onlyLogs = [f for f in onlyFiles if f.endswith(extension)]
+        assert onlyLogs
+        for log in onlyLogs:
+            mime = mimetypes.guess_type(log)
```
hannes-ucsc: I think this only looks at the file name, which would be pointless.

gzip files always begin with the magic header `'\x1f\x8b'`. IMO, that would be sufficient to verify that they are gzip.

src/toil/leader.py
```diff
@@ -50,15 +52,54 @@ def __init__(self, jobStore):
...
+            # we don't have anywhere to write the logs, return now
+            return
+        fullName = createName(path, mainFileName, extension)
+        with writeFn(fullName, 'w') as f:
+            f.writelines(l + '\n' for l in jobLogList)
+        for alternateName in jobNames[1:]:
+            # There are chained jobs in this output - indicate this with a symlink
+            # of the job's name to this file
+            name = createName(path, alternateName, extension)
+            os.symlink(fullName, name)
```
hannes-ucsc: The symlink should be relative. Absolute symlinks would break if the user moved the directory around, or tarred it and then untarred to a different location.

PR #1320, opened 2016-11-16
Fix copy-paste typo (resolves #1211)

### LINKED ISSUE #1211: Minor copy-paste error in worker.py
I just stumbled across this:

https://github.com/BD2KGenomics/toil/blob/5ec9a2cd1fb081bd59a93a7aa0023d19cab6750b/src/toil/worker.py#L496

``` python
    #Close redirected stderr and replace with the original standard error.
    os.dup2(origStdOut, 2)
```

That should read `os.dup2(origStdErr, 2)` if you want to actually restore the original stderr to fd 2. I can't think of a way that this would ever cause a problem, though.

natanlao:
Resolves #1211

Restores original stderr to fd 2

### DIFF
--- a/src/toil/worker.py
+++ b/src/toil/worker.py
@@ -493,7 +493,7 @@ def main():
     os.dup2(origStdOut, 1)
     
     #Close redirected stderr and replace with the original standard error.
-    os.dup2(origStdOut, 2)
+    os.dup2(origStdErr, 2)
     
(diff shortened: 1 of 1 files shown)


natanlao:
Jenkins, test this please

natanlao:
Jenkins, test this please

PR #1329, opened 2016-11-21
Resolve #1321, cleans up leader

### LINKED ISSUE #1321: Fixing service deadlocks
We currently have three deadlock scenarios possible  in Toil by the current design (that I'm aware of!), all relate to service scheduling.

 In the first two cases (a) all the jobs in the cluster end up running services and (b) the cluster has reached its maximum possible size (either statically or through autoscaling). The two cases:

(Sub case 1) There remain non-service jobs in the queue 

(Sub case 2) There are no non-service jobs in the queue and there are service jobs in the queue.

Note, it should not be possible to reach a deadlock and have no jobs in the queue at all and have all the nodes running services.

I propose the following solution to the two sub-cases:

For (1) have a maximum services parameter such that at most x (or a proportion of the max) services are running. 

For (2), I do not see a clean solution, and propose recognising the case and throwing an exception advising the user to up the maximum number of concurrent services/increase the cluster size.

The result is that it should not be possible to deadlock in either case - although a crash is undesirable, it is definitely better than a deadlock.

The third case is more subtle - it occurs when the autoscaler estimates the number of nodes to complete a batch based on completed jobs and fails to recognise that there are running service jobs that will never finish without increasing the cluster size. To fix this I propose to ensure that run time of current jobs is factored in to the queue wait time so that the longer that services wait the more machines will be provisioned, helping to ensure that non-service jobs are given space on the cluster more quickly.

benedictpaten:


benedictpaten:
Closing this, as am creating new PR against master.

hannes-ucsc:
You don't *have* to create a new PR, you can just reset the PR branch to some other commit and force push.

PR #1337, opened 2016-11-22
make fileStore a protected field for Job (resolves #1224)

### LINKED ISSUE #1224: All jobs should have access to the underlying fileStore
From #1198, [comment](https://github.com/BD2KGenomics/toil/pull/1198#discussion_r84343466). Right now, only `ServiceJob` and `JobFunctionWrappingJob` set `self.fileStore`. We should make this consistent across jobs.

jessebrennan:
The purpose of this PR is to give all jobs access to the underlying fileStore but as a protected class attribute. Before, serviceJob and jobFunctionWrapingJob needed access to the fileStore and the thought was to give access to all jobs as a class attribute. But this creates two sorts of truths—the jobStore that a user passes in to run, and the one associated with the class. By making it protected, it can be safely ignored and is available if the user knows what they are doing. ServiceJob and jobFunctionWrappingJob have a property to mask the protected class to keep one source of truth since they need explicit access to the class attribute.

Changes:
- The protected attribute `_fileStore` is added to `Job` and is set in Job._runner
- `Job.Defer` now uses correct fileStore attribute.
- a `fileStore` property is used in `JobFunctionWrappingJob` and `ServiceJob` to mask `Job._fileStore` to keep one source of truth

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -293,7 +293,7 @@ class Job(JobLikeObject):
         # entire return value.
         self._rvs = collections.defaultdict(list)
         self._promiseJobStore = None
-        self.fileStore = None
+        self._fileStore = None
 
     def run(self, fileStore):
         """
@@ -671,9 +671,9 @@ class Job(JobLikeObject):
 
         :param dict kwargs: The keyword arguments to the function
         """
-        require( self.fileStore is not None, 'A deferred function may only be registered with a '
+        require(self._fileStore is not None, 'A deferred function may only be registered with a '
                                              'job while that job is running.')
-        self.fileStore._registerDeferredFunction(DeferredFunction.create(function, *args, **kwargs))
+        self._fileStore._registerDeferredFunction(DeferredFunction.create(function, *args, **kwargs))
 
 
     ####################################################
@@ -1280,7 +1280,9 @@ class Job(JobLikeObject):
                filestore
         :return:
         """
-        # Run the job
+        # Make fileStore available as an attribute during run() ...
+        self._fileStore = fileStore
+        # ... but also pass it to run() as an argument for backwards compatibility.
         returnValues = self._run(jobGraph, fileStore)
         # Serialize the new jobs defined by the run method to the jobStore
         self._serialiseExistingJob(jobGraph, jobStore, returnValues)
@@ -1391,9 +1393,13 @@ class JobFunctionWrappingJob(FunctionWrappingJob):
     instance (see :func:`toil.job.Job.run`), it is made a variable of the wrapping job \
     called fileStore.
     """
+
+    @property
+    def fileStore(self):
+        return self._fileStore
+
     def run(self, fileStore):
         userFunction = self._getUserFunction()
-        self.fileStore = fileStore
         rValue = userFunction(*((self,) + tuple(self._args)), **self._kwargs)
         return rValue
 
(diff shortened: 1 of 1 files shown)


src/toil/job.py
```diff
@@ -1381,6 +1378,10 @@ class JobFunctionWrappingJob(FunctionWrappingJob):
     instance (see :func:`toil.job.Job.Run`), it is made a variable of the wrapping job \
     called fileStore.
     """
+    @property
+    def fileStore(self):
+        return self._fileStore
+
     def run(self, fileStore):
         userFunction = self._getUserFunction()
         self.fileStore = fileStore
```
hannes-ucsc: This is obviously redundant. Same in ServiceJob.

src/toil/job.py
```diff
@@ -1271,7 +1268,7 @@ def _runner(self, jobGraph, jobStore, fileStore):
         :return:
         """
         # Run the job
-        returnValues = self._run(jobGraph, fileStore)
+        returnValues = self.run(fileStore)
```
hannes-ucsc: Shouldn't this also set `self._fileStore`?

src/toil/job.py
jessebrennan: I had thought this was the agreed solution. Are you suggesting that I add the public fileStore property to the Job class?

jessebrennan:
@hannes-ucsc please review

hannes-ucsc:
I'd prefer to review after branch is rebased and conflicts have been resolved.

src/toil/job.py
```diff
@@ -289,6 +289,8 @@ def __init__(self, memory=None, cores=None, disk=None, preemptable=None, unitNam
         # entire return value.
         self._rvs = collections.defaultdict(list)
         self._promiseJobStore = None
+        # All jobs have access to protected fileStore field. Especially needed for defer
```
hannes-ucsc: Remove comment. First sentence is obvious. Second sentence is anecdotal.

src/toil/job.py
```diff
@@ -1270,6 +1270,8 @@ def _runner(self, jobGraph, jobStore, fileStore):
                filestore
         :return:
         """
+        # store fileStore for use in JobFunctionWrappingJob, ServiceJob, and defer
```
hannes-ucsc: Change to `# Make fileStore available as an attribute during run() ...`

src/toil/job.py
```diff
@@ -1270,6 +1270,8 @@ def _runner(self, jobGraph, jobStore, fileStore):
...
+        self._fileStore = fileStore
         # Run the job
```
hannes-ucsc: Replace with `# ... but also pass it to run() as an argument for backwards compatibility.`

jessebrennan:
@cket I think this is ready to go. Please review.

benedictpaten:
Can we get this in @hannes-ucsc @cket

cket:
Jenkins, test this please

jessebrennan:
@cket ready for review.

cket:
@jessebrennan thanks!

jessebrennan:
SO this "fix" is completely wrong (as I understand it now). It was based on the assumption that there would be multiple sources of "truth" for the actual filestore. This is *not* the case. I am going to reopen the original issue.

PR #1338, opened 2016-11-23
Resolves #1321 fixing service deadlocks 

### LINKED ISSUE #1321: Fixing service deadlocks

benedictpaten:


benedictpaten:
jenkins, test this please

benedictpaten:
Not sure why the last tests timed out. Any clues @cket @hannes-ucsc ?

benedictpaten:
Seems like this branch is slower. I will do some performance probing.

hannes-ucsc:
Due to the parallel nature of the tests it is hard to determine which test timed out. We should change run_tests.sh to pipe output through `sed -e 's/^/WHATTEST/`:

https://github.com/BD2KGenomics/toil/issues/1342

Just by staring at the output, I can tell that CGCloudRestartTest was running while timeout occurred but that all other tests had passed. @cket remind me to explain my reasoning. It's possible that that tests simply hung.

benedictpaten:
Okay, the tests passed this time, I'm squashing the commits then (assuming it passes), I'd like @cket or @hannes-ucsc  to review and get this in. I want to work on testing of spark services next - @fnothaft what's happened with the docs pr?

src/toil/leader.py
```diff
@@ -182,49 +527,81 @@ def issueJobs(self, jobs):
...
+        Issue a service job, putting it on a queue if the maximum number of service
+        jobs to be scheduled has been reached.
+        """
+        if jobNode.preemptable:
+            self.preemptableServiceJobsToBeIssued.append(jobNode)
+        else:
+            self.serviceJobsToBeIssued.append(jobNode)
+        self.issueQueingServiceJobs()
+    
+    def issueQueingServiceJobs(self):
```
cket: Wouldn't this make more sense in the service manager? It took me a while to find this code

src/toil/leader.py
```diff
@@ -18,162 +18,507 @@
...
-                              args=(jobStore, self._stop))
-        self._worker.start()
+####################################################
+##Following class represents the leader
+####################################################
 
-    @staticmethod
-    def statsAndLoggingAggregator(jobStore, stop):
+# Represents a job and its requirements as issued to the batch system
+IssuedJob = namedtuple("IssuedJob", "jobStoreID memory cores disk preemptable")
```
cket: This isn't used anywhere, I think it was removed in the job name PR. Maybe reintroduced in a merge error?

src/toil/provisioners/clusterScaler.py
```diff
@@ -265,9 +265,7 @@ def addCompletedJob(self, job, wallTime):
         """
         Adds the shape of a completed job to the queue, allowing the scalar to use the last N
         completed jobs in factoring how many nodes are required in the cluster.
-
-        :param toil.job.JobNode job: The memory, core and disk requirements of the completed job
-
+        :param IssuedJob job: The memory, core and disk requirements of the completed job
```
cket: Merge error? The old type hint was correct but the comment was wrong. 'job' is now just a jobNode object not a tuple.

src/toil/provisioners/clusterScaler.py
```diff
@@ -302,36 +298,66 @@ def __init__(self, scaler, preemptable):
...
+                
+                # Average runtime of recently completed jobs
+                historicalAvgRuntime = sum(map(lambda jS : jS.wallTime, recentJobShapes))
+            
+                # Ratio of avg. runtime of currently running and completed jobs
+                runtimeCorrection = float(currentAvgRuntime)/historicalAvgRuntime if currentAvgRuntime > historicalAvgRuntime and numberOfRunningJobs >= estimatedNodes else 1.0
+                
+                # Make correction, if necessary (only do so if cluster is busy and average runtime is higher than historical
+                # average)
+                if runtimeCorrection != 1.0:
```
cket: This would be unnecessary if we limited the ratio of services issued, right? It seems like there are a lot of scenarios where this would be triggered without being in a deadlock.

src/toil/common.py
```diff
@@ -84,6 +84,11 @@ def __init__(self):
         self.betaInertia = 1.2
         self.scaleInterval = 10
         self.preemptableCompensation = 0.0
+        
+        # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
+        self.maxPreemptableServiceJobs = sys.maxint 
+        self.maxServiceJobs = sys.maxint
```
benedictpaten: Correct, I did not enable this by default because I was worried about it making weird side effects. I prefer this conservative approach until we have tested it at scale.

src/toil/provisioners/clusterScaler.py
benedictpaten: Yup, fixing.

src/toil/provisioners/clusterScaler.py
benedictpaten: This will be triggered a bunch, but I still think it's a better way to adjust our prior on how long we expect the cluster to run. Remember when we first start up the cluster it can take ages before it has a decent set of jobs to properly estimate the runtime of queued jobs. This heuristic will tend to increase the cluster size to tackle this case.

benedictpaten:
@cket providing the tests complete, are you okay with doing a merge?

cket:
@benedictpaten yes, you answered my main question about the runtime based scaling. Once we have real autoscaling data we can start adjusting any scaling algorithms, if needed

benedictpaten:
Good, agree about putting in better defaults for max services down the road.

PR #1340, opened 2016-11-24
CWL: migrate to latest cwltool

chapmanb:
Work in progress to update CWL support to use the latest version of
cwltool, as work to fix any ongoing CWL support issues in Toil.

This passes many standard runs including bcbio CWL, but currently fails
5 CWL conformance tests I'm stuck at how to fix; starting this branch/PR
to solicit help from @mr-c @tetron and any other CWL folks.

The PR includes these updates:

- Dump Jobs to pickles with pickle avoiding the HIGHEST_PROTOCOL.
  The cPickle/HIGHEST_PROTOCOL will not serialize CWLWorkflow objects
  which contain sub-workflow steps and avro/schema-salad classes
  (they fail with `Can't pickle <type 'instancemethod'>: attribute
  lookup __builtin__.instancemethod failed find problem attribute`)
- Ensure generalized resolver used for loading tools
- logging debugging with new pickling and Job object representation.
- Update to latest CWL conformance testing

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -65,6 +65,8 @@ support for the stable v1.0 specification, only lacking the following features:
   a good workaround.
 - `File literals <http://www.commonwl.org/v1.0/CommandLineTool.html#File>`_ that
   specify only ``contents`` to a File without an explicit file name.
+- Complex file inputs -- from ExpressionTool or a default value, both of which do not yet
+  get cleanly staged into Toil file management.
 
 To run in local batch mode, provide the CWL file and the input object file::
 
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ def runSetup():
                 'gcs_oauth2_boto_plugin==1.9',
                 botoRequirement],
             'cwl': [
-                'cwltool==1.0.20160714182449']},
+                'cwltool==1.0.20161221171240']},
         package_dir={'': 'src'},
         packages=find_packages(where='src',
                                # Note that we intentionally include the top-level `test` package for
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -26,8 +26,10 @@ import cwltool.main
 import cwltool.workflow
 import cwltool.expression
 import cwltool.builder
+import cwltool.resolver
+import cwltool.stdfsaccess
 from cwltool.pathmapper import adjustFiles
-from cwltool.process import shortname, adjustFilesWithSecondary, fillInDefaults
+from cwltool.process import shortname, adjustFilesWithSecondary, fillInDefaults, compute_checksums
 from cwltool.utils import aslist
 import schema_salad.validate as validate
 import schema_salad.ref_resolver
@@ -121,6 +123,9 @@ def resolve_indirect(d):
         return res
 
 def getFile(fileStore, dir, fileTuple, index=None, export=False, primary=None, rename_collision=False):
+    # File literal outputs with no path, from writeFile
+    if fileTuple is None:
+        raise cwltool.process.UnsupportedRequirement("CWL expression file inputs not yet supported in Toil")
     fileStoreID, fileName = fileTuple
 
     if rename_collision is False:
@@ -158,6 +163,10 @@ def writeFile(writeFunc, index, x):
     # Toil fileStore references are tuples of pickle and internal file
     if isinstance(x, tuple):
         return x
+    # File literal outputs with no path, we don't write these and will fail
+    # with unsupportedRequirement when retrieving later with getFile
+    elif x.startswith("_:"):
+        return None
     else:
         if x not in index:
             if not urlparse.urlparse(x).scheme:
@@ -170,6 +179,13 @@ def writeFile(writeFunc, index, x):
                 cwllogger.error("Got exception '%s' while copying '%s'", e, x)
                 raise
         return index[x]
+
+def computeFileChecksums(fs_access, f):
+    # File literal inputs with no path, no checksum
+    if isinstance(f, dict) and f.get("location", "").startswith("_:"):
+        return f
+    else:
+        return compute_checksums(fs_access, f)
 
 def addFilePartRefs(p):
     """Provides new v1.0 functionality for referencing file parts.
@@ -221,7 +237,7 @@ class CWLJob(Job):
         builder.timeout = 0
         builder.resources = {}
         req = tool.evalResources(builder, {})
-        self.cwltool = tool
+        self.cwltool = remove_pickle_problems(tool)
         # pass the default of None if basecommand is empty
         unitName = self.cwltool.tool.get("baseCommand", None)
         if isinstance(unitName, (list, tuple)):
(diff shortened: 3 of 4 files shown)


chapmanb:
You can run the currently failing tests by first setting up a testing environment. I do this with miniconda:
```
cd toil
wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh
bash Miniconda2-latest-Linux-x86_64.sh -b -p venv
venv/bin/python setup.py install
venv/bin/pip install -e .[cwl]
venv/bin/pip install sphinx==1.4.1 mock==1.0.1 pytest==2.8.3 stubserver==1.0.1
```
Then you run with:
```
cd src/toil/test/cwl
export PATH=/path/to/toil/venv/bin:$PATH
python -m pytest -vv cwlTest.py::CWLTest::test_run_conformance
```
Here is a gist of the current failures: https://gist.github.com/chapmanb/daf599e02e7061cd63a7956b2710f08e

They appear to be primarily due to path/location and other file representation changes. I've been having trouble following all the migrations there so am not sure what to fix there. Here's my take on the five failures:

- 39/87 -- I'm not sure why `path` is not setup correctly here (it has location in the CWL file) or how best to do it.
- 83/87 and 86/87 -- These look like a bug in the tests and DockerRequirement should be under hints instead of requirements. I can submit a PR upstream if this can be confirmed.
- 84/87 -- cwltoil's output file representation is different, I'm not sure how best to get the one that the CWL tests expect.
- 85/87 -- appears to be another consequence of `path` not setup as expected so might be solved by fixes for 39 and 83.

Thanks for any help/suggestions to finalize this and get us to synchronized cwltool/Toil

mr-c:
@chapmanb What version of `cwltest` are you using?

chapmanb:
Michael -- I just tested with the latest versions of cwltest (1.0.20161124105442) and cwltool (1.0.20161123190203) and get the same errors. Is there any version I should be using, or specific versions to avoid?

mr-c:
@chapmanb Just checking -- I can reproduce the same errors 

Note, one can run a particular CWL conformance test via

```
~/src/toil/src/toil/test/cwl/spec$ ./run_test.sh RUNNER=cwltoil DRAFT=v1.0 -n39
```

mr-c:
@chapmanb here's a patch to fix 83 & 86

``` diff
diff --git a/src/toil/cwl/cwltoil.py b/src/toil/cwl/cwltoil.py
index 8a09ca7..2605f61 100755
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -655,7 +655,7 @@ def main(args=None, stdout=sys.stdout):
             outobj = toil.restart()
         else:
             basedir = os.path.dirname(os.path.abspath(options.cwljob or options.cwltool))
-            builder = t._init_job(job, basedir=basedir)
+            builder = t._init_job(job, basedir=basedir, use_container=use_container)
             (wf1, wf2) = makeJob(t, {}, use_container=use_container, preserve_environment=options.preserve_environment, tmpdir=os.path.realpath(outdir))
             cwltool.builder.adjustDirObjs(builder.job, locToPath)
             cwltool.builder.adjustFileObjs(builder.job, locToPath)
```

mr-c:
85 is due to `cwltoil`'s lack of support for file literals

mr-c:
84 is due to reporting the results in a different order -- `cwltest` expects directory results to be sorted in a certain way (which is incorrect behavior)

mr-c:
For 39 it appears that files only specified as default values for inputs don't get staged into the global file store

chapmanb:
Michael -- awesome, thanks for all the help. This gets me much closer as I pushed fixes for 3 of the failures and now only fail on:

- 39 -- The default does appear to get staged into the global file store but the value is not propagated. I've been unsuccessful in understanding how this process is meant to happen for default values (or even regular values).
- 84 -- It sounds like the right thing to do is fix cwltest here, is that right? Do you know what needs to happen or should I dig into that as well.

Thanks again for helping push this further along.

src/toil/batchSystems/singleMachine.py
```diff
@@ -132,13 +132,12 @@ def _getDebugCmd(self, jobCommand):
...
             if job_store_locator.startswith("file:") or os.path.exists(job_store_locator):
                 import cPickle
-                import marshal as pickler
                 job_store_locator = job_store_locator.replace("file:", "")
                 job_cmd_file = os.path.join(job_store_locator, "tmp", job_store_id, "job")
                 with open(job_cmd_file) as in_handle:
-                    job = pickler.load(in_handle)
-                if job.get("command"):
-                    command_parts = job["command"].split()
+                    job = cPickle.load(in_handle)
```
cket: These changes could be unnecessary if #1343 is implemented

mr-c:
@chapmanb For 84, yes we will need to fix cwltest -- the simple fix would be to sort the keys before comparing

cket:
@chapmanb how close is this PR? Is there anything I can do on the Toil side to help out?

chapmanb:
Christopher -- thanks so much. I was hoping that #1368 would get merged, then my plan was to update this PR to include that. From there I can try to attack the last two failing CWL tests. For failing test 84, @mr-c do you have the time to fix that or should I dig into cwltest? For 39, I'll need to corner @tetron or @mr-c or find a way to skip that case for now. It would be great to have this merged and start real testing again with Toil + CWL.

cket:
@chapmanb thanks for the reminder - I've been working on #1368 & it should be merged later today, the tests just seem to be failing due to pypi issues atm. Sorry about blocking you with that, I didn't realize!

chapmanb:
Christopher -- thank you for working on this. Did you see my comment in #1368 regarding baseCommand often being a list instead of a single string? I wanted to be sure that case got handled as well before merging. Thanks again.

src/toil/job.py
```diff
@@ -1057,7 +1058,7 @@ def _serialiseJob(self, jobStore, jobsToJobGraphs, rootJobGraph):
         # for the mechanism which unpickles the job and executes the Job.run
         # method.
         with jobStore.writeFileStream(rootJobGraph.jobStoreID) as (fileHandle, fileStoreID):
-            cPickle.dump(self, fileHandle, cPickle.HIGHEST_PROTOCOL)
+            pickle.dump(self, fileHandle)
```
chapmanb: cPickle doesn't correctly dump the CWL objects, even without HIGHEST_PROTOCOL. Sorry it's frustrating, it complains about `Can't pickle <type 'instancemethod'>`

chapmanb:
Christopher and Hannes;
I've rebased against the latest master, updated to the latest cwltool, and fixed cwltoil to pass all CWL conformance tests. From my side this is ready to merge. We're now explicitly failing and document the complex file inputs we don't support, and otherwise have fixes for everything else broken during the update.

Happy to answer any questions, thanks for reviewing and helping with this and looking forward to finally getting it in.

cket:
@chapmanb the jenkins builds are failing with 
```
The required version of setuptools (>=18.5) is not available,
and can't be installed while this script is running. Please 
install a more recent version first, using
'easy_install -U setuptools'. (Currently using setuptools 3.3 (/usr/lib/python2.7/dist-packages))
```
Can setup tools not be updated by the dependency mechanism?

chapmanb:
Christopher -- ugh, sorry about the install problems. I updated to the very latest cwltool which I believe has a fix for this issue. In the process I was able to identify the underlying object causing pickling issues and now have a cwltoil-only workaround that does now require modifying pickling of jobs. Fingers crossed that the new build works cleanly.

Issue #1344, opened 2016-11-29
Sporadic caching failures


cket:
Seen on master, unfortunately.
http://jenkins.cgcloud.info/job/toil-branches/branch/master/286/console
It seems more common on the AWS job store, I will update this ticket with links for those failures once I've found them.

hannes-ucsc:
Seems to be deterministic.

hannes-ucsc:
http://jenkins.cgcloud.info/job/toil-branches/branch/master/286/testReport/src.toil.test.src.fileStoreTest/CachingFileStoreTestWithFileJobStore/testAsyncWriteWithCaching/history/

arkal:
This test is the problem @hannes-ucsc . WE tried monkey patching but it's still apparently really non-determinisitc. I say for now we just skip this till we can figure out how we can test something that is inherently so non-deterministic.

arkal:
it always passes on my laptop, and before i pushed it to jenkins, it passed 10 times in a row on a jenkins slave i had spun up

cket:
@arkal How do you know this is a problem with the test itself and not the caching code?

arkal:
because i've walked through the logs on the pass and fail cases and from the print statements i know the behaviour is exactly as expected.

arkal:
The non-determinism comes from the fact that it takes longer to start a new job than it does to finish writing the file to the filestore asynchronously.

arkal:
but maybe it has something to do with the code and i'll look into it as well.

cket:
@arkal where/how did you monkey patch the test?

PR #1345, opened 2016-11-29
Skip failing async caching test 

cket:
connected to #1344


cket:
@arkal figured out that the failure from #1344 is caused by the test not caching code itself. The monkey patch in the test was too broad in scope, causing an unintentional usage of the monkey patched method that made it non-deterministic. I'm merging this now just to unblock other work until his permanent fix is ready.

arkal:
The fix is in #1346

PR #1347, opened 2016-11-30
Enable metadata credential caching (resolves #688)

### LINKED ISSUE #688: Cache temporary IAM credentials obtained from instance metadata
The patch needs to be injected in the leader and the worker. The latter would activate it for user scripts. S3AM should get it, too. So naturally, this should go into bd2k-python-lib.

cket:


cket:
jenkins, test this please

cket:
@hannes-ucsc can you review this? I tried to insure the monkey patch was enabled on the leader and worker processes whenever boto was available.

PR #1362, opened 2016-12-02
Fix uninformative assert with require (resolves #1232)

### LINKED ISSUE #1232: toil/resource.py throws uninformative assert
It'd be useful to know what the `dirPath` is, at least:

```
  File "/root/toil-scripts/.env/bin/toil-bwa", line 11, in <module>
    load_entry_point('toil-scripts==2.1.0a1', 'console_scripts', 'toil-bwa')()
  File "build/bdist.linux-x86_64/egg/toil_scripts/bwa_alignment/bwa_alignment.py", line 292, in main
  File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 390, in wrapJobFn
  File "/usr/local/lib/python2.7/dist-packages/toil/job.py", line 1246, in __init__
  File "/usr/local/lib/python2.7/dist-packages/toil/resource.py", line 381, in forModule
    assert os.path.isdir(dirPath)
AssertionError
```

jessebrennan:
resolves #1232

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -396,7 +396,7 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
         log.debug("Module dir is %s", dirPath)
         require(os.path.isdir(dirPath),
                 'Bad directory path %s for module %s. Note that hot-deployment does not support \
-                .egg-link files yet.', dirPath, name)
+                .egg-link files yet, or scripts located in the root directory.', dirPath, name)
         fromVirtualEnv = inVirtualEnv() and dirPath.startswith(sys.prefix)
         return cls(dirPath=dirPath, name=name, fromVirtualEnv=fromVirtualEnv)
 
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -394,7 +394,9 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
                 assert dirPathTail == package
             dirPath = os.path.sep.join(filePath)
         log.debug("Module dir is %s", dirPath)
-        assert os.path.isdir(dirPath)
+        require(os.path.isdir(dirPath),
+                'Bad directory path %s for module %s. Note that hot-deployment does not support \
+                .egg-link files yet.', dirPath, name)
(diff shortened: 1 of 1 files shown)


jessebrennan:
@hannes-ucsc ready for review

src/toil/resource.py
```diff
@@ -378,7 +378,8 @@ def forModule(cls, name):
                 dirPathTail = filePath.pop()
                 assert dirPathTail == package
             dirPath = os.path.sep.join(filePath)
-        assert os.path.isdir(dirPath)
+        require(os.path.isdir(dirPath),
+                'Bad directory path %s for module %s. Maybe it contains egg-links?', dirPath, name)
```
hannes-ucsc: `… Note that hot-deployment does not support .egg-link files yet.`

jessebrennan: right, that's why the error warns them that maybe they are using egg-links in their directory. If this is not clear I can change it.

hannes-ucsc: In 

> Maybe it contains egg-links?

it's unclear what "it" refers to. It can't be the value of `dirPath` because we've just established that that's not a directory, so it can't contain things.

To make it clear, I'm asking you to replace `Maybe it contains egg-links? ` with `Note that hot-deployment does not support .egg-link files yet.`

benedictpaten:
@hannes-ucsc @cket Ditto, let's get this in.

cket:
This might be a good opportunity to fix #1455 as well

jessebrennan:
I'll skip it for now...

src/toil/resource.py
```diff
@@ -394,7 +394,9 @@ def forModule(cls, name):
                 assert dirPathTail == package
             dirPath = os.path.sep.join(filePath)
         log.debug("Module dir is %s", dirPath)
-        assert os.path.isdir(dirPath)
+        require(os.path.isdir(dirPath),
+                'Bad directory path %s for module %s. Note that hot-deployment does not support \
+                .egg-link files yet.', dirPath, name)
```
cket: Can you add in something along the lines of "or scripts located in the root directory" so users who run into #1455 will have some idea whats going on? Keep the ticket open though so we can look at a more permanent solution later.

jessebrennan:
@cket okay I changed it.

Issue #1343, opened 2016-11-28
Replace _getDebugCmd with new job name implementation


cket:
We should consider setting Job's jobName attribute in CWLToil to reflect the CWL step. The unitName attribute could be the sample name.

chapmanb:
Christopher -- happy to migrate to the Toil approach for doing this. Is there documentation about what I should be setting for debugging? Just these two attributes for any classes derived from Job? Or are there other changes.

cket:
Just those two attributes are sufficient. #1197 improved debugging native Toil jobs by making Job.__str__ return a reasonable name for the job, including the name of the method or class that defines them. The Job class automatically populates the 'jobName' attribute with this information. Users can also optionally set the 'unitName' attribute to whatever might help them with debugging, usually the sample's ID. Once these are set, calling str(job) will return the jobName followed by the unitName (if it's set), followed by the job's job store ID.

CWLToil is a special case since the steps aren't native Toil jobs so while the 'jobName' attribute is automatically set to a class name like CWLWorkflow, CWLGather, etc. they aren't particularly useful. The 'jobName' attribute should be set explicitly to a CWL analog (command?), but you can rely on the Job constructor to properly handle the 'unitName' parameter if you pass it when you construct a new CWL job.

cket:
@chapmanb mind if I take this over? I just ran into an unpicking error thrown by _getDebugCmd so I'm probably going to implement this rather than fix the existing method

cket:
jenkins, test this please

chapmanb:
Christopher -- absolutely, thanks for taking this over. Apologies for being so slow and not yet having time to work on this. Much appreciated.

cket:
No worries! I have an first pass at it in #1368

PR #1368, opened 2016-12-06
Use job name improve CWL tool debugging (connected to #1343)

cket:


cket:
jenkins, test this please

cket:
@chapmanb @mr-c would either of you mind taking a look at how I'm attempting to label CWL jobs? Any suggestions on which CWL tool attributes to use for naming would be appreciated!

chapmanb:
Christopher;
Thanks again for taking this on. This looks good for the jobName. I'd suggest using the executed command for the other debugging slot:
```
self.unitName = " ".join(cwltool.tool.get("baseCommand", []))
```
We don't have access to sample information at this point since it we're dealing with the workflow definition (in CWL) not the samples that are passing through, so don't have a way to set that. If Toil somehow knows to set unitName correctly we could instead concatenate the command information to the `id` and leave sample setting in unitName as is.

Thanks again for doing this. Once you get it integrated I'll work on rebasing it against my cwltool branch update.

cket:
jenkins, test this please

cket:
jenkins, test this please

cket:
@chapmanb it now handles both strings & lists for the baseCommand attribute - do you want to take another look before I merge it?

src/toil/cwl/cwltoil.py
```diff
@@ -221,12 +221,22 @@ def __init__(self, tool, cwljob, **kwargs):
         builder.timeout = 0
         builder.resources = {}
         req = tool.evalResources(builder, {})
+        self.cwltool = tool
+        # pass the default of None if basecommand is empty
+        unitName = self.cwltool.tool.get("baseCommand", None)
+        if isinstance(unitName, list):
```
chapmanb: You might want `isinstance(unitName, (list, tuple)):` here, just in case.

cket: Good catch, thanks.

cket:
@chapmanb thanks, I'll make that change & we should have this merged today!

PR #1380, opened 2016-12-13
Improve Python 3 compatibility

zero323:
This PR includes following changes:

- Adds [`six`](https://pypi.python.org/pypi/six) as a dependency.
- Replaces Python incompatible imports with equivalent `six` imports.
- Replaces `print` statements with `print` function calls.
- Replaces old-style ocatal literals (`0...`) with new-style literals (`0o...`).

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import print_function
 import os
 import textwrap
 
@@ -52,7 +53,7 @@ motd = heredoc('''
 # Prepare motd to be echoed in the Dockerfile using a RUN statement that uses bash's print
 motd = ''.join(l + '\\n\\\n' for l in motd.splitlines())
 
-print heredoc('''
+print(heredoc('''
     FROM ubuntu:14.04
 
     RUN echo "deb http://repos.mesosphere.io/ubuntu/ trusty main" \
@@ -101,4 +102,4 @@ print heredoc('''
 
     RUN echo '[ ! -z "$TERM" -a -r /etc/motd ] && cat /etc/motd' >> /etc/bash.bashrc \
         && printf '{motd}' > /etc/motd
-''')
+'''))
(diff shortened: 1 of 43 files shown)


cket:
Jenkins, test this please

benedictpaten:
Cool. Can someone please summarize for lagging edge folks like me how close this would get us to proper Python3.0 support? Importantly, do go Python3.0 would we need to break 2.7x support?

cket:
@zero323 thanks for taking this on! It looks like there just a couple import problems on Jenkins:
`StringIO has no attribute 'StringIO'` in CWLTest.test_run_revsort and  `TypeError: 'Module_six_moves_urllib_parse' object is not callable` in the CGCloud provisioner tests. Can you see the Jenkins build outputs, by the way?

zero323:
@cket My pleasure.

>  It looks like there just a couple import problems on Jenkins

I'll take a look at this when I have a spare moment. I have a family emergency so I typed most of that by heart in the waiting rooms so I might have mess something up.

> Can you see the Jenkins build outputs, by the way?

I don't think i have access.  I get:

> Access Denied
> 
> Please login to access job toil-pull-requests

no matter if I am marked as logged or not.

zero323:
@cket I pushed some fixes but it looks like it requires authorization to execute tests.

@benedictpaten In general it is possible to achieve Python 3 compatibility without loosing Python 2 support. It requires some trade-offs and obviously you cannot use new language features. 

Regarding "how close" I have to admit I have no idea. At first glance there shouldn't be many blockers but as for now I am evaluating `toil` so I wouldn't depend on my opinion to much. The biggest advantage here is that, once this patch is applied, you can start testing against Python 3.x and see if things still work.

cket:
Jenkins, ok to test

cket:
@zero323 Jenkins should run automatically on this PR now@

cket:
Jenkins, test this please

cket:
@zero323 I think those test failures were sporadic, hopefully they all pass this time around

cket:
@zero323 looks like the tests are all passing - thanks again! I'll take a closer look at this ASAP but I may not be able to get to it before the holidays

benedictpaten:
Let's get this in.

zero323:
@cket Thank you. I'll try to identify further blockers when I have some spare time.

cket:
@zero323 awesome, I'm looking forward to getting full support! Thanks again.

PR #1386, opened 2016-12-14
Install `screen` in appliance (fixes #1384)

### LINKED ISSUE #1384: Install screen in appliance

natanlao:


cket:
There are weird problems trying to use screen inside Docker https://github.com/docker/docker/issues/728. The workaround is to make sure the TERM envar is set then start `script` before running screen. To make the workaround simpler, @natanlao can you set TERM=linux in the Dockerfile for the appliance? There is an ENV docker statement that will set it for the life of the container

cket:
I'll merge after the master build finishes

Issue #1389, opened 2016-12-15
Testing docker_call


benedictpaten:
This issue depends on successful completion of: #1387 

Tests for docker_call should accomplish:

Testing failure of containers
Detecting existing "ghost" containers

DailyDreaming:
This seems to have been completed.

PR #1403, opened 2016-12-19
Add docker_call to core Toil (resolves #1387)

### LINKED ISSUE #1387: Move Docker call functions in Toil-lib into Toil
See https://github.com/BD2KGenomics/toil-lib/blob/master/src/toil_lib/programs.py

Proposal is that docker_call function (and necessary functions) moves to src/toil/lib/docker.py

jvivian:
PEP for imports
Fix adding custom docker_parameters
Remove all arguments for docker command line options from function args
Add parsing for container name in base_docker_call
Remove mock support (as decreed by the council of dev curmudgeons)
Add docstring example
No tests added

Screwed up the branch name, sorry. Make sure to delete branch after merging.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -596,6 +596,85 @@ Example::
 
             toil.exportFile(outputFileID, 'file:///some/other/local/path')
 
+
+Using Docker containers in Toil
+-------------------------------
+
+Docker containers are commonly used with Toil. The combination of Toil and Docker
+allows for pipelines to be fully portable between any platform that has both Toil
+and Docker installed. Docker eliminates the need for the user to do any other tool
+installation or environment setup.
+
+In order to use Docker containers with Toil, Docker must be installed on all
+workers of the cluster. Instructions for installing Docker can be found on the
+`Docker`_ website.
+
+.. _Docker: https://docs.docker.com/engine/getstarted/step_one/
+
+When using CGCloud or Toil-based autoscaling, Docker will be automatically set up
+on the cluster's worker nodes, so no additional installation steps are necessary.
+Further information on using Toil-based autoscaling can be found in the `Toil
+autoscaling documentation <Autoscaling>`.
+
+In order to use docker containers in a Toil workflow, the container can be built
+locally or downloaded in real time from an online docker repository like Quay_. If
+the container is not in a repository, the container's layers must be accessible on
+each node of the cluster.
+
+.. _Quay: quay.io
+
+When invoking docker containers from within a Toil workflow, it is strongly
+recommended that you use :func:`dockerCall`, a toil job function provided in
+``toil.lib.docker``. ``dockerCall`` provides a layer of abstraction over using the
+``subprocess`` module to call Docker directly, and provides container cleanup on
+job failure. When docker containers are run without this feature, failed jobs can
+result in resource leaks.
+
+In order to use ``dockerCall``, your installation of Docker must be set up to run
+without ``sudo``. Instructions for setting this up can be found here_.
+
+.. _here: https://docs.docker.com/engine/installation/linux/ubuntulinux/#/create-a-docker-group
+
+An example of a basic ``dockerCall`` is below:
+
+    dockerCall(job=job,
+                tool='quay.io/ucsc_cgl/bwa',
+                work_dir=job.fileStore.getLocalTempDir(),
+                parameters=['index', '/data/reference.fa'])
+
+``dockerCall`` can also be added to workflows like any other job function:
+
+     from toil.job import Job
+ 
+     align = Job.wrapJobFn(dockerCall,
+                           tool='quay.io/ucsc_cgl/bwa',
+                           work_dir=job.fileStore.getLocalTempDir(),
+                           parameters=['index', '/data/reference.fa']))
+
+     if __name__=="__main__":
+         options = Job.Runner.getDefaultOptions("./toilWorkflowRun")
+         options.logLevel = "INFO"
+         Job.Runner.startToil(align, options)
+
+`cgl-docker-lib`_ contains ``dockerCall``-compatible Dockerized tools that are
+commonly used in bioinformatics analysis. 
+
+.. _cgl-docker-lib: https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/README.md
+
+The documentation provides guidelines for developing your own Docker containers
+that can be used with Toil and ``dockerCall``. In order for a container to be
+compatible with ``dockerCall``, it must have an ``ENTRYPOINT`` set to a wrapper
+script, as described in cgl-docker-lib containerization standards. Alternately,
+the entrypoint to the container can be set using the docker option
+``--entrypoint``. The container should be runnable directly with Docker as:
+
+    $ docker run <docker parameters> <tool name> <tool parameters>
+
+For example:
+
+    $ docker run -d quay.io/ucsc-cgl/bwa -s -o /data/aligned /data/ref.fa'
+
+
 .. _service-dev-ref:
 
 Services
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -14,7 +14,6 @@
 
 from __future__ import absolute_import, print_function
 
-import cPickle
 import collections
 import importlib
 import inspect
@@ -29,6 +28,10 @@ from abc import ABCMeta, abstractmethod
 from argparse import ArgumentParser
 from contextlib import contextmanager
 from io import BytesIO
+
+# Python 3 compatibility imports
+from six.moves import cPickle
+from six import iteritems, string_types
 
 from bd2k.util.exceptions import require
 from bd2k.util.expando import Expando
@@ -894,7 +897,7 @@ class Job(JobLikeObject):
         """
         Sets the values for promises using the return values from this job's run() function.
         """
-        for path, promiseFileStoreIDs in self._rvs.iteritems():
+        for path, promiseFileStoreIDs in iteritems(self._rvs):
             if not path:
                 # Note that its possible for returnValues to be a promise, not an actual return
                 # value. This is the case if the job returns a promise from another job. In
@@ -1327,7 +1330,7 @@ class FunctionWrappingJob(Job):
                     # ... and finally fall back to a default value.
                     value = default
             # Optionally, convert strings with metric or binary prefixes.
-            if dehumanize and isinstance(value, basestring):
+            if dehumanize and isinstance(value, string_types):
                 value = human2bytes(value)
             return value
 
--- a/src/toil/lib/docker.py
+++ b/src/toil/lib/docker.py
@@ -1,0 +1,248 @@
+"""
+    Module for calling Docker. Assumes `docker` is on the PATH.
+
+    Contains two user-facing functions: dockerCall and dockerCheckOutput
+
+    Uses Toil's defer functionality to ensure containers are shutdown even in case of job or pipeline failure
+
+    Example of using dockerCall in a Toil pipeline to index a FASTA file with SAMtools:
+        def toil_job(job):
+            work_dir = job.fileStore.getLocalTempDir()
+            path = job.fileStore.readGlobalFile(ref_id, os.path.join(work_dir, 'ref.fasta')
+            parameters = ['faidx', path]
+            dockerCall(job, tool='quay.io/ucgc_cgl/samtools:latest', work_dir=work_dir, parameters=parameters)
+"""
+import base64
+import logging
+import subprocess
+
+import os
+from bd2k.util.exceptions import require
+
+_logger = logging.getLogger(__name__)
+
+
+def dockerCall(job,
+               tool,
+               parameters=None,
+               workDir=None,
+               dockerParameters=None,
+               outfile=None,
+               defer=None):
+    """
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
+    This function blocks until the subprocess call to Docker returns
+
+    :param toil.Job.job job: The Job instance for the calling function.
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools:latest).
+    :param list[str] parameters: Command line arguments to be passed to the tool.
+    :param str workDir: Directory to mount into the container via `-v`. Destination convention is /data
+    :param list[str] dockerParameters: Parameters to pass to Docker. Default parameters are `--rm`,
+            `--log-driver none`, and the mountpoint `-v work_dir:/data` where /data is the destination convention.
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param file outfile: Pipe output of Docker call to file handle
+    :param int defer: What action should be taken on the container upon job completion?
+           FORGO (0) will leave the container untouched.
+           STOP (1) will attempt to stop the container with `docker stop` (useful for debugging).
+           RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`. This is the default behavior if defer is set to None.
+    """
+    _docker(job, tool=tool, parameters=parameters, workDir=workDir, dockerParameters=dockerParameters,
+            outfile=outfile, checkOutput=False, defer=defer)
+
+
+def dockerCheckOutput(job,
+                      tool,
+                      parameters=None,
+                      workDir=None,
+                      dockerParameters=None,
+                      defer=None):
+    """
+    Returns the stdout from the Docker invocation (via subprocess.check_output)
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
+    This function blocks until the subprocess call to Docker returns
+
+    :param toil.Job.job job: The Job instance for the calling function.
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools:latest).
+    :param list[str] parameters: Command line arguments to be passed to the tool.
+    :param str workDir: Directory to mount into the container via `-v`. Destination convention is /data
+    :param list[str] dockerParameters: Parameters to pass to Docker. Default parameters are `--rm`,
+            `--log-driver none`, and the mountpoint `-v work_dir:/data` where /data is the destination convention.
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param int defer: What action should be taken on the container upon job completion?
+           FORGO (0) will leave the container untouched.
+           STOP (1) will attempt to stop the container with `docker stop` (useful for debugging).
+           RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`. This is the default behavior if defer is set to None.
+    :returns: Stdout from the docker call
+    :rtype: str
+    """
+    return _docker(job, tool=tool, parameters=parameters, workDir=workDir,
+                   dockerParameters=dockerParameters, checkOutput=True, defer=defer)
+
+
+def _docker(job,
+            tool,
+            parameters=None,
+            workDir=None,
+            dockerParameters=None,
+            outfile=None,
+            checkOutput=False,
+            defer=None):
+    """
+    :param toil.Job.job job: The Job instance for the calling function.
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools).
+    :param list[str] parameters: Command line arguments to be passed to the tool.
+    :param str workDir: Directory to mount into the container via `-v`. Destination convention is /data
+    :param list[str] dockerParameters: Parameters to pass to Docker. Default parameters are `--rm`,
+            `--log-driver none`, and the mountpoint `-v work_dir:/data` where /data is the destination convention.
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param file outfile: Pipe output of Docker call to file handle
+    :param bool checkOutput: When True, this function returns docker's output.
+    :param int defer: What action should be taken on the container upon job completion?
+           FORGO (0) will leave the container untouched.
+           STOP (1) will attempt to stop the container with `docker stop` (useful for debugging).
+           RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`. This is the default behavior if defer is set to None.
+    """
+    if parameters is None:
+        parameters = []
+    if workDir is None:
+        workDir = os.getcwd()
+
+    # Setup the outgoing subprocess call for docker
+    baseDockerCall = ['docker', 'run']
+    if dockerParameters:
+        baseDockerCall += dockerParameters
+    else:
+        baseDockerCall += ['--rm', '--log-driver', 'none', '-v',
+                           os.path.abspath(workDir) + ':/data']
+
+    # Ensure the user has passed a valid value for defer
+    require(defer in (None, FORGO, STOP, RM),
+            'Please provide a valid value for defer.')
+
+    # Get container name which is needed for _dockerKill
+    try:
+        if any('--name' in x for x in baseDockerCall):
+            if any('--name=' in x for x in baseDockerCall):
+                containerName = [x.split('=')[1] for x in baseDockerCall if '--name' in x][0]
+            else:
+                containerName = baseDockerCall[baseDockerCall.index('--name') + 1]
+        else:
+            containerName = _getContainerName(job)
+    except ValueError:
+        containerName = _getContainerName(job)
+        baseDockerCall.extend(['--name', containerName])
+    except IndexError:
+        raise RuntimeError("Couldn't parse Docker's `--name=` option, check parameters: " + str(dockerParameters))
+
+    # Defer the container on-exit action
+    if '--rm' in baseDockerCall and defer is None:
+        defer = RM
+    if '--rm' in baseDockerCall and defer is not RM:
+        _logger.warn('--rm being passed to docker call but defer not set to dockerCall.RM, defer set to: ' + str(defer))
+    job.defer(_dockerKill, containerName, action=defer)
+    # Defer the permission fixing function which will run after this job concludes.
+    # We call this explicitly later on in this function, but we defer it as well to handle unexpected job failure.
+    job.defer(_fixPermissions, tool, workDir)
+
+    # Make subprocess call
+    call = baseDockerCall + [tool] + parameters
+    _logger.info("Calling docker with " + repr(call))
+
+    if outfile:
+        subprocess.check_call(call, stdout=outfile)
+    else:
+        if checkOutput:
+            return subprocess.check_output(call)
+        else:
+            subprocess.check_call(call)
+
+
+FORGO = 0
+STOP = 1
+RM = 2
+
+
+def _dockerKill(containerName, action):
+    """
+    Kills the specified container.
+    :param str containerName: The name of the container created by docker_call
+    :param int action: What action should be taken on the container?  See `defer=` in
+           :func:`docker_call`
+    """
+    running = _containerIsRunning(containerName)
+    if running is None:
+        # This means that the container doesn't exist.  We will see this if the container was run
+        # with --rm and has already exited before this call.
+        _logger.info('The container with name "%s" appears to have already been removed.  Nothing to '
+                  'do.', containerName)
+    else:
+        if action in (None, FORGO):
+            _logger.info('The container with name %s continues to exist as we were asked to forgo a '
+                      'post-job action on it.', containerName)
+        else:
+            _logger.info('The container with name %s exists. Running user-specified defer functions.',
+                         containerName)
+            if running and action >= STOP:
+                _logger.info('Stopping container "%s".', containerName)
+                subprocess.check_call(['docker', 'stop', containerName])
+            else:
+                _logger.info('The container "%s" was not found to be running.', containerName)
+            if action >= RM:
+                # If the container was run with --rm, then stop will most likely remove the
+                # container.  We first check if it is running then remove it.
+                running = _containerIsRunning(containerName)
+                if running is not None:
+                    _logger.info('Removing container "%s".', containerName)
+                    subprocess.check_call(['docker', 'rm', '-f', containerName])
+                else:
+                    _logger.info('The container "%s" was not found on the system.  Nothing to remove.',
+                                 containerName)
+
+
+def _fixPermissions(tool, workDir):
+    """
+    Fix permission of a mounted Docker directory by reusing the tool to change ownership.
+    Docker natively runs as a root inside the container, and files written to the
+    mounted directory are implicitly owned by root.
+
+    :param list baseDockerCall: Docker run parameters
+    :param str tool: Name of tool
+    :param str workDir: Path of work directory to recursively chown
+    """
+    baseDockerCall = ['docker', 'run', '--log-driver=none',
+                      '-v', os.path.abspath(workDir) + ':/data', '--rm', '--entrypoint=chown']
+    stat = os.stat(workDir)
+    command = baseDockerCall + [tool] + ['-R', '{}:{}'.format(stat.st_uid, stat.st_gid), '/data']
+    subprocess.check_call(command)
+
+
+def _getContainerName(job):
+    return '--'.join([str(job),
+                      job.fileStore.jobID,
+                      base64.b64encode(os.urandom(9), '-_')]).replace("'", '').replace('_', '')
+
+
+def _containerIsRunning(container_name):
+    """
+    Checks whether the container is running or not.
+    :param container_name: Name of the container being checked.
+    :returns: True if running, False if not running, None if the container doesn't exist.
+    :rtype: bool
+    """
+    try:
+        output = subprocess.check_output(['docker', 'inspect', '--format', '{{.State.Running}}',
+                                          container_name]).strip()
+    except subprocess.CalledProcessError:
+        # This will be raised if the container didn't exist.
+        _logger.debug("'docker inspect' failed. Assuming container %s doesn't exist.", container_name,
+                   exc_info=True)
+        return None
+    if output == 'true':
+        return True
+    elif output == 'false':
+        return False
+    else:
+        raise RuntimeError("Got unexpected value for State.Running (%s)" % output)
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ The 'pypi' target publishes the current commit of Toil to PyPI after enforcing t
 copy and the index are clean.
 
 The 'docker' target builds the Docker images that make up the Toil appliance. You may set the
-TOIL_DOCKER_REGISTRY variable to override the default registry that the 'docker_push' target pushes
+TOIL_DOCKER_REGISTRY variable to override the default registry that the 'push_docker' target pushes
 the appliance images to, for example:
 
 	TOIL_DOCKER_REGISTRY=quay.io/USER make docker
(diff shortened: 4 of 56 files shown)


benedictpaten:
@cket will you review? I'm also looking.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
             elif isinstance(kwargs.get(r), PromisedRequirement):
                 foundPromisedRequirement = True
         return foundPromisedRequirement
+
+
+def docker_call(job,
+                tool,
+                parameters=None,
+                work_dir='.',
```
fnothaft: Small nit: if we're moving this code, I feel like it'd be nice for the default for this to be `None` instead of `'.'`, and if the user passes in `None`, then to call `os.getcwd()` to get the working directory, etc.

fnothaft: Not a huge deal, just feel like it's cleaner this way.

cket:
@jvivian looks good! Is there a ticket for removing the function from Toil-lib as well?

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+                docker_parameters=None,
+                outfile=None,
+                check_output=False,
```
fnothaft: If we're refactoring this code, I propose not having `check_output` as a flag, but instead have two functions. Ideally, the `check_output` version of `docker_call` would match up with `check_output`'s return signature.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+                defer=None):
+    """
+    Calls Docker for a particular tool with the specified parameters
+
+    Example of using docker_call in a Toil pipeline:
```
fnothaft: Nit: suggest `Example of using docker_call in a Toil pipeline to index a FASTA file with SAMtools:` instead.

fnothaft: Since this example references cgl-docker-lib indirectly via quay, might make sense (or not) to link to said repo.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    if parameters is None:
+        parameters = []
+
+    # Setup the outgoing subprocess call for docker
+    base_docker_call = ['docker', 'run']
+    if docker_parameters:
+        base_docker_call += docker_parameters
+    else:
+        base_docker_call += ['--rm', '--log-driver', 'none', '-v',
```
fnothaft: We should document that these parameters are passed to docker in the docstring for this function.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+    """
+    docker_call.FORGO = 0
+    docker_call.STOP = 1
+    docker_call.RM = 2
+
+    if parameters is None:
+        parameters = []
+
+    # Setup the outgoing subprocess call for docker
+    base_docker_call = ['docker', 'run']
```
fnothaft: Would it be useful to allow the user to override the docker command name?

fnothaft: If no, we should document in the docstring that it is assumed that `docker` is on the path.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+    try:
+        if any('--name=' in x for x in base_docker_call):
+            container_name = [x.split('=')[1] for x in base_docker_call if '--name=' in x][0]
+        else:
+            container_name = base_docker_call[base_docker_call.index('--name') + 1]
+    except ValueError:
+        container_name = _get_container_name(job)
+        base_docker_call.extend(['--name', container_name])
+    except IndexError:
+        raise RuntimeError("Couldn't parse Docker's `--name=` option, check parameters: " + str(base_docker_call))
```
fnothaft: s/parameters/docker_parameters/g

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    # Defer the container on-exit action
+    if '--rm' in base_docker_call and defer is None:
```
fnothaft: What's the behavior of the container/defer function if `--rm` is passed and the defer function is not RM?

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+        defer = docker_call.RM
+    job.defer(_docker_kill, container_name, action=defer)
+    # Defer the permission fixing function.  We call this explicitly later on in this function, but
+    # we defer it as well to handle unexpected job failure.
+    job.defer(_fix_permissions, base_docker_call, tool, work_dir)
```
fnothaft: I feel like this behavior should be documented.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    # Make subprocess call
+    call = base_docker_call + [tool] + parameters
+    job.fileStore.logToMaster("Calling docker with {}.".format(' '.join(call)))
```
fnothaft: Channeling my inner Hannes: prefer `repr(call)` to `' '.join(call)`.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    if outfile:
```
fnothaft: Should `require(outfile is None or not check_output)`, no?

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+        subprocess.check_call(call, stdout=outfile)
+    else:
+        if check_output:
+            return subprocess.check_output(call)
+        else:
+            subprocess.check_call(call)
+
+    # Fix root ownership of output files
+    _fix_permissions(base_docker_call, tool, work_dir)
```
fnothaft: Isn't this deferred?

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+def _docker_kill(container_name, action):
+    """
+    Kills the specified container.
+
+    :param str container_name: The name of the container created by docker_call
+    :param int action: What action should be taken on the container?  See `defer=` in
+           :func:`docker_call`
+    """
+    docker_call.FORGO = 0
```
fnothaft: Do these need to be defined in two places? I'd prefer not to.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+    docker_call.STOP = 1
+    docker_call.RM = 2
+
+    running = _container_is_running(container_name)
+    if running is None:
+        # This means that the container doesn't exist.  We will see this if the container was run
+        # with --rm and has already exited before this call.
+        logger.info('The container with name "%s" appears to have already been removed.  Nothing to '
```
fnothaft: Unrelated to this PR, I suppose, but I feel like we should rename `logger` to `_logger`.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+                    'do.', container_name)
+    else:
+        if action in (None, docker_call.FORGO):
+            logger.info('The container with name %s continues to exist as we were asked to forgo a '
+                        'post-job action on it.', container_name)
+            return
+        else:
+            logger.info('The container with name %s exists. Running user-specified defer functions.',
+                        container_name)
+            if running and action >= docker_call.STOP:
```
fnothaft: If we are going to use `>=` checks on the value of action, the importance of order should be documented wherever the `docker_call` constants are defined.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+    if running is None:
+        # This means that the container doesn't exist.  We will see this if the container was run
+        # with --rm and has already exited before this call.
+        logger.info('The container with name "%s" appears to have already been removed.  Nothing to '
+                    'do.', container_name)
+    else:
+        if action in (None, docker_call.FORGO):
+            logger.info('The container with name %s continues to exist as we were asked to forgo a '
+                        'post-job action on it.', container_name)
+            return
```
fnothaft: This `return` is unnecessary.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+                # container.  We first check if it is running then remove it.
+                running = _container_is_running(container_name)
+                if running is not None:
+                    logger.info('Removing container "%s".', container_name)
+                    try:
+                        subprocess.check_call(['docker', 'rm', '-f', container_name])
+                    except subprocess.CalledProcessError:
+                        logger.exception("'docker rm' failed.")
+                else:
+                    logger.info('The container "%s" was not found on the system.  Nothing to remove.',
```
fnothaft: I feel like this should be a `warn` not an `info`.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+            if action >= docker_call.RM:
+                # If the container was run with --rm, then stop will most likely remove the
+                # container.  We first check if it is running then remove it.
+                running = _container_is_running(container_name)
+                if running is not None:
+                    logger.info('Removing container "%s".', container_name)
+                    try:
+                        subprocess.check_call(['docker', 'rm', '-f', container_name])
+                    except subprocess.CalledProcessError:
+                        logger.exception("'docker rm' failed.")
```
fnothaft: I would give the `container_name` in this logging statement as well.

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    :param toil.Job.job job: The Job instance for the calling function.
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools).
+    :param list[str] parameters: Command line arguments to be passed to the tool.
+    :param str work_dir: Directory to mount into the container via `-v`. Destination convention is /data
+    :param list[str,] docker_parameters: Parameters to pass to Docker. Default parameters are `--rm`,
+            `--log-driver none`, and the mountpoint `-v work_dir:/data` where /data is the destination convention.
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param file outfile: Pipe output of Docker call to file handle
+    :param bool check_output: When True, this function returns docker's output.
```
fnothaft: If you have a tool that spams the output, you likely don't want to get the output back.

src/toil/job.py
```diff
@@ -14,22 +14,21 @@
```
fnothaft: +1

fnothaft:
Also, what about unit tests?

src/toil/job.py
jvivian: *shudder* I feel so dirty

src/toil/job.py
```diff
@@ -1722,3 +1721,185 @@ def convertPromises(kwargs):
...
+
+    :param container_name: Name of the container being checked.
+    :returns: True if running, False if not running, None if the container doesn't exist.
+    :rtype: bool
+    """
+    try:
+        output = subprocess.check_output(['docker', 'inspect', '--format', '{{.State.Running}}',
+                                          container_name]).strip()
+    except subprocess.CalledProcessError:
+        # This will be raised if the container didn't exist.
```
fnothaft: Can you document that inline?

src/toil/job.py
jvivian: +1 I'm perfectly fine with this, it's only set to '.' because it was requested.

src/toil/job.py
jvivian: It does under `dockerParameters` - should it be moved to the main docstring?

src/toil/job.py
jvivian: I see a couple places this would cause a `CalledProcessError` in `_dockerKill`, but whether it matters depends on if deferred functions failing affect the pipeline at all. If they don't, then there's no side-effects. 

Since it's likely unintended behavior on the user's behalf, it probably makes sense to either warn them if STOP / FORGO and `--rm` are both being used, raise an error, or remove `--rm` if either STOP / FORGO are passed.

src/toil/job.py
jvivian: Ya, when the deferred code was added a comment was left ,"We call this explicitly later on in this function, but we defer it as well to handle unexpected job failure."  so I left thinking it was intentional, but if the deferred function is always called then it's unnecessary to also include it at the end of the function.

src/toil/job.py
jvivian: I'm changing this to explicitly associate the action with one of the signal options.

jvivian:
@fnothaft - Unittests #1389. @cket said after review for this, tests will be added to this PR before merging.

fnothaft:
> @fnothaft - Unittests #1389. @cket said after review for this, tests will be added to this PR before merging.

What's the rationale for splitting the code and tests up? If we're going to add them later, we might as well add them now...

jvivian:
We won't be splitting the code / tests into separate PRs, it's just that Benedict assigned the ticket to himself so I'm leaving it for him to finish.  If he wants to assign it to me I'll go ahead and add tests.

jvivian:
Jenkins, test this please

jvivian:
@cket — Ready for review

src/toil/lib/docker.py
```diff
@@ -0,0 +1,238 @@
...
+           using `docker rm -f`.
+           The default value is None and that shadows docker_call.FORGO, unless --rm is being passed in.
+    :returns: Stdout from the docker call
+    :rtype: str
+    """
+    return Docker(job, tool=tool, parameters=parameters, workDir=workDir,
+                  dockerParameters=dockerParameters, checkOutput=True, defer=defer)
+
+
+def Docker(job,
```
fnothaft: Why's this uppercase? Only class names should be uppercase.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,238 @@
...
+import subprocess
+
+from bd2k.util.exceptions import require
+
+_logger = logging.getLogger(__name__)
+
+
+def dockerCall(job, tool, parameters=None, workDir=None, dockerParameters=None, outfile=None, defer=None):
+    """
+    Wrapper call to Docker. See docstring for Docker()
```
fnothaft: This method throws a `CalledProcessError` if the Docker invocation has a non-zero exit code, right? Let's document that.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,238 @@
...
+    :param file outfile: Pipe output of Docker call to file handle
+    :param int defer: What action should be taken on the container upon job completion?
+           docker_call.FORGO (0) will leave the container untouched.
+           docker_call.STOP (1) will attempt to stop the container with `docker stop` (useful for
+           debugging).
+           docker_call.RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`.
+           The default value is None and that shadows docker_call.FORGO, unless --rm is being passed in.
+    """
+    Docker(job, tool=tool, parameters=parameters, workDir=workDir, dockerParameters=dockerParameters,
```
fnothaft: Unimportant nit: if we're splitting lines, my preference is to have one line per method argument. Not sure if anyone else cares.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,238 @@
...
+           outfile=outfile, checkOutput=False, defer=defer)
+
+
+def dockerCheckOutput(job, tool, parameters=None, workDir=None, dockerParameters=None, defer=None):
+    """
+    Wrapper call to Docker. See docstring for Docker()
```
fnothaft: As above, we should document the behavior when the Docker call terminates with an error.

src/toil/lib/docker.py
fnothaft: Also, I feel like this method should be private. What's the rationale for having it be public?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,238 @@
...
+    if outfile:
+        subprocess.check_call(call, stdout=outfile)
+    else:
+        if checkOutput:
+            return subprocess.check_output(call)
+        else:
+            subprocess.check_call(call)
+
+
+Docker.FORGO = 0
```
fnothaft: Add docstrings for these constants?

src/toil/lib/docker.py
jvivian: +1

src/toil/lib/docker.py
jvivian: My caMeLCaSE foo is shoddy, boofed this one. +1 to private.

src/toil/job.py
```diff
@@ -14,22 +14,21 @@
...
+import base64
 import cPickle
 import collections
 import importlib
 import inspect
 import logging
 import os
-import sys
+import subprocess
 import time
```
benedictpaten: Are the added imports necessary?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,251 @@
...
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param file outfile: Pipe output of Docker call to file handle
+    :param int defer: What action should be taken on the container upon job completion?
+           docker_call.FORGO (0) will leave the container untouched.
+           docker_call.STOP (1) will attempt to stop the container with `docker stop` (useful for
+           debugging).
+           docker_call.RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`.
+           The default value is None and that shadows docker_call.FORGO, unless --rm is being passed in.
+    """
```
benedictpaten: Bit confused. Isn't `--rm` passed in as a parameter by default? Would it not be clearer to say:

"The default value is None, which shadows docker_call.FORGO. However, by default `--rm` is passed as a parameter (see parameters argument), causing the default behavior to be equivalent to docker_call.RM"

Even this is a bit confusing, surely it would be simpler to make the default docker_call.RM and not pass in `--rm` as a parameter? Or am I missing something?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,251 @@
...
+            `--log-driver none`, and the mountpoint `-v work_dir:/data` where /data is the destination convention.
+             These defaults are removed if docker_parmaters is passed, so be sure to pass them if they are desired.
+    :param int defer: What action should be taken on the container upon job completion?
+           docker_call.FORGO (0) will leave the container untouched.
+           docker_call.STOP (1) will attempt to stop the container with `docker stop` (useful for
+           debugging).
+           docker_call.RM (2) will stop the container and then forcefully remove it from the system
+           using `docker rm -f`.
+           The default value is None and that shadows docker_call.FORGO, unless --rm is being passed in.
+    :returns: Stdout from the docker call
```
benedictpaten: Ditto the comment above about `--rm`

src/toil/lib/docker.py
```diff
@@ -0,0 +1,251 @@
...
+
+def dockerCheckOutput(job,
+                      tool,
+                      parameters=None,
+                      workDir=None,
+                      dockerParameters=None,
+                      defer=None):
+    """
+    Wrapper call to Docker. See docstring for docker()
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
```
benedictpaten: I'm not sure I understand what "CheckOutput" means. Worth adding a sentence to explain?

src/toil/lib/docker.py
benedictpaten: Perhaps I'm confused about --rm and using docker_call.RM. I guess @almussel doc PR should address this.

src/toil/job.py
jvivian: Nope, forgot to remove it when I ported the code to src/lib/docker.py

src/toil/lib/docker.py
jvivian: I'd agree with making `docker_call.RM` the default since we pass in `--rm` by default.  

@arkal - that cool with you?

src/toil/lib/docker.py
jvivian: I'll add a sentence to `dockerCheckOutput`.  `check_output` is a subprocess derivative of `check_call`, it just returns the stdout as a string.  It was split as a separate function so the return signature would be consistent.

src/toil/lib/docker.py
jvivian: I'll also add to the _docker() docstring why we use Toils' `defer`

almussel:
I've added docs to this. Please review!

docs/developing.rst
```diff
@@ -598,6 +598,59 @@ Example::
...
+``docker_call``, a toil job function provided in ``toil.lib.docker``. ``docker_call``
+provides a layer of abstraction over using the ``subprocess`` module to call Docker directly,
+and simplifies and streamlines the development process by providing container cleanup on job failure.
+
+In order to use ``docker_call``, your installation of Docker must be set up to run
+without ``sudo``. Instructions for setting this up can be found here_.
+
+.. _here: https://docs.docker.com/engine/installation/linux/ubuntulinux/#/create-a-docker-group
+
+``docker_call`` takes the full name of the dockerized tool, such as ``quay.io/ucsc_cgl/samtools``,
```
cket: This will be easier to maintain if we can point directly to the docker_call docstring using :func:\`docker_call\` instead of repeating information here.

docs/developing.rst
```diff
@@ -598,6 +598,59 @@ Example::
...
+on all workers of  cluster in order to run a workflow involving Docker containers.
+Instructions for installing Docker can be found on the `Docker`_ website.
+
+.. _Docker: https://docs.docker.com/engine/getstarted/step_one/
+
+When using Toil-based autoscaling, Docker will be automatically set up on the cluster,
+so no additional installation steps are necessary. Further information on using
+autoscaling can be found `here`_.
+
+.. _here: https://github.com/BD2KGenomics/toil/blob/master/docs/running.rst
```
cket: You can link directly to the autoscaling header in sphinx, you don't have to link to the raw rst.

docs/developing.rst
```diff
@@ -598,6 +598,59 @@ Example::
 
 .. _service-dev-ref:
 
+
+Using Docker containers in Toil
+-------------------------------
+
+Docker containers are commonly used with Toil to maintain the portability of
+workflows. In order to use Docker containers with Toil, Docker must be installed
+on all workers of  cluster in order to run a workflow involving Docker containers.
```
benedictpaten: Typo, missing "the"

docs/developing.rst
```diff
@@ -598,6 +598,59 @@ Example::
...
+.. _Docker: https://docs.docker.com/engine/getstarted/step_one/
+
+When using Toil-based autoscaling, Docker will be automatically set up on the cluster,
+so no additional installation steps are necessary. Further information on using
+autoscaling can be found `here`_.
+
+.. _here: https://github.com/BD2KGenomics/toil/blob/master/docs/running.rst
+
+`cgl-docker-lib`_ contains Toil-compatible Dockerized tools that are commonly
+used in bioinformatics analysis, as well as documentation on how to develop
```
benedictpaten: What does "Toil-compatible" mean?

benedictpaten: We should provide a little more description here of the cgl-docker-lib containers to justify its existence.

docs/developing.rst
```diff
@@ -598,6 +598,59 @@ Example::
...
+docker_call has several input variables specifying how the container should be run, including
+whether it should be run detached, if directories should be mounted to it, and what should happen
+to the container upon job completion. Docker's output can be returned by docker call or directed
+into a file. An example of a basic ``docker_call`` is below:
+
+    docker_call(job=job,
+                tool='quay.io/ucsc_cgl/bwa',
+                work_dir=job.fileStore.getLocalTempDir(),
+                parameters=['index', '/data/reference.fa'])
+
```
benedictpaten: As docker_call is a job function, it's worth providing an example of how it can directly be added as job to a workflow, I think.

docs/developing.rst
benedictpaten: Are their any essential docker container conventions that we rely upon?

almussel: @benedictpaten What kind of description of cgl-docker-lib containers are you looking for? The fact that they can be used for writing bioinformatics pipelines? I'm not sure they really need additional justification here.

benedictpaten: I mean what does "toil-compatible" mean? Can you explain?

almussel:
@cket @benedictpaten The documentation is ready for re-review.

benedictpaten:
I'm 1+ on this once the tests passes.

fnothaft:
This'll need a rebase as there's a conflict in `src/toil/job.py`.

cket:
Thanks @almussel - also +1 when the tests are passing

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+    if running is None:
+        # This means that the container doesn't exist.  We will see this if the container was run
+        # with --rm and has already exited before this call.
+        _logger.info('The container with name "%s" appears to have already been removed.  Nothing to '
+                  'do.', containerName)
+    else:
+        if action in (None, _docker.FORGO):
+            _logger.info('The container with name %s continues to exist as we were asked to forgo a '
+                      'post-job action on it.', containerName)
+            return
```
fnothaft: Unnecessary return.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+            if action >= _docker.RM:
+                # If the container was run with --rm, then stop will most likely remove the
+                # container.  We first check if it is running then remove it.
+                running = _containerIsRunning(containerName)
+                if running is not None:
+                    _logger.info('Removing container "%s".', containerName)
+                    try:
+                        subprocess.check_call(['docker', 'rm', '-f', containerName])
+                    except subprocess.CalledProcessError:
+                        _logger.exception("'docker rm' failed.")
```
fnothaft: I'm not convinced that we want to eat the exception here. What's the reasoning behind catching this?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+                         containerName)
+            if running and action >= _docker.STOP:
+                _logger.info('Stopping container "%s".', containerName)
+                subprocess.check_call(['docker', 'stop', containerName])
+            else:
+                _logger.info('The container "%s" was not found to be running.', containerName)
+            if action >= _docker.RM:
+                # If the container was run with --rm, then stop will most likely remove the
+                # container.  We first check if it is running then remove it.
+                running = _containerIsRunning(containerName)
```
fnothaft: JFYI, _containerIsRunning throws an assert, so if you don't check here, you might throw an assert back to the user.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def _dockerKill(containerName, action):
+    """
+    Kills the specified container.
+    :param str containerName: The name of the container created by docker_call
+    :param int action: What action should be taken on the container?  See `defer=` in
+           :func:`docker_call`
+    """
+    running = _containerIsRunning(containerName)
```
fnothaft: Ditto here RE throwing an assert back to the user.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
+import base64
+import logging
+import subprocess
+
+import os
```
jvivian: Pycharm's optimize imports put it there, not sure why.  I agree it makes sense to move it among the core library imports.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+
+def dockerCall(job,
+               tool,
+               parameters=None,
+               workDir=None,
+               dockerParameters=None,
+               outfile=None,
+               defer=None):
+    """
+    Wrapper call to Docker. See docstring for _docker()
```
jvivian: Would you prefer I copy and paste the `_docker` docstring into both `dockerCall` and `dockerCheckOutput` and then remove all but the `:param` stuff from `_docker` ?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
+
+    :param toil.Job.job job: The Job instance for the calling function.
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools).
```
jvivian: I could also just append `:latest` to `quay.io/ucsc_cgl/samtools` as the example if you'd think that sufficient.

src/toil/lib/docker.py
jvivian: To clarify, do you mean wrap in a try/except block and catch the `AssertionError`?

src/toil/lib/docker.py
jvivian: @arkal says it is to minimize the traceback the user gets.

src/toil/lib/docker.py
jvivian: Removed

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+    if outfile:
+        subprocess.check_call(call, stdout=outfile)
+    else:
+        if checkOutput:
+            return subprocess.check_output(call)
+        else:
+            subprocess.check_call(call)
+
+
+_docker.FORGO = 0
```
jvivian: @arkal had a suggestion — What would you think about making a class `Docker` or `ToilDocker` that has a call method, checkOutput method, etc? This way FORGO / RM / STOP can be attributes of the class and.

src/toil/lib/docker.py
fnothaft: I had a comment lower down about making the `_docker` documentation into module level docs. LMK if that makes sense.

src/toil/lib/docker.py
fnothaft: That SGTM

src/toil/lib/docker.py
fnothaft: If you just drop the `_docker` prefix here, the "constants" (hahahaha python IKR!!!!!) will show up as just in the module (i.e., `from toil.lib.docker import FORGO`), which I think is the correct way to do this.

src/toil/lib/docker.py
fnothaft: I'd have to read through the code a bit more, but I'd either wrap in a try/except and catch, or change the AssertionError downstream into a RuntimeError or something of the like.

src/toil/lib/docker.py
jvivian: Ok, that works for me.

src/toil/lib/docker.py
fnothaft: As a user of the API, my general feeling is that I'd prefer to get a `RuntimeException` back if this failed (or at least the option to get an exception back), rather than having the exception get eaten silently. Like, I'm thinking of how we'd use these in the Spark service jobs to kill the Spark/HDFS containers (which run in a daemon-ish mode). If I try to kill those and I fail, then the nodes are in a weird state and I likely can't run a new Spark/HDFS container on those nodes (resource conflicts, IIRC). That's my general intuition.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,255 @@
...
+def _docker(job,
+            tool,
+            parameters=None,
+            workDir=None,
+            dockerParameters=None,
+            outfile=None,
+            checkOutput=False,
+            defer=None):
+    """
+    Calls Docker for a particular tool with the specified parameters. Assumes `docker` is on the PATH.
```
jvivian: SGTM

src/toil/lib/docker.py
jvivian: I'll do the latter, that's a bit less cludgy.

src/toil/lib/docker.py
jvivian: That makes sense to me. @arkal , care to comment or you OK with me removing it?

docs/developing.rst
```diff
@@ -596,6 +596,80 @@ Example::
...
+between any platform that has both Toil and Docker installed. Docker eliminates the need for
+the user to do any other tool installation or environment setup.
+
+In order to use Docker containers with Toil, Docker must be installed
+on all workers of the cluster in order to run a workflow involving Docker containers.
+Instructions for installing Docker can be found on the `Docker`_ website.
+
+.. _Docker: https://docs.docker.com/engine/getstarted/step_one/
+
+When using Toil-based autoscaling, Docker will be automatically set up on the cluster,
```
almussel: @jvivian @cket can we get clarification on this?

docs/developing.rst
```diff
@@ -596,6 +596,80 @@ Example::
...
+
+`cgl-docker-lib`_ contains ``dockerCall``-compatible Dockerized tools that are commonly
+used in bioinformatics analysis. 
+
+.. _cgl-docker-lib: https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/README.md
+
+The documentation provides guidelines for developing your own Docker containers that can
+be used with Toil and ``dockerCall``. In order for a container to be compatible with
+``dockerCall``, it must have an ``ENTRYPOINT`` set to a wrapper script, as described
+in cgl-docker-lib containerization standards, such that the container can be run directly
```
almussel: The standards are in the readme linked in the previous paragraph, but I can link again if you want?

docs/developing.rst
```diff
@@ -596,6 +596,80 @@ Example::
...
+         Job.Runner.startToil(align, options)
+
+`cgl-docker-lib`_ contains ``dockerCall``-compatible Dockerized tools that are commonly
+used in bioinformatics analysis. 
+
+.. _cgl-docker-lib: https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/README.md
+
+The documentation provides guidelines for developing your own Docker containers that can
+be used with Toil and ``dockerCall``. In order for a container to be compatible with
+``dockerCall``, it must have an ``ENTRYPOINT`` set to a wrapper script, as described
```
almussel: Just checked the ``docker run`` docs and all I'm seeing is ``--entrypoint``, which should work, yes.

jvivian:
@benedictpaten — Tests pass and @fnothaft 's previous review comments have been addressed. Once doc changes are in this should be g2g.

docs/developing.rst
```diff
@@ -596,6 +596,80 @@ Example::
...
+
+.. _Docker: https://docs.docker.com/engine/getstarted/step_one/
+
+When using Toil-based autoscaling, Docker will be automatically set up on the cluster,
+so no additional installation steps are necessary. Further information on using
+autoscaling can be found in the `Toil autoscaling documentation <Autoscaling>`.
+
+In order to use docker containers in a Toil workflow, the container can be built locally or
+downloaded in real time from an online docker repository like Quay_. If the container is local,
+it must be built on each node of the cluster.
```
almussel: I think it's fair to say that anyone using a private repository is responsible for figuring that out.

docs/developing.rst
cket: This is true for both cgcloud and the appliance based autoscaling. CGCloud-based autoscaling is still supported, btw.

src/toil/lib/docker.py
```diff
@@ -0,0 +1,246 @@
...
+
+def dockerCall(job,
+               tool,
+               parameters=None,
+               workDir=None,
+               dockerParameters=None,
+               outfile=None,
+               defer=None):
+    """
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
```
benedictpaten: Should talk about the blocking behavior of this command here, i.e. that this command blocks until the docker container dies, right?

src/toil/lib/docker.py
```diff
@@ -0,0 +1,246 @@
...
+
+def dockerCheckOutput(job,
+                      tool,
+                      parameters=None,
+                      workDir=None,
+                      dockerParameters=None,
+                      defer=None):
+    """
+    Returns the stdout from the Docker invocation (via subprocess.check_output)
+    Throws CalledProcessorError if the Docker invocation returns a non-zero exit code
```
benedictpaten: Ditto, here a few sentences to explain the difference between this and dockerCall would probably help the newcomer.

benedictpaten:
Okay, we're basically all good. I like the test - it's nice and simple. I requested a couple more sentences of documentation, then I'll merge, either @jvivian  or @almussel please add. Thanks.

benedictpaten:
Oh yeah, and at same time please squash down the commits and rebase on master. Let's get this merged. I will then write some more tests for services and docker containers. Thanks

jvivian:
Continued in https://github.com/BD2KGenomics/toil/pull/1438 due to general incompetence.

benedictpaten:
#1438 is merged. Closing this.

PR #1404, opened 2016-12-20
Issues/documentation

Allkoman:
'make_file_to_srt' should be 'make_file_to_sort' at the docs file


cket:
@Allkoman thanks for reporting this - I think we've already fixed this typo in the master branch's documentation here: http://toil.readthedocs.io/en/latest/. Once we do our next stable release of Toil the default readthedocs page should show the fixed documentation. Feel free to reopen this as an issue if this didn't solve the problem!

PR #1417, opened 2016-12-30
Move internal log messages to DEBUG

### LINKED ISSUE #1061: Move logging message for internal Toil job operations to DEBUG
cc @arkal @hannes-ucsc 

This PR is spun out of the discussion in #1048. In CWL runs we see a large number of logging messages devoted to internal Toil processes output to INFO:

```
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'h/y/tmpMJXEcB.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'j/W/tmp2cwAzR.tmp'.
Got message from job at time 07-13-2016 10:15:21: Successfully deleted local copies of file with ID 'G/2/tmpGT1jJU.tmp'.
```

From the discussion, the work plan is to only report these messages in INFO if called through the Toil API, and if called from internal Toil (or Toil/CWL) processes report them as DEBUG.

chapmanb:
During CWL runs, there are extra messages in the INFO log
relating to internal processes, which obscure useful messages. This
moves these to DEBUG:

- Internal jobs releated to CWL
- Cache access and read messages
- Deletion of cached files
- Adding files to be deleted from cache

Fixes #1061

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -491,7 +491,7 @@ class CachingFileStore(FileStore):
                                            disk=diskUsed,
                                            humanRequestedDisk=bytes2human(jobReqs),
                                            requestedDisk=jobReqs))
-            self.logToMaster(logString, level=logging.INFO)
+            self.logToMaster(logString, level=logging.DEBUG)
             if diskUsed > jobReqs:
                 self.logToMaster("Job used more disk than requested. Please reconsider modifying "
                                  "the user script to avoid the chance  of failure due to "
@@ -628,7 +628,7 @@ class CachingFileStore(FileStore):
         # userPath. Cache operations can only occur on local files.
         with self.cacheLock() as lockFileHandle:
             if fileIsLocal and self._fileIsCached(fileStoreID):
-                logger.info('CACHE: Cache hit on file with ID \'%s\'.' % fileStoreID)
+                logger.debug('CACHE: Cache hit on file with ID \'%s\'.' % fileStoreID)
                 assert not os.path.exists(localFilePath)
                 if mutable:
                     shutil.copyfile(cachedFileName, localFilePath)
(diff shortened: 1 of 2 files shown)


src/toil/fileStore.py
```diff
@@ -1603,7 +1607,8 @@ def open(self, job):
                                            disk=diskUsed,
                                            humanRequestedDisk=bytes2human(jobReqs),
                                            requestedDisk=jobReqs))
-            self.logToMaster(logString, level=logging.INFO)
+            self.logToMaster(logString, level=(logging.DEBUG if self.jobName.startswith(self.debugJobNames)
```
cket: I don't think this logic is necessary. We can just make this statement DEBUG outright since the next log message is level WARNING but only logs when the job uses more disk than requested.

src/toil/fileStore.py
```diff
@@ -1603,7 +1607,8 @@ def open(self, job):
...
+                                               else logging.INFO))
             if diskUsed > jobReqs:
```
cket: I would possibly change this to be if/else since this message and the one above are redundant

cket:
@chapmanb thanks for the PR, just a couple minor changes!

src/toil/fileStore.py
chapmanb: Thanks for looking at the PR. What I'm trying to do with the different levels is move the internal events (in CWL, things like workflow and job invocations) to the DEBUG log, while leaving the higher level outputs (actual commands being run) in the INFO output. This is a super useful output to have for tasks since it lets you measure progress of a workflow but gets overwhelmed by all the internal messages. The WARNING below is also useful but a separate thing you'd want reported. Hopefully that helps explain the logic behind having two different levels here.

src/toil/fileStore.py
chapmanb: I think it is different than the warning message, and tried to explain the logic above. Thanks again for the comments.

src/toil/fileStore.py
cket: Ok, I think I understand the intent now. In that case, there are a couple ways to easily improve tracking of the workflow. I'm still in favor of decreasing this statement to DEBUG since its main purpose is propagating the disk usage message which I think is only of general interest when the usage is greater than expected, which is handled by the next log statement.

Instead, I think increasing this message to INFO (potentially with the internal CWL message masking) will do what you are looking for: https://github.com/BD2KGenomics/toil/blob/36d22d1171075de4918b40a317dbdce921ea5f31/src/toil/leader.py#L423 

Bumping this message to INFO and improving it by changing ```jobNode.jobStoreID``` to just ``jobNode`` could also help track the workflow's progress: https://github.com/BD2KGenomics/toil/blob/36d22d1171075de4918b40a317dbdce921ea5f31/src/toil/leader.py#L519

With those changes you'll get messages when the task is ready to be run & when it is finished running.

chapmanb:
Christopher -- great suggestions, thank you. I swapped over to logging job start/stop in INFO for non-internal CWL job using the pointers in leader you suggested. The disk usage metrics are now all in DEBUG. This is great, the output is now much cleaner and we have messages when jobs start and finish. Thank you again and please let me know if you have any other suggestions.

benedictpaten:
+1 Excellent change

PR #1423, opened 2017-01-10
Clean up Mesos logging 

### LINKED ISSUE #1422: Remove mesos offer log messages
@arkal found that in a recent autoscaling test 90+% of the log messages were debug messages about declining mesos offers.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-127)
┆Issue Number: TOIL-127

cket:
Resolves #1422. Removes a log message that was repeated len(queued jobs) * len(offers) - when we have thousands of jobs and dozens of simultaneous offers this gets insane. Especially since we get a fresh batch of offers a multiple times a minute.


cket:
As discussed in #1422, we need a different approach for this issue.

PR #1433, opened 2017-01-13
Enhance logs output when Toil runs out of disk (resolves #1428)

### LINKED ISSUE #1428: Better error messages when Toil runs out of disk
For both caching and non-caching file store

arkal:
resolves#1428


cket:
@arkal is this ready for review or are you still working on it?

PR #1451, opened 2017-01-23
Remove CGCloud provisioner, refactor integration tests

cket:
Release of Toil 3.5 exposed dependency issues in the integration tests. The version of toil-scripts we rely on only 'works' (due to range checking, not actual compatibility issues) with Toil < 3.5.


cket:
jenkins, test this please

PR #1460, opened 2017-01-26
Increase wait time before failing test 

### LINKED ISSUE #1452: Gridengine runJobs test too aggressive
Only gives the batch system 2 seconds to issue the jobs given to it. This isn't always enough.

cket:
Resolves #1452


cket:
jenkins, test this please

PR #1473, opened 2017-02-01
Handle race between async writeGlobalFile and exportFile

### LINKED ISSUE #1469: S3 multipart uploader crashed for large file
I tried to use the AWS jobStore's exportFile() to export a 30G file to an S3 bucket.  Got this error (same code worked for smaller files)

https://github.com/BD2KGenomics/toil-vg/issues/110#issuecomment-276095314

cket:
Resolves #1469 

First commit just adds test.

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -248,7 +248,7 @@ class FileStore(object):
         return self.jobStore.importFile(srcUrl, sharedFileName=sharedFileName)
 
     def exportFile(self, jobStoreFileID, dstUrl):
-        self.jobStore.exportFile(jobStoreFileID, dstUrl)
+        raise NotImplementedError()
 
     # A utility method for accessing filenames
     def _resolveAbsoluteLocalPath(self, filePath):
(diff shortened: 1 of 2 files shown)


arkal:
Let me know if you come across any difficulties

cket:
Perfect, the test reproduces the error that @glennhickey reported. I'll push the fix.

cket:
jenkins, test this please

src/toil/fileStore.py
```diff
@@ -721,6 +721,13 @@ def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=None):
                                                               0.0, False)
         return localFilePath
 
+    def exportFile(self, jobStoreFileID, dstUrl):
+        while jobStoreFileID in self._pendingFileWrites:
```
arkal: Can i pull a @hannes-ucsc on this and say that we have a CachingFileStore.harbingerFile.waitOnDownload method in place exactly for this reason?

cket: Maybe I'm not understanding the docs, but I don't think that does what I want. I want something along the lines of .waitOnUpload()

arkal: It's basically the same thing..... waitOnDownload waits for the harbinger to get deleted, signifying the file has been transferred to/from the jobstore. Can possibly  rename it to `waitOnTransfer`

cket: I see. So far it seems like this approach would still be cleaner since it is further abstracted from the actual caching logic. I don't have to refactor/rename any methods and I don't have to mess with the harbinger/lock file at all, I can just perform my check with the fileID directly.

arkal: Cleaner, maybe. But i feel that since the wait is inherently dependent on caching, i feel that would be the correct way to do it.

cket: Fair enough. If it comes down to personal preference I'll keep the code as is since it makes the most sense to me

PR #1474, opened 2017-02-02
Optionally tag auto-scaled EC2 instances with user-specified tags (resolves #1341)

### LINKED ISSUE #1341: Optionally tag auto-scaled EC2 instances with user-specified tags
Add `--tag key=value` to the ~~`toil create-cluster`~~ `toil launch-cluster` entry point and use the provided key and value to tag all instances, the leader and subsequently auto-scaled worker instances.

Note that auto-scaling (i.e. the creation of worker nodes) usually happens in a different process, not the ~~`toil create-cluster`~~ `toil launch-cluster` one.  It happens in the leader process, when the user script is run on the leader. When a new worker instance has been created in the leader, simply copy the leader's tags to newly created worker instance. I know of at least one standard tag that should not be copied. I think the name of that tag is `Name` and the EC2 console uses it for the column of the same name in the list view of all instances.

For argparse, specify `action=append` such that the user can specify it multiple times. Also add a short form `-T`:

`-T foo=bar -T x=y`

jessebrennan:
User can specify tags in the command line when running `toil launch-cluster` which will automatically be added to all node as the cluster auto scales. The form for tags is:

`toil launch-cluster -T key1=value1 --tag key2=value2 ...`

Multiple tags can be specified.

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -22,6 +22,7 @@ import time
 import sys
 
 # Python 3 compatibility imports
+from six import iteritems
 from six.moves import xrange
 
 from bd2k.util import memoize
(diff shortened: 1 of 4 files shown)


jessebrennan:
Currently no tests run with user specified tags nor check that tags are successfully copied to all children.

Otherwise it's ready for review @cket.

cket:
Great, thanks @jessebrennan!

PR #1475, opened 2017-02-02
Report 'theoretical' cluster usage in addition to actual usage

### LINKED ISSUE #1416: Report cluster utilization in terms of Job's resource requirements, not the actual usage
This will help us improve autoscaling efficiency since we can't control how accurate the resource requirements actually are.

cket:
Resolves #1416 

The cluster statistics currently report the cluster's actual resource usage. This is helpful to users but for Toil development, especially in relation to the autoscaler, we only care about how efficiently Toil is managing the resources that a user requests regardless of whether the user is using them. This lets us distinguish between efficiency losses due to Toil scheduling/autoscaling and losses due to a user requesting more resources than needed.

### DIFF
--- a/docs/contributing/source.rst
+++ b/docs/contributing/source.rst
@@ -113,3 +113,23 @@ this may come up empty. To fix it, run the following:
 
 .. _Docker: https://www.docker.com/products/docker
 .. _Quay: https://quay.io/
+
+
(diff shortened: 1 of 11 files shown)


cket:
Jenkins, test this please

cket:
Jenkins, test this please

PR #1478, opened 2017-02-03
Issues/1465 1466 Fix verification that new checkpoint jobs are leaf vertices

### LINKED ISSUE #1465: Detection of checkpoint jobs that are not leaf vertices identifies leaf vertices incorrectly
The Toil documentation is clear that a Job with checkpoint=True must be a leaf in the initial workflow graph. This makes sense, as it ensures the job can be reliably cleaned-up and restarted in the case that it or one of its successors fails. Yet, the current code to identify a leaf vertex misses some cases.

Below is an excerpt of the current code from  `toil.job.Job.checkNewCheckpointsAreLeafVertices`:

```python
for y in filter(lambda x : x.checkpoint, jobs):
    if y not in roots: # The roots are the prexisting jobs
        if len(y._children) != 0 and len(y._followOns) != 0 and len(y._services) != 0:
            raise JobGraphDeadlockException("New checkpoint job %s is not a leaf in the job graph" % y)
```

The following line checks that a checkpoint Job has children and follow-ons and services for the exception to be raised:

```python
        if len(y._children) != 0 and len(y._followOns) != 0 and len(y._services) != 0:
```

Instead, I think this line should be as follows:
```python
        if len(y._children) != 0 or len(y._followOns) != 0 or len(y._services) != 0:
```

If the checkpoint Job has any children or any follow-ons or any services, then it is not a leaf and the exception should be raised.

The before and after behavior of this change may be verified with the following simple workflow.
With the current code, the workflow runs to completion.
With the modification above, the exception is raised due to the checkpoint job that is not a leaf.
```python
from argparse import ArgumentParser
from toil.job import Job

def jobFn(job, value):
    job.fileStore.logToMaster(value)

def createWorkflow():
    rootJob = Job.wrapJobFn(jobFn, "Parent")
    childJob = rootJob.addChildJobFn(jobFn, "Child", checkpoint=True) # Checkpoint a non-leaf job
    grandChildJob = childJob.addChildJobFn(jobFn, "Grandchild")
    return rootJob

def main():
    parser = ArgumentParser()
    options = parser.parse_args()

    rootJob = createWorkflow()

    Job.Runner.startToil(rootJob, options)

if __name__ == '__main__':
    main()
```

### LINKED ISSUE #1466: Detection of checkpoint jobs that are not leaf vertices omits the workflow root job
The current code to detect checkpoint jobs that are not leaf vertices does not check the workflow root job.

Below is an excerpt from `toil.job.Job.checkNewCheckpointsAreLeafVertices`:
```python
for y in filter(lambda x : x.checkpoint, jobs):
    if y not in roots: # The roots are the prexisting jobs
        if len(y._children) != 0 and len(y._followOns) != 0 and len(y._services) != 0:
            raise JobGraphDeadlockException(“New checkpoint job %s is not a leaf in the job graph” % y)
```
The line in question:
```python
    if y not in roots: # The roots are the prexisting jobs
```

This line **_is correct and necessary_** during normal/ongoing operation, as well as when resuming a workflow. Yet, it also prevents detection of a workflow root vertex that is marked as a checkpoint yet is not a leaf vertex at the time the workflow is first started. (This case is illustrated in the sample workflow below.)

Both the initial/start-up case and the ongoing/resume case come through this same code path, and in general they can and likely should be handled the same. However, in this particular corner case, it creates an issue.

Two potential approaches come to mind for fixing this:
- Change the current code path to detect the difference between initial and ongoing states and handle them differently.
I don’t like this option because it adds otherwise unnecessary complexity to the common case. It would complicate both the production code and the testing. Also, I have not noticed precedent for this approach in the current code base.

- Leave the current code path as it is, and identify another place in the code that the check for this specific case (initial, root checkpoint that is not a leaf) could be performed.

The second option could be implemented as follows:

Refactor the leaf check into a new method:
```python
    @staticmethod
    def _isLeafVertex(job):
        return len(job._children) == 0 and len(job._followOns) == 0 and len(
            job._services) == 0
```

The existing code is unchanged except for this substitution:
```python
for y in filter(lambda x : x.checkpoint, jobs):
    if y not in roots: # The roots are the prexisting jobs
        if not Job._isLeafVertex(y):
            raise JobGraphDeadlockException("New checkpoint job %s is not a leaf in the job graph" % y)
```

And one if-statement is added to _serializeFirstJob (first statement in the method below)
```python
def _serialiseFirstJob(self, jobStore):
    """
    Serialises the root job. Returns the wrapping job.

    :param toil.jobStores.abstractJobStore.AbstractJobStore jobStore:
    """
    # Special case check for an initial root job that is a checkpoint yet is not a leaf vertex.
    # This is checked for all other jobs by Job.checkNewCheckpointsAreLeafVertices
    if self.checkpoint and not Job._isLeafVertex(self):
        raise JobGraphDeadlockException(
            "Root checkpoint job %s is not a leaf in the job graph" % self)

    # Create first jobGraph
    jobGraph = self._createEmptyJobGraphForJob(jobStore=jobStore, predecessorNumber=0)
    # Write the graph of jobs to disk
    self._serialiseJobGraph(jobGraph, jobStore, None, True)
    jobStore.update(jobGraph)
    # Store the name of the first job in a file in case of restart. Up to this point the
    # root job is not recoverable. FIXME: "root job" or "first job", which one is it?
    jobStore.setRootJob(jobGraph.jobStoreID)
    return jobGraph
```

Although this is a corner case, it could happen and would be a problem for users if it did.
This check is performed for every other job vertex it the workflow, it seems we should do it for the workflow root vertex as well.
I prefer to not have a check for the essentially the same condition in two different locations in the code, yet at least the two locations are in the same class (Job), rather than disparate parts of the code. To me, this seems to be the better of the two options identified.
Do you have thoughts/recommendations for better ways to fix this?

The before and after behavior of this change may be verified with the following simple workflow.
With the current code, the workflow runs to completion.
With the modification above, the exception is raised due to the root checkpoint job that is not a leaf.
```python
from argparse import ArgumentParser
from toil.job import Job

def jobFn(job, value):
    job.fileStore.logToMaster(value)

def createWorkflow():
    rootJob = Job.wrapJobFn(jobFn, "Parent", checkpoint=True) # Checkpoint root non-leaf job
    childJob = rootJob.addChildJobFn(jobFn, "Child")
    grandChildJob = childJob.addChildJobFn(jobFn, "Grandchild")
    return rootJob

def main():
    parser = ArgumentParser()
    options = parser.parse_args()

    rootJob = createWorkflow()

    Job.Runner.startToil(rootJob, options)

if __name__ == '__main__':
    main()
```

mikebaumann:
Fix verification that new checkpoint jobs are leaf vertices (resolves #1465, resolves #1466)

Changed to 1) correctly identify a leaf vertex and 2) to include checking of the workflow root job.

Note: This pull request is also listing a merge-related commit from updating my fork.
However, this merge-related commit did not contribute any of the changes displayed for this pull request. I reviewed the changes shown for this pull request, and they are a correct and complete set for the two fix-related commits that are intended for this pull request.
I think I know why that happened and I will avoid including extraneous commits in future pull requests.

### DIFF
--- a/src/toil/test/src/jobTest.py
+++ b/src/toil/test/src/jobTest.py
@@ -275,7 +275,6 @@ class JobTest(ToilTest):
 
         self.runNewCheckpointIsLeafVertexTest(createWorkflow)
 
-    @unittest.skip("Enable this test when issue #1466 has been fixed.")
     def testNewCheckpointIsLeafVertexRootCase(self):
         """
         Test for issue #1466: Detection of checkpoint jobs that are not leaf vertices
--- a/src/toil/test/src/jobTest.py
+++ b/src/toil/test/src/jobTest.py
@@ -254,6 +254,113 @@ class JobTest(ToilTest):
                 and (fNode, tNode) not in childEdges and (fNode, tNode) not in followOnEdges):
                 checkFollowOnEdgeCycleDetection(fNode, tNode)
 
+    def testNewCheckpointIsLeafVertexNonRootCase(self):
+        """
+        Test for issue #1465: Detection of checkpoint jobs that are not leaf vertices
+        identifies leaf vertices incorrectly
+
+        Test verification of new checkpoint jobs being leaf verticies,
+        starting with the following baseline workflow:
+
+            Parent
+              |
+            Child # Checkpoint=True
+
+        """
+
+        def createWorkflow():
+            rootJob = Job.wrapJobFn(simpleJobFn, "Parent")
+            childCheckpointJob = rootJob.addChildJobFn(simpleJobFn, "Child", checkpoint=True)
+            return rootJob, childCheckpointJob
+
+        self.runNewCheckpointIsLeafVertexTest(createWorkflow)
+
+    @unittest.skip("Enable this test when issue #1466 has been fixed.")
+    def testNewCheckpointIsLeafVertexRootCase(self):
+        """
+        Test for issue #1466: Detection of checkpoint jobs that are not leaf vertices
+                              omits the workflow root job
+
+        Test verification of a new checkpoint job being leaf vertex,
+        starting with a baseline workflow of a single, root job:
+
+            Root # Checkpoint=True
+
+        """
+
+        def createWorkflow():
+            rootJob = Job.wrapJobFn(simpleJobFn, "Root", checkpoint=True)
+            return rootJob, rootJob
+
+        self.runNewCheckpointIsLeafVertexTest(createWorkflow)
+
+    def runNewCheckpointIsLeafVertexTest(self, createWorkflowFn):
+        """
+        Test verification that a checkpoint job is a leaf vertex using both
+        valid and invalid cases.
+
+        :param createWorkflowFn: function to create and new workflow and return a tuple of:
+
+                                 0) the workflow root job
+                                 1) a checkpoint job to test within the workflow
+
+        """
+
+        print('Test checkpoint job that is a leaf vertex')
+        self.runCheckpointVertexTest(*createWorkflowFn(),
+                                     expectedException=None)
+
+        print('Test checkpoint job that is not a leaf vertex due to the presence of a service')
+        self.runCheckpointVertexTest(*createWorkflowFn(),
+                                     checkpointJobService=TrivialService("LeafTestService"),
+                                     expectedException=JobGraphDeadlockException)
+
+        print('Test checkpoint job that is not a leaf vertex due to the presence of a child job')
+        self.runCheckpointVertexTest(*createWorkflowFn(),
+                                     checkpointJobChild=Job.wrapJobFn(
+                                         simpleJobFn, "LeafTestChild"),
+                                     expectedException=JobGraphDeadlockException)
+
+        print('Test checkpoint job that is not a leaf vertex due to the presence of a follow-on job')
+        self.runCheckpointVertexTest(*createWorkflowFn(),
+                                     checkpointJobFollowOn=Job.wrapJobFn(
+                                         simpleJobFn,
+                                         "LeafTestFollowOn"),
+                                     expectedException=JobGraphDeadlockException)
+
+    def runCheckpointVertexTest(self,
+                                workflowRootJob,
+                                checkpointJob,
+                                checkpointJobService=None,
+                                checkpointJobChild=None,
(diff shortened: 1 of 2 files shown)


cket:
Jenkins, test this please

mikebaumann:
Hi team,

While reviewing this pull request, please consider the question of implementation options for issue #1466, as posed in the description of that issue.

Currently, the fix for #1466 is implemented very simply. Based on your broader perspective and experience, you may identify and prefer other implementation options. If so, just let me know and I'll get to it!

Thanks!

Mike

cket:
@mikebaumann thanks for the PR! The implementation looks good to me and I'm happy with the tests' coverage as well. Although there are only 2 new tests each test really runs 4 test cases, right? We generally try make each test as granular as possible, but in this case I think grouping the 4 cases together makes sense too.

I left one minor inline comment, and if I'm being very picky I would also replace the PRs title with the first sentence in your description instead.

src/toil/test/src/jobTest.py
```diff
@@ -254,6 +254,112 @@ def checkFollowOnEdgeCycleDetection(fNode, tNode):
...
+        valid and invalid cases.
+
+        :param createWorkflowFn: function to create and new workflow and return a tuple of:
+
+                                 0) the workflow root job
+                                 1) a checkpoint job to test within the workflow
+
+        """
+
+        print('Test checkpoint job that is a leaf vertex')
```
mikebaumann: Hi CJ,

Thanks for the helpful comment.
Yes, logging is the preferred way to output informational text.
As background, when adding these tests originally, I checked whether the test (jobTest.py) used logging or print for informational/diagnostic messages. I found no use of logging and a few print statements. I then incorrectly inferred that this is the CGL/Toil convention. Yet, based on viewing other tests since then, I see that logging is clearly the convention, as it should be.
I added the commonly used test logging setup in JobTest.setUpClass, then changed the print statements I had added to this test to logging statements instead.
I did not change the other few, preexisting print statements in jobTest.py as they are not associated with this PR. I would be happy to update these preexisting uses of print as well, yet that is probably best not done as part of this in this PR, IMHO.

cket:
Jenkins, ok to test

cket:
Jenkins, test this please

Issue #1470, opened 2017-01-31
toil AWS Provisioner failing without TOIL_AWS_AMI set


rborkow:
Hi - 

Just getting started with toil (3.5.1), and attempted to set up a cluster as per the "Autoscaling" documentation:

`toil launch-cluster -p aws <cluster-name> \
     --nodeType=m3.large \
     --keyPairName=<keypair>`

Consistently fails while trying to launch a non-preemptable leader as follows:
`Traceback (most recent call last):
  File "/usr/local/bin/toil", line 9, in <module>
    load_entry_point('toil==3.5.1', 'console_scripts', 'toil')()
  File "/usr/local/lib/python2.7/site-packages/toil/utils/toilMain.py", line 33, in main
    module.main()
  File "/usr/local/lib/python2.7/site-packages/toil/utils/toilLaunchCluster.py", line 53, in main
    keyName=config.keyPairName, spotBid=spotBid)
  File "/usr/local/lib/python2.7/site-packages/toil/provisioners/aws/awsProvisioner.py", line 357, in launchCluster
    create_ondemand_instances(ctx.ec2, image_id=cls._discoverAMI(ctx),
  File "/usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/bd2k/util/__init__.py", line 45, in new_f
    r = f( *args )
  File "/usr/local/lib/python2.7/site-packages/toil/provisioners/aws/awsProvisioner.py", line 120, in _discoverAMI
    assert len(coreOSAMI) == 1
AssertionError
`

Searching through the AWS console, there doesn't appear to be to be an AMI matching the criteria this function defaults to - description containing 'stable 1068.9.0' with an owner '679593333241'. 

Updating the description string to 'stable 1235.6.0' will allow cluster creation to proceed after going to a URL to accept licensing terms.

cket:
@rborkow what EC2 zone are you running in? Unfortunately, AMIs are local to every zone so it is possible that the zone you are running in doesn't have the CoreOS AMI we rely on.

rborkow:
Got it -- 'us-west-1'

cket:
Ahh, thats unfortunate. I thought I had checked that region for a matching AMI - I hope they aren't deprecating it. Let me know if 1235.6.0 is working for you - it might make sense for us update to that version.

rborkow:
[...]

PR #1446, opened 2017-01-19
Handle AWS request limit exceeded

cket:
Resolves #1445

PR #1512, opened 2017-02-14
AWS autoscaling: arbitrary VPCs

### LINKED ISSUE #1508: AWS runs: Identifying clusterName in VPC with multiple security groups
CJ and all;
I've made progress in setting up AWS Toil CWL runs and am currently running into an issue when the AWSProvisioner class bootstraps. The origin of the problem is here:

https://github.com/BD2KGenomics/toil/blob/master/src/toil/provisioners/aws/awsProvisioner.py#L51

where it tries to infer the cluster name from security groups in the VPC. If you have any other security groups present in the VPC (default, plus any specialized ones for other applications) you get this error:
```
ip-10-0-0-210.ec2.internal 2017-02-13 19:09:09,739 MainThread INFO toil.common: Using AWS provisioner.
I0213 19:09:09.777951   152 sched.cpp:1987] Asked to stop the driver
I0213 19:09:09.778194   222 sched.cpp:1187] Stopping framework '90df95d0-8968-43ea-8994-d125f78ce8a1-0001'
  File "/usr/local/bin/cwltoil", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 727, in main
    outobj = toil.start(wf1)
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 675, in start
    self._setProvisioner()
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 712, in _setProvisioner
    self._provisioner = AWSProvisioner(self.config, self._batchSystem)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/aws/awsProvisioner.py", line 52, in __init__
    self.ctx = self._buildContext(clusterName=self.clusterName)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/aws/awsProvisioner.py", line 97, in _buildContext
    return Context(availability_zone=zone, namespace=cls._toNameSpace(clusterName))
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/aws/awsProvisioner.py", line 221, in _toNameSpace
    assert isinstance(clusterName, str), repr(clusterName)
AssertionError: ['toil-bcbio', 'default', 'bcbio_cluster_sg']
```
Are there approaches we can use to input the clusterName, or otherwise query for it, so we avoid the need to infer this from security_groups? Thanks for any suggestions/directions on how best to approach this.

chapmanb:
CJ -- this follows up on our discussion in #1470 around using accounts with non-default VPCs. It's back compatible with the previous approach but provides a new option so a user can specify a VPC to launch in, then all other steps happen within that VPC. With this in place I can spin up and start running autoscaling CWL jobs, then get blocked by a bug in cwltool which is also fixed here. Thanks much for the help getting this in.

- Allow specifying a VPC subnet when launching a cluster. The previous
  code assumed use of a default VPC which is not true for all AWS
  accounts.
- VPC subnet passed to all node creation scripts so everything occurs
  in same configured VPC.
- Using specific VPCs in boto requires sending security groups by ID
  instead of name (https://github.com/boto/boto/issues/350).
- Retrieve cluster name from instance tags instead of security groups
  (fixes #1508)
- Place spot instances in current VPC availability zone.
- Handle spot instances where we can't find pricing history for
  an instance type.
- Use public_dns_name instead of ip_address for ssh. This allows
  tweaking AWS ssh options in `~/.ssh/config` via `Host *.amazonaws.com`
- Update cwltool to latest stable tested version, fixing issue
  with schema-salad in current Toil Docker images.

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -26,6 +26,7 @@ from six import iteritems
 from six.moves import xrange
 
 from bd2k.util import memoize
+import boto.ec2
 from boto.ec2.blockdevicemapping import BlockDeviceMapping, BlockDeviceType
 from boto.exception import BotoServerError, EC2ResponseError
 from cgcloud.lib.ec2 import (ec2_instance_types, a_short_time, create_ondemand_instances,
@@ -48,7 +49,7 @@ class AWSProvisioner(AbstractProvisioner):
     def __init__(self, config, batchSystem):
         super(AWSProvisioner, self).__init__(config, batchSystem)
         self.instanceMetaData = get_instance_metadata()
-        self.clusterName = self.instanceMetaData['security-groups']
+        self.clusterName = self._getClusterNameFromTags(self.instanceMetaData)
         self.ctx = self._buildContext(clusterName=self.clusterName)
         self.spotBid = None
         assert config.preemptableNodeType or config.nodeType
@@ -62,6 +63,18 @@ class AWSProvisioner(AbstractProvisioner):
         self.keyName = self.instanceMetaData['public-keys'].keys()[0]
         self.masterPublicKey = self.setSSH()
         self.tags = self._getLeader(self.clusterName).tags
+
+    def _getClusterNameFromTags(self, md):
+        """Retrieve cluster name from current instance tags
+        """
+        instance = self._getClusterInstance(md)
+        return str(instance.tags["Name"])
+
+    def _getClusterInstance(self, md):
+        zone = getCurrentAWSZone()
+        region = Context.availability_zone_re.match(zone).group(1)
+        conn = boto.ec2.connect_to_region(region)
+        return conn.get_all_instances(instance_ids=[md["instance-id"]])[0].instances[0]
 
     @staticmethod
     def retryPredicate(e):
@@ -115,7 +128,7 @@ class AWSProvisioner(AbstractProvisioner):
         logger.info('SSH ready')
         kwargs['tty'] = sys.stdin.isatty()
         command = args if args else ['bash']
-        cls._sshAppliance(leader.ip_address, *command, **kwargs)
+        cls._sshAppliance(leader.public_dns_name, *command, **kwargs)
 
     def _remainingBillingInterval(self, instance):
         return awsRemainingBillingInterval(instance)
@@ -207,7 +220,7 @@ class AWSProvisioner(AbstractProvisioner):
     @classmethod
     def rsyncLeader(cls, clusterName, args, zone=None):
         leader = cls._getLeader(clusterName, zone=zone)
-        cls._rsyncNode(leader.ip_address, args)
+        cls._rsyncNode(leader.public_dns_name, args)
 
     @classmethod
     def _rsyncNode(cls, ip, args, applianceName='toil_leader'):
@@ -266,7 +279,7 @@ class AWSProvisioner(AbstractProvisioner):
     def _waitForNode(cls, instance, role):
         # returns the node's IP
         cls._waitForIP(instance)
-        instanceIP = instance.ip_address
+        instanceIP = instance.public_dns_name
         cls._waitForSSHPort(instanceIP)
         cls._waitForSSHKeys(instanceIP)
         # wait here so docker commands can be used reliably afterwards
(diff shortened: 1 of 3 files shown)


cket:
@chapmanb great, thanks for the PR! Unfortunately the jenkins build is failing with that old setuptools issue:
```
    The required version of setuptools (>=18.5) is not available,
    and can't be installed while this script is running. Please
    install a more recent version first, using
    'easy_install -U setuptools'.
```

chapmanb:
CJ -- ugh, sorry. I guess this setuptools pinning is happening somewhere inside cwltool. @tetron do you know offhand if this is resolved? If so, I'll test and get this PR updated with a more recent cwltool.

chapmanb:
CJ -- I re-thought this and it would make sense to get the fixes for AWS VPC in and then tackle the CWL issues separately. That would help with debugging the CWL failures I'm seeing on Toil runs since I wouldn't have to first patch for all these VPC issues. I dropped the setup.py changes and hopefully the new tests will run cleanly. Thanks much for helping with this.

cket:
@chapmanb it looks like this approach might not work with new account types?
```
boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
<Response><Errors><Error><Code>SecurityGroupLimitExceeded</Code><Message>You have exceeded the number of VPC security groups allowed per instance.</Message></Error></Errors><RequestID>b33cdea3-2dc3-4aa9-8a9d-8c2c8bf1ac2a</RequestID></Response>
```

chapmanb:
CJ;
Sorry about the issues, I pushed a fix that I hope will resolve this. When we were looking up security group IPs within the current VPC (or default) it didn't limit by name of the cluster. I'd guess when running on a system with larger numbers of security groups available this triggered the problem. Hopefully the updated approach will run cleanly now through the tests.

cket:
@chapmanb looks like the fix worked! Have you tried the cluster utilities (ssh-, rsync-, and destroy-cluster) on your account with this branch? I'm curious to see whether they will work with an older account.

The branch also has some conflicts that need to be addressed due to #1446, hopefully they shouldn't be too tricky.

chapmanb:
CJ -- awesome. This does work on my account with all the commands: start, ssh, destroy-cluster. This PR gets me to the point I can start running the bcbio CWL tests. I'm keeping notes on progress here: https://github.com/chapmanb/bcbio-nextgen/tree/master/scripts/ansible#running-with-toil

I rebased against the current master and resolved the merge conflict so hopefully tests will pass cleanly again. Thanks again for the help getting this merged.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -409,13 +425,18 @@ def expectedShutdownErrors(e):
...
             for attempt in retry(timeout=300, predicate=expectedShutdownErrors):
                 with attempt:
-                    try:
-                        ctx.ec2.delete_security_group(name=clusterName)
-                    except BotoServerError as e:
-                        if e.error_code == 'InvalidGroup.NotFound':
-                            pass
-                        else:
-                            raise
+                    sg_torm = None
```
cket: Can you rename this variable to make its purpose a little more clear?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -409,13 +425,18 @@ def expectedShutdownErrors(e):
...
+                    for sg in ctx.ec2.get_all_security_groups():
+                        if sg.name == clusterName:
```
cket: I'm wondering why this is the only check you do on the security group here...

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -637,6 +665,12 @@ def groupNotFound(e):
                 with attempt:
                     # the following authorizes all port access within the web security group
                     web.authorize(ip_protocol='tcp', from_port=0, to_port=65535, src_group=web)
+        out_sgs = []
+        for sg in ctx.ec2.get_all_security_groups():
+            if sg.name == name:
+                if vpc_id is None or sg.vpc_id == vpc_id:
```
cket: ... compared to here?

chapmanb:
CJ -- thanks for the code review. I pushed a fix to check for VPC ID matches during deletion as well and to use camelCase (it's so painful to code python this way). Hopefully that makes everything fit cleanly.

cket:
@chapmanb great, thanks again for the PR! I also prefer snake case but I try to repress it when I'm working on Toil

PR #1513, opened 2017-02-16
CWL: update to latest cwltool

chapmanb:
Updates cwltool and schema-salad to help with debugging fixes for CWL
runs on AWS.

- Handles new status return tuple from single_job_executor
- Updates `adjust*Objs` import path from builder to pathmapper

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,8 @@ def runSetup():
                 'gcs_oauth2_boto_plugin==1.9',
                 botoRequirement],
             'cwl': [
-                'cwltool==1.0.20161221171240']},
+                'cwltool==1.0.20170217172322',
+                'cwltest>=1.0.20170214185319']},
         package_dir={'': 'src'},
         packages=find_packages(where='src',
                                # Note that we intentionally include the top-level `test` package for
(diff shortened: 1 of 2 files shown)


chapmanb:
CJ -- I'm currently stuck with this setuptools error for cwltool (`The required version of setuptools (>=18.5) is not available`) and can't figure out how to work around it. The underlying source of the error is html5lib but I can't seem to replicate this locally as the newer versions of rdflib, which we now explicitly ask for in requirement, no longer need html5lib. So I don't know where the origin of the dependency is and don't have a workflow to fix it since I can't replicate.

Is it possible to add updating setuptools to the build script? This could happen alongside the pip install and at least ensure we have an up to date version of setuptools to skip this check. Thanks for considering this.

cket:
@chapmanb I'll look into it - IIRC we had some problems with newer versions of setuptools so I'll have to track those issues down first

chapmanb:
CJ -- thank you. I finally see what the underlying issue is. cwltool uses ez_setup which forces the 18.5 dependency and the error:

https://github.com/common-workflow-language/cwltool/blob/master/setup.py#L2

I'll submit a PR to cwltool to remove this from the install and just require having setuptools pre-installed. That might be a better workaround if updating here is hard.

chapmanb:
CJ -- brilliant, the latest fixes to cwltool worked and tests now pass. Thanks for the help with this and the VPC PR, looking forward to restarting more AWS testing with all these in place.

PR #1514, opened 2017-02-20
cwltoil: sync path/location on filestore retrieval

chapmanb:
Files retrieved from a filestore reference the staged file in the CWL
object `path` but not in `location` resulting in differing references
to original and staged files depending on how cwltool references the
object. This synchronizes the `location` and `path` to avoid the
confusion.


cket:
Thanks for the fix @chapmanb

PR #1517, opened 2017-02-22
Add notes on launching a local Mesos/Toil cluster

cket:
@chapmanb I have a rough draft of the steps needed to launch a local cluster. Let me know if you have any problems with the setup or any other feedback!


chapmanb:
CJ -- this is so helpful, thank you for putting it together. I'll work on setting up my local machine to be able to test based on this. +1 for including it in the docs.

PR #1521, opened 2017-02-23
ImportFile should return a FileID object

jessebrennan:
Currently jobStore.importFile returns a string which is inconsistent with other methods for uploading a file to the jobStore. This PR instead makes importFile return a FileID object.

This involved adding a getSize attribute to abstractJobStore which returns the file size (in bytes) from the specified URL. This is necessary because FileID has a size attribute.


jessebrennan:
Jenkins test this please.

jessebrennan:
@cket this puppy is ready for review! Do you want me to squash?

PR #1529, opened 2017-02-27
Allow redirecting stdin to a container, several other docker fixes

adderan:
- NonCachingFileStore needs the jobID for compatibiltiy with
  dockerCall
- class Job needs fileStore for comppatibility with deferred job
  functions

These are some features that it would be nice to have so that Cactus can use toil's dockerCall function, and take advantage of deferred jobs.

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -1591,6 +1591,7 @@ class NonCachingFileStore(FileStore):
         self.jobStore = jobStore
         self.jobGraph = jobGraph
         self.jobName = str(self.jobGraph)
+        self.jobID = sha1(self.jobName).hexdigest()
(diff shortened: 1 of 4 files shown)


adderan:
Actually, I didn't realize that the docker pipe functionality existed. This PR might not be necessary.

cket:
@adderan sorry, I didn't see you assigned to the NonCachingFileStore ticket you opened so I fixed it in a separate PR, and I think #1337 fixed the FileStore issue. 

Let me know what you're thinking about the Docker changes in this PR - is the chaining functionality sufficient?

adderan:
Yes, I've switched over to using the pipe chain and that seems to be working well. It looks like both of the other bugs have been fixed now, so I'll close this.

PR #1536, opened 2017-03-01
Fix service deadlock detection (resolves #1509)

### LINKED ISSUE #1509: Service deadlock detection triggers when there is no service deadlock
The only state the service deadlock detector has is a) the service jobs that are issued during the last "potential deadlock" and b) the time that the "potential deadlock" occurred.

But the deadlock detector doesn't reset the "potential deadlock" information after it knows a deadlock didn't occur. Here's the situation I run into:

- Service job A gets issued. While it's starting, (for e.g. 5 seconds) service job A is the only job issued, so the "potential deadlock" code gets triggered.
- A bunch of jobs B get issued that depend on service A. This takes, say, an hour.
- The last job of batch B gets completed.
- A bunch of jobs C get issued.

The issue is that, between batch B completing and batch C getting issued, even though this may take only a fraction of a second, the service deadlock code can run. If it runs at just the right time, it detects that service job A is the only job issued, just like the last "potential deadlock", and terminates the workflow because it's been an hour since the last "potential deadlock".

I have a fix for this, but I want to double-check that I actually understand the purpose of the service deadlock detection. As of right now I don't think it's too useful, because it works solely on the basis of *issued* jobs. Consider the situation where two cores are available, and 2 service jobs and 1 regular job are issued, but only the 2 service jobs are running. No forward progress can be made, but currently the code wouldn't detect this as a "service deadlock". Is that intentional?

joelarmstrong:
The current service deadlock detection doesn't actually detect most (any?) deadlocks, because it works on the basis of issued jobs. Instead it just occasionally crashes the workflow when there aren't any deadlocks, because it doesn't reset when "potential deadlock" situations are gone. It also calls deadlocks when services are still starting up.

This patch:
- Works on the basis of *running*, not issued or starting, jobs
- Resets when deadlock situations are no longer detected

Stylistically, I'm not totally happy with the deep level of if conditions, and the duplicated code on the elses, but it does speed things up since it skips some expensive checks the majority of the time.

Unfortunately it's difficult to create a proper test for this. As a stopgap, here is a script that shows the wrong behavior on the old version and the correct behavior on the new version:

```
python deadlockExample.py jobStore --deadlockWait 60 --maxCores 1
```
This has a deadlock because there is one service job, and one core available. The new version detects the deadlock, and the old version doesn't.

```
python deadlockExample.py jobStore --deadlockWait 10 --maxCores 4
```
The old version incorrectly detects a deadlock, because the service takes longer than 10 seconds to start up. The new version finishes the workflow.

```python
#!/usr/bin/env python2.7
import time
from argparse import ArgumentParser

from toil.job import Job
from toil.common import Toil

class WaiterService(Job.Service):
    def __init__(self, startupTime, shutdownTime):
        super(WaiterService, self).__init__()
        self.startupTime = startupTime
        self.shutdownTime = shutdownTime

    def start(self, job):
        time.sleep(self.startupTime)

    def stop(self, job):
        time.sleep(self.shutdownTime)

    def check(self):
        return True

def start(job):
    job.addService(WaiterService(30, 35))
    job.addChildJobFn(next)

def next(job):
    time.sleep(10)
    job.addChildJobFn(next2)

def next2(job):
    time.sleep(10)

def main():
    parser = ArgumentParser()
    opts = parser.parse_args()
    with Toil(opts) as toil:
        if opts.restart:
            toil.restart()
        else:
            toil.start(Job.wrapJobFn(start))

if __name__ == '__main__':
    main()
```

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -35,6 +35,7 @@ from toil.provisioners.clusterScaler import ClusterScaler
 from toil.serviceManager import ServiceManager
 from toil.statsAndLogging import StatsAndLogging
 from toil.jobGraph import JobNode
+from toil.job import ServiceJobNode
 from toil.toilState import ToilState
 
 logger = logging.getLogger( __name__ )
@@ -488,28 +489,29 @@ class Leader:
         """
         Checks if the system is deadlocked running service jobs.
         """
-        # If there are no updated jobs and at least some jobs issued
-        if len(self.toilState.updatedJobs) == 0 and self.getNumberOfJobsIssued() > 0:
+        totalRunningJobs = len(self.batchSystem.getRunningBatchJobIDs())
+        totalServicesIssued = self.serviceJobsIssued + self.preemptableServiceJobsIssued
+        # If there are no updated jobs and at least some jobs running
+        if totalServicesIssued >= totalRunningJobs and len(self.toilState.updatedJobs) == 0 and totalRunningJobs > 0:
+            serviceJobs = filter(lambda x : isinstance(x, ServiceJobNode), self.jobBatchSystemIDToIssuedJob.values())
+            runningServiceJobs = set(filter(lambda x : self.serviceManager.isRunning(x), serviceJobs))
 
-            # If all scheduled jobs are services
-            assert self.serviceJobsIssued + self.preemptableServiceJobsIssued <= self.getNumberOfJobsIssued()
-            if self.serviceJobsIssued + self.preemptableServiceJobsIssued == self.getNumberOfJobsIssued():
-
-                # Sanity check that all issued jobs are actually services
-                for jobNode in self.jobBatchSystemIDToIssuedJob.values():
-                    assert jobNode.jobStoreID in self.toilState.serviceJobStoreIDToPredecessorJob
-
-                # An active service job is one that is not in the process of terminating
-                activeServiceJobs = filter(lambda x : self.serviceManager.isActive(x), self.jobBatchSystemIDToIssuedJob.values())
-
-                # If all the service jobs are active then we have a potential deadlock
-                if len(activeServiceJobs) == len(self.jobBatchSystemIDToIssuedJob):
-                    # We wait self.config.deadlockWait seconds before declaring the system deadlocked
-                    if self.potentialDeadlockedJobs != activeServiceJobs:
-                        self.potentialDeadlockedJobs = activeServiceJobs
-                        self.potentialDeadlockTime = time.time()
-                    elif time.time() - self.potentialDeadlockTime >= self.config.deadlockWait:
-                        raise DeadlockException("The system is service deadlocked - all issued jobs %s are active services" % self.getNumberOfJobsIssued())
+            # If all the running jobs are active services then we have a potential deadlock
+            if len(runningServiceJobs) == totalRunningJobs:
+                # We wait self.config.deadlockWait seconds before declaring the system deadlocked
+                if self.potentialDeadlockedJobs != runningServiceJobs:
+                    self.potentialDeadlockedJobs = runningServiceJobs
+                    self.potentialDeadlockTime = time.time()
+                elif time.time() - self.potentialDeadlockTime >= self.config.deadlockWait:
+                    raise DeadlockException("The system is service deadlocked - all %d running jobs are active services" % totalRunningJobs)
+            else:
+                # We have observed non-service jobs running, so reset the potential deadlock
+                self.potentialDeadlockedJobs = []
+                self.potentialDeadlockTime = 0
+        else:
+            # We have observed non-service jobs running, so reset the potential deadlock
+            self.potentialDeadlockedJobs = []
+            self.potentialDeadlockTime = 0
 
 
     def issueJob(self, jobNode):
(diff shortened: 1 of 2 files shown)


src/toil/leader.py
```diff
@@ -488,28 +489,29 @@ def checkForDeadlocks(self):
...
-            # If all scheduled jobs are services
-            assert self.serviceJobsIssued + self.preemptableServiceJobsIssued <= self.getNumberOfJobsIssued()
-            if self.serviceJobsIssued + self.preemptableServiceJobsIssued == self.getNumberOfJobsIssued():
-
-                # Sanity check that all issued jobs are actually services
-                for jobNode in self.jobBatchSystemIDToIssuedJob.values():
-                    assert jobNode.jobStoreID in self.toilState.serviceJobStoreIDToPredecessorJob
-
-                # An active service job is one that is not in the process of terminating
-                activeServiceJobs = filter(lambda x : self.serviceManager.isActive(x), self.jobBatchSystemIDToIssuedJob.values())
```
cket: Just for clarification, this is where the problems were arising right? Not terminating isn't a strong enough condition, we also need to check if the jobs are actually running

src/toil/leader.py
```diff
@@ -488,28 +489,29 @@ def checkForDeadlocks(self):
...
-                    elif time.time() - self.potentialDeadlockTime >= self.config.deadlockWait:
-                        raise DeadlockException("The system is service deadlocked - all issued jobs %s are active services" % self.getNumberOfJobsIssued())
+        totalRunningJobs = len(self.batchSystem.getRunningBatchJobIDs())
+        totalServicesIssued = self.serviceJobsIssued + self.preemptableServiceJobsIssued
+        # If there are no updated jobs and at least some jobs running
+        if totalServicesIssued >= totalRunningJobs and len(self.toilState.updatedJobs) == 0 and totalRunningJobs > 0:
+            serviceJobs = filter(lambda x : isinstance(x, ServiceJobNode), self.jobBatchSystemIDToIssuedJob.values())
+            runningServiceJobs = set(filter(lambda x : self.serviceManager.isRunning(x), serviceJobs))
+
+            # If all the running jobs are active services then we have a potential deadlock
```
cket: Can we add a sanity check here for ``len(runningServiceJobs) <= totalRunningJobs``?

src/toil/leader.py
joelarmstrong: Yep, that is where the services that were still starting counted toward a potential deadlock. So often any service that took longer than `deadlockWait` seconds to start up would cause an exception to be thrown.

There were basically three separate problems, 1) not resetting "potential deadlock" information, 2) counting issued and non-terminated services and 3) missing deadlocks because a single issued non-service job would stop a potential deadlock, even if that job couldn't possibly run because all the cores were occupied by services. So this is where 2) arose.

src/toil/leader.py
```diff
@@ -488,28 +489,29 @@ def checkForDeadlocks(self):
...
+            if len(runningServiceJobs) == totalRunningJobs:
+                # We wait self.config.deadlockWait seconds before declaring the system deadlocked
+                if self.potentialDeadlockedJobs != runningServiceJobs:
+                    self.potentialDeadlockedJobs = runningServiceJobs
+                    self.potentialDeadlockTime = time.time()
+                elif time.time() - self.potentialDeadlockTime >= self.config.deadlockWait:
+                    raise DeadlockException("The system is service deadlocked - all %d running jobs are active services" % totalRunningJobs)
+            else:
+                # We have observed non-service jobs running, so reset the potential deadlock
+                self.potentialDeadlockedJobs = []
```
joelarmstrong: Yep, you're right, it's cleaner if it's a set.

joelarmstrong:
I can address these, thanks for the helpful feedback. Would you prefer an additional commit, or a single squashed commit? I'm OK with either.

cket:
I prefer an additional commit if it's the same to you

joelarmstrong:
Hmm, not sure the test failure could be related. Has that test been flaky lately?

cket:
@joelarmstrong no, all the tests have been pretty solid actually

joelarmstrong:
I think this failure (and all other jenkins-it failures on the subsequent pull requests) is because we ran out of EBS space. Jenkins went a bit apeshit making about a thousand 50 GB snapshots on March 2nd:
screen shot 2017-03-05 at 00 43 29-fullpage
And now we haven't been able to launch spot instances (any instances?!?!?!?) since the 2nd because of EBS limits. (The root volume is technically an EBS volume.) So the test failure is completely unrelated--but someone may want to check out why Jenkins did that. I'm deleting all the snapshots I think are suspect.

joelarmstrong:
Hmm, maybe it's actually just a generic Toil bug (caused by https://github.com/BD2KGenomics/toil/commit/91782051244d91c8eb0f7ffddc44738845cf4ca9?). I assumed it was Jenkins because I saw it in one of the Owner tags, but I just saw one of my own instances in the Owner tag, as well. Maybe the root volumes aren't being cleaned up properly.

cket:
@joelarmstrong that explains a lot, thanks for bringing that up! I opened #1545 to follow up on the cause of this issue.

cket:
Jenkins, test this please

cket:
@joelarmstrong great, thanks again for the PR

PR #1539, opened 2017-03-02
Support shared temp directory in distributed system (resolves #1462)

### LINKED ISSUE #1462: support shared temp directory in distributed system
Related to #1397

workers will attempt to kill pids belonging to jobs other workers if the same directory onthe same file system is used as TMPDIR across multiple nodes. This can be circumvented by adding a `machine-id` check to the process.

jessebrennan:
This PR adds two mutually exclusive command line options:
- `--nodeIDFile <file>`  and
- `--nodeIDCommand <command>`.

`--nodeIDFile` indicates a file that contains a unique nodeID while `--nodeIDCommand` indicates a commands that can be run on a node to consistently get a nodeID.

Giving a node an ID will force the node to check that any jobs it may delete are jobs that it had originally created. This becomes necessary when the temp/work directory is shared among nodes. If neither of these two options are specified, then Toil assumes that the temp directory is not shared and will delete any job that it is not currently running.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -20,6 +20,7 @@ import re
 import sys
 import tempfile
 import time
+import subprocess
 from argparse import ArgumentParser
 from threading import Thread
 
@@ -109,6 +110,10 @@ class Config(object):
         self.retryCount = 0
         self.maxJobDuration = sys.maxint
         self.rescueJobsFrequency = 3600
+
+        # Node ID
+        self.nodeIDCommand = None
+        self.nodeIDFile = None
 
         #Misc
         self.disableCaching = False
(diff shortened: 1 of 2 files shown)


jessebrennan:
Jenkins test this please.

jessebrennan:
Jenkins test this please

src/toil/common.py
```diff
@@ -1107,3 +1142,28 @@ def getFileSystemSize(dirPath):
...
+
+    :param str nodeIDFile: absolute path to a file with nodeID
+    :param str nodeIDCommand: a command to be run which returns the nodeID
+    :return: a unique strin
+    :rtype: str
+    """
+    if nodeIDCommand is not None:
+        try:
+            output = subprocess.check_output(nodeIDCommand.split())
+        except subprocess.CalledProcessError as e:
```
jessebrennan: Because we need to catch the output and save the uuid. check_call only gives you the return value while check_output give the output. I was a little unsure about throwing this exception here, but I could not think of a more elegant way to exit.

jessebrennan:
@cket Lets talk about a test soon so we can merge this puppy. I don't remember where we left off.

jessebrennan:
@ejacox I'm pretty sure this was working when I left off and just needed a test. Now it is superseded by #1573. I propose we close.

PR #1542, opened 2017-03-03
Fix incorrect memory usage and "wait" stats reporting (resolves #1541)

### LINKED ISSUE #1541: Memory usage stats are incorrect
I just discovered a copy-paste bug in jobTree's stats this morning, and lo and behold, it's still in toil as well.

https://github.com/BD2KGenomics/toil/blob/43484f818126155f20dfe4f7b173f0e97c19d457/src/toil/lib/bioio.py#L179

```python
    me = resource.getrusage(resource.RUSAGE_SELF)
    childs = resource.getrusage(resource.RUSAGE_CHILDREN)
    totalCPUTime = me.ru_utime+me.ru_stime+childs.ru_utime+childs.ru_stime
    totalMemoryUsage = me.ru_maxrss+ me.ru_maxrss
```

That last line should obviously read `me.ru_maxrss + childs.ru_maxrss`. The stats basically reflected twice the worker python process RSS, rather than the sum of its RSS and the RSS of any terminated child processes. I'm testing my fix right now, to make sure this actually does vaguely what is advertised. Will make a PR in a bit.

joelarmstrong:
The reported "wait" time (wall-clock time - CPU time) was also almost always negative, because it was calculating CPU time - wall-clock time instead.

NB: the memory usage is still not perfect, because the only data available for the children is the maximum size of the single largest child, not (as previously implied by the docstring) the entire process tree:
>    ru_maxrss (since Linux 2.6.32)
>          This is the maximum resident set size used (in kilobytes).  For RUSAGE_CHILDREN, this is the resident set size of the largest child, not the maximum resident set size of the process tree.

Unfortunately, it's significantly more difficult to portably get the maximum RSS for the entire process tree. We would probably need to spawn a monitoring thread if we wanted to be fully accurate. But this should be correct in the vast majority of cases (it will only be wrong if people spawn two or more children at the same time), and it's certainly an improvement over the old behavior.

### DIFF
--- a/src/toil/lib/bioio.py
+++ b/src/toil/lib/bioio.py
@@ -171,12 +171,14 @@ def system(command):
     subprocess.check_call(command, shell=isinstance(command, string_types), bufsize=-1)
 
 def getTotalCpuTimeAndMemoryUsage():
-    """Gives the total cpu time and memory usage of itself and its children.
+    """
+    Gives the total cpu time of itself and all its children, and the maximum RSS memory usage of
+    itself and its single largest child.
     """
     me = resource.getrusage(resource.RUSAGE_SELF)
     childs = resource.getrusage(resource.RUSAGE_CHILDREN)
-    totalCPUTime = me.ru_utime+me.ru_stime+childs.ru_utime+childs.ru_stime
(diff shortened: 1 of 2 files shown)


cket:
Jenkins, test this please

PR #1547, opened 2017-03-08
Improve cloud documentation

cket:
Jenkins, skip this please


docs/install/cloud.rst
```diff
@@ -5,15 +5,84 @@
...
+    $ screen
+
+Simply running ``screen`` within ``script`` will get things working properly again.
+
+Finally, you can execute remote commands with the following syntax::
+
+    $ toil ssh-cluster remoteCommand CLUSTER-NAME-HERE
+
+It is not advised that you run your Toil workflow using remote execution like this
+unless `nohup <https://linux.die.net/man/1/nohup>`_ is used to insure the process
```
cket: I didn't want to recommend `screen` in this situation since it requires the ``script`` workaround. It is misleading to imply that only `nohup` would work though, I'll fix that

cket:
@joelarmstrong all good points, thanks for the feeback!

PR #1556, opened 2017-03-11
CWL: stage docker files inside Toil work directory

chapmanb:
When running cwltoil into the Toil Docker appliance, as on AWS, ensure
that the jobs spawned by CWL use directories relative to the Toil work
directory. This allows docker-in-docker file mapping to work correctly
since the default `/var` directories are not exposed. Thanks to @cket

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -712,7 +712,11 @@ def main(args=None, stdout=sys.stdout):
             outobj = toil.restart()
         else:
             basedir = os.path.dirname(os.path.abspath(options.cwljob or options.cwltool))
-            builder = t._init_job(job, basedir=basedir, use_container=use_container)
+            # ensure docker containers run inside the base workdir
+            docker_kwargs = {"docker_outdir": os.path.join(basedir, "var", "spool", "cwl"),
+                             "docker_stagedir": os.path.join(basedir, "var", "lib", "cwl"),
+                             "docker_tmpdir": os.path.join(basedir, "tmp")}
+            builder = t._init_job(job, basedir=basedir, use_container=use_container, **docker_kwargs)
             (wf1, wf2) = makeJob(t, {}, use_container=use_container, preserve_environment=options.preserve_environment, tmpdir=os.path.realpath(outdir))
             try:
                 if isinstance(wf1, CWLWorkflow):
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -267,9 +267,11 @@ class CWLJob(Job):
         inpdir = os.path.join(fileStore.getLocalTempDir(), "inp")
         outdir = os.path.join(fileStore.getLocalTempDir(), "out")
         tmpdir = os.path.join(fileStore.getLocalTempDir(), "tmp")
+        stagedir = os.path.join(fileStore.getLocalTempDir(), "stage")
         os.mkdir(inpdir)
         os.mkdir(outdir)
         os.mkdir(tmpdir)
+        os.mkdir(stagedir)
 
         # Copy input files out of the global file store, ensure path/location synchronized
         index = {}
@@ -284,10 +286,15 @@ class CWLJob(Job):
                                                             basedir=os.getcwd(),
                                                             outdir=outdir,
                                                             tmpdir=tmpdir,
+                                                            stagedir=stagedir,
+                                                            docker_tmpdir=tmpdir,
+                                                            docker_outdir=outdir,
+                                                            docker_stagedir=stagedir,
                                                             tmpdir_prefix="tmp",
                                                             **opts)
         if status != "success":
             raise cwltool.errors.WorkflowException(status)
+
         cwltool.pathmapper.adjustDirObjs(output, locToPath)
         cwltool.pathmapper.adjustFileObjs(output, locToPath)
         cwltool.pathmapper.adjustFileObjs(output, functools.partial(computeFileChecksums,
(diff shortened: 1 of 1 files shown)


cket:
@chapmanb if I'm understanding this change correctly it will require the user to set ``basedir`` to the same value as Toil's workdir, right? I would prefer if we could avoid that, if possible. I think the right approach would be to have jobs determine the proper location to write to via ``Job.FileStore.getLocalTempDir()``, which returns a per-job temp directory.

Let me know if you have any questions, thanks for taking this on!

chapmanb:
Cj;
Thanks for the suggestion. I'm not sure how to implement this, or get access to the mounted directories Toil wants to run inside the Docker appliance automatically. At the point in the code we're needing to access these we're still setting up to run jobs so we don't have access to a configured filestore to call `getLocalTempDir()`.

Inside the run, I think we're already calling what you're asking about:

https://github.com/BD2KGenomics/toil/blob/master/src/toil/cwl/cwltoil.py#L267

but I'm honestly just swinging wildly in the dark here and am not even sure what directory we should be including. I thought it was the base directory but am probably off. If you have a better way to implement/test I'm happy to go with whatever works.

cket:
@chapmanb I'm not sure what the right approach here is either. The goal is to write all files to the Toil workdir but that location is determined on file store instantiation and is specific to each worker. 

It looks like many of the paths in CWLToil are set before launching the Toil run - is that correct? Are those paths assumed to exist on all workers as well? 

What do you think about leaving those fields (fs_access, basedir) unset until Job.run, where the fields can be filled in appropriately with paths from `getLocalTempDir`?

chapmanb:
Cj;
Thanks for the suggestion. I've been trying to implement this and seem to be moving in the right direction but got stuck again. I added in this and can get the Docker runs to stage inside the work directory, but then run into the issue that Docker creates root owned files and the worker fails trying to clean them up from externally after the job completes:
```
kunkel 2017-03-15 13:48:57,893 MainThread INFO toil.leader: Issued job 'file:///home/chapmanb/bio/toil/src/toil/test/cwl/revtool.cwl' rev M/v/job5CSX04 with job batch system ID: 1 and cores: 1, disk: 2.0 G, and memory: 1.0 G
  File "/home/chapmanb/bio/toil/venv/bin/_toil_worker", line 11, in <module>
    load_entry_point('toil==3.7.0a1', 'console_scripts', '_toil_worker')()
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/site-packages/toil-3.7.0a1-py2.7.egg/toil/worker.py", line 558, in main
    shutil.rmtree(localWorkerTempDir)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 256, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/home/chapmanb/bio/toil/venv/lib/python2.7/shutil.py", line 254, in rmtree
    os.rmdir(path)
OSError: [Errno 13] Permission denied: '/tmp/user/1001/toil-bac1856b-ba75-410a-a4ba-a319535cadb4/tmpcnx1Mz/3aff26a1-864c-41b8-8f81-9412c70a44b4/tSDKh5x/tmp/user/1001/toil-bac1856b-ba75-410a-a4ba-a319535cadb4/tmpcnx1Mz/3aff26a1-864c-41b8-8f81-9412c70a44b4/tqqOgNL/out'
```
Have you seen this problem and have an idea how to best work around it? I'm stuck on that and also confused why the staging/output directories get double nested (`/tmp/user/1001/toil-bac1856b-ba75-410a-a4ba-a319535cadb4/tmpcnx1Mz/3aff26a1-864c-41b8-8f81-9412c70a44b` is twice in the output directory:
```
/tmp/user/1001/toil-bac1856b-ba75-410a-a4ba-a319535cadb4/tmpcnx1Mz/3aff26a1-864c-41b8-8f81-9412c70a44b4/tSDKh5x/tmp/user/1001/toil-bac1856b-ba75-410a-a4ba-a319535cadb4/tmpcnx1Mz/3aff26a1-864c-41b8-8f81-9412c70a44b4/tqqOgNL/out'
```
This PR will fail because of this issue but I wanted to get more eyes on it in case you had any idea how best to handle the cleanup issue, or an idea about the annoyingly nested directories. Thanks again for all the help working on this.

cket:
@chapmanb I'm not sure about the nested directories, but Toil's dockerCall function does have a work around for the root-owned output files: https://github.com/BD2KGenomics/toil/blob/master/src/toil/lib/docker.py#L216. How is cwlToil calling out to docker? Does CWL Tool do that for us?

chapmanb:
Cj;
cwltool does all the Docker runs so doesn't use `dockerCall` explicitly. Thanks for the pointer to the useful chmod functionality. I tried manually patching the cwltoil job run to call this and although it gets run we still run into the same error. I know this last commit doesn't help resolve anything but I thought I'd post in case you spotted any obvious mistakes I made or have other ideas how to fix this.

If we're totally stumped, would you be open to making the toil appliance correctly mount the `/var` directories cwltool uses by default so we can skip over trying to dig into this more? Just trying to think of workarounds to get this working. Thanks much.

cket:
@chapmanb thats a good idea, I opened #1580 so we can keep testing and #1579 + #1581 to track the current issues we are facing

chapmanb:
CJs fixes resolved this issue so closing the PR, apologies for leaving it hanging.

ejacox:
Thank you for closing it @chapmanb

PR #1560, opened 2017-03-14
Improve efficiency of Mesos batch system

### LINKED ISSUE #1531: Mesos batchSystem is O(n log n) in the number of job types per request per offer batch
The mesos BatchSystem in Toil sorts its queued job "types" (i.e. resource requirements) before going through the received offers and assigning jobs to be issued. This means responding to an offer batch is O(n log n) in the number of job types, if I'm reading the method correctly. Under normal conditions this is totally fine, but when resource requirements are dynamically assigned, say based on a multiple of the input file size, there can be tens of thousands of very slightly different job types, causing the leader to slow to a crawl and the workers to be under-utilized.

This is pretty easy to solve from our end: we created a thin layer over Toil.Job that rounds all resource requirements up to the next 100MB interval. So our problem is solved, but in the interest of making sure this doesn't catch anyone else by surprise, here are a few options off the top of my head, any/all/none of these could be applied:

- Document this catch in the docs for the Mesos batch system
- Apply some rounding of resource requirements within toil itself
- (way more work) Keep the queued job types in sorted order in the first place, and use binary search to find the first job type that would fit into each offer. Maybe this would require multiple sorted lists?

### LINKED ISSUE #1554: Update Toil Appliance Docker client to 1.12.3
And update Toil Appliance dev documentation with Docker version caveats.

cket:
Resolves #1531 
Resolves #1554

### DIFF
--- a/src/toil/batchSystems/mesos/__init__.py
+++ b/src/toil/batchSystems/mesos/__init__.py
@@ -17,6 +17,7 @@ from Queue import Queue
 from collections import namedtuple
 from functools import total_ordering
 from bisect import bisect
+from threading import Lock
 
 TaskData = namedtuple('TaskData', (
     # Time when the task was started
@@ -38,31 +39,35 @@ class JobQueue(object):
         self.queues = {}
         # list of jobTypes in decreasing resource expense
         self.sortedTypes = []
+        self.jobLock = Lock()
 
     def insertJob(self, job, jobType):
-        # logn insertion
-        if jobType not in self.queues:
-            index = bisect(self.sortedTypes, jobType)
-            self.sortedTypes.insert(index, jobType)
-            self.queues[jobType] = Queue()
-
-        self.queues[jobType].put(job)
+        with self.jobLock:
+            if jobType not in self.queues:
+                index = bisect(self.sortedTypes, jobType)
+                self.sortedTypes.insert(index, jobType)
+                self.queues[jobType] = Queue()
+            self.queues[jobType].put(job)
 
     def sorted(self):
-        return self.sortedTypes
+        return list(self.sortedTypes)
 
     def jobIDs(self):
-        # O(N)
         return [job.jobID for queue in self.queues.values() for job in list(queue.queue)]
 
     def nextJobOfType(self, jobType):
-        # constant lookup, pop off first element
-        # also remove from taskMap
-        job = self.queues[jobType].get(block=False)
-        return job
+        with self.jobLock:
+            job = self.queues[jobType].get(block=False)
+            if self.queues[jobType].empty():
+                del self.queues[jobType]
+                self.sortedTypes.remove(jobType)
+            return job
 
     def typeEmpty(self, jobType):
-        return self.queues.get(jobType, Queue()).empty()
+        # without a lock we could get a false negative from this method
+        # if it were called while nextJobOfType was executing
+        with self.jobLock:
+            return self.queues.get(jobType, Queue()).empty()
 
 
 @total_ordering
--- a/src/toil/batchSystems/mesos/__init__.py
+++ b/src/toil/batchSystems/mesos/__init__.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import absolute_import
+
+from Queue import Queue
 from collections import namedtuple
 from functools import total_ordering
 from bisect import bisect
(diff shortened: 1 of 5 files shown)


cket:
Jenkins, test this please

cket:
@joelarmstrong despite the test failure (Docker pull issue) this is ready to test if you're interested!
Jenkins, test this please

joelarmstrong:
Yes, definitely very interested in testing this! Thanks for the quick turnaround. I'll try this out on my next test run (probably in a day or so).

src/toil/batchSystems/mesos/__init__.py
```diff
@@ -27,22 +31,110 @@
...
+        return self.sortedTypes
+
+    def jobIDs(self):
+        # O(N)
+        return [job.jobID for queue in self.queues.values() for job in list(queue.queue)]
+
+    def nextJobOfType(self, jobType):
+        # constant lookup, pop off first element
+        # also remove from taskMap
+        job = self.queues[jobType].get(block=False)
```
joelarmstrong: Unless I missed something, this will end up leaking jobTypes/Queues (keeping them around in `sortedTypes` and `queues`) even after all the jobs of a given type have been run. The leaked memory isn't really an issue, but the leaked entries in `sortedTypes` will cause a slowdown. The scheduler will have to iterate over each empty jobType. If there are a ton of different jobTypes, most of which only have one or two jobs associated with them, the leader will get slower and slower as time goes on.

Fixing this would be easy--just remove the jobType once its queue runs out--but would probably require a lock on the list & hash to avoid losing jobs.

src/toil/batchSystems/mesos/__init__.py
```diff
@@ -27,22 +31,110 @@
...
+        # logn insertion
+        if jobType not in self.queues:
+            index = bisect(self.sortedTypes, jobType)
+            self.sortedTypes.insert(index, jobType)
+            self.queues[jobType] = Queue()
+
+        self.queues[jobType].put(job)
+
+    def sorted(self):
+        return self.sortedTypes
```
joelarmstrong: `sortedTypes` can be modified by the leader thread while it's being iterated by the scheduler. I think this may not actually cause any problems right now (worst case a jobType is skipped temporarily or considered twice), but it might be nice to make this future-proof by doing `return list(self.sortedTypes)`, or at least noting the possibility of simultaneous modification in a comment in case this causes problems later.

src/toil/batchSystems/mesos/__init__.py
```diff
@@ -27,22 +31,110 @@
...
+class JobQueue(object):
+
+    def __init__(self):
+        # mapping of jobTypes to queues of jobs of that type
+        self.queues = {}
+        # list of jobTypes in decreasing resource expense
+        self.sortedTypes = []
+
+    def insertJob(self, job, jobType):
+        # logn insertion
```
joelarmstrong: Nitpick: I think this is O(n), because list insertion is worst-case O(n) (the elements after the insertion need to be shifted). That said, I think it's probably not worth using a tree or something to get O(log n), O(n) is most likely totally fine, because shifting the elements is insanely fast. I doubt the asymptotics really matter with less than millions of job types.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -21,6 +21,7 @@
 import pwd
 import socket
 import time
+from bisect import bisect
 from collections import defaultdict
 from operator import attrgetter
```
joelarmstrong: Additional nitpick: these three imports (bisect, defaultdict, attrgetter) aren't used anywhere in this file according to pyflakes. I think the part that used defaultdict was removed in this PR, and I assume the others are from a first draft of the new code.

joelarmstrong:
PS: just started to test this, and after a few hours this seems to be working really well. I saw a bunch of leader CPU being used (though not totally CPU-bound) keeping ~700 cores occupied using the old scheduler, and with the new one, it's using at least an order of magnitude less. It's not a perfect comparison, the number of total jobs is slightly different, but anecdotally it looks good.

Also, this PR fixed a bug I just noticed with the old Mesos scheduler, where preemptable jobs were being preferred over non-preemptable jobs by the old scheduler.

cket:
Jenkins, test this please

cket:
jenkins, test this please

cket:
jenkins, test this please

cket:
jjenkins, test this please

cket:
Jenkins, test this please

cket:
@joelarmstrong thanks for the suggestions, I implemented them all & I think this is ready to merge! Do you want to take another pass before I do?

cket:
Also, the repeated testing was to debug the new jenkins server, not this PR

joelarmstrong:
Great! Sure, I'll take a quick look, should be done within an hour or so.

Makes sense re: jenkins, I noticed jenkins-it had some issues, and figured it all had to do with the AWS account moves.

joelarmstrong:
Nice, LGTM!

Issue #1561, opened 2017-03-14
Set node count can throw AttributeError 


cket:
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/987/testReport/junit/src.toil.test.provisioners.aws.awsProvisionerTest/AWSAutoscaleTest/testSpotAutoScale/

diekhans:
That doesn't mean it isn't still a bug.

cricketsloan:
Agreed that it could still be a bug.  But given that large chunks of the code are rewritten, it would need to be reproduced. Do you have evidence that it still exists?

diekhans:
I don't necessary disagree with closing, only the justification
that it is a year stale from someone who is gone.

I don't know if the related code has been modified or not.
Suggest just ask Lon if he suspects the code has been change and
the problem would have been fixed.

Given there is no test case with the bug, `appears to be no
longer a problem' is great by me!  If it is a bug, it will come
back.

PR #1566, opened 2017-03-14
Retry Azure importFile, improve Boto throttlePredicate 

### LINKED ISSUE #1564: Retry on Azure importFile timeout

cket:
Connected to #1561 
Resolves #1564

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -278,9 +278,11 @@ class AzureJobStore(AbstractJobStore):
     @classmethod
     def _readFromUrl(cls, url, writable):
         blob = cls._parseWasbUrl(url)
-        blob.service.get_blob_to_file(container_name=blob.container,
-                                      blob_name=blob.name,
-                                      stream=writable)
+        for attempt in retry_azure():
+            with attempt:
+                blob.service.get_blob_to_file(container_name=blob.container,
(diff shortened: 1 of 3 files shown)


src/toil/jobStores/azureJobStore.py
```diff
@@ -278,9 +278,11 @@ def getSize(cls, url):
     @classmethod
     def _readFromUrl(cls, url, writable):
         blob = cls._parseWasbUrl(url)
-        blob.service.get_blob_to_file(container_name=blob.container,
-                                      blob_name=blob.name,
-                                      stream=writable)
+        for attempt in retry_azure():
+            with attempt:
+                blob.service.get_blob_to_file(container_name=blob.container,
```
joelarmstrong: There's no good solution to this, but just want to point out that this might rarely end up writing a corrupted file with the same part represented two or more times, if you get something like ECONNRESET in the middle of a stream. I can't think of any good way around that, so I think sticking with this behavior is OK. AWSJobStore seems to do something very similar in `FileInfo.downloadStream`, and I assume it hasn't really caused problems, so it's probably only a theoretical problem.

Otherwise, LGTM!

cket: Thats a good point. I'm also not aware of a better solution for that behavior but it's a good thing to keep in mind if we do start seeing problems.

Issue #1397, opened 2016-12-16
OSError: [Errno 1] Operation not permitted


evan-wehi:
Hello,

I received the following stack trace:

```
w/W/jobG9NqPx    Traceback (most recent call last):
w/W/jobG9NqPx      File "/home/thomas.e/home/dev/toil/src/toil/worker.py", line 335, in main
w/W/jobG9NqPx        with fileStore.open(job):
w/W/jobG9NqPx      File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
w/W/jobG9NqPx        return self.gen.next()
w/W/jobG9NqPx      File "/home/thomas.e/home/dev/toil/src/toil/fileStore.py", line 1581, in open
w/W/jobG9NqPx        self.findAndHandleDeadJobs(self.workFlowDir)
w/W/jobG9NqPx      File "/home/thomas.e/home/dev/toil/src/toil/fileStore.py", line 1699, in findAndHandleDeadJobs
w/W/jobG9NqPx        if not cls._pidExists(jobState['jobPID']):
w/W/jobG9NqPx      File "/home/thomas.e/home/dev/toil/src/toil/fileStore.py", line 397, in _pidExists
w/W/jobG9NqPx        os.kill(pid, 0)
w/W/jobG9NqPx    OSError: [Errno 1] Operation not permitted
```

Looking at the code, it seems it is trying to test if the PID exists. Perhaps the PID has been reused (not sure how quickly that might happen).  It seems that perhaps this exception should be caught as well.

This is from the current master, not an official release.

Thanks,
Evan.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-122)
┆Issue Number: TOIL-122

cket:
@evan-wehi what OS did this happen on? I think you're right, that exception should probably be caught too

evan-wehi:
centOS version 7.5 (I think)

uname -a
Linux torquelord.hpc.wehi.edu.au<http://torquelord.hpc.wehi.edu.au> 3.10.0-327.28.3.el7.x86_64 #1 SMP Thu Aug 18 19:05:49 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux



--
Evan Thomas
Research Computing Scientist
Computational Biology
The Walter and Eliza Hall Institute of Medical Research
[...]

PR #1573, opened 2017-03-16
Torque and workDir fixes (resolves #1462, #1628)

### LINKED ISSUE #1462: support shared temp directory in distributed system

andreyto:
Proposing two sets of changes. These were caused by not being able to execute `toil-sort-example.py` from the Toil manual on our cluster.

Our cluster:

- runs Torque. The `qsub` job scripts need to use `/bin/sh` instead of `/bin/bash`, otherwise `-V` variables get zapped by BASH profiles sourced when a job starts. This is the same as [this](https://github.com/roryk/ipython-cluster-helper/commit/33c9eb1fdeb5816c87a5c0b685291673ac685cad)

- has a tiny `/tmp` file system on each node (2GB each), and no other local disks that we can use. This is a very common setup on the clusters. Because of that, we need to use a shared FS for all scratch work, including as a `--workDir` when running Toil. This results in workers on different nodes erasing each other's live job directories in a call to `fileStore.findAndHandleDeadJobs()`.

With the proposed changes, `toil-sort-example.py` runs under Torque with `--workDir` on a shared FS.

More details are in the commit messages.

Thanks,
Andrey

### DIFF
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -20,6 +20,7 @@ import subprocess
 import time
 import math
 import sys
+import shlex
 import xml.etree.ElementTree as ET
 import tempfile
 
@@ -138,6 +139,15 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             
             if reqline:
                 qsubline += ['-l',','.join(reqline)]
+            
+            # All other qsub parameters can be passed through the environment (see man qsub).
+            # No attempt is made to parse them out here and check that they do not conflict
+            # with those that we already constructed above
+            arglineEnv = os.getenv('TOIL_TORQUE_ARGS')
+            if arglineEnv is not None:
+                logger.debug("Additional Torque arguments appended to qsub from "\
+                        "TOIL_TORQUE_ARGS env. variable: {}".format(arglineEnv))
+                qsubline += shlex.split(arglineEnv)
 
             # "Native extensions" for TORQUE (see DRMAA or SAGA)
             nativeConfig = os.getenv('TOIL_TORQUE_ARGS')
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -134,6 +134,8 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             if reqlineEnv is not None:
                 logger.debug("Additional Torque resource requirements appended to qsub from "\
                         "TOIL_TORQUE_REQS env. variable: {}".format(reqlineEnv))
+                if ("mem=" in reqlineEnv) or ("nodes=" in reqlineEnv) or ("ppn=" in reqlineEnv):
+                    raise ValueError("Incompatible resource arguments ('mem=', 'nodes=', 'ppn='): {}".format(reqlineEnv))
 
                 reqline.append(reqlineEnv)
             
@@ -145,18 +147,12 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             # with those that we already constructed above
             arglineEnv = os.getenv('TOIL_TORQUE_ARGS')
             if arglineEnv is not None:
-                logger.debug("Additional Torque arguments appended to qsub from "\
-                        "TOIL_TORQUE_ARGS env. variable: {}".format(arglineEnv))
+                logger.debug("Native Torque options appended to qsub from TOIL_TORQUE_ARGS env. variable: {}".\
+                        format(arglineEnv))
+                if ("mem=" in arglineEnv) or ("nodes=" in arglineEnv) or ("ppn=" in arglineEnv):
+                    raise ValueError("Incompatible resource arguments ('mem=', 'nodes=', 'ppn='): {}".format(arglineEnv))
                 qsubline += shlex.split(arglineEnv)
 
-            # "Native extensions" for TORQUE (see DRMAA or SAGA)
-            nativeConfig = os.getenv('TOIL_TORQUE_ARGS')
-            if nativeConfig is not None:
-                logger.debug("Native TORQUE options appended to qsub from TOIL_TORQUE_RESOURCES env. variable: {}".format(nativeConfig))
-                if ("mem=" in nativeConfig) or ("nodes=" in nativeConfig) or ("ppn=" in nativeConfig):
-                    raise ValueError("Incompatible resource arguments ('mem=', 'nodes=', 'ppn='): {}".format(nativeConfig))
-            qsubline.extend(nativeConfig.split())
-            
             return qsubline
 
         def generateTorqueWrapper(self, command):
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -91,6 +91,8 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
                     logger.debug('Exit Status: ' + status)
                     return int(status)
                 if 'unknown job id' in line.lower():
+                    # some clusters configure Torque to forget everything about just
+                    # finished jobs instantly, apparently for performance reasons
                     logger.debug('Batch system no longer remembers about job {}'.format(torqueJobID))
                     # return assumed success; status files should reveal failure
                     return 0
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -90,6 +90,11 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
                     status = line.split(' = ')[1]
                     logger.debug('Exit Status: ' + status)
                     return int(status)
+                if 'unknown job id' in line.lower():
+                    logger.debug('Batch system no longer remembers about job {}'.format(torqueJobID))
+                    # return assumed success; status files should reveal failure
+                    return 0
+            return None
 
         """
         Implementation-specific helper methods
@@ -106,7 +111,7 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             # qsubline = ['qsub', '-V', '-j', 'oe', '-o', '/dev/null',
             #             '-e', '/dev/null', '-N', 'toil_job_{}'.format(jobID)]
 
-            qsubline = ['qsub', '-V', '-N', 'toil_job_{}'.format(jobID)]
+            qsubline = ['qsub', '-S', '/bin/sh', '-V', '-N', 'toil_job_{}'.format(jobID)]
 
             if self.boss.environment:
                 qsubline.append('-v')
@@ -116,10 +121,20 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             reqline = list()
             if mem is not None:
                 memStr = str(mem / 1024) + 'K'
-                reqline += ['-l mem=' + memStr]
+                reqline.append('mem=' + memStr)
 
             if cpu is not None and math.ceil(cpu) > 1:
-                qsubline.extend(['-l ncpus=' + str(int(math.ceil(cpu)))])
+                reqline.append('ncpus=' + str(int(math.ceil(cpu))))
+
+            # Other resource requirements can be passed through the environment (see man qsub)
+            reqlineEnv = os.getenv('TOIL_TORQUE_REQS')
+            if reqlineEnv is not None:
+                logger.debug("Additional Torque resource requirements appended to qsub from TOIL_TORQUE_REQS env. variable: {}".format(reqlineEnv))
+
+                reqline.append(reqlineEnv)
+            
+            if reqline:
+                qsubline += ['-l',','.join(reqline)]
 
             # "Native extensions" for TORQUE (see DRMAA or SAGA)
             nativeConfig = os.getenv('TOIL_TORQUE_ARGS')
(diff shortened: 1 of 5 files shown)


cket:
@andreyto thanks for the PR. It looks like one of the fixes is a duplicate of https://github.com/BD2KGenomics/toil/pull/1539, correct? 

Jenkins, test this please.

andreyto:
@cket yes, it appears one fix is solving the same problem as #1539, which I did not notice. Not sure though why the solution has to be that complex as in #1539, and require the user to be aware of a possible race condition and enter non-default arguments. My solution is very simple (one import and one line of code, no user arguments), should just work on a shared FS, and make no harm on a local FS. The option for customizing node id generation in #1539 can be useful, however, in some extreme cases (like no MAC on a node).

cket:
@andreyto I agree it's very complex, you can see our discussion here on how we came to that solution https://github.com/BD2KGenomics/toil/issues/1462. 

The combination of both approached might be the best solution, my main concern is that uuid.getNode could hurt local FS users in cases where 1) the random UUID is returned or 2) a random mac address is chosen nondeterministically when there are multiple addresses for the machine. The getNode documentation isn't entirely clear on case 2, it's possible that the address is chosen deterministically.

andreyto:
@cket Yes, finding a universally available method to get an ID is a conundrum. I can see that Python3 code for `getnode()` tries first calling some [_uuid_generate_time](https://svn.python.org/projects/python/tags/r31/Lib/uuid.py). 

How about reading from `/proc/sys/kernel/random/boot_id`? I just found its description here (http://0pointer.de/blog/projects/ids.html):
```
A random ID that is regenerated on each boot. As such it can be used to identify the local machine's current boot. It's universally available on any recent Linux kernel. It's a good and safe choice if you need to identify a specific boot on a specific booted kernel.
```
If a node is rebooted, there has to be some mechanism to clean up its dead jobs because the `boot_id` will change. But I assume that this case is already taken care of elsewhere in Toil code because the node can also simply go down forever, and Toil still should be able to notice and re-run the jobs?

Anyway, whatever the implementation is, my key wish is that the cause of the workDir on a shared FS should be taken care of automatically. Or at least, be described very conspicuously in the documentation. It took me quite a bit of effort to figure out why the example was failing with cryptic exceptions.

cket:
@andreyto boot id sounds promising, I'll read up on it. Thanks!

cket:
I had some more time to check it out & I think using boot id is a great idea, are you interested in implementing that behavior?

andreyto:
Yes, sure. Should I essentially replace a call to `uuid.getnode()` with a call to a method `getNodeID()` that tries to read `/proc/sys/kernel/random/boot_id`? How about the method first tries the `/var/lib/dbus/machine-id`, like the 0pointer advises, then tries boot_id, if that is not available finally uses `uuid.getnode()`? Do I need to provide an escape hatch for the user to override this sequence with a file name or an expression like in #1539? Or is that never likely be needed?

Also, have you considered what will happen if Docker is used to execute most tasks? Is worker process executed on the host, and runs docker for each task?

cket:
That sounds good to me! You don't have to worry about an escape hatch, we can rebase #1539 onto your changes to provide that functionality. 

I don't think Docker will be an issue - the worker processes aren't containerized so we shouldn't run into that scenario.

andreyto:
I have squashed the commit with the new method for getting node ID with my previous commit, and force-pushed into GitHub. Hopefully, this is having the desired effect that you are seeing the updated pull request.

cket:
Jenkins, ok to test

src/toil/common.py
```diff
@@ -565,6 +565,62 @@ def addGroup(headingString, bodyString):
...
+                if len(nodeID.split()) == 1:
+                    logger.debug("Obtained node ID {} from file {}".\
+                            format(nodeID,idSourceFile))
+                    break
+                else:
+                    logger.warning(("Node ID {} from file {} contains spaces."+\
+                            " Will try next method to get node ID").\
+                            format(nodeID,idSourceFile))
+    else:
+        import uuid
```
cket: Can you make this a module level import?

src/toil/batchSystems/torque.py
```diff
@@ -113,10 +117,20 @@ def prepareQsub(self, cpu, mem, jobID):
...
                 memStr = str(mem / 1024) + 'K'
-                reqline += ['-l mem=' + memStr]
+                reqline.append('mem=' + memStr)
 
             if cpu is not None and math.ceil(cpu) > 1:
-                qsubline.extend(['-l ncpus=' + str(int(math.ceil(cpu)))])
+                reqline.append('ncpus=' + str(int(math.ceil(cpu))))
+
+            # Other resource requirements can be passed through the environment (see man qsub)
+            reqlineEnv = os.getenv('TOIL_TORQUE_REQS')
```
cket: Can you add a brief section on this variable under "Environment variable options" in /toil/docs/running/running.rst?

src/toil/common.py
```diff
@@ -565,6 +565,62 @@ def addGroup(headingString, bodyString):
...
+    import os
+    idSourceFiles = extraIDFiles + ["/var/lib/dbus/machine-id",
+        "/proc/sys/kernel/random/boot_id"]
+    for idSourceFile in idSourceFiles:
+        if os.path.exists(idSourceFile):
+            try:
+                with open(idSourceFile,"r") as inp:
+                    nodeID = inp.readline().strip()
+            except EnvironmentError:
+                logger.warning(("Exception when trying to read ID file {}. "+\
```
andreyto: @cket This warning is only triggered if the file does exist, but trying to read it results in an OS exception (e.g. file has disappeared between two method calls, or permissions restrict it from reading - weird stuff, in other words). On MacOS, these files do not exist, so the warning is not triggered. Do you still want the warning to be turned into a debug message?

src/toil/common.py
```diff
@@ -565,6 +565,62 @@ def addGroup(headingString, bodyString):
...
+                            format(nodeID,idSourceFile))
+    else:
+        import uuid
+        nodeIDs = []
+        for i_call in range(2):
+            nodeID = str(uuid.getnode()).strip()
+            assert len(nodeID.split())==1,("Node ID {} from uuid.getnode() "+\
+                    "contains spaces").format(nodeID)
+            nodeIDs.append(nodeID)
+        assert nodeIDs[0] == nodeIDs[1], ("Different node IDs {} received from"+\
```
andreyto: @cket OK, will do. However, this will also mean cryptic failures, or, worse, incorrect outputs, if they are using a shared file system for tmp, and the node ID is not determined by any of the techniques.

andreyto:
@cket Once I make the changes, do you want them rebased and squashed like the last time?

cket:
@andreyto no, an additional commit with the changes is OK by me

src/toil/common.py
cket: Good point. I think you're right then, this is better as a warning

src/toil/common.py
cket: That's true, but hopefully the warning will give them enough insight to fix the problem

mr-c:
Is there anything I can do to assist? At EBI we are also experiencing the issues in #1628 and are eager to see this fixed :-)

andreyto:
I forgot about it. Will try to work on it this week.

On Wed, May 24, 2017 at 8:13 AM, Michael R. Crusoe <notifications@github.com

mr-c:
It is unfortunate that this missed the 3.8.0 release. Perhaps we should split the fixes up into smaller parts? I am happy to test and polish the non-torque pieces

andreyto:
I have completed all the changes that were requested from the original pull request (I believe so, at least). In the meantime, a conflicting change has been made to add Torque env var. I have kept it my way, as two vars - I explained in the commit message why. Your test harness `make test` fails on my system because apparently use of Docker is hard wired, and my system with Torque does not have Docker. But your sort example from the manual runs OK under Torque on my system.

ejacox:
@evan-wehi, will this satisfy #1397?

src/toil/common.py
```diff
@@ -568,6 +569,59 @@ def addGroup(headingString, bodyString):
                            "argparse.ArgumentParser" % parser.__class__)
 
 
+def getNodeID(extraIDFiles=[]):
+    """Return unique ID of the current node (host).
```
jessebrennan: For consistency, docstrings begin on a separate line from the quotes.

src/toil/common.py
```diff
@@ -568,6 +569,59 @@ def addGroup(headingString, bodyString):
...
+    the same node at least until the next OS reboot.
+
+    The last resort method is uuid.getnode() that in some rare OS configurations may return a random ID each time it is
+    called. However, this method should never be reached on a Linux system, because reading from 
+    /proc/sys/kernel/random/boot_id will be tried prior to that. If uuid.getnode() is reached, it will be called twice,
+    and exception raised if the values are not identical.
+
+    :param list extraIDFiles: Optional list of additional file names to try reading node ID before trying default 
+    methods. ID should be a single word (no spaces) on the first line of the file.
+
```
jessebrennan: Could you delete this white space?

src/toil/common.py
```diff
@@ -568,6 +569,59 @@ def addGroup(headingString, bodyString):
...
+                    logger.debug("Obtained node ID {} from file {}".format(nodeID,idSourceFile))
+                    break
+                else:
+                    logger.warning(("Node ID {} from file {} contains spaces. Will try next method to get node ID").\
+                            format(nodeID,idSourceFile))
+    else:
+        nodeIDs = []
+        for i_call in range(2):
+            nodeID = str(uuid.getnode()).strip()
+            if len(nodeID.split())==1:
```
jessebrennan: Could you put spaces around `==`?

src/toil/common.py
```diff
@@ -568,6 +569,59 @@ def addGroup(headingString, bodyString):
...
+        nodeIDs = []
+        for i_call in range(2):
+            nodeID = str(uuid.getnode()).strip()
+            if len(nodeID.split())==1: 
+                nodeIDs.append(nodeID)
+            else:
+                logger.warning("Node ID {} from uuid.getnode() contains spaces".format(nodeID))
+        nodeID = ""
+        if (len(nodeIDs) == 2):
+            if (nodeIDs[0] == nodeIDs[1]):
```
jessebrennan: Likewise the parentheses on these if statements.

src/toil/test/src/nodeIDTest.py
```diff
@@ -0,0 +1,34 @@
...
+        for i in range(10):
+            nodeID = getNodeID()
+            if i > 0:
+                self.assertEquals(nodeID,prevNodeID)
+            prevNodeID = nodeID
+
+    def testSkipNonExistingFile(self):
+        import uuid, os
+        no_file = "/etc/totally_non_existing_"+str(uuid.uuid4())
+        assert not os.path.exists(no_file), "Could not make up a missing file name"
```
jessebrennan: `self.assertFalse` may give cleaner error message though in this case it wouldn't make much difference.

andreyto: Folks, I do not know when I will have time to address your style changes - this is a second review, and a new set of style requirements. Since those are just code formatting suggestions, perhaps one of the project owners will be willing to make them and commit.

jessebrennan:
We'll take care of it no problem. Thanks for the contribution.

evan-wehi:
@ejacox Yes, I believe this will fix #1397. That was caused by the dead job clean trying to kill PIDs that were actually in another node

jessebrennan:
Jenkins, test this please.

jessebrennan:
Jenkins, test this please.

ejacox:
Jenkins, test this please.

andreyto:
@jessebrennan Jenkins does not let me access testing results - asks to login even if I am already logged in. Jenkins tests were passing before your latest changes, but maybe something has changed in the testing environment unrelated to the changes. Let me know if I can help.

mr-c:
@andreyto et al.
```
======== 2 passed, 570 deselected, 2 pytest-warnings in 1203.86 seconds ========
Build timed out (after 240 minutes). Marking the build as failed.
```

mr-c:
Jenkins, test this please.

mr-c:
Tests have passed :fireworks: ! @cket @jessebrennan can you merge?

ejacox:
There were testing problems on Friday. Boto was returning 400 errors.

Thank you @mr-c and @evan-wehi for helping this PR along.

andreyto:
Thanks!  @jessebrennan, good catch with my mutable default!

ejacox:
Thank your for this PR @andreyto

ionox0:
Just to follow up on this we are still experiencing this issue with Toil 3.19. We do use a --workDir folder that is shared across all nodes.

Could it be possible that even with a unique identifier for the node, that two jobs running on the same node could delete each others tmp file? Because they will still have the same ID in that case

berguner:
We are also experiencing a similar problem on our cluster. We have are using beeGFS shared file system for all the worker nodes and the TMPDIR is also located in this FS. The failed jobs sometimes finishes successfully in the second attempt. Here is the error we get:

`WARNING:toil.leader:r/j/jobe3n311c3    subprocess.CalledProcessError: Command '['du', '-s', '/scratch/users/berguener/tmpdir/toil-db817b2d-5165-4aab-95f6-242a975c5904-14556eaa535a0508c8a6b3bf00000017/tmpdkbz4mcb/919d098d-9fcc-4381-9a85-5ae0a32788c9']' returned non-zero exit status 1.`

mr-c:
@berguner can you move this to an issue? I think a closed PR will be forgotten again. Thank you for the report.

Issue #1598, opened 2017-03-28
Checkpoint jobs are considered "totally failed" after a successful restart


joelarmstrong:
I can consistently get failures if I do the following:
1. Run a workflow containing a checkpointed subtree with a retry count of 0, wait for the workflow to get to the checkpointed subtree, and kill the workflow (causing some job failures in the checkpointed subtree)
2. Restart the workflow

On step 2, the workflow completes the checkpointed subtree fine, but then bails out because it considers the checkpointed job to be totally failed. But restarting the workflow after that works fine, and runs to completion.

If it matters, the checkpointed job in question has an associated service. I think this is probably something wrong with the "cleanup" of the jobStore when resuming a workflow.

benedictpaten:
You can assign to me - this is definitely my bug.

joelarmstrong:
Well, regardless of who it's assigned to, I have a minimal example which should make things a bit easier. That minimal example actually revealed another bug: it looks like checkpointed subtrees are restarted even if the retry count is set to zero. But obviously that's a little less serious.

@cket, would it be make sense if I made a PR for the unit test I have that reproduces this, maybe marking it xfail? I think the bug(s) are probably easy to find given a reproducible example, but I don't think I'll have time to get around to fixing it for a while, since it's not really a blocker.

cket:
A PR with a unit test would be great! You don't have to mark it expected fail since we will want it to be passing before we merge it in.

joelarmstrong:
I went ahead and fixed this (hopefully correctly), because it was about to cost us a few hundred otherwise by needlessly restarting a successful mammals run.

benedictpaten:
@joelarmstrong  We can close this, right? Looks like your pr was committed.

joelarmstrong:
It hasn't been merged yet, my PR has a bug where a checkpoint can never run out of retries. The fix is simple, but I haven't gotten around to it yet.

PR #1599, opened 2017-03-29
Checkpointed restart bugs (#1598)

joelarmstrong:
This adds unit tests for the two bugs described in #1598. Hopefully they both fail!

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -266,6 +266,7 @@ class Leader:
                         # unless it has more than 1 try.
                         elif jobGraph.checkpoint is not None and jobGraph.remainingRetryCount > 1:
                             jobGraph.setupJobAfterFailure(self.config)
+                            self.jobStore.update(jobGraph)
                             logger.warn('Job: %s is being restarted as a checkpoint after the total '
                                         'failure of jobs in its subtree.', jobGraph.jobStoreID)
                             self.issueJob(JobNode.fromJobGraph(jobGraph))
@@ -910,8 +911,9 @@ class Leader:
                     # Remove it from the set of jobs with active successors
                     self.toilState.successorCounts.pop(predecessorJob.jobStoreID)
 
-                    # Pop stack at this point, as we can get rid of its successors
-                    predecessorJob.stack.pop()
+                    if predecessorJob.jobStoreID not in self.toilState.hasFailedSuccessors:
+                        # Pop stack at this point, as we can get rid of its successors
+                        predecessorJob.stack.pop()
 
                     # Now we know the job is done we can add it to the list of updated job files
                     assert predecessorJob not in self.toilState.updatedJobs
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -260,7 +260,12 @@ class Leader:
                             continue
 
                         # If the job is a checkpoint and has remaining retries then reissue it.
-                        elif jobGraph.checkpoint is not None and jobGraph.remainingRetryCount > 0:
+                        # The logic behind using > 1 rather than > 0 here: Since this job has
+                        # been tried once (without decreasing its retry count as the job
+                        # itself was successful), and its subtree failed, it shouldn't be retried
+                        # unless it has more than 1 try.
+                        elif jobGraph.checkpoint is not None and jobGraph.remainingRetryCount > 1:
+                            jobGraph.setupJobAfterFailure(self.config)
                             logger.warn('Job: %s is being restarted as a checkpoint after the total '
                                         'failure of jobs in its subtree.', jobGraph.jobStoreID)
                             self.issueJob(JobNode.fromJobGraph(jobGraph))
--- a/src/toil/test/src/checkpointTest.py
+++ b/src/toil/test/src/checkpointTest.py
@@ -30,42 +30,70 @@ class CheckpointTest(ToilTest):
         # to check that the job wasn't retried short of parsing the
         # log.
         with self.assertRaises(FailedJobsException):
-            Job.Runner.startToil(Checkpointed(succeedOnRetry=True), options)
+            Job.Runner.startToil(CheckRetryCount(numFailuresBeforeSuccess=1), options)
 
-    def testCheckpointedRestart(self):
+    def testCheckpointRetriedOnce(self):
+        """A checkpoint job with a retryCount of 1 should be retried exactly once."""
+        options = Job.Runner.getDefaultOptions(self._getTestJobStorePath())
+        options.retryCount = 1
+
+        # This should succeed (the checkpoint should be retried once, then succeed)
+        try:
+            Job.Runner.startToil(CheckRetryCount(numFailuresBeforeSuccess=1), options)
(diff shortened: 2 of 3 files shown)


src/toil/leader.py
```diff
@@ -260,7 +260,12 @@ def innerLoop(self):
                             continue
 
                         # If the job is a checkpoint and has remaining retries then reissue it.
-                        elif jobGraph.checkpoint is not None and jobGraph.remainingRetryCount > 0:
+                        # The logic behind using > 1 rather than > 0 here: Since this job has
+                        # been tried once (without decreasing its retry count as the job
+                        # itself was successful), and its subtree failed, it shouldn't be retried
+                        # unless it has more than 1 try.
+                        elif jobGraph.checkpoint is not None and jobGraph.remainingRetryCount > 1:
```
joelarmstrong: OK, @cket, @benedictpaten, I now remember why this is correct. The "remainingRetryCount" is set to AbstractJobStore._defaultTryCount() [in all the jobStores (e.g. fileJobStore)](https://github.com/BD2KGenomics/toil/blob/b27a8efadb82089a2799415d8a506bd61758aead/src/toil/jobStores/fileJobStore.py#L96). Then [it's passed as `tryCount` to the jobGraph](https://github.com/BD2KGenomics/toil/blob/b27a8efadb82089a2799415d8a506bd61758aead/src/toil/jobGraph.py#L122). And indeed it is a try count: [_defaultTryCount is set to retryCount + 1](https://github.com/BD2KGenomics/toil/blob/b27a8efadb82089a2799415d8a506bd61758aead/src/toil/jobStores/abstractJobStore.py#L950).

I'd vote to rename it remainingTryCount.

Maybe some of the other job types are tried one too many times as well?

src/toil/worker.py
```diff
@@ -259,10 +259,6 @@ def badWorker():
                 else:
                     jobGraph.command = jobGraph.checkpoint
 
-                # Reduce the retry count
-                assert jobGraph.remainingRetryCount >= 0
-                jobGraph.remainingRetryCount = max(0, jobGraph.remainingRetryCount - 1)
```
joelarmstrong: This part is what was causing the "failure" on a successful resume. The retry count was decremented during "cleanup" of a restarted checkpoint job--even if the checkpoint was restarted because the workflow was restarted. This is now handled by the `setupJobAfterFailure` line in the leader.

cket:
@joelarmstrong should I hold off on reviewing this further?

joelarmstrong:
Yes, hold off please, I suspect that this broke checkpointed restart with more than 0 retryCount. I had a pipeline retry a checkpoint more than 200 times. But I haven't had a chance to look into it quite yet. I'll add another unit test once I fix it, just to confirm it actually works this time. Sorry about that!

Should have a fix for this by next Monday, if we're still meeting then.

ejacox:
@joelarmstrong was this fixed?

joelarmstrong:
Nah, there's still a remaining issue where the checkpointed jobs get retried forever. I can take a second look tomorrow morning, it's probably something stupid.

joelarmstrong:
OK, this is finally ready to be reviewed/merged. There were a few additional bugs in the leader that the new tests uncovered.

src/toil/test/src/checkpointTest.py
```diff
@@ -0,0 +1,102 @@
...
+
+class CheckpointTest(ToilTest):
+    def testCheckpointNotRetried(self):
+        """A checkpoint job with a retryCount of 0 should not be retried."""
+        options = Job.Runner.getDefaultOptions(self._getTestJobStorePath())
+        options.retryCount = 0
+        # We set the workflow to *succeed* if the checkpointed job is
+        # retried despite the retry count of 0. So, if the workflow
+        # fails, this test succeeds.
+        # This ends up being a bit confusing, but it's the easiest way
```
benedictpaten: We should clarify here that it is only true that a root job with a retry count of 0 should be tried only once. The behavior of a sequence with length n of successive checkpoint jobs, each with a retry count of 1 and each a child of the previous job in the sequence is that the last job in the sequence will potentially be retried n times, right?

src/toil/worker.py
benedictpaten: I'm a bit confused, how did remainingRetryCount every get below 0? The advantage of doing this here is there is less I/O on the leader, and yes, its a bit squishy with the max statement

src/toil/test/src/checkpointTest.py
joelarmstrong: I should've been clearer in the docstring -- in this test it's not that the job itself has a remainingRetryCount of 0, it's that the *workflow* has a retryCount of 0. In that case, checkpoint jobs should never be retried (yet before this PR, they were).

For your second question, yes, that behavior is correct and it should be what happens now (though I haven't tried that specifically).

src/toil/worker.py
joelarmstrong: The remainingRetryCount never got below 0 (at least I don't think so). The assert was removed when I moved the retry-count decrementing to the leader. I could have left the assert in, but didn't feel that there was much point now that the leader handles things.

When a checkpoint fails, the "cleanup" of its successors is only performed on the next run of the checkpoint. That's also the case on resume: the cleanup of the checkpoint is only ever handled by the checkpoint.

So if you have this situation, say with a retryCount of 1 for the workflow:

- workflow runs, CheckpointJob runs and creates a bunch of ChildJobs
- one ChildJob totally fails and kills the workflow
- on resume, CheckpointJob is run and removes all its successors. It reduces its remainingRetryCount to 0. It then runs from a fresh start and creates a bunch of ChildJobs.
- All its ChildJobs succeed. The workflow finishes.
- The workflow is considered totally failed because a totally failed job is detected (CheckpointJob).
- On restart you start from CheckpointJob yet again, wasting all the (successful) work done earlier. If it succeeds this time, everything is OK.

The behavior is similar with a higher retryCount, but to trigger this bug the checkpoint has to be restarted retryCount - 1 times.

I totally agree that this PR causes more blocking I/O on the leader. We could instead call this code to "cleanup" the checkpoint as part of the general jobStore "cleanup" when resuming, which would avoid leader I/O. But this was a huge blocker for me, so just needed to fix it ASAP at the time.

src/toil/leader.py
joelarmstrong: Nah, it's a *try* count, despite the name. So this code should be correct.

We could rename to remainingTryCount. But I think that should be a new PR. Plus it breaks jobStore resumability, and if we're going to do that, then we might as well rename many things at the same time to make it less painful.

benedictpaten: Okay, got confused by the comment. I agree, let's put this in a new PR, and go through systematically and fix the issue of try vs. retry.

src/toil/worker.py
benedictpaten: Yes, to make this clean this should really be fixed on startup, plus it keeps the I/O on the leader in one place. Up to you, we can also merge this in now and then do a review later.

PR #1602, opened 2017-03-31
DRMAA batch system (resolves #1558)

### LINKED ISSUE #1558: Use of qacct on GridEngine
I'm trying to run toil on a GridEngine cluster but am facing some difficulties. It appears that toil submits jobs correctly but fails to notice when they finish, i.e. even very simple pipelines hang indefinitely. I believe that this is related to #1114.

My understanding from the previous discussion is that toil monitors the output of `qacct` to determine whether jobs have completed successfully. However, I don't have access to `qacct` on the local cluster, which means that this inevitably fails. As far as I can tell this essentially working as designed but I was wondering what the rationale behind the use of `qacct` is and whether an alternative scheme that uses `qstat` instead would be feasible. That would have the benefit of being more widely available but I don't know enough about the toil internals to understand what the potential issues might be.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-142)
┆Issue Number: TOIL-142

humburg:
This PR introduces `AbstractDRMAABatchSystem`, a replacement for `AbstractGridEngineBatchSystem`, that uses the [DRMAA API](https://www.drmaa.org/) to communicate with the underlying system. This alleviates the need to query the state system via command line tools and allows GridEngine like BatchSystems to delegate more of the work to a common base class. 

A few system specific tasks, that aren't covered by the API, still remain. An implementation for SGE is included in this PR. The same could be done for the other BatchSystems that support DRMAA (LSF, Torque) but I'm not currently in a position to test these.

For SGE there are two main benefits. Firstly, the new implementation does not rely on the use of `qacct`, which is not guaranteed to be configured properly, as discussed in #1558. Secondly, users now have some control over how resource requirements are communicated to SGE. Site specific configurations may lead to different requirements that Toil currently can't accommodate.

The documentation still has to be updated to reflect the changes but I'd like to get some feedback before doing much more work on this.


cket:
Jenkins, test this please. @humburg thanks for the PR! Unfortunately I'm traveling for the next week so I won't be able to review this closely until I get back. @benedictpaten may have some feedback on this, and @cmarkello I think this would help with the Grid Engine compatibility problems you ran into earlier.

cjfields:
Just a comment: it might be worth checking into the various batch systems (SLURM, LSF, Torque, SGE) to see whether DRMAA is easily supported, if it's version specific, and whether the batch system needs to be compiled with DRMAA support prior to deployment.  Otherwise, replacing batch system-specific code with something that relies completely on DRMAA could cause issues for 'legacy' systems that can't easily implement this support (or maybe don't want to implement it, for other reasons).  I recall this being an issue with earlier versions of PBS/Torque (required compile-time parameter that wasn't enabled by default and was really buggy), though I think the situation is much better with later versions.

Saying that, I think it would be really great to have a DRMAA-specific implementation that doesn't rely on specific batch system-specific underlying idiosyncrasies.  I'd be willing to test it from the Torque and SLURM end.

humburg:
@cjfields As best as I can tell the situation is as follows.
* SGE: Distributed with the C library since version 6.0.
* Torque: Distributed with the C library since version 2.4.2 (at least) but may need some configuration changes to work. Based on the information [here](http://apps.man.poznan.pl/trac/pbs-drmaa) it doesn't seem like compile time flags are required.
* LSF: Supported since version 8.3 via open source project (on [GitHub](https://github.com/PlatformLSF/lsf-drmaa)). I assume this has to be installed separately.
* SLURM: It isn't entirely clear to me what the situation is here. A DRMAA implementation for SLURM is available (http://apps.man.poznan.pl/trac/slurm-drmaa). This has to be installed separately. However, the latest release is 3 years old and there are also at least two related projects on GitHub (https://github.com/natefoo/slurm-drmaa and https://github.com/FredHutch/slurm-drmaa), making me wonder whether there are issues that may still be unresolved. This may need some investigation.

cjfields:
If it's not telling, I've worked w/ a few HPC sysadmins who can be (possibly rightfully) recalcitrant to upgrading or adding additional functionality across an in-production cluster environment.  

@humburg I suppose my point is that replacing batch system specific code with something that relies on DRMAA would also add DRMAA support as an extra dependency depending on the system, which wasn't required before.  That is unless I'm mistaken, and the SGE-specific fixes you are implementing won't require DRMAA and upgrading to SGE v6 to get built-in DRMAA support.  Otherwise this may be a problem deploying on some systems.

humburg:
@cjfields Apologies for the late response, I was away for a few days. I completely understand that changes to HPC systems can be problematic. In fact, this whole thing started because my local sysadmin was unwilling to make the changes to our SGE cluster that would be required to make Toil work. For SGE specifically, it seems to me that DRMAA support has been around for long enough (10+ years) that the (vast?) majority of clusters in use are likely to have it anyway. This is less clear for some of the other platforms that support DRMAA.

brainstorm:
@humburg, I used to work closely with the main developer of SLURM-drmaa, Mariusz Mamonski while I was doing my MSc thesis. Unfortunately, he passed away 2-3 years ago, so I would go for @natefoo's fork since I'm pretty sure he and the Galaxy team are taking good care of DRMAA nowadays.

diekhans:
apologies that this PR hasn't been acted on.  We are very interested in seeing if this can simplify out HPC support.  @humburg , would you be able to work with us on this?

mr-c:
Can we have both DRMAA & keep the existing implementations for now? Then a decision one way or the other can be made for each batch system.

cjfields:
What @mr-c said sounds reasonable to me, this would definitely help getting the changes in

diekhans:
yes, there is no need to remove existing ones until DRMAA is demonstrated replacing them.   Long term, this would be a great simplification.

humburg:
@diekhans Thanks for looking at this. I've since moved to a new job and no longer have access to an HPC cluster that I could test this on. So, unfortunately, I don't think I can be much help anymore.

DailyDreaming:
This seems unlikely to be revived.  Closing this unless someone posts an objection.

Issue #1453, opened 2017-01-25
jobStore.importFile returns a string not a fileStoreID


ArtRand:
Is this intentional? It means that if you use the importFIle method then pass the result to a job and invoke the .size method (for resource allocation), you get an error...

jessebrennan:
@cket this is an easy fix, but I need confirmation. Is this intentional? Will changing this break any existing code?

cket:
@jessebrennan this isn't intentional, if it breaks existing code then that also needs to be fixed

arkal:
@ifiddes also reported this to me a few days ago. It's not intentional and might not be an easy fix for non-fileJobStore implementations for importFile. My temp fix for a file that exists on the same machine was
```
from toil.fileStore import fileID
x = FileID.forPath(toil.importFile('file:///path/to/file')
```
This won't work as well for something on s3 but you could find a similar hack

ArtRand:
I have a similar hack that works for S3 and local jobStores:
https://github.com/ArtRand/marginAlign/blob/toil_dev_signal/src/margin/toil/localFileManager.py#L145
 Ingestion of files with .importFIle and .exportFile can be a little dodgy in my experience (i.e. works sometimes).

PR #1605, opened 2017-04-02
Return an integer rather than string size for HTTP imports

joelarmstrong:
Didn't create an issue for this. Hopefully that's OK for such a trivial change. Somewhat related to #1453.


cket:
Jenkins, test this please

Issue #1607, opened 2017-04-03
Improve error when passing a promised object to another child job function


jvivian:
Never fully resolved in: https://github.com/BD2KGenomics/toil/pull/912

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-150)
┆Issue Number: TOIL-150

DailyDreaming:
Stale, and now unclear what the issue was.  Closing, please reopen if still an issue.

PR #1608, opened 2017-04-04
Create an unfulfilled promise sentinel (resolves #517, #1607)

### LINKED ISSUE #517: Better error needed if job is not encapsulated by the user

joelarmstrong:
Quick stab at improving the error message. The current EOFError is really hard to debug, especially since the traceback leads to the unpickling of the job as a whole rather than the unpickling of the promise. This error message points back to the fulfilling job (which presumably hasn't run yet).

Unfortunately, the unpickler goes a bit haywire when you throw an exception during class creation, and the exception gets stacked in on itself with the tuple returned by `Promise.__reduce__`. The output looks like this (`C` is passed a promise from sibling `B`):

```
=========> Failed job 'C' o/h/jobFxT0UY 
---TOIL WORKER OUTPUT LOG---
INFO:toil:Running Toil version 3.7.0a1-3d8272872287e7b0bd2725aa0f857062c69c0a50.
WARNING:toil.resource:Can't find resource for leader path '/home/joel/toil-fresh'
WARNING:toil.resource:Can't localize module ModuleDescriptor(dirPath='/home/joel/toil-fresh', name='testPromiseWiringError', fromVirtualEnv=False)
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/worker.py", line 325, in main
    job = Job._loadJob(jobGraph.command, jobStore)
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/job.py", line 873, in _loadJob
    return cls._unpickle(userModule, fileHandle, jobStore.config)
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/job.py", line 897, in _unpickle
    runnable = unpickler.load()
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/job.py", line 1677, in __new__
    return cls._resolve(*args)
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/job.py", line 1689, in _resolve
    value = cPickle.load(fileHandle)
  File "/home/joel/toil-fresh/venv/local/lib/python2.7/site-packages/toil/job.py", line 1768, in __setstate__
    "this job as a child/follow-on of {jobName}.".format(jobName=jobName))
RuntimeError: (RuntimeError("This job was passed a promise that wasn't yet resolved when it ran. The job 'B' that fulfills this promise hasn't yet finished. This means that there aren't enough constraints to ensure the current job always runs after 'B'. Consider adding a follow-on indirection between this job and its parent, or adding this job as a child/follow-on of 'B'.",), <class 'toil.job.Promise'>, ('file:/home/joel/toil-fresh/jobStore', '5/1/tmpddadqK.tmp'))
ERROR:toil.worker:Exiting the worker because of a failed job on host jabberwocky
WARNING:toil.jobGraph:Due to failure we are reducing the remaining retry count of job 'C' o/h/jobFxT0UY with ID o/h/jobFxT0UY to 0
<=========
```

I've tried several different things to fix that, but no matter whether the exception is thrown in `__init__`, `__setstate__`, or `__new__`, you get the same tuple-y behavior. If someone has an idea of how to fix that, or can think of a better way to phrase the error message, go for it!

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -524,7 +524,9 @@ class Job(JobLikeObject):
         if self._promiseJobStore is None:
             raise RuntimeError('Trying to pass a promise from a promising job that is not a ' +
                                'predecessor of the job receiving the promise')
-        jobStoreFileID = self._promiseJobStore.getEmptyFileStoreID()
+        with self._promiseJobStore.writeFileStream() as (fileHandle, jobStoreFileID):
+            promise = UnfulfilledPromiseSentinel(str(self), False)
+            cPickle.dump(promise, fileHandle, cPickle.HIGHEST_PROTOCOL)
         self._rvs[path].append(jobStoreFileID)
         return self._promiseJobStore.config.jobStore, jobStoreFileID
 
@@ -1745,3 +1747,22 @@ class PromisedRequirement(object):
             elif isinstance(kwargs.get(r), PromisedRequirement):
                 foundPromisedRequirement = True
         return foundPromisedRequirement
+
+
+class UnfulfilledPromiseSentinel(object):
+    """This should be overwritten by a proper promised value. Throws an
+    exception when unpickled."""
+    def __init__(self, fulfillingJobName, unpickled):
+        self.fulfillingJobName = fulfillingJobName
+
+    @staticmethod
+    def __setstate__(stateDict):
+        """Only called when unpickling. This won't be unpickled unless the
+        promise wasn't resolved, so we throw an exception."""
+        jobName = stateDict['fulfillingJobName']
+        raise RuntimeError("This job was passed a promise that wasn't yet resolved when it "
+                           "ran. The job {jobName} that fulfills this promise hasn't yet "
+                           "finished. This means that there aren't enough constraints to "
+                           "ensure the current job always runs after {jobName}. Consider adding a "
+                           "follow-on indirection between this job and its parent, or adding "
+                           "this job as a child/follow-on of {jobName}.".format(jobName=jobName))


cket:
Thanks for the PR @joelarmstrong!

PR #1614, opened 2017-04-10
Remove "refresh" thread and just set S3 user script ACL to public-read (resolves #1600, #1516)

### LINKED ISSUE #1600: AWS token expiry due to (presumable) leader thread contention
The temporary IAM credentials used to download the hot-deployed user script can expire if the leader's _refreshUserScript thread doesn't get scheduled within a frame of 2 minutes. This will cause jobs to fail basically as fast as the leader can schedule them:

https://joel.armstro.ng/graftoil/dashboard/snapshot/FZqfN5f18jJPGO4FU1NzuW0LnupObmMe

(the failed jobs are marked as non-preemptable after failure because of a temporary modification I made.)

Thankfully this is pretty much self-correcting once the refresh thread does get scheduled, though every task prior to that needs to run, fail and be rescheduled. But do we really need that refresh thread at all? I'm not too surprised that it missed a 2 minute window, python threads [aren't scheduled in an even remotely sane manner](http://www.dabeaz.com/python/UnderstandingGIL.pdf).

Sorry if this is a stupid question--I don't know much about AWS. But couldn't we use [AWS Signature v4 and presigned URLs](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) to sidestep this whole credential refresh issue? It seems like boto2 has supported this since 2014. I'm eager to get rid of any unneeded threads.

joelarmstrong:
Just throwing this out there, not really 100% sure that this is the best solution.

The refresh thread isn't really guaranteed to be scheduled within its 2 minute window, especially given the number of other threads that are running in the leader process. This would remove the refresh thread entirely, and set the ACL of the user script object to `public-read` instead. Unfortunately I can't seem to get a straight answer on the consistency of the PUT object?acl operation. It seems to take effect right away, but I'm not totally sure.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -874,33 +874,6 @@ class Toil(object):
             userScriptResource = userScript.saveAsResourceTo(self._jobStore)
             logger.debug('Injecting user script %s into batch system.', userScriptResource)
             self._batchSystem.setUserScript(userScriptResource)
-            thread = Thread(target=self._refreshUserScript,
-                            name='refreshUserScript',
-                            kwargs=dict(userScriptResource=userScriptResource))
-            thread.daemon = True
-            thread.start()
-
-    def _refreshUserScript(self, userScriptResource):
-        """
-        Periodically refresh the user script in the job store to prevent credential
-        expiration from causing the public URL to the user script to expire.
-        """
-        while True:
-            # Boto refreshes IAM credentials if they will be expiring within the next five
-            # minutes, but it will only check the expiry if and when credentials are needed to
-            # sign an actual AWS request. This means that we should be refreshing the user script
-            # at least every 5 minutes. Note that refreshing the user script in the job store
-            # involves an S3 request requiring credentials and therefore also triggers refreshing
-            # the IAM role credentials. In the worst case, refresh() is called 5 minutes plus
-            # epsilon before IAM credential expiration. The resource is refreshed three minutes
-            # after that, leaving two minutes plus epsilon generating a new signed URL, this time
-            # with refreshed IAM role credentials. This consideration only applies to AWS and
-            # Boto2, of course. See https://github.com/BD2KGenomics/toil/issues/1372.
-            time.sleep(3 * 60)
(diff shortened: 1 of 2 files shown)


cket:
I would expect that the operation is eventually consistent, so this might also be racey. I'll look into it, it might make sense for the workers to just retry during hot deployment until the script is available.

cket:
On second thought, we only need the ACL to have taken affect by the time the IAM credentials expire, right? I think that's probably safe

joelarmstrong:
Sadly, even if a link is public, you can't read it with an expired IAM token in the parameters. I did some testing and it looks like the credential expiry is checked before the ACL for the object is checked, which results in an early 403. In this PR, I make sure the link isn't authenticated with any credentials to avoid that. So if the PUT is eventually consistent, this is a race. I'll try to test this a bit today by rapidly creating objects & changing ACLs on them to see if I can trigger the eventually consistent behavior. It won't give a definitive answer, but it should at least provide some data.

(The lack of credentials also means there is no expiration on the public URL.)

ejacox:
@joelarmstrong, can key.set_canned_acl('public-read') be called when the job store is created?

joelarmstrong:
Unfortunately I don't think so, because the file name (i.e. "key") isn't known at that time. We could potentially work around that by making sure the hot-deployed user script always has the same name, like "user-script", and set that to public-read on creation.  But then getPublicURL() would continue to have the same problem (though I'm not sure it's actually used anywhere else).

I could contact AWS and ask what the consistency on the "PUT acl" operation is. It seems to take effect immediately because it hasn't given me any trouble yet, but that doesn't mean all that much.

ejacox:
What do you suggest? Merge as is? Wait and think about it? You could also leave in the old code. Would that give two chances to avoid race conditions?

joelarmstrong:
I would say merge as-is, I haven't run into any issues with it on my side and I use hot-deployment pretty heavily. The consistency issues are mostly theoretical.

Leaving in the old code won't help. It just refreshes the AWS credentials, but boto will do that automatically if needed.

Issue #1629, opened 2017-04-15
batch system health checks?


joelarmstrong:
This may be only relevant in the Mesos case, but if a batch system that runs on a separate thread breaks, the leader doesn't seem to notice. Mesos aborted my toil framework last night, but the leader didn't notice, and left several instances running idle till I looked this morning.

From syslog:
```
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: E0415 08:54:23.268079    15 process.cpp:2105] Failed to shutdown socket with fd 13: Transport endpoint is not connected
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: I0415 08:54:23.268388    12 master.cpp:1284] Framework f9c1e779-505b-4fdd-8cd2-3bb0c225f0fa-0019 (toil) at scheduler-f42d8bfd-0d9b-4a7b-abac-84451a0fe9a6@172.31.32.56:45241 disconnected
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: I0415 08:54:23.268424    12 master.cpp:2725] Disconnecting framework f9c1e779-505b-4fdd-8cd2-3bb0c225f0fa-0019 (toil) at scheduler-f42d8bfd-0d9b-4a7b-abac-84451a0fe9a6@172.31.32.56:45241
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: I0415 08:54:23.268456    12 master.cpp:2749] Deactivating framework f9c1e779-505b-4fdd-8cd2-3bb0c225f0fa-0019 (toil) at scheduler-f42d8bfd-0d9b-4a7b-abac-84451a0fe9a6@172.31.32.56:45241
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: I0415 08:54:23.268558    12 master.cpp:1297] Giving framework f9c1e779-505b-4fdd-8cd2-3bb0c225f0fa-0019 (toil) at scheduler-f42d8bfd-0d9b-4a7b-abac-84451a0fe9a6@172.31.32.56:45241 0ns to failover
Apr 15 08:54:23 ip-172-31-32-56.us-west-2.compute.internal docker[2277]: I0415 08:54:23.268739    14 hierarchical.cpp:382] Deactivated framework f9c1e779-505b-4fdd-8cd2-3bb0c225f0fa-0019
[...]

Issue #1553, opened 2017-03-08
Toil's Mesos executor doesn't handle temporary network partitions properly


joelarmstrong:
I had a worker node disconnect from, then reconnect to the Mesos master. All its jobs were marked as TASK_LOST, but they didn't seem to get shut down properly, and jobs that attempted to restart on that same node got this error:

```
f066954f-b182-47d2-8cb1-60c88b8ec238    Traceback (most recent call last):
f066954f-b182-47d2-8cb1-60c88b8ec238      File "/usr/local/lib/python2.7/dist-packages/toil/worker.py", line 336, in main
f066954f-b182-47d2-8cb1-60c88b8ec238        with fileStore.open(job):
f066954f-b182-47d2-8cb1-60c88b8ec238      File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
f066954f-b182-47d2-8cb1-60c88b8ec238        return self.gen.next()
f066954f-b182-47d2-8cb1-60c88b8ec238      File "/usr/local/lib/python2.7/dist-packages/toil/fileStore.py", line 488, in open
f066954f-b182-47d2-8cb1-60c88b8ec238        self.cleanCache(jobReqs)
f066954f-b182-47d2-8cb1-60c88b8ec238      File "/usr/local/lib/python2.7/dist-packages/toil/fileStore.py", line 1111, in cleanCache
f066954f-b182-47d2-8cb1-60c88b8ec238        assert self.jobID not in cacheInfo.jobState
f066954f-b182-47d2-8cb1-60c88b8ec238    AssertionError
```
presumably because the KtServerService was still running.

Excerpts:

Mesos master log at disconnection:
```
I0308 08:18:29.357434    11 master.cpp:1245] Agent f15072d4-b9e6-44df-a669-66e986da45f7-S205 at slave(1)@172.31.39.251:5051 (ip-172-31-39-251.us-west-2.compute.internal) disconnected
I0308 08:18:29.357456    11 master.cpp:2784] Disconnecting agent f15072d4-b9e6-44df-a669-66e986da45f7-S205 at slave(1)@172.31.39.251:5051 (ip-172-31-39-251.us-west-2.compute.internal)
I0308 08:18:29.357470    11 master.cpp:2803] Deactivating agent f15072d4-b9e6-44df-a669-66e986da45f7-S205 at slave(1)@172.31.39.251:5051 (ip-172-31-39-251.us-west-2.compute.internal)
[...]

PR #1630, opened 2017-04-17
Upgrade to Mesos 1.0.1

joelarmstrong:
1.0.1 fixes a [pretty critical bug in Mesos](https://issues.apache.org/jira/browse/MESOS-5943) which causes the scheduler to occasionally crash when it receives enough task status updates at once that a read chunk ends in the middle of a request.

This is somewhat relevant to #1629, where I ran into this crash, and #1553, where the "partitions" (really just disconnections) *may* have the same root cause, though I'm not totally sure. It does have a very similar pattern (an random RST from the server after a bunch of large packets get POSTed at once). It doesn't really fix either of those issues in principle, but it might in practice.

This PR doesn't upgrade past the 1.0 series, because beyond that the meaning of the task updates changes slightly. I'm working on proper reconciliation, because right now I see lost tasks popping up constantly. As part of that I will probably upgrade to 1.2 while trying to keep backward compatibility with 1.0.

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -28,7 +28,7 @@ dependencies = ' '.join(['libffi-dev',  # For client side encryption for 'azure'
                          'wget',
                          'curl',
                          'openssh-server',
-                         'mesos=1.0.0-2.0.89.ubuntu1404',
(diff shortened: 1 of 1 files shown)


cket:
Jenkins, test this please

PR #1637, opened 2017-04-21
Return tuple instead of None in _getJobDetailsFromScontrol 

### LINKED ISSUE #1636: Slurm batch system -- TypeError: 'NoneType' object is not iterable
I just got an issue posted to me by a user of my pipeline, whose cluster runs Slurm. As far as I can tell from the traceback, this seems to be an issue with the Slurm batch system, and not my pipeline itself:

https://github.com/ComparativeGenomicsToolkit/Comparative-Annotation-Toolkit/issues/52

```
04-20 16:30:45 toil.statsAndLogging WARNING Got message from job at time 04-20-2017 16:30:45: Job used more disk than requested. Cconsider modifying the user 
script to avoid the chance of failure due to incorrectly requested resources. Job J/H/jobLBFkRk/g/tmpJLhGTC.tmp used 280.16% (11.2 GB [12032901120B] used, 4.0
 GB [4294967296B] requested) at the end of its run.
Exception in thread Thread-141:
  File "/n/home01/lassance/.conda/envs/ENV_PROGRESSIVECACTUS/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/n/home01/lassance/.conda/envs/ENV_PROGRESSIVECACTUS/lib/python2.7/site-packages/toil/batchSystems/slurm.py", line 179, in run
    activity |= self.checkOnJobs()
  File "/n/home01/lassance/.conda/envs/ENV_PROGRESSIVECACTUS/lib/python2.7/site-packages/toil/batchSystems/slurm.py", line 160, in checkOnJobs
    status = self.getJobExitCode(self.slurmJobIDs[jobID])
  File "/n/home01/lassance/.conda/envs/ENV_PROGRESSIVECACTUS/lib/python2.7/site-packages/toil/batchSystems/slurm.py", line 235, in getJobExitCode
    state, rc = self._getJobDetailsFromScontrol(slurmJobID)
TypeError: 'NoneType' object is not iterable
```

Anyone have any ideas?

cket:
Resolves #1636

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -134,7 +134,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                 # slurm_load_jobs error: Invalid job id specified
                 # There is no job information, so exit.
                 if len(values)>0 and values[0] == 'slurm_load_jobs':
-                    return None
+                    return (None, None)
                 
                 # Output is in the form of many key=value pairs, multiple pairs on each line
                 # and multiple lines in the output. Each pair is pulled out of each line and


cket:
@ifiddes does this fix make sense to you? I don't have a way to replicate the failure but I believe that this is the cause

ifiddes:
Makes sense to me, but I don't know enough about the toil code to really say for sure. I can ask if @lassancejm can test it first if you want.

lassancejm:
Just relaunched the job with which I saw this error. Will post the outcome as soon as it's done. 

JML

cket:
@lassancejm how did it go? Part of the problem with validating this fix is that I don't think the error is easily reproducible. I'm pretty confident that the error was caused by this bug though.

lassancejm:
Well, so far so good,  as I have not seen the error message yet (the job is still running but after this long, I observed it at least  twice or more previously)

I guess it is safe to assume that the correction fixed the problem.

cket:
@lassancejm great, thanks for testing this!

PR #1641, opened 2017-04-21
Improve cluster state tracking

cket:
Work in progress, opening to get an idea of what tests I broke

### DIFF
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -108,6 +108,10 @@ class AbstractProvisioner(object):
             # propagate any errors raised in the threads execution
             thread.join(timeout=0)
 
+    def _nodeFilter(self):
+        # TODO: take executor info & get instance object which can then be passed to _considerNodes
+        pass
+
     def _gatherStats(self, preemptable):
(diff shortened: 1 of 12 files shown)


cket:
Jenkins, test this please

cket:
jenkins, test this please

cket:
@ejacox this should be ready to go!

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -52,7 +46,7 @@ class AbstractProvisioner(object):
     __metaclass__ = ABCMeta
 
 
-    def __init__(self, config=None, batchSystem=None):
+    def __init__(self, config=None):
         """
```
jessebrennan: Why doesn't this take in batchSystem as an argument anymore? Also the docstring below needs to be changed to reflect that. I realize this has already been merged, but I'm hoping @cket can help explain this design choice.

cket: The batch system was only referenced in the higher level logic that I moved into the cluster scaler class so there was no need to keep it in this class as well. But yeah the doc string should be updated

PR #1649, opened 2017-04-22
Increase service job timeout to five minutes (resolves #1648)

### LINKED ISSUE #1648: Increase default deadlock wait time
Currently, the default deadlock wait time is [one minute](https://github.com/BD2KGenomics/toil/blob/master/src/toil/common.py#L97). This can cause spurious deadlock failures (e.g., https://github.com/BD2KGenomics/toil-lib/pull/89) if the service jobs that are deadlocking have to pull down large containers/resources. I'd suggest bumping this to 2.5–5 minutes, esp. since most jobs that we are running in Toil are relatively long lived.

fnothaft:


fnothaft:
Closing as unnecessary; see discussion on #1648.

PR #1653, opened 2017-04-24
CWL: add recent nodejs to Appliance Dockerfile

chapmanb:
Allows complex javascript manipulations on AWS runs using the appliance.
The nodejs default version in Ubuntu 14.04 is not new enough so uses
the stable nodejs distribution from NodeSource
(https://github.com/nodesource/distributions)


cket:
@chapmanb looks like we are getting exit code 100 trying to install nodeJS. Any idea what is causing that?

chapmanb:
Sorry it looks like it's complaining about the https apt repository:
```
The method driver /usr/lib/apt/methods/https could not be found.
```
I swapped over to use http, rebased and pushed. Fingers crossed that'll fix it. Thanks for the heads up

Issue #1646, opened 2017-04-22
Azure template needs to add the user to the `docker` group


adamnovak:
By default the user can `sudo docker` but not `docker`, and this breaks `toil-vg`.

The toil-vg docs recommend:

```
sudo usermod -aG docker $USER
```

We might have to change it to add the main user and not whoever the setup script runs as (is it the same user?)

PR #1654, opened 2017-04-24
Update Azure template

### LINKED ISSUE #1644: Azure template doesn't work because of upped setuptools version requirement
Toil on Azure (at least) now needs setuptools 8.0, but the Azure template doesn't install it. This produces an error in the install logs, and Toil doesn't work right:

```
Installing branch master of BD2KGenomics/toil for Toil.
+ sudo pip install --pre 'git+https://github.com/BD2KGenomics/toil@master#egg=toil[mesos,azure]'
Collecting toil[azure,mesos] from git+https://github.com/BD2KGenomics/toil@master#egg=toil[mesos,azure]
  Cloning https://github.com/BD2KGenomics/toil (to master) to /tmp/pip-build-PwL3l5/toil
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-PwL3l5/toil/setup.py", line 110, in <module>
        version = importVersion()
      File "/tmp/pip-build-PwL3l5/toil/setup.py", line 90, in importVersion
        new = version_template.expand_()
      File "version_template.py", line 131, in expand_
        return ''.join("%s = %s\n" % (k, repr(resolve(k))) for k, v in variables.items())
      File "version_template.py", line 131, in <genexpr>
        return ''.join("%s = %s\n" % (k, repr(resolve(k))) for k, v in variables.items())
      File "version_template.py", line 127, in resolve
        v = v()
      File "version_template.py", line 85, in dockerShortTag
        return shortVersion()
      File "version_template.py", line 46, in shortVersion
        return _version(shorten=True)
      File "version_template.py", line 50, in _version
        return '-'.join(filter(None, [distVersion(),
      File "version_template.py", line 63, in distVersion
        raise RuntimeError("Setuptools version 8.0 or newer required. Update by running "
    RuntimeError: Setuptools version 8.0 or newer required. Update by running 'pip install setuptools --upgrade'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-PwL3l5/toil/
```

A command needs to be added to the template's setup script to install the correct pip version.

If we don't want future dependency changes to break the Azure templates, we should put them under CI.

adamnovak:
This closes #1644 and #1646, and also lets you deploy some Python packages to your Azure nodes (for things that Toil can't hot-deploy because they need binaries or something).

### DIFF
--- a/contrib/azure/agent-0.json
+++ b/contrib/azure/agent-0.json
@@ -140,6 +140,12 @@
         "description": "Branch on Github to pull cluster setup scripts and Toil from."
       }
     },
+    "pythonPackages": {
+      "type": "string",
+      "metadata": {
+        "description": "Extra Python package specifiers to install, like 'pysam>=1.0'. Space separated."
+      }
+    },
     "omsStorageAccount": {
       "type": "string",
       "metadata": {
--- a/contrib/azure/agent-gt0.json
+++ b/contrib/azure/agent-gt0.json
@@ -140,6 +140,12 @@
         "description": "Branch on Github to pull cluster setup scripts and Toil from."
       }
     },
+    "pythonPackages": {
+      "type": "string",
+      "metadata": {
+        "description": "Extra Python package specifiers to install, like 'pysam>=1.0'. Space separated."
+      }
+    },
     "omsStorageAccount": {
       "type": "string",
       "metadata": {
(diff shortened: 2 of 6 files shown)


adamnovak:
@cket Do you know what's wrong with Jenkins here? I don't think I did anything terrible enough to completely obliterate its test results. I've only touched the Azure templates, which aren't tested by CI.

cket:
Yeah, it looks like we hit the EC2 limit in the new account. I requested an increase and in the meantime this should pass now. Jenkins, test this please.

jessebrennan:
+1 this kept happening to me as well.

PR #1661, opened 2017-05-01
Add `import os`

susannasiebert:
I'm getting an error `NameError: global name 'os' is not defined` when using cwltoil. This PR should fix this error.


cket:
@susannasiebert thanks for the PR! Jenkins, test this please

PR #1665, opened 2017-05-02
Lower timeout for domain binding in Toil Clean

### LINKED ISSUE #1660: toil clean can take a while on AWS
For large job stores, `toil clean` can take tens of minutes to hours, at least on AWS. I think that may mean it's deleting every file individually. Maybe in these situations the job store could just drop the bucket / domain?

cket:
Resolves #1660

### DIFF
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -778,7 +778,10 @@ class AWSJobStore(AbstractJobStore):
                 retry timeout expires.
         """
         log.debug("Binding to job store domain '%s'.", domain_name)
-        for attempt in retry_sdb(predicate=lambda e: no_such_sdb_domain(e) or sdb_unavailable(e)):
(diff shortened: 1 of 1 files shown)


cket:
@joelarmstrong previously, Toil Clean would retry discovery of a SDB domain for up to 5 minutes, but if the job store is corrupted or incomplete then the domain(s) won't exist & we can needlessly wait for up to 10 minutes. 

In the ticket you brought up the possibility of dropping an entire bucket without deleting the contents, but I don't think S3 supports that option, as far as I can tell  http://docs.aws.amazon.com/AmazonS3/latest/dev/delete-or-empty-bucket.html#delete-bucket. 

There is a mention of using lifecycle rules to expire all files & let s3 handle the deletion, but the minimum expiration date is 24 hours.

joelarmstrong:
d'oh! You're definitely right about having to delete all objects, [the REST API docs even state so explicitly](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketDELETE.html). My bad, I didn't look thoroughly enough. Thanks for the patch to speed this up in some cases, though!

PR #1668, opened 2017-05-08
Set mesosMasterAddress default to private IP (resolves #1664)

### LINKED ISSUE #1664: mesosMasterAddress should default to leader's private IP, not localhost

```
import socket
import fcntl
import struct

def get_ip_address(ifname):
        s.fileno(),
        0x8915,  # SIOCGIFADDR
    )[20:24])

get_ip_address('eth0')
```

jessebrennan:
This sets the default address for the Mesos master to the private IP of the node where the workflow is launched. 

Because workflows are always launched from the leader node, the private IP is precisely what we need in 99% of circumstances and thus makes a sensible default.

First commit is the actual change. A second commit will follow that will adjust the tests to make use of the new feature.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -37,32 +37,13 @@ from toil import logProcessContext
 from toil.lib.bioio import addLoggingOptions, getLogLevelString, setLoggingFromOptions
 from toil.realtimeLogger import RealtimeLogger
 
+from socket import gethostbyname, gethostname
+
 logger = logging.getLogger(__name__)
 
 # This constant is set to the default value used on unix for block size (in bytes) when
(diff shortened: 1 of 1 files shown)


jessebrennan:
Jenkins test this please

joelarmstrong:
I don't mean to be contrary, but in its current state this PR doesn't do anything but add an unused import and an extra level of indirection when creating the string 'localhost:5050'. The previous commits had something that looked promising, but the net effect of all the commits is that the behavior is exactly the same as before, so I think we should keep #1664 open for now.

ejacox:
Please be contrary. I will open it back up.

Issue #1656, opened 2017-04-26
[cwl] Directory inputs and outputs in pipelines


qiukunlong:
Do you have  any plan to support Directory inputs and outputs in pipelines?

tetron:
Directory input/output support is pending in #1810

Issue #1459, opened 2017-01-26
return Directory as output returns a array of files using cwl


gijzelaerr:
hi! I'm trying to write a cwl workflow that outputs a directory:

https://github.com/gijzelaerr/vermeerkat-cwl/blob/master/h5toms.cwl#L16
https://github.com/gijzelaerr/vermeerkat-cwl/blob/dir_output_toil_bug/workflow.cwl#L9

using cwlref-runner this works correctly, it results in a directory in my output folder. But using toil it will return a flat list of files, it loses the nested structure. Not sure if i'm doing it wrong, but the cwlref-runner behavior is what i would expect and want.

thanks!

cket:
@chapmanb do you have any ideas about this?

humburg:
Support for directory inputs/outputs would indeed be great.

stefanoberri:
I would like to give it a go at implementing the Directory support. I am a bit worried that @chapmanb says " I don't have the cwltool/Toil expertise to know how to implement it". I am new to both tools as well, but I am happy to dive in. Help and guidance is welcome.

This issue is also reported in 1656
https://github.com/BD2KGenomics/toil/issues/1656

stefanoberri:
I have started working on these issues.
forked and created branch
https://github.com/stefanoberri/toil/tree/issues/1459-1656-cwl-directories

stefanoberri:
I have removed a couple of line that prevented Toil from proceed in case input was a Directory and it seem to be able to read the directory correctly. That seems too easy and I am not sure why the check was introduced in the first place by @chapmanb 
Any Idea? 
I will now look into *creating* directories.

chapmanb:
Stefano;
Thanks for looking at this. I added a lot of checks for things I didn't know how to test and ensure worked, so it could have been an necessarily conservative check. I had to reduce the complexity somewhat to get most of it ported, sorry if it was blocking you.

If things work and it passes the CWL test suite then it's no problem with me to remove that.

As an FYI, @tetron is planning to do work on revamping and better integrating cwltoil with cwltool changes and will be tackling directory support as part of this. He might be worth coordinating with as you move forward so you don't duplicate work. Thanks again.

ejacox:
I believe that @tetron was going to address this issue, too. Any comments @tetron?

stefanoberri:
Thank you @chapmanb for the reply.
[...]

PR #1677, opened 2017-05-12
CWL: Add support for InitalWorkDirRequirement

chapmanb:
The current toil integration with cwltool allows specification of
inputs into the initial working directory, so this enables this feature.
This is not generalized directory support (#1656 #1459), but does
allow writing inputs to the work directory making cross platform
CWL support for records easier.


mr-c:
Thank you @chapmanb !!!

PR #1681, opened 2017-05-16
Use hard links instead of copied files on import to file job store

### LINKED ISSUE #1680: Avoid copying on import to job store when possible
When using a file job store most of the time we can hard link files into the job store rather than create entire new copies

cket:
Resolves #1680


cket:
Jenkins test this please

cket:
Jenkins test this please

cket:
Failed due to exceeding number of s3 buckets. Jenkins, test this please

PR #1617, opened 2017-04-11
PBSPro support

brainstorm:
The commercial version of PBS only seems to support JSON output.
The oss version, only does XML.

This PR takes into account this quirk and tries to accomodate for different versions (privative and open).

Thanks @cjfields and @ohofmann for your support.

@cjfields Could you please try this version/commit in your NCSA system just in case I broke something?

@toil_mantainers, the `--disableCache` option is problematic, I would love advice on how to handle it: https://github.com/BD2KGenomics/toil/compare/master...brainstorm:pbspro?expand=1#diff-ec7c0fff4aad88175b4920cd3d1e5f8fR824

I'm aware it has been an issue in the past and @chapmanb managed to address it in PR https://github.com/BD2KGenomics/toil/pull/1090, but I'm not able to let the code below uncommented.

PR #1686, opened 2017-05-21
PBSPro (Torque's proprietary version) in addition to OSS's PBS

brainstorm:
This is a followup on PR #1617.

The child jobs' exit codes were not properly monitored by the leader. Flag differences between pbspro and the oss version are indeed confusing:

`qstat -x` means: XML on OSS version BUT provide exit codes on PBS pro... ¬_¬'''

Also the `qstat` polling 1 second polling is considered excessive by traditional HPC cluster operators, raised it to 10 seconds. 
 
Thanks to @cket and @hcraT for their invaluable insights on this issue over gitter: https://gitter.im/bd2k-genomics-toil/Lobby?at=5912e6b9ac693c532adb46d1

### DIFF
--- a/src/toil/batchSystems/torque.py
+++ b/src/toil/batchSystems/torque.py
@@ -31,13 +31,29 @@ logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
 
+def _pbsVersion():
+    """ Determines PBS/Torque version via pbsnodes
+    """
+    try:
+        subprocess.check_output(["pbsnodes", "--version"])
+    except subprocess.CalledProcessError as e:
+       if e.returncode != 0:
+            logger.error("Could not determine PBS/Torque version")
+       elif "PBSPro" in e.output:
+            return "pro"
+       else:
+            return "oss"
+        
 
 class TorqueBatchSystem(AbstractGridEngineBatchSystem):
-    
 
 
     # class-specific Worker
     class Worker(AbstractGridEngineBatchSystem.Worker):
+
+        
+        # Determine if we are in PBSPro or OSS variant of Torque/PBS batch system
+        version = _pbsVersion()
 
         """
         Torque-specific AbstractGridEngineWorker methods
@@ -49,12 +65,17 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             # Limit qstat to current username to avoid clogging the batch system on heavily loaded clusters
             #job_user = os.environ.get('USER')
             #process = subprocess.Popen(['qstat', '-u', job_user], stdout=subprocess.PIPE)
-            # -x shows exit status in PBSPro
-            process = subprocess.Popen(['qstat', '-x'], stdout=subprocess.PIPE)
+            # -x shows exit status in PBSPro, not XML output like OSS PBS
+            if version == "pro":
+                process = subprocess.Popen(['qstat', '-x'], stdout=subprocess.PIPE)
+            elif version == "oss":
+                process = subprocess.Popen(['qstat'], stdout=subprocess.PIPE)
+
+
             stdout, stderr = process.communicate()
 
             # qstat supports XML output which is more comprehensive, but PBSPro does not support it 
-            # so instead we stick with plain commandline "qstat" outputs
+            # so instead we stick with plain commandline qstat tabular outputs
             for currline in stdout.split('\n'):
                 items = currline.strip().split()
                 if items:
@@ -101,7 +122,10 @@ class TorqueBatchSystem(AbstractGridEngineBatchSystem):
             return result
 
         def getJobExitCode(self, torqueJobID):
-            args = ["qstat", "-x", "-f", str(torqueJobID).split('.')[0]]
+            if version == "pro":
+                args = ["qstat", "-x", "-f", str(torqueJobID).split('.')[0]]
+            elif version == "oss":
+                args = ["qstat", "-f", str(torqueJobID).split('.')[0]]
 
             process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
             for line in process.stdout:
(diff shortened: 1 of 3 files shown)


cket:
@brainstorm I'm glad you figured this out, it'll be awesome to get this patch in. Jenkins, test this please!

brainstorm:
I'm always torn with that "Permission denied" from Jenkins here in Toil land  :_(

mr-c:
Jenkins, test this please!

cket:
Jenkins, ok to test

ejacox:
Jenkins, ok to test

brainstorm:
From the current [jenkins run](http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1138/console):

```
_______________ ERROR collecting src/toil/batchSystems/torque.py _______________
src/toil/batchSystems/torque.py:47: in <module>
    class TorqueBatchSystem(AbstractGridEngineBatchSystem):
src/toil/batchSystems/torque.py:51: in TorqueBatchSystem
    class Worker(AbstractGridEngineBatchSystem.Worker):
src/toil/batchSystems/torque.py:55: in Worker
    version = _pbsVersion()
src/toil/batchSystems/torque.py:37: in _pbsVersion
    subprocess.check_output(["pbsnodes", "--version"])
/usr/lib/python2.7/subprocess.py:566: in check_output
    process = Popen(stdout=PIPE, *popenargs, **kwargs)
/usr/lib/python2.7/subprocess.py:710: in __init__
    errread, errwrite)
/usr/lib/python2.7/subprocess.py:1335: in _execute_child
    raise child_exception
E   OSError: [Errno 2] No such file or directory
```

According to [src/test/__init__.py](https://github.com/BD2KGenomics/toil/blob/master/src/toil/test/__init__.py#L334), `pbsnodes` should be in the test system/image's PATH. I'm not seeing why it is not finding it while running the tests. It is indeed present on our cluster:

```
In [2]: subprocess.check_output(["pbsnodes", "--version"])
Out[2]: 'pbs_version = PBSPro_13.1.0.160576\n'
```

chapmanb:
Roman;
I'm not sure if PBS is present on the test machine, it looks like that decorator would just skip tests if it doesn't find pbsnodes.

I think the underlying issue is that the `version` generation happens at the class level so isn't isolated to a function call, so it's leaking into cases where you're not actually using PBS.

The way I'd implement is to add the `_pbsVersion()` as an internal function on the `Worker`, and have it cache the output as a class attribute by setting `self._version` as well as returning it. Then on a second lookup you can check if `self._version` exists and return that instead of calling out again.

I think that will clean up the check to only happen when the PBS specific machinery gets instantiated. Hope that fixes it for you.

brainstorm:
Jenkins, ok to test

ejacox:
Hey @brainstorm. Did you resolve the pbsnodes issue?  The Jenkins UI has been down since yesterday. I am working on resolving the problem.

brainstorm:
@ejacox Almost there, taking advice from @chapmanb and moving `_pbsVersion` to the Worker level so that other non-torque tests do not fail on `_version`.

/cc @ohofmann

brainstorm:
Yet I'm a bit confused since I'm getting file not found errors whenever I try to define/call `_pbsVersion` from either inside Worker class or higher in the class hierarchy (as @ejacox commited a while ago) :/

src/toil/batchSystems/torque.py
```diff
@@ -31,29 +31,62 @@
...
+            except subprocess.CalledProcessError as e:
+               if e.returncode != 0:
+                    logger.error("Could not determine PBS/Torque version")
+               elif "PBSPro" in e.output:
+                    return "pro"
+               else:
+                    return "oss"
+        
+        # Determine if we are in PBSPro or OSS variant of Torque/PBS batch system
+        self._version = _pbsVersion()
```
mr-c: Maybe a self caching method (always call `_pbsVersion()`, which caches the result in `self._version`) instead of referring to the class level attribute (`self._version`) as the former will only get called when needed but the later is getting called inappropriately?

ejacox:
Hey @brainstrom, are you able to run tests locally? If not, why not?

brainstorm:
@ejacox The main reason I'm going for jenkins testing is that for some reason testing on our cluster takes a lot of time (~30-40min or more on a dedicated interactive node):

```
$ pytest batchSystems/batchSystemTest.py
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.13, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: /home/005/rg3930/dev/toil, inifile: setup.cfg
plugins: cov-2.3.1, marks-0.4, mock-1.1
collected 46 items

batchSystems/batchSystemTest.py ssssss........sssssssssssssssssssss....F.
```

So I assumed/self-conviced myself that your jenkins env is the authoritative way to test toil "safely" plus I just wanted to get this PR done/closed ;)

I'll try to debug/test locally without polluting this PR with more commits.

Thanks @mr-c for the hint(s), I was looking into `@properties` before taking a break, retaking this now...

ejacox:
Perfect! I was hoping you could get a faster turn around. 

You can also try something like:
```make test tests=<test_name>```

For example:
```make test tests=src/toil/test/sort/sortTest.py::SortTest::testSort```

**Comment by brainstorm** [comment]
Thanks @ejacox! Problem with make is that it doesn't seem to be friends with my anaconda environments (seems to require a python venv instead):

```
$ make test tests=src/toil/test/batchSystems/batchSystemTest.py::TorqueBatchSystemTest::testRunJobs
which: no docker in (... PATH ...)
Makefile:93: Cannot find 'docker' executable. Docker-related targets will be skipped.
A virtualenv must be active.
make: *** [check_venv] Error 1
```

`pytest` takes quite a while, but seems to complain the right way, so I'll stick with it for now...

brainstorm:
Alright thanks all @ejacox, @mr-c and @chapmanb, I just rebased the mess so this should be ready to ship 🚀  after jenkins (tests were passing already).

PR #1687, opened 2017-05-22
CWL: symlink jobStore files instead of copying

### LINKED ISSUE #1627: [CWL] Needs method to avoid copying files when a shared filesystem is available
Paths of `File` and `Directory` objects residing on an identified shared file system should be exempted from copying and just passed directly to the node.

To identify these shared filesystem a `cwltoil` command line option specifying the root path could be added.

Inspiration: 
> hcraT @hcraT 12:59
> Hi I'm writing a pipeline in cwl to run some analyses on biological data. I was able to use toil to run some tests.
> However I realized that toil copies my input files that are quite big. It is possible to avoid this behavior?
> I run my jobs on a batch system.
> I actually fed in as input to the pipeline symbolic links to the actual files.
> Thanks

https://gitter.im/bd2k-genomics-toil/Lobby?at=58f09d818fcce56b20ff77e4

> Pavlo Lutsik @lutsik 14:08
> I echo @hcraT. I probably have a very similar gridEngine-based setup, and copying of large files was a pain. I also implemented a similar workaround: simply eliminated the File type, working with string paths only. I had to adapt all the cwl wrappers and write my own cleanup steps, but the payoff in terms of performance was tremendous.

https://gitter.im/bd2k-genomics-toil/Lobby?at=58f0ada6f22385553d3e440f

chapmanb:
getFile would copy all input files from the jobStore previous due to a
refactoring bug: the `copy` function argument was removed but still
referenced; it picked up the `copy` module and was always true.

This moves to using symlinks to avoid the extra file copy. In doing so
it exposed some new issues, also fixed in this PR:

- Needed to correctly set basename and location in the File objects to
  avoid resolving the original Toil name instead of the actual file name.

- We don't support writable InitialWorkDirRequirement in CWL, but still
  support the more common readable case.

Fixes #1627

### DIFF
--- a/docs/running/running.rst
+++ b/docs/running/running.rst
@@ -65,6 +65,9 @@ support for the stable v1.0 specification, only lacking the following features:
   be enumerated as Files.
 - `File literals`_ that specify only ``contents`` to a File without an explicit
   file name.
+- Writable `InitialWorkDirRequirement
+  <http://www.commonwl.org/v1.0/CommandLineTool.html#InitialWorkDirRequirement>`
+  objects. Standard readable inputs do work.
 - Complex file inputs – from ExpressionTool or a default value, both of which do
   not yet get cleanly staged into Toil file management.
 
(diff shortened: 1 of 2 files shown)


cket:
@chapmanb looks good, thanks for fixing this!

PR #1694, opened 2017-05-29
hard links: resolve inputs with symlinks

chapmanb:
CJ;
Thank you for the work on hard links and avoiding file copying. I was testing this and ran into a problem when the input files were symlinks themselves. This resolves those to get it working with these cases as well. Thanks again.

Input files for hard linking into the filestore fail if they contain
relative symlinks in the input, since the hard links contain only
the original symlinks which are no longer valid:
```
$ ln -s base.txt input.txt
$ ls -lh
total 0
-rw-rw-r-- 1 chapmanb chapmanb 0 May 29 05:32 base.txt
lrwxrwxrwx 1 chapmanb chapmanb 8 May 29 05:32 input.txt -> base.txt
$ mkdir store
$ ln input.txt store/output.txt
$ ls -lh store/
total 0
lrwxrwxrwx 2 chapmanb chapmanb 8 May 29 05:32 output.txt -> base.txt
```
Supplement to #1681

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -169,7 +169,7 @@ class FileJobStore(AbstractJobStore):
                 os.close(fd)
                 os.unlink(absPath)
                 try:
-                    os.link(self._extractPathFromUrl(url), absPath)
+                    os.link(os.path.realpath(self._extractPathFromUrl(url)), absPath)
                 except OSError:
                     shutil.copyfile(self._extractPathFromUrl(url), absPath)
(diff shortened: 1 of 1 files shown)


ejacox:
Jenkins, test this please.

brainstorm:
This should pass now after #1681 got merged, right? From what I see in the jenkins logs, it seems to be 403 Forbidden SDB issue?

Jenkins, test this please.

ejacox:
Your right @brainstorm, it passed.

ejacox:
Thank you for the fix @chapmanb .

PR #1703, opened 2017-06-08
Fix hot-deployment with python-3-supporting libraries (fixes #1678)

### LINKED ISSUE #1678: Hotdeployment fails when certain python-3-supporting libraries are in the virtualenv
Some libraries, e.g. jinja2, have source files with python-3-only syntax, using the `async` keyword. This isn't a problem when loading them using python 2, because those source files just aren't loaded when using python 2.

But toil seems to be parsing every python file in the entire virtualenv when preparing to hotdeploy it. This ends up crashing because the `async` keyword isn't recognized:

```
Registered with framework ID 98da91a2-889a-4c12-a798-8128c2832c13-0011
Saving user script ModuleDescriptor(dirPath='/venv/lib/python2.7/site-packages', name='cactus.progressive.cactus_progressive', fromVirtualEnv=True) as a resource
  File "/venv/lib/python2.7/site-packages/jinja2/asyncsupport.py", line 22
    async def concat_async(async_gen):
            ^
SyntaxError: invalid syntax

  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/venv/lib/python2.7/site-packages/cactus/progressive/cactus_progressive.py", line 496, in <module>
    main()
  File "/venv/lib/python2.7/site-packages/cactus/progressive/cactus_progressive.py", line 461, in main
    halID = toil.restart()
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 699, in restart
    self._setupHotDeployment()
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 874, in _setupHotDeployment
    userScriptResource = userScript.saveAsResourceTo(self._jobStore)
  File "/usr/local/lib/python2.7/dist-packages/toil/resource.py", line 445, in saveAsResourceTo
    return self._getResourceClass().create(jobStore, self._resourcePath)
  File "/usr/local/lib/python2.7/dist-packages/toil/resource.py", line 85, in create
    with cls._load(leaderPath) as src:
  File "/usr/local/lib/python2.7/dist-packages/toil/resource.py", line 316, in _load
    zipFile.writepy(path)
  File "/usr/lib/python2.7/zipfile.py", line 1402, in writepy
    self.write(fname, arcname)
  File "/usr/lib/python2.7/zipfile.py", line 1118, in write
    st = os.stat(filename)
OSError: [Errno 2] No such file or directory: '/venv/lib/python2.7/site-packages/jinja2/asyncsupport.pyc'
```

Is there a reason to use `writepy` on each .py file rather than just naively dumping the entire virtualenv to a zip? It looks like it's trying to compile the file before zipping.

joelarmstrong:
As a side-effect now *every* file in the hot-deployed directory or virtualenv is deployed, not just Python files. I'm open to changing that so that the new code still has the old behavior, but I don't know if there's really a good reason to keep the old behavior. Non-python files can be included in virtualenvs, and it's conceivable that those might need to be hot-deployed as well.

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -28,7 +28,7 @@ from io import BytesIO
 from pydoc import locate
 from tempfile import mkdtemp
 from urllib2 import HTTPError
-from zipfile import ZipFile, PyZipFile
+from zipfile import ZipFile
 
 # Python 3 compatibility imports
 from bd2k.util.retry import retry
@@ -265,8 +265,8 @@ class FileResource(Resource):
 class DirectoryResource(Resource):
     """
     A resource read from a directory on the leader. The URL will point to a ZIP archive of the
-    directory. Only Python script/modules will be included. The directory may be a package but it
-    does not need to be.
+    directory. All files in that directory (and any subdirectories) will be included. The directory
+    may be a package but it does not need to be.
     """
 
     @classmethod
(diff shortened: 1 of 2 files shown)


ejacox:
If the additional files are small I don't see any problem with including them.

joelarmstrong:
Yes, I think so as well. But maybe there's a case to be made that when deploying user scripts (not virtualenvs) we should include only .py files--to avoid deploying a lot of unnecessary data when a user does something like:
```
./simpleToilScript.py ./huge-data.bam
```
though maybe that doesn't happen all that often

The tests should hopefully pass now, sorry about that, I only ran the hot-deployment tests and didn't know those unit tests even existed.

joelarmstrong:
Looks like this failed because of an unrelated cache race. Maybe we should turn debug logging on for the CI tests? Otherwise it's pretty hard to figure out what went wrong--and these tests are flaky, so it's usually not easy to reproduce the error locally.

joelarmstrong:
Not sure how to restart the build. This failed because of a flaky test.

Jenkins, test this please.

joelarmstrong:
This (second) flaky test is fixed in #1720. Let's try this one again in the meantime. Jenkins, test this please.

PR #1720, opened 2017-06-18
Fix race in ClusterScaler test (resolves #1719)

### LINKED ISSUE #1719: Race in ClusterScaler test
Encountered here: http://jenkins.cgcloud.info/job/toil-pull-requests/2868/testReport/junit/src.toil.test.provisioners.clusterScalerTest/ClusterScalerTest/testClusterScalingWithPreemptableJobs/

The basic problem is that the preemptable and non-preemptable scalers are handled by two different threads. There isn't any tracking of what nodes belong to what scaler, so the mock provisioner just removes nodes from both lists. When both scaler threads attempt to remove nodes at the same time, one will try to remove from an empty list.

joelarmstrong:
The test was flaky because both the preemptable and non-preemptable
scalers would attempt to remove from the same list of workers. This
keeps track of which nodes belong to which scaler, which should fix
the problem.

### DIFF
--- a/src/toil/test/provisioners/clusterScalerTest.py
+++ b/src/toil/test/provisioners/clusterScalerTest.py
@@ -203,7 +203,7 @@ class MockBatchSystemAndProvisioner(AbstractScalableBatchSystem, AbstractProvisi
         # for jobs we create two parallel instances of the following class
         self.config = config
         self.secondsPerJob = secondsPerJob
-        self.delegates = [self.Delegate(), self.Delegate()]
+        self.delegates = [self.Delegate(preemptable=False), self.Delegate(preemptable=True)]
         self.provisioner = self
         self.batchSystem = self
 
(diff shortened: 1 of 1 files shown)


src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -354,17 +342,20 @@ def stop(self):
                     return time.time() - self.startTime
 
             for i in xrange(numNodes):
```
jessebrennan: PyCharm bugs me if `i` is used instead of `_`.

src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -308,23 +307,12 @@ def getNumberOfJobsIssued(self):
...
 
         def getNodes(self):
-            return {address: NodeInfo(coresTotal=0, coresUsed=0, requestedCores=1,
-                                      memoryTotal=0, memoryUsed=0, requestedMemory=1,
-                                      workers=1 if w.busyEvent.is_set() else 0)
-                    for address, w in enumerate(self.workers)}
-
-        # AbstractProvisioner functionality
-
-        def setNodeCount(self, numNodes):
```
jessebrennan: setNodeCount doesn't need to be implemented?

joelarmstrong: The setNodeCount method got moved away from the provisioner a while ago, but I guess no one removed the dead code in this test. I only noticed it was dead when changing the way `_removeNodes` was called.

src/toil/test/provisioners/clusterScalerTest.py
joelarmstrong: Oops, you're totally right, good catch. Sadly this got merged already--let me see if github handles pushing to the same branch or if fixing that would require a new PR.

jessebrennan:
Makes sense. I just realized my review was after the merge lol. Oh well!

PR #1724, opened 2017-06-19
Revert "Set mesosMasterAddress default to private IP (resolves #1664)"

### LINKED ISSUE #1664: mesosMasterAddress should default to leader's private IP, not localhost

jessebrennan:
This PR was not ready to be merged.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -20,7 +20,6 @@ import re
 import sys
 import tempfile
 import time
-import socket
 from argparse import ArgumentParser
 from threading import Thread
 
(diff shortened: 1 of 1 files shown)


jessebrennan:
@ejacox I was thinking we could revert this PR and then I could resume work on the old one to figure out what may be causing a failure.

ejacox:
Sounds good @jessebrennan.

jessebrennan:
@ejacox This is weird. How could reverting back to a previous version of master cause a failure? Unless either master was broken before, or something is wrong with Jenkins.

jessebrennan:
Jenkins, test this please

jessebrennan:
@ejacox looks like this is passing as expected and can be merged. I will continue the fix on the other branch.

jessebrennan:
In offline discussion with @ejacox we decided to just leave as is and the new PR with override.

PR #1726, opened 2017-06-21
Skip provisioner utils test if rsync version is outdated or unavailable (resolves #1725)

### LINKED ISSUE #1725: Running integration tests locally on OSX can cause mysterious fail with rsync `--protect-args`
Tests assume that the machine running them has `rsync` installed and that the option `--protect-args` is a valid option. 

The version of `rsync` that ships with OSX by default is 2.6.9 which does not have this option (it is added in 3.0.0).

Adding a warning into the test to inform the user that they need to update their version of rsync and/or skipping the test if rsync is too low of version are solutions.

jessebrennan:
Previously if you had an outdated version of rsync installed (<3.0.0, like the one that comes by default with OSX) running the provisioner utils tests would fail.

This PR adds a decorator that checks if the user has rsync installed and makes sure that the version is sufficient, skipping the tests otherwise.

### DIFF
--- a/src/toil/test/__init__.py
+++ b/src/toil/test/__init__.py
@@ -20,6 +20,7 @@ import os
 import re
 import shutil
 import signal
+import subprocess
 import tempfile
 import threading
 import time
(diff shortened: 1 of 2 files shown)


jessebrennan:
@ejacox This passes locally. Can you review? I'm not 100% sure that I did the decorator correctly—I just copied the style of others.

ejacox:
You can test the decorator by changing the program name or version and checking that it skips the test.

jessebrennan:
Indeed I have tested it locally by using different versions of rsync and changing the name so as to not be recognized by the test. All cases behave as expected although instead of skipping the test and printing an error message, the test just passes quickly.

src/toil/test/__init__.py
```diff
@@ -245,6 +246,24 @@ def _mark_test(name, test_item):
         return MarkDecorator(name)(test_item)
 
 
+def needs_rsync(test_item):
```
jessebrennan: How about `needs_rsync3`?

src/toil/test/__init__.py
```diff
@@ -245,6 +246,24 @@ def _mark_test(name, test_item):
...
+    """
+    Use as a decorator before test classes or methods that depend on any features used in rsync
+    version 3.0.0+
+    """
```
jessebrennan: I'm not understanding your comment here

Issue #1700, opened 2017-06-07
Uninformative integration testing failures


natanlao:
Exhibit A: http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1145/console (#1572 4f52f1f02c8afdbf1c304e1b377c220fa76fd201)
Exhibit B: http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1039/console (#1506 84807b483eb4e7b81c4c60f51d4d84be3a1bf948)

Jenkins recognizes that testing suite fails but cannot identify which test failed. No traceback in console output.

In the case of Exhibit B, suite fails because of infinite loop that causes suite to time out - the problematic test, however, is not identified.

jessebrennan:
More examples:
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1172/console (#1392)
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1176/console (#1572)
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1177/console (#1724)
- http://jenkins.cgcloud.info/job/toil-pull-requests/2815/console (#1668)

Is this happening on all branches?

This is show-stopping for me as is. I'm not really sure what causes this to happen or how to go about trying to fix it.

jessebrennan:
No longer show stopping. I think I have found the source of the problem in at least some of the cases. If when running integration tests on specifies only a particular test such as 
```
$ make integration_test tests=src/toil/test/utils/utilsTest.py::UtilsTest::testAWSProvisionerUtils
```
the test may pass successfully, but `make` will fail overall. I suspect that this is because multiple threads are launched to run each test from a list within run_tests.py and since only one test is specified, the threads with nothing to do fail.

This theory also makes sense with another one of the cases I was experiencing the error. I had commented out one of the tests and `make` would "fail" despite none of the tests failing.

jessebrennan:
This isn't completely fixed in all cases. When a test hangs we still don't have an adequate reason as to why. This is being addressed in #1738 and should be in soon.

jessebrennan:
This is resolved by #1738.

PR #1731, opened 2017-06-23
Set mesosMasterAddress default to private IP (resolves #1664)

### LINKED ISSUE #1664: mesosMasterAddress should default to leader's private IP, not localhost

jessebrennan:
Note this is the same as #1668 but reopened.

### DIFF
--- a/src/toil/batchSystems/options.py
+++ b/src/toil/batchSystems/options.py
@@ -15,6 +15,12 @@ from __future__ import absolute_import
 
 from .registry import batchSystemFactoryFor, defaultBatchSystem, uniqueNames
 
+import socket
+
+
+def getLocalIP():
+    # may return localhost on some systems (not osx and coreos) https://stackoverflow.com/a/166520
+    return socket.gethostbyname(socket.gethostname())
 
 def _parasolOptions(addOptionFn):
     addOptionFn("--parasolCommand", dest="parasolCommand", default=None,
@@ -90,7 +96,7 @@ def setDefaultOptions(config):
     config.linkImports = False
 
     # mesos
-    config.mesosMasterAddress = 'localhost:5050'
+    config.mesosMasterAddress = '%s:5050' % getLocalIP()
 
     # parasol
     config.parasolCommand = 'parasol'
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -20,7 +20,6 @@ import re
 import sys
 import tempfile
 import time
-import socket
 import uuid
 from argparse import ArgumentParser
 from threading import Thread
@@ -35,7 +34,6 @@ from bd2k.util.humanize import bytes2human
 from toil import logProcessContext
 from toil.lib.bioio import addLoggingOptions, getLogLevelString, setLoggingFromOptions
 from toil.realtimeLogger import RealtimeLogger
-from toil.batchSystems.options import setOptions as setBatchOptions
 from toil.batchSystems.options import addOptions as addBatchOptions
 from toil.batchSystems.options import setDefaultOptions as setDefaultBatchOptions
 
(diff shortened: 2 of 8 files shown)


jessebrennan:
Jenkins, test this please.

joelarmstrong:
Sort of off-topic, but this test failure looks like a repeat of #1700. It looks like there's some (unrelated) infinite-loop problem with the GoogleJobStore, but it's hard to pin it down. I wonder if, instead of using the jenkins timeout, which unceremoniously and silently kills everything, we could use a [pytest timeout](https://pypi.python.org/pypi/pytest-timeout), so we know which test failed? That would help narrow down the possibilities--these flaky tests are super annoying and we should track them down if possible.

jessebrennan:
Joel that's a great idea,  I didn't even know that was an option!

Also are you suggesting that my code is not responsible for the failure?

On Jun 23, 2017 9:48 PM, "Joel Armstrong" <notifications@github.com> wrote:

Sort of off-topic, but this test failure looks like a repeat of #1700
<https://github.com/BD2KGenomics/toil/issues/1700>. It looks like there's
some (unrelated) infinite-loop problem with the GoogleJobStore, but it's
hard to pin it down. I wonder if, instead of using the jenkins timeout,
which unceremoniously and silently kills everything, we could use a pytest
timeout <https://pypi.python.org/pypi/pytest-timeout>, so we know which
test failed? That would help narrow down the possibilities--these flaky
tests are super annoying and we should track them down if possible.

—
You are receiving this because you authored the thread.

Reply to this email directly, view it on GitHub
<https://github.com/BD2KGenomics/toil/pull/1731#issuecomment-310813970>,
or mute
the thread

joelarmstrong:
On closer inspection it looks like the failure is not in the Google jobStore tests but in the sort tests, so maybe this is caused by your code after all. My bad! I think the point about pytest timeouts still stands, though.

(I think the failure is probably caused by [the mesos test support libs](https://github.com/BD2KGenomics/toil/blob/master/src/toil/batchSystems/mesos/test/__init__.py#L68), which specify a master IP of 127.0.0.1. Mesos slaves reject any master that advertises a different IP than it's contacted on. Changing those to use the same "get external IP" logic will probably fix things.)

jessebrennan:
Jenkins, test this please

jessebrennan:
Jenkins test this please

jessebrennan:
Well cool! Looks like this is working now too! @ejacox ready for review

ejacox:
It looks like a merge took out the call to getLocalIP.

jessebrennan:
@ejacox Hah! okay, I'm on it. Thanks

jessebrennan:
It looks like all of the integration test failures are due to the IAM outage

jessebrennan:
HAH! Okay I'm going to squash and rebase this terribly ugly branch

jessebrennan:
Ahhhh... Down to just four beautiful commits. This is ready for review @ejacox

jessebrennan:
I ran the failed test (src.toil.test.provisioners.aws.awsProvisionerTest.AWSAutoscaleTest.testSpotAutoScale) locally and it passed fine. My suspicion is that the failure was transient, but I don't have a good reason why.

Also it is suspicious that _this_ test failed given that it was modified to test the new default value of `--mesosMaster`. I don't think the failure was related however because if it was, the staticAutoScaleTest would have probably failed as well. 

I will test again. Jenkins test this please.

jessebrennan:
@ejacox merge?

PR #1738, opened 2017-06-26
Add timeout for running pytest (resolves #1735)

### LINKED ISSUE #1735: Timeout tests with PyTest instead of Jenkins for better debugging hanging tests
Currently, if a test hangs Jenkins eventually shuts it down. Unfortunately this leaves no indication of which test was running or what may have been causing it to hang.

By making tests timeout from [PyTest timeout](https://pypi.python.org/pypi/pytest-timeout) we can get a full trace as well as information on which test was hanging.

This will also probably be enough to fix #1700.

Idea credit to [Joel](https://github.com/BD2KGenomics/toil/pull/1731#issuecomment-310813970).

jessebrennan:
This amends the call to `pytest` in run_test.py by adding the option `--timeout=600` which gives any particular test 10 minutes to run. I imagine this may be enough time for all of the tests. If it is not, we can increase the overall timeout by adjusting the option above, or any particularly lengthy tests can be given their own timeout by adding the decorator `@pytest.mark.timeout(XXX)` where `XXX` is some number of seconds.

This passes tests locally, but pytest-timeout may need to be installed on Jenkins for this to work.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -41,7 +41,7 @@ test_suites = {
 
 
 def run_tests(keywords, index, args):
-    args = [sys.executable, '-m', 'pytest', '-vv',
+    args = [sys.executable, '-m', 'pytest', '-vv', '--timeout=600',
             '--junitxml', 'test-report-%s.xml' % index,
             '-k', keywords] + args
     log.info('Running %r', args)
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ check_build_reqs:
 
 
 prepare: check_venv
-	$(pip) install sphinx==1.5.5 mock==1.0.1 pytest==2.8.3 stubserver==1.0.1
+	$(pip) install sphinx==1.5.5 mock==1.0.1 pytest==2.8.3 stubserver==1.0.1 pytest-timeout==1.2.0
 
 
 check_venv:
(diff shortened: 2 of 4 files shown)


jessebrennan:
Jenkins, test this please.

jessebrennan:
Boto throws 400 error. Let's try again. Jenkins, test this please.

jessebrennan:
@ejacox I moved this to sprint 7. It is essentially done—just waiting for tests to pass since boto has been throwing 400s. Getting this in the next release will make future development much easier.

run_tests.py
```diff
@@ -41,7 +41,7 @@
 
 
 def run_tests(keywords, index, args):
-    args = [sys.executable, '-m', 'pytest', '-vv',
+    args = [sys.executable, '-m', 'pytest', '-vv', '--timeout=600',
```
ejacox: What does setting this value do?

ejacox: Is it automatically read by pytest?

jessebrennan:
@ejacox this is ready to merge!

run_tests.py
jessebrennan: Yep! see [here]( https://pypi.python.org/pypi/pytest-timeout).

PR #1743, opened 2017-06-29
Try waiting for worker before checking volume (resolves #1742)

### LINKED ISSUE #1742: Sporadic integration test failure due to KeyError: '/dev/xvda'
The test to fail is AWSStaticAutoscaleTest.testAutoScale.

Example of the trace
```
  File "/mnt/ephemeral/workspace/toil-pull-request-integration-tests/src/toil/test/provisioners/aws/awsProvisionerTest.py", line 249, in testAutoScale
    self._test(spotInstances=False)
  File "/mnt/ephemeral/workspace/toil-pull-request-integration-tests/src/toil/test/provisioners/aws/awsProvisionerTest.py", line 128, in _test
    self.launchCluster()
  File "/mnt/ephemeral/workspace/toil-pull-request-integration-tests/src/toil/test/provisioners/aws/awsProvisionerTest.py", line 278, in launchCluster
    rootBlockDevice = workers[0].block_device_mapping["/dev/xvda"]
KeyError: '/dev/xvda'
```

This is happening across multiple branches. This was most likely introduced in #1572 by not waiting for the worker to be running before checking if the volume is mounted.

jessebrennan:
This forces the worker to wait before we check for the volume. Since the failure is sporadic, this will be sort of difficult to test. 

The wait function used (from CGCloud) may or may not be sufficient. Another approach is to retry accessing the volume until success. If this first approach doesn't work that will be next.

### DIFF
--- a/src/toil/test/provisioners/aws/awsProvisionerTest.py
+++ b/src/toil/test/provisioners/aws/awsProvisionerTest.py
@@ -23,6 +23,7 @@ import time
 import pytest
 from boto.ec2.blockdevicemapping import BlockDeviceType
 from boto.exception import EC2ResponseError
+from cgcloud.lib.ec2 import wait_instances_running
 
 from toil.provisioners.aws.awsProvisioner import AWSProvisioner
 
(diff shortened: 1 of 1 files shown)


jessebrennan:
@ejacox well it passed. I think we should merge and if we keep getting those sporadic failures then we can try a more sophisticated fix. Thoughts?

jessebrennan:
👍

jessebrennan:
@ejacox have you noticed any `KeyError: '/dev/xvda'` failures recently? Do you think this fix was totally effective?

joelarmstrong:
Just got one yesterday:

http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1260/testReport/junit/src.toil.test.provisioners.aws.awsProvisionerTest/AWSStaticAutoscaleTest/testSpotAutoScale/

I don't think this patch fixed any of the sporadic errors, because `worker.block_device_mapping` is exactly the same before the wait as it is after. (It's not a dynamic call: `BlockDeviceMapping` is just a subclass of `dict`.)

I think the wait is a good idea, though. You would just have to update the instance metadata after the wait, either by using [the values that `wait_instances_running` yields](https://github.com/BD2KGenomics/cgcloud/blob/f89a416b3153e332e47ca5f5476fe73049b35bf7/lib/src/cgcloud/lib/ec2.py#L231), or by calling [`worker.update()`](http://boto.cloudhackers.com/en/latest/ref/ec2.html#boto.ec2.instance.Instance.update).

PR #1744, opened 2017-07-03
Allow clusters with multiple node types

adderan:
Not ready for review yet - this is kind of a mess after resolving merge conflicts. Opening PR to see test results.

### DIFF
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -110,7 +110,7 @@ class MesosBatchSystem(BatchSystemSupport,
         self.hostToJobIDs = {}
 
         # see self.setNodeFilter
-        self.nodeFilter = []
+        self.nodeFilter = set()
 
         # Dict of launched jobIDs to TaskData objects
         self.runningJobMap = {}
@@ -149,10 +149,15 @@ class MesosBatchSystem(BatchSystemSupport,
         self.lastTimeOfferLogged = 0
         self.logPeriod = 30  # seconds
 
+        self.ignoredNodes = set()
+
         self._startDriver()
 
     def setUserScript(self, userScript):
         self.userScript = userScript
+
+    def ignoreNode(self, nodeAddress):
+        self.ignoredNodes.add(nodeAddress)
 
     def issueBatchJob(self, jobNode):
         """
@@ -382,6 +387,11 @@ class MesosBatchSystem(BatchSystemSupport,
         unableToRun = True
         # Right now, gives priority to largest jobs
         for offer in offers:
+            if offer.hostname in self.ignoredNodes:
+                log.debug("Declining offer %s because node %s is designated for termination" %
+                        (offer.id.value, offer.hostname))
+                driver.declineOffer(offer.id)
+                continue
             runnableTasks = []
             # TODO: In an offer, can there ever be more than one resource with the same name?
             offerCores, offerMemory, offerDisk, offerPreemptable = self._parseOffer(offer)
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -110,7 +110,7 @@ class MesosBatchSystem(BatchSystemSupport,
         self.hostToJobIDs = {}
 
         # see self.setNodeFilter
-        self.nodeFilter = set()
+        self.nodeFilter = []
 
         # Dict of launched jobIDs to TaskData objects
         self.runningJobMap = {}
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -159,6 +159,9 @@ class MesosBatchSystem(BatchSystemSupport,
     def ignoreNode(self, nodeAddress):
         self.ignoredNodes.add(nodeAddress)
 
+    def unignoreNode(self, nodeAddress):
+        self.ignoredNodes.remove(nodeAddress)
+
     def issueBatchJob(self, jobNode):
         """
         Issues the following command returning a unique jobID. Command is the string to run, memory
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -112,7 +112,8 @@ class Leader:
         # Map of batch system IDs to IsseudJob tuples
         self.jobBatchSystemIDToIssuedJob = {}
 
-
+        # Number of preempetable jobs currently being run by batch system
+        self.preemptableJobsIssued = 0
 
         # Tracking the number service jobs issued,
         # this is used limit the number of services issued to the batch system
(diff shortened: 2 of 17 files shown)


adderan:
Looks like the failure is in JobServiceTest, which is also failing non-deterministically on the master branch. So I think this PR is ready for review whenever.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -149,11 +149,16 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
...
+        self.ignoredNodes = set()
+
         self._startDriver()
 
     def setUserScript(self, userScript):
         self.userScript = userScript
 
+    def ignoreNode(self, nodeAddress):
+        self.ignoredNodes.add(nodeAddress)
+
```
joelarmstrong: I like this ignored node stuff. But could we add a method to *remove* ignored nodes to AbstractBatchSystem, and call it from clusterScaler at the same time it removes a node from its ignoredNodes set? (clusterScaler.py line 597 as of right now.) It's not out of the question that an instance could be marked to be ignored, terminate, then hours later a new instance might be launched that happens to get the same IP.

src/toil/leader.py
```diff
@@ -572,22 +572,24 @@ def getNumberOfJobsIssued(self, preemptable=None):
...
 
-    def getNumberAndAvgRuntimeOfCurrentlyRunningJobs(self):
+    def getNumberAndAvgRuntimeOfCurrentlyRunningJobs(self, nodeShape=None):
         """
         Returns a tuple (x, y) where x is number of currently running jobs and y
         is the average number of seconds (as a float)
         the jobs have been running for.
         """
         runningJobs = self.batchSystem.getRunningBatchJobIDs()
+        runningJobs = {jobID:runningJobs[jobID] for jobID in runningJobs if getNodeShapeForJob(jobShape=self.jobBatchSystemIDToJobNode[jobID], nodeShapes=self.provisioner.nodeShapes) == nodeShape}
```
joelarmstrong: getNodeShapeForJob isn't defined anywhere.

joelarmstrong: Similarly, the jobBatchSystemIDToJobNode attribute doesn't appear to be set anywhere either. Actually, it looks like this entire function is dead code, which is why these undefined variables don't cause an error. Can we remove this? (and, potentially, the version in clusterScalerTest?)

src/toil/provisioners/clusterScaler.py
```diff
@@ -626,18 +691,14 @@ def _getInfo(allMesosNodes, ip):
...
-            raise RuntimeError('Provisioner could not terminate all autoscaled nodes. There are '
-                               '%s nodes left in the cluster, %s of which were statically provisioned' % (totalNodes, len(self.getStaticNodes(preemptable)))
-                               )
-        elif totalNodes < len(self.scaler.provisioner.getStaticNodes(preemptable)):  # ignore static nodes
-            raise RuntimeError('Provisioner incorrectly terminated statically provisioned nodes.')
-
+        for nodeShape in self.nodeShapes:
+            preemptable = nodeShape.preemptable
+            nodeType = self.nodeShapeToType[nodeShape]
+            totalNodes = self.setNodeCount(nodeType=nodeType, numNodes=0, preemptable=preemptable, force=True)
```
joelarmstrong: pyflakes says `local variable 'totalNodes' is assigned to but never used`. It looks like the call to self.setNodeCount is necessary, but maybe we can just ignore the return value?

src/toil/provisioners/clusterScaler.py
```diff
@@ -468,49 +512,63 @@ def setNodeCount(self, numNodes, preemptable=False, force=False):
...
+                self.ignoredNodes.add(node.privateIP)
+                self.scaler.leader.batchSystem.ignoreNode(node.privateIP)
+
+            if not force:
+                # Filter out nodes with jobs still running. These
+                # will be terminated in _removeIgnoredNodes later on
+                # once all jobs have finished, but they will be ignored by
+                # the batch system and cluster scaler from now on
+                nodesToTerminate = [(node,nodeInfo) for (node,nodeInfo) in nodesToTerminate if nodeInfo is not None and nodeInfo.workers < 1]
+            #nodesToTerminate = nodesToTerminate[:numNodes]
```
joelarmstrong: stray commented out code, looks like this can be removed

adderan:
Jenkins, test this please.

adderan:
Jenkins, test this please.

jessebrennan:
@adderan I've run into this $#!+ before where run_tests.py is out of sync with your actual tests and it's a pain in the @$$. I think this should fix it. 

Also you may want to rebase #1773.

jessebrennan:
@joelarmstrong I think this is ready now.

joelarmstrong:
Testing this now. It may take a few days.

src/toil/provisioners/clusterScaler.py
```diff
@@ -126,87 +131,91 @@ def addToReservation():
...
+                            # If the job would fit, but is longer than the total node allocation
+                            # extend the node allocation
+                            elif y.nReservation == None and x == nodeReservation:
+                                # Extend the node reservation to accommodate jS
+                                y.nReservation = NodeReservation(nodeShape)
+                            reservedSuccessfully = True
+
+                        else: # Does not fit, reset
+                            x = y.nReservation
+                            t = 0
```
ejacox: This could be broken up into sub functions to make it more readable.

joelarmstrong:
Really sorry for taking so long with this. I tracked down one small bug -- I think it only triggers in very special cases -- where reservedSuccessfully could get set even though the job never fits entirely in a reservation. (It occurs when the job fits the reservation so far (from `x` to `y`), but when there are still more reservations left to go before the reservations cover the job's full runtime.)

I've pushed a fix for that (now it seems to be scaling up well). But in the process I actually found a few bugs in the original binpacking algorithm, I think--not new to this PR, though. Basically, the "wallTime" for reservations doesn't always seem to be set correctly. But since the existing binpacking algorithm has been proven to work well before (even if accidentally!), I think this PR is good to go. Any bugs in the binpacking algorithm already existed before.

I'm refactoring the binpacking a bit, to try to track down these bugs and so it's relatively testable--but that doesn't need to go in this PR, I think this is OK to merge as is.

src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -153,47 +159,113 @@ def testClusterScaling(self):
...
+        
+        mock = MockBatchSystemAndProvisioner(config, secondsPerJob=2.0)
+        clusterScaler = ClusterScaler(mock, mock, config)
+        clusterScaler.start()
+        mock.start()
+
+        try:
+            #Add small jobs
+            map(lambda x: mock.addJob(jobShape=smallNode), range(numJobs))
+            map(lambda x: mock.addJob(jobShape=mediumNode), range(numJobs))
```
adderan: This is what broke the last build, should be list(map(

ejacox: OK. Thanks @adderan

PR #1745, opened 2017-07-03
Allow batch creation of jobs

adderan:
Create SDB items representing the job graph with a batch insertion, to avoid large numbers of transactions when there are many child jobs.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -993,7 +993,7 @@ class Job(JobLikeObject):
         """
         # set _config to determine user determined default values for resource requirements
         self._config = jobStore.config
-        return jobStore.create(JobNode.fromJob(self, command=command,
+        return jobStore.getJobGraph(JobNode.fromJob(self, command=command,
                                                predecessorNumber=predecessorNumber))
 
     def _makeJobGraphs(self, jobGraph, jobStore):
@@ -1005,6 +1005,9 @@ class Job(JobLikeObject):
             jobs = map(lambda successor:
                 successor._makeJobGraphs2(jobStore, jobsToJobGraphs), successors)
             jobGraph.stack.append(jobs)
+
+        #Write the jobs to the jobStore all at once
+        jobStore.batchCreate(jobsToJobGraphs.values())
         return jobsToJobGraphs
 
     def _makeJobGraphs2(self, jobStore, jobsToJobGraphs):
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -993,7 +993,7 @@ class Job(JobLikeObject):
         """
         # set _config to determine user determined default values for resource requirements
         self._config = jobStore.config
-        return jobStore.getJobGraph(JobNode.fromJob(self, command=command,
+        return jobStore.create(JobNode.fromJob(self, command=command,
                                                predecessorNumber=predecessorNumber))
 
     def _makeJobGraphs(self, jobGraph, jobStore):
@@ -1005,9 +1005,6 @@ class Job(JobLikeObject):
             jobs = map(lambda successor:
                 successor._makeJobGraphs2(jobStore, jobsToJobGraphs), successors)
             jobGraph.stack.append(jobs)
-
-        #Write the jobs to the jobStore all at once
-        jobStore.batchCreate(jobsToJobGraphs.values())
         return jobsToJobGraphs
 
     def _makeJobGraphs2(self, jobStore, jobsToJobGraphs):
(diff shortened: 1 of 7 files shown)


src/toil/job.py
```diff
@@ -1005,6 +1005,9 @@ def _makeJobGraphs(self, jobGraph, jobStore):
             jobs = map(lambda successor:
                 successor._makeJobGraphs2(jobStore, jobsToJobGraphs), successors)
             jobGraph.stack.append(jobs)
+
+        #Write the jobs to the jobStore all at once
+        jobStore.batchCreate(jobsToJobGraphs.values())
```
joelarmstrong: This serializes the job graphs, but they're still overwritten serially in _serialiseJobs below, right? I don't think the jobGraph command is set before this point, so the overwrite is still necessary as written. (Moving the lines which set .command up to here would probably work.)

src/toil/jobStores/aws/jobStore.py
```diff
@@ -248,17 +248,26 @@ def _awsJobToItem(self, job):
...
-                assert self.jobsDomain.put_attributes(job.jobStoreID, item)
-        return job
+                assert self.jobsDomain.put_attributes(jobGraph.jobStoreID, item)
+        return jobGraph
+
+    def batchCreate(self, jobGraphs):
+        items = {job.jobStoreID:self._awsJobToItem(job) for job in jobGraphs}
+        for attempt in retry_sdb():
+            with attempt:
+                assert self.jobsDomain.batch_put_attributes(items)
```
joelarmstrong: I think this needs to be chunked into 25-item bunches to work properly. I don't know if boto chunks the items for you, but my guess is that it doesn't.

src/toil/jobStores/fileJobStore.py
```diff
@@ -87,17 +87,25 @@ def destroy(self):
...
-        self.update(job)
         return job
+        
+    def create(self, jobNode):
+        jobGraph = self.getJobGraph(jobNode)
+        # Write job file to disk
+        self.update(jobGraph)
+        return jobGraph
+
+    def batchCreate(self, jobGraphs):
```
joelarmstrong: Nitpick: I think it may be better to put the various "not actually batching" batchCreates as a default batchCreate method in abstractJobStore which just loops over the jobs and calls `self.create`. That way we could get rid of the other 4 batchCreates that don't really do multiple writes at once.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -361,7 +361,19 @@ def callback(f2):
             self.assertFalse(master.exists(jobOnMaster.jobStoreID))
             # TODO: Who deletes the shared files?
 
-
+        def testBatchCreate(self):
+            master = self.master
+            masterRequirements = dict(memory=12, cores=34, disk=35, preemptable=True)
+            jobGraphs = []
+            for i in range(10):
```
joelarmstrong: It's better to do 50 or 100, IMO, to test the batch size limits. SimpleDB limit is 25, not sure about other job stores. Better to go over than to be too conservative.

src/toil/job.py
```diff
@@ -993,7 +993,7 @@ def _createEmptyJobGraphForJob(self, jobStore, command=None, predecessorNumber=0
         """
         # set _config to determine user determined default values for resource requirements
         self._config = jobStore.config
-        return jobStore.create(JobNode.fromJob(self, command=command,
+        return jobStore.getJobGraph(JobNode.fromJob(self, command=command,
                                                predecessorNumber=predecessorNumber))
```
joelarmstrong: I hate making style comments, but this one jumped out at me: the indent isn't aligned anymore.

adderan:
Ready for review.

src/toil/jobStores/googleJobStore.py
```diff
@@ -113,12 +113,12 @@ def destroy(self):
...
-        job = JobGraph(jobStoreID=jobStoreID, unitName=jobNode.name, jobName=jobNode.job,
-                       command=jobNode.command, remainingRetryCount=self._defaultTryCount(),
-                       logJobStoreFileID=None, predecessorNumber=jobNode.predecessorNumber,
-                       **jobNode._requirements)
+        job = self.getJobGraph(jobNode)
         self._writeString(jobStoreID, cPickle.dumps(job, protocol=cPickle.HIGHEST_PROTOCOL))
+        if hasattr(self, "_batchedJobGraphs") and self._batchedJobGraphs is not None:
+            self._batchedJobGraphs.append(job)
+        else:
+            self._writeString(jobStoreID, cPickle.dumps(job, protocol=cPickle.HIGHEST_PROTOCOL))
```
ejacox: Where is the context manager? Does Google do batch creation? This code is broken at the moment. Maybe just make a note to add the batch ability if Google has it.

ejacox:
Thank you @adderan

PR #1750, opened 2017-07-08
Fix AWS readFileStream deadlock on partial reads (resolves #1748)

### LINKED ISSUE #1748: readFileStream deadlocks when not read through completely
I think I was aware of this at one point, but using readFileStream without reading all the way to EOF will totally deadlock the worker process on anything but the file jobStore. This is pretty unintuitive (surely this is one of the only use cases for streaming the file?). It may be especially confusing to new users who aren't used to toil deadlocks.

Unfortunately, it's not easy to fix. The writing thread will be blocked on write, so there's no way to wake it up and tell the writing thread to quit. Python threads can't (safely) be killed from their parents. The thread *should* have dropped the GIL while blocked on the write, so we might barely get away with killing its PID. But I'd suggest adjusting the ReadableThread/WritableThread classes to use processes rather than threads. They already use pipes, so very little would need to change. There would be a bit of added overhead from the extra fork(), but surely that's worth getting rid of some of these GIL nightmares.

joelarmstrong:
This should also fix the problem for jobStores other than AWS, but I haven't tested it on anything else yet. I added a unit test to catch this error in all jobStores, so if the tests pass everything should be OK.

The basic idea is to always close the file on exit in the reader thread, which wakes the writer up and causes an exception in its thread. Thanks to @diekhans for the suggestion.

### DIFF
--- a/src/toil/jobStores/utils.py
+++ b/src/toil/jobStores/utils.py
@@ -119,13 +119,6 @@ class WritablePipe(object):
                 os.close(readable_fh)
 
 
-# FIXME: Unfortunately these two classes are almost an exact mirror image of each other.
-# Basically, read and write are swapped. The only asymmetry lies in how shutdown is handled. I
-# tried generalizing but the code becomes inscrutable. Until I (or someone else) has a better
-# idea how to solve this, I think its better to have code that is readable at the expense of
-# duplication.
-
-
 class ReadablePipe(object):
     """
     An object-oriented wrapper for os.pipe. Clients should subclass it, implement
@@ -201,10 +194,14 @@ class ReadablePipe(object):
         raise NotImplementedError()
 
     def _writer(self):
-        with os.fdopen(self.writable_fh, 'w') as writable:
-            # FIXME: another race here, causing a redundant attempt to close in the main thread
-            self.writable_fh = None  # signal to parent thread that we've taken over
-            self.writeTo(writable)
+        try:
+            with os.fdopen(self.writable_fh, 'w') as writable:
+                self.writeTo(writable)
+        except IOError as e:
+            # The file may have been closed by the reading thread,
+            # which is OK.
+            if e.strerror != 'Broken pipe':
+                raise
 
     def __init__(self):
         super(ReadablePipe, self).__init__()
@@ -220,17 +217,11 @@ class ReadablePipe(object):
         return self.readable
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        try:
-            if exc_type is None:
-                if self.thread is not None:
-                    # reraises any exception that was raised in the thread
-                    self.thread.join()
-        finally:
-            self.readable.close()
-            # The responsibility for closing the writable end is generally that of the writer
-            # thread. To cover the small window before the writer takes over we also close it here.
-            writable_fh = self.writable_fh
-            if writable_fh is not None:
-                # FIXME: This is still racy. The writer thread could close it now, and someone
-                # else may immediately open a new file, reusing the file handle.
-                os.close(writable_fh)
+        # Close the write end of the pipe. The writing thread may
+        # still be writing to the other end, but this will wake it up
+        # if that's the case.
+        self.readable.close()
+        if exc_type is None:
+            if self.thread is not None:
+                # reraises any exception that was raised in the thread
+                self.thread.join()
--- a/src/toil/jobStores/utils.py
+++ b/src/toil/jobStores/utils.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import errno
 from abc import ABCMeta
 from abc import abstractmethod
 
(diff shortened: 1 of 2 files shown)


joelarmstrong:
OK, turns out that closing the pipe doesn't work quite right in all situations--and there is no way to send a signal to the thread in a portable way. And switching to processes would require a lot of extra machinery in the jobStores, because they all rely on shared memory between the reading and writing side.

Instead I'm using sockets, which at least have a guarantee of interrupting writes on shutdown. This shouldn't be much slower than pipes (which isn't saying much) if UNIX sockets are available.

diekhans:
On some version of unix, pipes are implemented as unix domain
sockets, performance should not be a concern.

However, what when wrong with closing the read end of the pipe?
That should always result in the write failing; either through
a status or an EPIPE single if enabled.

The lack of the ability to kill a thread was always something
that seemed weird about POSIX threads.

joelarmstrong:
After reading your comment I realized I'd been closing the *write* end of the pipe all this time--no wonder it didn't work properly! (It did accidentally work sometimes, because the threads share fds.) I even had the pipe man page in front of me.

Anyway, it now works properly using pipes and not sockets :)

src/toil/jobStores/utils.py
```diff
@@ -220,17 +217,11 @@ def __enter__(self):
...
-        finally:
-            self.readable.close()
-            # The responsibility for closing the writable end is generally that of the writer
-            # thread. To cover the small window before the writer takes over we also close it here.
-            writable_fh = self.writable_fh
-            if writable_fh is not None:
-                # FIXME: This is still racy. The writer thread could close it now, and someone
-                # else may immediately open a new file, reusing the file handle.
-                os.close(writable_fh)
+        # Close the write end of the pipe. The writing thread may
```
ejacox: Do you mean "Close the read end of the pipe."?

joelarmstrong: Yes, oops! Thanks for catching that.

src/toil/jobStores/utils.py
```diff
@@ -220,17 +217,11 @@ def __enter__(self):
...
-            writable_fh = self.writable_fh
-            if writable_fh is not None:
-                # FIXME: This is still racy. The writer thread could close it now, and someone
-                # else may immediately open a new file, reusing the file handle.
-                os.close(writable_fh)
+        # Close the read end of the pipe. The writing thread may
+        # still be writing to the other end, but this will wake it up
+        # if that's the case.
+        self.readable.close()
+        if exc_type is None:
```
diekhans: It is unclear why the thread is not joined if there is an exception.  Shouldn't the thread always be joined if active?

src/toil/jobStores/utils.py
```diff
@@ -201,10 +194,14 @@ def writeTo(self, writable):
...
-            # FIXME: another race here, causing a redundant attempt to close in the main thread
-            self.writable_fh = None  # signal to parent thread that we've taken over
-            self.writeTo(writable)
+        try:
+            with os.fdopen(self.writable_fh, 'w') as writable:
+                self.writeTo(writable)
+        except IOError as e:
+            # The other side of the pipe may have been closed by the
+            # reading thread, which is OK.
+            if e.strerror != 'Broken pipe':
```
diekhans: this should check the errno, not the human readable string.

src/toil/jobStores/utils.py
joelarmstrong: This is from the original code--I didn't want to screw with the behavior too much. That said, I also thought this was a bit weird.

I can only guess, but I think the logic here is that joining the thread may raise a new exception, masking the one from the main thread. The other thread is, in theory, doomed to complete or die at this point, so we shouldn't end up leaking the thread.

How about this? We could just get rid of this if, and wrap the join in a try/except that reraises the child exception if exc_type is None and `pass`es otherwise. I think that would be cleaner, and it would guarantee that we don't leak threads.

src/toil/jobStores/utils.py
joelarmstrong: Agreed

diekhans:
That sounds like better approach, and easier to understand too.

joelarmstrong:
Looks like the connection to Quay was a little flaky, causing the integration tests to fail because of a timeout. (I wonder if Quay would be more reliable if we had a paid plan?) There isn't much we can do about it, because the timeout was within Docker itself.

Jenkins, test this please.

ejacox:
Jenkins test this please

joelarmstrong:
Thanks for retesting this Edwin. (For the record there was another failure of the integration tests, this time possibly due to a spot request being priced too low).

@diekhans, ready for re-review if you want to take a look. Thanks for the very helpful comments.

joelarmstrong:
Poking @diekhans -- can you re-review this? Should be much cleaner now.

ejacox:
Thank you @joelarmstrong

PR #1758, opened 2017-07-14
Add local testing targets (resolves #1741)

### LINKED ISSUE #1741: Add 'make tests' for running locally.
'make test' and 'make integration_test' are configured to run under Jenkins. 

New targets should be added to run tests locally. They should call pytest directly, skipping the docker target and calling run_tests.py.

Consider renaming the existing tests to 'jenkins_<test_name>'.

jessebrennan:
This PR adds two new targets to the makefile for testing locally:
- `test_offline` and
- `integration_test_local`.

The `test_offline` target gives a cleaner interface to run unit tests without running the docker/appliance tests. Previously, one had to set the environment variable `TOIL_DOCKER_REGISTRY` to the empty string and _then_ run `make test`. Now, one can simply run `make test_offline` and get the results that they expect. Help messages were changed to clarify the new process. Name taken from 

The `integration_test_local` target now runs in series and output is not captured. This can still use some improvement however because running all of the tests in series takes a looong time and produces a _lot_ of output.

Other changes:
- target that are _only_ designed to be run on Jenkins have updated names (previously this was `test_parallel` and `integration_test`) prepended with `jenkins_`. The old target names still exist temporarily, but will be phased out soon. Probably no one was using the to develop locally anyway.
- targets that meant to run locally: `test`, `test_offline`, and `integration_test_local` are now all run using pytest's `-s` option which allows the output to be printed to the stdout like one expects.

Things that still need to happen (though not necessarily right now):
- add to the documentation information on how to use these targets, etc.
- The "jenkins only" targets should depend upon the `check_running_on_jenkins` targets. This was held off because it's possible someone _is_ using these targets locally and this would be disruptive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,7 @@ integration_test_local: check_venv check_build_reqs sdist push_docker
 # FIXME when they are removed add check_running_on_jenkins to the jenkins targets
 test_parallel: jenkins_test_parallel
 
-test_integration: jenkins_test_integration
+integration_test: jenkins_test_integration
 
 # This target is designed only for use on Jenkins
 jenkins_test_parallel: check_venv check_build_reqs docker
@@ -288,7 +288,7 @@ check_cpickle:
 		check_cpickle \
 		develop clean_develop \
 		sdist clean_sdist \
-		test test_all test_appliance test_parallel integration_test \
+		test test_offline test_parallel integration_test \
 		jenkins_test_parallel jenkins_test_integration \
 		pypi clean_pypi \
 		docs clean_docs \
--- a/Makefile
+++ b/Makefile
@@ -47,9 +47,9 @@ particular test, e.g.
 The 'test_local' target is similar to 'test' but it skips the docker dependent tests and their
 setup.
 
-The 'integration_test_local' target runs toil's integration tests. This are more thorough but also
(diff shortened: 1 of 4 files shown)


jessebrennan:
@ejacox Sorry I had never pinged. This is ready for review.

ejacox:
Jenkins, test this please.

jessebrennan:
Jenkins test this please

jessebrennan:
I cannot explain why integration tests are not passing. The last substantive commit, 75fbb8a passed fine and the most recent commit, which was directly after, was only documentation.

I ran integrations test locally with complete success using the target that jenkins is configured to use: `integration_test`.

Have Jenkins configurations changed or something? @ejacox any ideas?

jessebrennan:
ah.. Possibly related to "slow registry operiations" currently reported for quay at [status.quay.io](http://status.quay.io) ?

jessebrennan:
@ejacox why did that fix things? No target that was called by Jenkins used that option as far as I knew. 

Also I thought that option would be useful for local testing.

ejacox:
-s disables capturing of test output, which is used to tell Jenkins if the tests passed I think

mr-c:
Can this get merged?

mr-c:
Woohoo, thanks @jessebrennan & @ejacox !

PR #1746, opened 2017-07-06
Fixes toil_sort_example.py to run with autoscaling

jessebrennan:
This PR revamps the [toil-sort-example.py](https://github.com/BD2KGenomics/toil/blob/master/docs/running/toil-sort-example.py) included in the docs so that it works with autoscaling. It now also uses the Toil context manager.

I tested this locally as well as on AWS (with autoscaling) with success. For insurance however, it would be nice if someone else could test it too. The procedure is:
1. Lauch cluster
  `toil launch-cluster MY-sort-test --keyPairName <your-key-pair-name-here> --nodeType t2.micro --zone us-west-2a`
2. copy the script to the leader
  `toil rsync-cluster MY-sort-test toil-sort-example.py :/home`
3. ssh to leader
  `toil ssh-cluster My-sort-test`
4. run the script
  `python /home/toil-sort-example.py aws:us-west-2:sort-test-jobstore --provisioner aws --nodeType=t2.small --batchSystem mesos --mesosMaster <LOCAL IP>` 

This should work as is, could maybe use further improvements such as more and (more in depth) comments.

I also encountered some minor documentation errors that are fixed in a separate commit. They're not really worth their own PR and are tangentially related to the main fix.

As a side note, I was stuck with a weird permissions bug while testing locally where the permissions on the source file where changed behind the scenes, presumably because of a hard-link optimization. I had an issue all ready to file, but this bug disappeared just as soon as it was isolated.

**EDIT:** added nodeType to run command (I had typed the commands from memory before).

PR #1760, opened 2017-07-17
Add option to link files when importing into fileJobStore (resolves #1755)

### LINKED ISSUE #1755: Weird permissions error when using importFile with file jobStore
I encountered this bug last week, wrote a ticket for it, was about to submit—and then it mysteriously vanished. Well now it's back.

The importFile can change the permissions of the local copy of the file when using the file jobStore This is my best diagnosis although I'm not certain.

When using the file JobStore, [importFile](https://github.com/BD2KGenomics/toil/blob/90bd990f5f0eb2ccc8c2018c244ff43ea2898135/src/toil/jobStores/fileJobStore.py#L165) uses a slight hack/optimization: if the file is being imported from the same disk, instead of copying it, a hard link is made. This shouldn't be a problem except that permissions are stored on the inode.

_THIS_ wouldn't be a problem if not for the fact that somewhere the permissions are changed to read only (I cannot find where in the code).

The code that is causing this error for me is my current version of the [toil-sort-example.py.txt](https://github.com/BD2KGenomics/toil/files/1140489/toil-sort-example.py.txt) (I had to add the .txt so that github would let me upload it) that I was revamping for #1722.

The problem arises from using exportFile because the exporting the file with the same name back to the source directory because we don't have permissions to overwrite it.

It would be awesome if someone else could validate this. by error trace looks like:
>Traceback (most recent call last):
  File "toil-sort-example.py", line 170, in <module>
    main()
  File "toil-sort-example.py", line 165, in main
    toil.exportFile(sorted_file_id, sort_file_url)
  File "/Users/jesse/Documents/genomics/toil/src/toil/common.py", line 900, in exportFile
    self._jobStore.exportFile(jobStoreFileID, dstUrl)
  File "/Users/jesse/Documents/genomics/toil/src/toil/jobStores/abstractJobStore.py", line 327, in exportFile
    self._exportFile(otherCls, jobStoreFileID, dstUrl)
  File "/Users/jesse/Documents/genomics/toil/src/toil/jobStores/fileJobStore.py", line 190, in _exportFile
    shutil.copyfile(self._getAbsPath(jobStoreFileID), self._extractPathFromUrl(url))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 13] Permission denied: '/Users/jesse/Documents/genomics/toil/docs/running/file_to_sort.txt'

jessebrennan:
Previously when using importFile function hard-links were made if possible in the fileJobStore. Caching code changes permissions of the imported file to read-only which can have unexpected results. 

This PR changes importFile to copy by default instead. It also add a command line option `--linkImports` which will cause linking if desired. The help message explains the permissions side affect.

Also I did a few minor style changes in nearby areas.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -164,15 +164,19 @@ class FileJobStore(AbstractJobStore):
     # Functions that deal with temporary files associated with jobs
     ##########################################
 
-    def _copyOrLink(self, url, path):
+    def _copyOrLink(self, srcURL, destPath):
         # linking is not done be default because of issue #1755
+        srcPath = self._extractPathFromUrl(srcURL)
         if self.linkImports:
             try:
-                os.link(os.path.realpath(self._extractPathFromUrl(url)), path)
+                os.link(os.path.realpath(srcPath), destPath)
             except OSError:
-                shutil.copyfile(self._extractPathFromUrl(url), path)
+                shutil.copyfile(srcPath, destPath)
+            else:
+                # make imported files read-only if they're linked for protection
+                os.chmod(destPath, 0o444)
         else:
-            shutil.copyfile(self._extractPathFromUrl(url), path)
+            shutil.copyfile(srcPath, destPath)
 
     def _importFile(self, otherCls, url, sharedFileName=None):
         if issubclass(otherCls, FileJobStore):
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -189,7 +189,6 @@ class FileJobStore(AbstractJobStore):
             else:
                 self._requireValidSharedFileName(sharedFileName)
                 path = self._getSharedFilePath(sharedFileName)
-                # linking is not done be default because of issue #1755
                 self._copyOrLink(url, path)
                 return None
         else:
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -60,6 +60,7 @@ class FileJobStore(AbstractJobStore):
         logger.debug("Path to job store directory is '%s'.", self.jobStoreDir)
         # Directory where temporary files go
         self.tempFilesDir = os.path.join(self.jobStoreDir, 'tmp')
+        self.linkImports = None
 
     def initialize(self, config):
         try:
@@ -70,6 +71,7 @@ class FileJobStore(AbstractJobStore):
             else:
                 raise
         os.mkdir(self.tempFilesDir)
+        self.linkImports = config.linkImports
         super(FileJobStore, self).initialize(config)
 
     def resume(self):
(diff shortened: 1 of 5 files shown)


jessebrennan:
@ejacox ready for review. @arkal Do you think you could take a look at this?

jessebrennan:
This should have a test. I will add in a test as I revamp the sort test (that is after all where I first discovered this bug)

src/toil/jobStores/fileJobStore.py
```diff
@@ -162,24 +164,29 @@ def jobs(self):
     # Functions that deal with temporary files associated with jobs
     ##########################################
 
+    def _copyOrLink(self, url, path):
+        # linking is not done be default because of issue #1755
+        if self.linkImports:
+            try:
+                os.link(os.path.realpath(self._extractPathFromUrl(url)), path)
```
arkal: I would explicitly make the file read-only at this point. This will solve 2 things
1) Prevent any unexpected modification of the file before it is pulled in by readGlobalFile
2) If the file(inode) cannot be made read-only It will crash the run much earlier than if we were to wait for it to be pulled down by readGlobalFile for the first time.

jessebrennan: @arkal actually the error only occured when trying to overwrite the original file with the new output by using exportFile. Changing the permissions here won't help with reason 2. (Reason 1 is enough for me though).

jessebrennan:
jenkins test this please

jessebrennan:
This was merged into #1746 which was serving as the test. Merging (plus some unrelated fixes) caused #1746 to pass tests, proving that this works. @ejacox it is ready for review and/or merge.

jessebrennan:
Or, since it was already merged into #1746 we could just let it get added when that branch is merged... I'm not sure what is the best git etiquette. My guess is merge separately.

ejacox:
I don't see the requested test. Otherwise, it look goods and we can merge it before the sort example.

jessebrennan:
I added a test to make sure linking works. Also the sort test ensures that permissions don't change out from under the user. @ejacox review/merge?

mr-c:
Since `cwltoil` uses `importFile` it'll be interesting to see how much of a speedup this provides. Might even make sense to make it default in that context..

jessebrennan:
@mr-c It was default before, but it causes some unexpected errors (see the issue thread, #1775) when running a very simple example.

mr-c:
@jessebrennan I'm not seeing a relation between this and #1775, did you mean a different issue?

mr-c:
Ah, you meant https://github.com/BD2KGenomics/toil/issues/1755 -- but that's about a pure Toil workflow. I haven't seen reports about this for CWL workflows..

jessebrennan:
Woops! sorry about that. yes #1775. Then perhaps CWL should change this option by default?

mr-c:
@jessebrennan Yep, I made an issue for that: https://github.com/BD2KGenomics/toil/issues/1786

PR #1762, opened 2017-07-18
automatically use the correct LSF units (resolves #1691)

### LINKED ISSUE #1691: LSF clusters have varying resource units -- toil needs to detect this and adapt accordingly
Use of the correct units is needed at 
https://github.com/BD2KGenomics/toil/blob/105aa24/src/toil/batchSystems/lsf.py#L38
> `mem = '' if mem is None else '-R "select[type==X86_64 && mem > ' + str(int(mem/ 1000000)) + '] rusage[mem=' + str(int(mem/ 1000000)) + ']" -M' + str(int(mem/ 1000000)) + '000'`

For example -- the trailing ` + '000'` needs to be left off when running on EBI's LSF clusters

mr-c:


mr-c:
@jessebrennan This is ready for review & merge

jessebrennan:
Let's ping @ejacox as well because he's the one with the power to merge. I'll gladly look it over though.

ejacox:
Merges are currently held up until release 3.9.0 is out. Hopefully today.

ejacox:
Looks good @mr-c. Thank you.

PR #1769, opened 2017-07-21
Sleep support for all state-querying functions in batch

brainstorm:
Context: Our cluster admins at NCI Australia were pointing out that there was too much polling in our PBSPro cluster when launching jobs (via qstat). Sub-second polling was overwhelming the queue system.

This patch should address/alleviate the issue by including more functions that were not participating in the collective siesta.

@ohofmann, @chapmanb


chapmanb:
Roman -- this is awesome, thank you. A couple of thoughts:

- Could the poll time be an option we can specify on the command line? I imagine different cluster admins will have different tolerance for polling and it allows users to balance responsiveness and polling.
- Could we apply this to other schedulers that Toil supports?

brainstorm:
Hi @ejacox! I've been trying to generalize this PR by using the `options.py` system on toil, but I wonder how to access those from the abstractEngine class? (the following line shouldn't work):

https://github.com/BD2KGenomics/toil/pull/1769/files#diff-795fe02487003fd6949460cf4d9ae4cbR381

ejacox:
Does that commit work for you? It seems that sleepSeconds should be an instance method in order to access self.config at runtime. Also, I moved the default value into the function body as default arguments are evaluated at compile time.

brainstorm:
@ejacox Thanks for the hint, those were my thoughts but `self` cannot be accessed in that context (instance method) either:

```
  File "/g/data/gx8/local/development/bcbio/anaconda/bin/cwltoil", line 11, in <module>
    load_entry_point('toil', 'console_scripts', 'cwltoil')()
  File "/home/005/rg3930/dev/toil/src/toil/cwl/cwltoil.py", line 764, in main
    outobj = toil.start(wf1)
  File "/home/005/rg3930/dev/toil/src/toil/common.py", line 718, in start
    return self._runMainLoop(rootJobGraph)
  File "/home/005/rg3930/dev/toil/src/toil/common.py", line 988, in _runMainLoop
    jobCache=self._jobCache).run()
  File "/home/005/rg3930/dev/toil/src/toil/leader.py", line 175, in run
    self.innerLoop()
  File "/home/005/rg3930/dev/toil/src/toil/leader.py", line 481, in innerLoop
    self.checkForDeadlocks()
  File "/home/005/rg3930/dev/toil/src/toil/leader.py", line 498, in checkForDeadlocks
    totalRunningJobs = len(self.batchSystem.getRunningBatchJobIDs())
  File "/home/005/rg3930/dev/toil/src/toil/batchSystems/abstractGridEngineBatchSystem.py", line 345, in getRunningBatchJobIDs
    return self.worker.getRunningJobIDs()
  File "/home/005/rg3930/dev/toil/src/toil/batchSystems/torque.py", line 68, in getRunningJobIDs
    self.boss.sleepSeconds()
  File "/home/005/rg3930/dev/toil/src/toil/batchSystems/abstractGridEngineBatchSystem.py", line 383, in sleepSeconds
    sleeptime = sleeptime or self.config.statePollingWait
NameError: global name 'self' is not defined
```

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -382,9 +377,14 @@ def getWaitDuration(self):
     def getRescueBatchJobFrequency(cls):
         return 30 * 60 # Half an hour
 
-    @classmethod
-    def sleepSeconds(cls):
-        return 1
+    def sleepSeconds(cls, sleeptime=None):
```
mr-c: Shouldn't this signature be `def sleepSeconds(self, sleeptime=None):`?
(`self`, not `cls`)

brainstorm: Ouch, doh, thanks @mr-c!

brainstorm:
Works on our cluster and docs seem fine too (cmdline `help` should suffice since not all flags are documented on readthedocs for brevity). 

Jenkins, test this please. ;P

setup.py
```diff
@@ -50,8 +50,8 @@ def runSetup():
                 'gcs_oauth2_boto_plugin==1.9',
                 botoRequirement],
             'cwl': [
-                'cwltool==1.0.20170413194156',
-                'schema-salad==2.5.20170328195758',
+                'cwltool>=1.0.20170811195303',
+                'schema-salad>=2.6.20170806163416',
```
mr-c: I bet the tests will succeed if you don't update these

brainstorm: Ugh, thanks @mr-c! ... I always wondered why Toil is hardpinning CWL so hard... since CWL 1.x, it doesn't break that often, does it? 

Every time I setup a new toil development environment I forget to tweak those back :/

brainstorm: I also wonder why, conversely, `'cwltest>=1.0.20170214185319'` is not pinned that strictly?

mr-c: We break things all the time in `cwltool`, so strict versioning is totally legitimate :-0

brainstorm:
@mr-c I'm not sure jenkins was failing due to the pinnings but seems to be AWS-related transient issue now? Seems that I've been asking for too many tests lately :-?:

```
<ErrorResponse xmlns="https://iam.amazonaws.com/doc/2010-05-08/">
  <Error>
    <Type>Sender</Type>
    <Code>Throttling</Code>
    <Message>Rate exceeded</Message>
  </Error>
  <RequestId>5438a413-8267-11e7-a16d-9f0feae8f51a</RequestId>
</ErrorResponse>
```

mr-c:
@brainstorm Every PR is getting those errors right now in `jenkins-it` :-/
The previous commit failed the CWL conformance tests, lets hope this one passes 'em :-)

brainstorm:
I can at least confirm a full clean run on our PBS cluster with `test_bcbio_cwl` with this PR in place :-S ... @ejacox Any indication on what's going on with Jenkins and AWS?

joelarmstrong:
Looks like the IAM rate limits on our account went *way* down--or else there's some stray process that is going crazy making IAM requests. I think someone with access is going to contact AWS support. The issue is tracked in #1837.

ejacox:
Don't worry about the jenkins-it tests for now. I didn't see why the other tests failed. I should have some time to take a look later.

ejacox:
Jenkins, test this please

ejacox:
Jenkins, test this please.

ejacox:
@brainstorm I changed all the defaults to 1 seconds and it worked. The tests can be adjusted if you want the default to be 10 seconds.

brainstorm:
Right on @ejacox, please merge as is, 1second is fine for most scenarios and does not change the current behaviour.

PR #1771, opened 2017-07-21
Fixed `job.addChild().encapsulate`

### LINKED ISSUE #1643: `job.addChild().encapsulate` doesn't work
A.addChild(B).encapsulate creates the relationship between jobs A and B before encapsulating B, thereby causing the encapsulated version of B to have 2 parents with unique roots (the original parent, and the synthetic root for the encapsulation subgraph) Adding a relationship between the synthetic encapsulating parent and the original parent fixes this (the if block in the snippet).
```
    def __init__(self, job):
        """
        :param toil.job.Job job: the job to encapsulate.
        """
        # Giving the root of the subgraph the same resources as the first job in the subgraph.
        Job.__init__(self, **job._requirements)
        if job._directPredecessors:
            for job_ in job._directPredecessors:
                job_.addChild(self)
        self.encapsulatedJob = job
        Job.addChild(self, job)
        # Use small resource requirements for dummy Job instance.
        self.encapsulatedFollowOn = Job(disk='1M', memory='32M', cores=0.1)
        Job.addFollowOn(self, self.encapsulatedFollowOn)
```

ejacox:
Resolves #1643. Calling `job.addChild().encapsulate` would create two parents. This PR uses the code suggested in #1643 and adds a test to confirm that only one parent exists.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -1483,6 +1483,9 @@ class EncapsulatedJob(Job):
         """
         # Giving the root of the subgraph the same resources as the first job in the subgraph.
         Job.__init__(self, **job._requirements)
+        if job._directPredecessors:
+            for job_ in job._directPredecessors:
+                job_.addChild(self)
         self.encapsulatedJob = job
         Job.addChild(self, job)
         # Use small resource requirements for dummy Job instance.
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -1473,6 +1473,12 @@ class EncapsulatedJob(Job):
         #  B will run after A and all its successors have completed, A and its subgraph of
         # successors in effect appear to be just one job.
 
+    If the job being encapsulated has predecessors (e.g. is not the root job), then the encapsulated
+    job will inherit these predecessors. If predecessors are added to the job being encapsulated
+    after the encapsulated job is created then the encapsulating job will NOT inherit these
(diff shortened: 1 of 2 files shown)


src/toil/job.py
```diff
@@ -1483,6 +1483,9 @@ def __init__(self, job):
         """
         # Giving the root of the subgraph the same resources as the first job in the subgraph.
         Job.__init__(self, **job._requirements)
+        if job._directPredecessors:
+            for job_ in job._directPredecessors:
+                job_.addChild(self)
         self.encapsulatedJob = job
```
benedictpaten: Excellent - this looks like the right thing to do. I would add a comment above #1486 to say:

Ensure that the encapsulated job has the same direct predecessors as the job being encapsulated.

benedictpaten: I also think we need a comment in the docstring to describe the behavior of encapsulated jobs with regards to predecessors, it should say:

"If the job being encapsulated has predecessors (e.g. is not the root job), then the encapsulated job will inherit these predecessors. If predecessors are added to the job being encapsulated after the encapsulated job is created then the encapsulating job will NOT inherit these predecessors automatically. Care should be exercised to ensure the encapsulated job has the proper set of predecessors."

ejacox:
Jenkins, test this please.

ejacox:
Jenkins, test this please.

ejacox:
Jenkins, test this please.

PR #1773, opened 2017-07-24
Allow monitoring toil clusters with a Grafana dashboard on the leader

adderan:
I'm open to suggestions on packaging the configuration files more cleanly. Dockerizing the services individually seems to be the best way to do this, but I could also deploy them to the leader through ssh, or through the startup script in awsUserData.


benedictpaten:
V. excited to see this. I don't see any documentation. Can you add @adderan @joelarmstrong ?

adderan:
This is ready for review, but it is rebased onto the autoscaling PR, so it will have to be merged after that one.

DailyDreaming:
Can't wait to start using this.  Thanks!

adderan:
This is ready, unless someone can think of a cleaner way to manage the containers. I think we were also waiting for people to test it. But it works for me.

jessebrennan:
Lon and and will test and report back.

ejacox:
@DailyDreaming and @jessebrennan, if this is ready, you need to approve it.

jessebrennan:
I could not get this running to a satisfying degree. I had to make some
modifications so that the docker images for grafana and prometheous would
be available from the remote machine (although from what I can tell this
would only be a issue for development / testing) .

Another thing was that it was very hard to get it to shutdown. ^C would
cause (some?) things to quit, but you'd be left with a blank terminal
session that could only be closed by closing the window.

Finally, although I could view a console at localhost:3000 I never actually
saw any data displayed there.

This could all just be user error, but I think the second issue I pointed
out should be addressed.

jessebrennan:
Was planning on running again to give some logs.

joelarmstrong:
I've gotten this to run without much of a hitch (it works brilliantly!!), but there are 2 problems I saw:

- number of queued/running jobs can go negative on failures. This is on me, I think, I have updated my dashboard config a few times but didn't send the incremental updates to Alden.
- in some situations, you can get a crash on restart (if an old mtail container isn't properly terminated, the new mtail container will fail to bind the port and die).

I have fixes in progress for both of these (easier to fix it than complain :)) but they're not ready yet.

adderan:
I think maybe I should have the containers running as init scripts so they can be restarted automatically if they fail. And then the logger would docker attach to the mtail container. Deploying them when the pipeline starts is kind of clunky, now that I think about it more.

adderan:
Okay, I think I've addressed the problems Jesse had. I've made it possible to configure the location of the containers with TOIL_DOCKER_REGISTRY so that it works the same way as the toil appliance, and the containers now terminate with only one Control-C. 

This should be ready to merge, but I think someone will need to set the toil-prometheus, toil-grafana, and toil-mtail repositories to public on Quay so that people can download the containers.

ejacox:
@adderan I made the repositories public. I tried this with the sort example locally. Everything seems to run fine, but I only get spinning little wheels in my dashboard. What I am doing wrong?

joelarmstrong:
@ejacox, @jessebrennan, when you're running locally and things are broken, are you running on Macs? As written this won't work on Macs, because `--net=host` is completely (and silently) broken on Macs. (To be fair, because Docker runs in a VM on Mac, it's not really possible to make --net=host work properly.)

We could probably rejigger things to use the default bridged mode (pointing to `docker.for.mac.localhost` rather than `localhost` on Macs). But that will break some things on Linux as well: on Linux the containers each have their own loopback interface, so if grafana tries to access prometheus at localhost:9090, that will point to the *grafana container* rather than the outside host. I think the way to address that is to point everything at the `docker0` gateway, but really this just adds a huge amount of complexity. Docker networking is hell.

ejacox:
Thanks @joelarmstrong. I suggest noting that the single machine mode doesn't work on a Mac and merging this once the updates are resolved. We can revisit the Mac issue later.

joelarmstrong:
Hope you don't mind: I pushed a bit of an update to the dashboard config (fixing some stacking issues in the queued/running jobs graph that were probably inherited from my crappy config) and fixed a typo which caused the "desired size" log update to get missed.

This all seems to be working pretty well for me.

ejacox:
Jenkins, test this please

ejacox:
Jenkins, test this please.

jessebrennan:
Travis failed due to a timeout. Since Jenkins passed I think this is safe to merge.

I haven't tested Alden's force quit fix, but it looked fine to me.

PR #1779, opened 2017-07-26
Fix docker tests failing on mac. resolves #1778

### LINKED ISSUE #1778: docker tests fail on Mac OS X
Example output:

```
WARNING:toil.leader:4/n/jobgAPa3z    docker: Error response from daemon: Mounts denied: 
WARNING:toil.leader:4/n/jobgAPa3z    The path /var/folders/0q/52rv8xtd7mjcgljsgmbkk5yr0000gq/T/toil-test-toil.test.lib.dockerTest.DockerTest-testNonCachingDockerClean-tempDir-86kF0J/data
WARNING:toil.leader:4/n/jobgAPa3z    is not shared from OS X and is not known to Docker.
WARNING:toil.leader:4/n/jobgAPa3z    You can configure shared paths from Docker -> Preferences... -> File Sharing.
WARNING:toil.leader:4/n/jobgAPa3z    See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
```

The challenge is that on macs `/var` is a symlink to `/private/var` and when you go through the GUI to add `/var` it will instead add `/private/var`, but `/private` is already on the list so it rejects the change.

ryanking:
Docker on mac requires whitelisting directories before mounting them into docker images. Temp files are created in `/var/folders/...`. `/var` is a symlink to `/var/private`. `/private` is in the whitelist, but since the canonical name is not being used docker doesn't recognize it.

The workaround of adding `/var/folders` fails because the GUI will only select the canonical name (`/private/var/folders`) and because `/private` is already on the list the change is rejected.

resolves #1778

### DIFF
--- a/src/toil/test/__init__.py
+++ b/src/toil/test/__init__.py
@@ -151,7 +151,7 @@ class ToilTest(unittest.TestCase):
         prefix = ['toil', 'test', strclass(cls)]
         prefix.extend(filter(None, names))
         prefix.append('')
-        temp_dir_path = tempfile.mkdtemp(dir=cls._tempBaseDir, prefix='-'.join(prefix))
+        temp_dir_path = os.path.realpath(tempfile.mkdtemp(dir=cls._tempBaseDir, prefix='-'.join(prefix)))
         cls._tempDirs.append(temp_dir_path)
         return temp_dir_path
 


ejacox:
Jenkins, test this please.

ejacox:
Transient failure?

ejacox:
Jenkins, test this please.

PR #1794, opened 2017-07-31
futurize stage 1 resolves #1784

### LINKED ISSUE #1784: Futurize stage 1 [python 2/3]
Context: https://github.com/BD2KGenomics/toil/issues/1780

The [futurize tool](http://python-future.org/automatic_conversion.html) can automatically convert code to a form that is compatible with both python 2 and 3. Their stage 1 conversions are intended to be totally safe, so we'll do those first.

ryanking:
In the plan to get to python2/3 compat, this is the first step.

I ran `futurize -1 -w src` and fixed a few things up.

The only thing that might be controversial is that I deleted some doctests in `src/toil/test/__init__.py`. Python's api changed here in a problematic way and I don't fully understand what the code is supposed to do. So I suggest removing and if there's a compelling reason to add it back.

### DIFF
--- a/src/toil/test/__init__.py
+++ b/src/toil/test/__init__.py
@@ -537,7 +537,7 @@ def timeLimit(seconds):
 # FIXME: move to bd2k-python-lib
 
 
-def make_tests(generalMethod, targetClass=None, **kwargs):
+def make_tests(generalMethod, targetClass, **kwargs):
     """
     This method dynamically generates test methods using the generalMethod as a template. Each
     generated function is the result of a unique combination of parameters applied to the
@@ -569,7 +569,7 @@ def make_tests(generalMethod, targetClass=None, **kwargs):
     >>> class Bar(Foo):
     ...     pass
 
-    >>> make_tests(Foo.has, targetClass=Bar, num={'one':1, 'two':2}, letter={'a':'a', 'b':'b'})
+    >>> make_tests(Foo.has, Bar, num={'one':1, 'two':2}, letter={'a':'a', 'b':'b'})
 
     >>> b = Bar()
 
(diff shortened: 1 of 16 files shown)


src/toil/test/__init__.py
```diff
@@ -585,27 +586,7 @@ def make_tests(generalMethod, targetClass=None, **kwargs):
     >>> hasattr(f, 'test_has__num_one__letter_a')  # should be false because Foo has no test methods
     False
 
-    >>> make_tests(Foo.has, num={'one':1, 'two':2}, letter={'a':'a', 'b':'b'})
```
ryanking: Tests deleted bc this functionality is not used anywhere else and I can't figure out (easily) how to make it work in python 3.

ejacox:
Jenkins, add to whitelist

src/toil/test/__init__.py
ejacox: Isn't it used to generate jobstore tests? What part is failing?

ejacox:
Jenkins, test this please.

src/toil/test/__init__.py
ryanking: This is the best illustration of the problem I've found: https://stackoverflow.com/questions/3589311/get-defining-class-of-unbound-method-object-in-python-3

Basically, python 3 gets rid of the concept of "unbound method" and instead just has "functions".

The `make_tests` method, when `targetClass` is None relies on this behavior. The only place that is used is in these tests, so it seems safe to delete? Perhaps we should make `targetClass` non-optional to make the change more clear.

ejacox: I agree. Making targetClass non-optional and removing those tests seems safe. Will this PR be ready to merge after that?

ryanking:
Yeah I should be ready after that, AFAIK.

ryanking:
Any insight into the failure? I am trying to reproduce locally, but no failures yet.

ejacox:
Jenkins, test this please.

ejacox:
Thank you for this PR @ryanking. It is very much appreciated.

PR #1798, opened 2017-08-03
CWL: avoid importing existing files into fileStore

chapmanb:
Check for presence of an existing file before deciding on upload into
the fileStore. This prevents double uploading an identical file shared
across multiple sample, which would trigger a later export issue when
we try to write two identical files with different fileStoreIDs to a
work directory. This fixes a bug introduced in the symlink changes
(a1d574ade864342f82045cd53217b09a3eccc219).

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -162,8 +162,9 @@ def getFile(fileStore, dir, fileTuple, index=None, export=False, primary=None, r
         srcPath = fileStore.readGlobalFile(fileStoreID)
         if srcPath != dstPath:
             if os.path.exists(dstPath):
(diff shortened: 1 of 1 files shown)


ejacox:
Checks passed. The only failure was a transient failure from an AWS test.

brainstorm:
@ejacox I guess we could just retrigger the build then?

Jenkins please test this.

ejacox:
Jenkins, test this please.

ejacox:
Is this ready? If the only failure is the aws utility test, I can merge it anyways.

chapmanb:
Edwin -- thanks so much. It's ready to go and would be great to merge. Once it's in I'll update the bioconda recipes as well. Much appreciated.

Issue #1768, opened 2017-07-21
Singularity support


ps-account:
Hi, a forked project already seems to include Singularity support, is it possible to include this in the main toil project? Many HPC providers only allow Singularity containers http://singularity.lbl.gov/

Example of Singularity support in toil_vg:  https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/singularity.py

Info on how to use Singularity via Toil in the manual would also be helpful.

Gr. Pim

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-178)
┆Issue Number: TOIL-178

pimpim:
The implementation could be similar to: https://github.com/nextflow-io/nextflow/issues/356

jessebrennan:
I don't know anything about singularity personally. @glennhickey Is their any reason we cannot just merge your singularity changes into Toil proper?

glennhickey:
Not really, it's completely independent of anything so it won't hurt.
Still, we don't have any unit tests (would need singularity installed on
Jenkins), and it's much slower when running Docker images than calling
docker directly.

glennhickey:
Sure, I'll make a PR.  Just to be clear though, we've only been using
Singularity to run command-line programs from Toil scripts from
single-machine batch runs, and not to deploy Toil itself.

(Speaking of singularities, has Jenkins finally become sentient?)

On Mon, Jul 31, 2017 at 3:18 PM, *Jenkins* <notifications@github.com>
*wrote*:

pimpim:
yes it's safer to use than Docker as it limits itself to userspace while still configurable to access accellerators, Infiniband, etc., so it's rapidly becoming the container-system of choice in big (and small) HPC centers. I'm not an expert on the inner workings but it seems similar enough in use to Docker that adding support for Singularity next to Docker should be feasible without heavy rewriting.

pimpim:
There is one additional remarkt wrt the implementation in toil_vg, though. If I understand correctly, Toil_vg is slower in accessing containers because it converts Docker images to Singularity images. It is however possible to directly access a Singularity hub, I think it is implemented like that in nextflow-io

pimpim:
I think the request as formulated to nextflow seems therefore a good way to implement singularity support in a workflow system:

[...]

PR #1805, opened 2017-08-04
Add (incomplete) singularity interface

glennhickey:
I said back in #1768 that I'd PR [toil-vg](https://github.com/vgteam/toil-vg)'s singularity module into Toil, so here it is.  It seems to run okay for our purposes of calling command line tools in containers, except that the cache used to run Docker images on the fly gets easily overwhelmed.  This is worked around by converting the images a priori (we could probably automate this if necessary). If Singularity ever made the cache size tunable, that'd probably work too. 

There's still a lot of work to make this a drop-in replacement for docker.py:
* There's no interface to inspect or stop running containers. 
* The test won't run on Jenkins because singularity's not on the jenkins slave image. (even then, I'm not sure since I can't get any Toil tests to run locally from the command line any more) 
* I wouldn't be surprised if there are nesting issues that would arise when running these within the docker appliance (or vice versa)

I think would also require some thought for an abstract container interface.  In toil-vg we wrap everything (https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_common.py#L87-L244) so it's easy to flip between docker/singularity/no-container.  I think something similar is needed in Toil to make this useful.  On a related note, singularity.py shares about 99% of its code with docker.py, so there may be room to refactor that a bit better too.

### DIFF
--- a/src/toil/lib/singularity.py
+++ b/src/toil/lib/singularity.py
@@ -1,0 +1,128 @@
+"""
+    Module for running Singularity (and Docker) images with Singularity.   
+    Derived from https://github.com/BD2KGenomics/toil/blob/master/src/toil/lib/docker.py
+    with which nearly all code is shared, save the command lines used to run the containers.
+
+    Docker images can be run by prefacing the input tool with docker:// In this case, Singularity
+    will download, convert, and cache the image on the fly.  This cache can be set with 
+    SINGULARITY_CACHEDIR, and defaults to the user's home directory.  This cache can be a major
+    bottleneck when repeatedly more different images than it can hold (not very many).  So for this
+    type of usage pattern (concurrent or short consecutive calls to different images), it is 
+    best to run Singularity images natively.
+    
+"""
+import base64
+import logging
+import subprocess
+import pipes
+import os
+from bd2k.util.exceptions import require
+
+_logger = logging.getLogger(__name__)
+
+
+def singularityCall(job,
+               tool,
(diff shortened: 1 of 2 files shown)


pimpim:
As far as abstract container interface is concerned, maybe at one time it could be Open-Container compatible. ( https://blog.docker.com/2017/07/demystifying-open-container-initiative-oci-specifications/ ) - though this is not of very high functional impact to e.g. Singularity ( https://twitter.com/gmkurtzer/status/887705065573687296 )

jessebrennan:
> In toil-vg we wrap everything (https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_common.py#L87-L244) so it's easy to flip between docker/singularity/no-container. I think something similar is needed in Toil to make this useful.

@glennhickey, is there a reason not to just add this class to Toil as well in this PR? I suppose it will need documentation, etc. however.

glennhickey:
No real reason not to add it that I can think of.  In addition to better
documentation, would just need a small look-over to make sure the interface
in general enough.  I'll take a look.

DailyDreaming:
Closing this due to:
- `1.` Toil is likely moving to `github.com/DataBiosphere` tomorrow and we're attempting to clean up PRs.
- `2.` This has sat here for nearly a year.
- `3.` This would need quite a bit more work and testing.
- '4.' `https://github.com/leukgen/toil_container`

PR #1808, opened 2017-08-05
Fix docker interface

paboldin:
#. don't add `--rm` is defer is not `RM`
#. always generate a correct name for docker container.

### DIFF
--- a/src/toil/lib/docker.py
+++ b/src/toil/lib/docker.py
@@ -16,6 +16,7 @@ import base64
 import logging
 import os
 import pipes
+import re
 import uuid
 
 from bd2k.util.exceptions import require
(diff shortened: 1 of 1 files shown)


mr-c:
Jenkins, okay to test

ejacox:
Jenkins, test this please.

paboldin:
I can't access test details. May you guys please share them with me?

mr-c:
@paboldin `src.toil.test.provisioners.aws.awsProvisionerTest.AWSStaticAutoscaleTest.testAutoScale` timed out (probably not your fault?)

ejacox:
Jenkins, add to whitelist.

ejacox:
Jenkins, test this please.

ejacox:
@DailyDreaming, should this be part of #1905?

DailyDreaming:
I recommend using the new apiDockerCall() in docker.py.  We can't change the functionality of rm vs RM.  I don't like the convention, nor the use of the previous defer variable side-by-side with the defer(function), but it is necessary to the previous expected functionality of subprocessDockerCall().  From dockerTest.py:

This is the truth table for the different combinations at the end of the test. R = Running, X = Does not exist, E = Exists but not running:
`#              None     FORGO     STOP    RM
#    rm        X         R         X      X
# detached     R         R         E      X
#  Neither     R         R         E      X`

For the naming convention, is there a particular reason that you wish to change it?

paboldin:
1. (Regarding the naming) I want to be able to define `__str__` of the Job so Toil correctly displays job's arguments.
2. Looks like the patch changing `defer` behaviour is not needed then.

Issue #1814, opened 2017-08-10
i2.8xlarge nodes don't work with aws provisioner


glennhickey:
`  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/aws/awsProvisioner.py", line 702, in _getBlockDeviceMapping
    ephemeral_name='ephemeral{}'.format(disk - 1))  # ephemeral counts start at 0
                                                                                IndexError: list index out of range`

As pointed out by @joelarmstrong, the problem lies in this list being too small https://github.com/BD2KGenomics/toil/blob/5c5d086bbf2b9d092400edd150e53a16d3345b21/src/toil/provisioners/aws/awsProvisioner.py#L706

PR #1815, opened 2017-08-10
Extend bdtKeys to support more ephemeral drives

glennhickey:
This is a fix for #1814, where nodes with > 4 ephemeral drives, such as i2.8xlarge, are not supported.  

There is probably a more elegant/future-proof fix to be made here, but this at least works for the node type in question for the time being.

### DIFF
--- a/src/toil/provisioners/aws/__init__.py
+++ b/src/toil/provisioners/aws/__init__.py
@@ -212,10 +212,9 @@ write_files:
         #!/bin/bash
         set -x
         ephemeral_count=0
-        possible_drives="/dev/xvdb /dev/xvdc /dev/xvdd /dev/xvde"
         drives=""
         directories="toil mesos docker"
-        for drive in $possible_drives; do
+        for drive in /dev/xvd{{b..z}}; do
             echo checking for $drive
             if [ -b $drive ]; then
                 echo found it
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -19,6 +19,7 @@ import logging
 import time
 
 import sys
+import string
 
 # Python 3 compatibility imports
 from _ssl import SSLError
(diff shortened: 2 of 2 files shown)


glennhickey:
Actually, I've just noticed that in my test `toil launch-cluster test1 --nodeType i2.8xlarge --keyPairName glenn.hickey@gmail.com`, while the node's started up, `launch-cluster` is still hanging with repeated messages like this (for about 10 min -- nothing much running on the node as far as top is concerned)

`Connection to ec2-52-37-224-108.us-west-2.compute.amazonaws.com closed.
INFO:toil.provisioners.aws.awsProvisioner:...Still waiting, trying again in 10sec...
Warning: Permanently added 'ec2-52-37-224-108.us-west-2.compute.amazonaws.com,52.37.224.108' (ECDSA) to the list of known hosts.
`

I think someone who knows what they're doing will have to look at this and test properly.

jessebrennan:
I could look at this later and test, but perhaps just following the instructions [here](http://toil.readthedocs.io/en/3.9.1/contributing/contributing.html#developing-with-the-toil-appliance) will be enough to make things work

joelarmstrong:
I think this is the good ol' toil appliance tag problem that you get whenever you make a new commit/have a dirty working tree. I get
```
Unable to find image 'quay.io/ucsc_cgl/toil:3.10.0a1-5c5d086bbf2b9d092400edd150e53a16d3345b21-dirty' locally
```

when I test this, for example. Setting TOIL_APPLIANCE_SELF should help. (maybe we should just get rid of this tagging thing entirely, except for releases?)

I found an issue where the instance won't actually *use* more than 4 disks, though. I'll just commit the fix for that to this branch, if that's OK.

ejacox:
Jenkins, test this please.

ejacox:
Jenkins, test this please

ejacox:
Jenkins, test this please

ejacox:
Jenkins test this please

benedictpaten:
Looks like this failed because of the flaky service test.. suggest @joelarmstrong  reviews?

joelarmstrong:
BTW, the only failing tests are the flaky ones, and the affected tests all pass on this run. On previous runs we ran into the jenkins-it push failure tracked in #1821, but this shows the jenkins-it tests running fine. So this is totally OK to force merge without re-testing, if that's possible.

jessebrennan:
> So this is totally OK to force merge without re-testing, if that's possible.

@ejacox

PR #1817, opened 2017-08-11
use latest pynacl

mr-c:
They ship a manylinux1 wheel which makes install/testing faster


mr-c:
Only test failure was `src/toil/test/src/jobServiceTest.py::JobServiceTest::testServiceWithCheckpoints FAILED`

ejacox:
The test runs fine for me.

mr-c:
Thanks!

PR #1820, opened 2017-08-13
Fix flaky JobService tests (resolves #1819)

### LINKED ISSUE #1819: JobServiceTest tests are very flaky
This is really, really annoying, because it causes the Jenkins tests to fail about 30% of the time. I feel like this may be related to #1598. The old, broken checkpoint behavior may have been papering over another bug, or this could be a new bug introduced by my changes. Either way, I think I'm best equipped to tackle this, so I'm assigning to myself.

Examples:

http://jenkins.cgcloud.info/job/toil-pull-requests/3088/testReport/junit/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceWithCheckpoints/

http://jenkins.cgcloud.info/job/toil-pull-requests/3072/testReport/junit/src.toil.test.src.jobServiceTest/JobServiceTest/testServiceParallelRecursive/

joelarmstrong:
Hopefully this gives us some relief from the flaky tests.

This arose because of the recent checkpoint fixes. Unfortunately updating the state of the jobGraphs from the leader caused issues in some cases: for a job that launches services, the leader sets `jobGraph.services` to an empty list as a leader-specific signal that its services have been launched. This is out-of-sync with how it should appear in the jobStore. Crucially, this means that the job state can't ever be updated from the leader. If the jobGraph's services list gets cleared in the jobStore while the services are running, then those services will be orphaned and can't be cleared on checkpoint restart. If the orphaned services then become totally failed, and then the workflow ultimately completes (because a checkpointed subtree restarted successfully), the workflow will be marked as failed despite having completed (and despite the root job having been deleted).

### DIFF
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -495,8 +495,26 @@ class AbstractJobStore(object):
             # Delete the job
             self.delete(jobGraph.jobStoreID)
 
+        jobGraphsReachableFromRoot = {id: getJob(id) for id in reachableFromRoot}
+
+        # Clean up any checkpoint jobs -- delete any successors it
+        # may have launched, and restore the job to a pristine
+        # state
+        jobsDeletedByCheckpoints = set()
+        for jobGraph in [jG for jG in jobGraphsReachableFromRoot.values() if jG.checkpoint is not None]:
+            if jobGraph.jobStoreID in jobsDeletedByCheckpoints:
+                # This is a checkpoint that was nested within an
+                # earlier checkpoint, so it and all its successors are
+                # already gone.
+                continue
+            logger.info("Restarting checkpointed job %s" % jobGraph)
+            deletedThisRound = jobGraph.restartCheckpoint(self)
+            jobsDeletedByCheckpoints |= set(deletedThisRound)
+        for jobID in jobsDeletedByCheckpoints:
+            del jobGraphsReachableFromRoot[jobID]
+
         # Clean up jobs that are in reachable from the root
-        for jobGraph in (getJob(x) for x in reachableFromRoot):
+        for jobGraph in jobGraphsReachableFromRoot.values():
             # jobGraphs here are necessarily in reachable from root.
 
             changed = [False]  # This is a flag to indicate the jobGraph state has
--- a/src/toil/worker.py
+++ b/src/toil/worker.py
@@ -252,46 +252,14 @@ def main():
         # The job is a checkpoint, and is being restarted after previously completing
         if jobGraph.checkpoint != None:
             logger.debug("Job is a checkpoint")
-            if len(jobGraph.stack) > 0 or len(jobGraph.services) > 0 or jobGraph.command != None:
-                if jobGraph.command != None:
-                    assert jobGraph.command == jobGraph.checkpoint
-                    logger.debug("Checkpoint job already has command set to run")
-                else:
-                    jobGraph.command = jobGraph.checkpoint
-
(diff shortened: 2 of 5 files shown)


joelarmstrong:
The test failure is caused by the issue tracked in #1821. It should be safe to merge. I suggest @benedictpaten take a quick look at this, and hopefully we can merge ASAP. The flaky tests are a huge barrier to collaboration, and this JobServiceTest issue changed our failure rate from (guessing) ~20% to ~60%.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -495,8 +495,26 @@ def getConnectedJobs(jobGraph):
             # Delete the job
             self.delete(jobGraph.jobStoreID)
 
+        jobGraphsReachableFromRoot = {id: getJob(id) for id in reachableFromRoot}
+
```
benedictpaten: Just a note, not to be solved now - but we should clean up this whole function so that it does not use getJobs() to load all jobs into memory at once - it is unnecessary - unless caching is enabled.

ejacox:
@benedictpaten is this ready to merge?

src/toil/worker.py
```diff
@@ -252,46 +252,14 @@ def badWorker():
...
-                    # that has been previously serialised but which as yet has no successors
-
-                    jobGraph.services = [] # Empty the services
-
-                    # Update the jobStore to avoid doing this twice on failure and make this clean.
-                    jobStore.update(jobGraph)
-
-            # Otherwise, the job and successors are done, and we can cleanup stuff we couldn't clean
-            # because of the job being a checkpoint
+            if len([i for l in jobGraph.stack for i in l]) > 0 or len(jobGraph.services) > 0:
```
benedictpaten: Suggest adding a comment above:

"If the checkpoint job has outstanding successors or services then its subtree is not complete and the checkpoint job must be restarted"

Or words to that effect

src/toil/jobStores/abstractJobStore.py
joelarmstrong: In practice the jobGraphs are already read into memory in the jobCache. So this does create n keys, but the values aren't copied, just referenced. So it's not all *that* bad, assuming the ids are small compared to the jobGraphs.

That said, I totally agree. The memory usage for the leader balloons if you have more than a few hundred thousand jobs to resume. We may have to make the jobCache an actual cache, and rewrite this bit so it doesn't store all jobs at once.

benedictpaten:
BTW, why do we see the jenkins-it build bit fail?

src/toil/jobStores/abstractJobStore.py
benedictpaten: I know, but the logic for the function should be that you can turn off the cache and have the jobs never fully resident in memory. Currently not using the cache is totally stupid.

ejacox:
Push docker failure. It could be related to the fix in #1822, which is failing because of the flaky tests problem.

ejacox:
Jenkins, test this please.

joelarmstrong:
Yes, exactly. The jenkins-it tests are flaky on their own, mostly due to issues out of our control, which #1822 is meant to work around.

Sorry again about all the annoying test failures! I got (un)lucky and never had this failure trigger on local or Jenkins tests before the merge.

ejacox:
The one failed test worked fine when I ran it separately. Merge this PR?

joelarmstrong:
Yeah, looks like the failure is an issue with the "temporary credential race" code. This PR doesn't affect that.

PR #1824, opened 2017-08-14
Issues/1459 1656 cwl directories

stefanoberri:
Two unit tests check reading from directories and writing to directories


tetron:
Thanks @stefanoberri !

PR #1827, opened 2017-08-15
Add missing quote to bsub invocation

mr-c:


mr-c:
Just the sketchy AWS tests are failing, otherwise all is good.

diekhans:
This bug would not have happened if subprocess shell=True were
not used.  It's also more robust and safer, the Python manual
explicitly warns against using it and there is rarely a reason
to need it.

PR #1828, opened 2017-08-15
Stop log level option defaults from clobbering each other (resolves #1316)

### LINKED ISSUE #1316: Log level must be explicitly set for RealtimeLogger to work
I have logging statements in my user code that log through the Toil RealtimeLogger. However, the RealtimeLogger doesn't get initialized unless I specify one of the Toil log level options (--logInfo, --logWarning, etc.). Without those I just get a message that it is disabled, even when I pass --realTimeLogging.

Instead, the real time logger should be enabled with the default log level if --realTimeLogging alone is passed.

adamnovak:
The default default of None for some of the options with a dest of
"logLevel" was overriding the explicitly set default. Adding
an explicit default to all of them solves the problem.

Also restores functionality of the --logOff option which didn't
work.

Fixes #1316 by having a non-None logLevel for --realTimeLogging to
use.


benedictpaten:
jenkins, test this please.

PR #1838, opened 2017-08-16
fix in NonCachingFileStore:open that may delete the parent directory of localTempDir (resolves #1825)

### LINKED ISSUE #1825: NonCachingFileStore open() may delete the parent directory of localTempDir in LSF job
I have a workflow that is running using `--disableCaching --batchSystem=lsf `. In the workflow, it will use scatter method, that means a set of jobs will be executed under the same parent directory. The symptom is that some job will fail to open `.jobState` file because its directory is removed by another job. 

In `NonCachingFileStore open()`, it will call `findAndHandleDeadJobs(workFlowDir)`. Howerver `workFlowDir` is assigned to be the parent directory of `localTempDir`. I guess there is some confusion about `localTempDir`: `localTempDir` is assigned to a sub directory of the input `localTempDir`, and the input `localTempDir` should become the `workFLowDir` of the current job.

Since `workFlowDir` is used in `NonCachingFileStore`, it is better to move `workFlowDir` from the base class `FileStore` to as a data member of `NonCachingFileStore`, and assign 'localTempDir` to `workFlowDir` solved my problem.

liuxf09:
The fix remove workFlowDir from the base class FileStore, and call findAndHandleDeadJobs(localTempDir) in NonCachingFileStore:open, before localTempDir is changed to sub directory.

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -106,7 +106,6 @@ class FileStore(object):
         self.jobStore = jobStore
         self.jobGraph = jobGraph
         self.localTempDir = os.path.abspath(localTempDir)
-        self.workFlowDir = os.path.dirname(self.localTempDir)
         self.jobName = self.jobGraph.command.split()[1]
         self.inputBlockFn = inputBlockFn
         self.loggingMessages = []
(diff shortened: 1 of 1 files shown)


liuxf09:
Sorry I am not familiar with this coil auto test framework. For case like `src.toil.test.src.fileStoreTest.NonCachingFileStoreTestWithFileJobStore.testDeferredFunctionRunsWithFailures`, is it supposed to fail ?

ejacox:
The jenkins-it failures were due to an AWS problem. They should work now. Maybe the other failures are transient, too.
Jenkins, test this please.

ejacox:
This might work for LSF, but it is failing with other batch systems. Closing.

PR #1843, opened 2017-08-17
Removed split

nazamy:
Split was causing white character to return such that casting to int was returning an error.


ejacox:
Jenkins, test this please.

ejacox:
Thank you @nazamy

PR #1847, opened 2017-08-27
Apply scale input parameter to memory

chapmanb:
Currently `--scale` only applies to cores, but memory also needs
to change along with cores, assuming the memory/core will be a
more stable scaling point than fixed machine memory.

This is a usage issue I ran into while running some real CWL workflows. They were initially designed for an 8 core system but I wanted to do some debugging/test re-runs of part of the pipeline on single core machines. You could adjust down scale which handled cores but did not fix memory. This moves them in parallel, making it easier to scale up and down.


DailyDreaming:
Closing this as it's out of date.

PR #1863, opened 2017-09-08
CWL: avoid early removal of the temporary outdir

chapmanb:
A recent change in 3.10.1 (https://github.com/BD2KGenomics/toil/commit/a909bada25940803f3dc9351fb8b6cd72e3769f0)
caused temporary output directories to get removed prior to finalizing a
Toil step, resulting in missing file errors when the outputs reference
files in the temporary directory. This fixes the issue by putting
them in the defined tempdir. cc @tetron for review.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -418,20 +418,21 @@ class CWLJob(Job):
         cwljob = resolve_indirect(self.cwljob)
         fillInDefaults(self.step_inputs, cwljob)
 
+        opts = copy.deepcopy(self.executor_options)
+        # Exports temporary directory for batch systems that reset TMPDIR
+        os.environ["TMPDIR"] = os.path.realpath(opts.pop("tmpdir", None) or fileStore.getLocalTempDir())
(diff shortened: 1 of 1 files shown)


mr-c:
Can we get a new test case?

tetron:
This might explain why bcbio and biowardrobe are broken @DailyDreaming @ejacox

chapmanb:
Peter -- that's exactly right. The errors you get out are a bit confusing since the job completes but then during Toil processing it complains about missing files that were in this directory. Thanks much for looking at this.

Michael -- I don't have a minimal test case but it causes this bcbio test case to fail: https://github.com/bcbio/test_bcbio_cwl/tree/master/somatic

Irrationone:
`temp_output_dir`'s (`out_tmpdir*`) seem to be sticking around after CWL step completion (with toil-cwl-runner), which is causing worker node disks to fill when running on Azure in the cloud (eventually resulting in out of space errors, since each step creates a ~500MB output file). I presume it has something to do with this pull request -- could they be deleted earlier? Thanks!

mr-c:
@Irrationone Can you make a separate issue for that? A comment on a closed PR is likely to get lost 🙂

PR #1864, opened 2017-09-08
CWL: run internal jobs locally, not on cluster

### LINKED ISSUE #1783: run CWL housekeeping jobs on the executor, don't send them to the batch scheduler
Run the housekeeping jobs on the launcher node and not via cluster submission:
 (`CWLWorkflow`,`ResolveIndirect`, `CWLGather`,`CWLScatter`, etc)

chapmanb:
The cwltoil integration includes non-processing based internal
jobs that handle dispatching and managing workflows. When running on a
cluster these jobs were being submitted to the scheduler. The larger
number of short submissions make busy schedulers unhappy and slow down
the runs.

This introduces a local batch queue along with the cluster submission
queue. Internal CWL jobs go to this, getting run locally instead of on
the cluster. It's a general fix that should apply across all of the
supported batch systems (SGE, Slurm, Torque, PBSPro).

Closes #1783

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -59,6 +59,10 @@ import six.moves.urllib.parse as urlparse
 
 cwllogger = logging.getLogger("cwltool")
 
+# Define internal jobs we should avoid submitting to batch systems and logging
+CWL_INTERNAL_JOBS = ("CWLJob", "CWLJobWrapper", "CWLWorkflow", "CWLScatter", "CWLGather",
+                     "ResolveIndirect")
+
 # The job object passed into CWLJob and CWLWorkflow
 # is a dict mapping to tuple of (key, dict)
 # the final dict is derived by evaluating each
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -24,11 +24,13 @@ from abc import ABCMeta, abstractmethod
 
 # Python 3 compatibility imports
 from six.moves.queue import Empty, Queue
+from future.utils import with_metaclass
 
 from bd2k.util.objects import abstractclassmethod
 
 from toil.batchSystems.abstractBatchSystem import BatchSystemSupport
-from future.utils import with_metaclass
+from toil.batchSystems import registry
+from toil.cwl import cwltoil
 
 logger = logging.getLogger(__name__)
 
@@ -284,7 +286,6 @@ class AbstractGridEngineBatchSystem(BatchSystemSupport):
         # much smaller
         self.maxCPU, self.maxMEM = self.obtainSystemConstants()
 
-        self.nextJobID = 0
         self.newJobsQueue = Queue()
         self.updatedJobsQueue = Queue()
         self.killQueue = Queue()
@@ -293,6 +294,8 @@ class AbstractGridEngineBatchSystem(BatchSystemSupport):
         self.worker = self.Worker(self.newJobsQueue, self.updatedJobsQueue, self.killQueue,
                               self.killedJobsQueue, self)
         self.worker.start()
+        self.localBatch = registry.batchSystemFactoryFor(registry.defaultBatchSystem())()(config, maxCores,
+                                                                                          maxMemory, maxDisk)
 
     def __des__(self):
         # Closes the file handle associated with the results file.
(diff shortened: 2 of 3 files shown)


brainstorm:
The ssh/aws errors seem to be transient? @ejacox, tempted to merge this right away, LGTM, may I?

ejacox:
One more try at the tests @brainstorm

ejacox:
Jenkins, test this please.

chapmanb:
Edwin -- I was treating this failure as unrelated but let me know if there is anything I can do to help avoid it. Thanks much to you and Roman for looking at this.

ejacox:
@chapmanb - We have been discussing this PR. We are concerned that running jobs locally will impact the leader performance. How large of a problem are these small dispatching and managing jobs?

benedictpaten:
I'm somewhat queasy about this change. 

I totally get the motivation for the PR, but I worry that it solves one problem (tiny jobs cluttering up the scheduler), to create another: more load on the leader program. In general, the design of toil is such that all efforts (and a lot of ugly code) is implemented to avoid creating load on the leader, because that becomes the bottleneck as you scale the number of worker nodes.

While Brad's done an awesome job of keeping the changes small, this change also makes special case code and state around CWL within the batch system class of several schedulers, which makes me throw up a little bit. Is there not a way to confine these changes to cwl-toil, by chaining the short jobs within larger jobs? There are potential (or at least theoretical) ways to do this that might ultimately be more scalable and general purpose by using the existing job API in a more effective way - what about extending the job class to implement extra "inbetween" logic for short lived jobs? This wouldn't fix every short-job case, but could potentially fix a number of them. 

In general, I want to get to the point where toil can run CWL pipelines at the scale of thousands of nodes and tens of thousands of cores, in short, I'm not sure changes like this are pushing in that direction - which is not to say that there are not many places where we need to radically improve the efficiency of the current CWL runner.

src/toil/cwl/cwltoil.py
```diff
@@ -59,6 +59,10 @@
 
 cwllogger = logging.getLogger("cwltool")
 
+# Define internal jobs we should avoid submitting to batch systems and logging
+CWL_INTERNAL_JOBS = ("CWLJob", "CWLJobWrapper", "CWLWorkflow", "CWLScatter", "CWLGather",
```
mr-c: Isn't `CWLJob` the actual running of the job itself and not just a synthetic job?

chapmanb: For batch systems CWLJob dispatches the actual job, so it doesn't get labelled as this step. So, the jobs themselves do still go the batch queue. Thanks for reviewing this.

chapmanb:
Benedict;
I understand your points about this not being the most elegant solution. I'd love if CWL were more of a first class citizen in Toil but my understanding is that these CWL shims are necessary now to map over CWL workflows, scatters and tool runs into the native Toil representations. @tetron can probably comment more on the architecture and if other setups would be feasible with additional work.

Practically, the current implementation of submitting these work steps leaves us unable to run any kind of real job on a production cluster using Toil. The overhead, and total number of jobs compared to work done, are just too high. We're at the point of trying to demonstrate that a handful of samples work so we can sort through jobStore issues (debugging, copying/linking files). Tens of thousands of samples would be sweet but we're still learning to crawl with our CWL jobs right now.

Did you have thoughts on who might available with the expertise and time to work on an improved solution? I'm open to any ideas to getting us unstuck and able to run cluster jobs with Toil. Thanks for all the discussion.

mr-c:
@benedictpaten Thank you for your thoughtful response.

I agree that it would be better for this mechanism to be more generic, though it does impact the batch systems with a shared queue more than others so the location of this change might be the correct one.

However the all of synthetic jobs targeted by this PR like `CWLWorkflow`,`ResolveIndirect`, `CWLGather`,`CWLScatter` either consume the output of many prior Jobs and/or spawn many follow-on jobs; thus making them ineligible for job bundling.

Because `CWLJobWrapper` exists  to spawn a job with the correct resource requirements based upon the inputs, it is also not a candidate for job bundling.




Inspired by this enhancement, in #1869 I am refactoring the LSF backend to adopt this shared interface, an upgrade that is long overdue.

tetron:
@benedictpaten I think the basic feature that's needed is a low latency work queue for very small tasks, sort of like a twelve items or less express line at the grocery store.  Perhaps implementation could be more elegant, but I think the basic idea is sound.

As @mr-c pointed out, the sorts of jobs we're talking about are mainly dynamic workflow elements: dynamic resource requirements, scatter/gather.  These inherently can't be combined with other jobs because their entire purpose is to either create a bunch of new jobs on the fly or depend on and consolidate the output of a set of upstream jobs.

chapmanb:
Benedict -- thanks again for all the thoughts, I appreciate the flexibility knowing this fix isn't exactly your vision. We're more on the side of nuts and bolts and making things work right now but happy to consider bigger changes in architecture as folks have time to work on those.

Regarding runtimes, the major difference in local cwltoil runtimes versus the base cwltool runner is from file copying in the job store. This adds overhead, especially as file sizes increase. I was hoping to explore an approach similar to `--linkImports` but using soft links instead of hard. The hard linking approach doesn't work for our use case since it also changes access permissions. This causes problems when you're using a globally installed set of input data (like reference genomes) where you don't have chmod permissions. I'm hoping symlinks can improve this, and get us a consistent runtime with raw cwltool.

benedictpaten:
Hey Brad, thanks, I'm also pragmatic, so we can merge this. 

I understand what Peter is saying,  I'm surprised these "join" jobs are actually that significant - @tetron if you eliminate edge cases like "scatter, but there is only one worker needed", etc. how many of these are present in the bcbio workflow?

Brad, What you say about the file staging makes total sense. This is a general issue about staging and access of files across workers on different systems that needs to be solved. We definitely need to address this.

Issue #1865, opened 2017-09-08
GA4GH-DREAM Challenge cwltoil pipeline can't find output file.


DailyDreaming:
Using the GA4GH-DREAM Challenge pipeline with cwltoil running biowardrobe fails (running locally).  It is the same error for multiple pipelines on the GA4GH-DREAM Challenge.  It complains about not finding a necessary output file, specifically:

In cwltoil.py, under the class ToilFsAccess(cwltool.stdfsaccess.StdFsAccess), the p variable in ToilFsAccess._abs attempts to find the file:

/tmp/toil-23e5f8c8-3b92-49f9-9ad2-fcfbdcad9499-21789b7fd99841a0b3dcc0c94da039ef/tmpILt07N/1d00e0e1-1614-4a87-bb20-da11c97b7563/t_EzUMS2PMlTF/tmp9oaTDF.tmp

file exists?: False
Instead, tmp9oaTDF.tmp is found here: [u'/tmp/tmplvtTBZ/tmp/s/g/tmp9oaTDF.tmp']
Error given:
    [job samtools-rmdup.cwl] Job error:
    ../../../tools/samtools-rmdup.cwl:99:3: Error collecting output for parameter 'rmdup_output':
    :1:1:   Did not find output file with glob pattern: '[u'tmp9oaTDF.tmp']'

I looked for the function with the error message and found the collect_output function in the draft2tool.py.  It does not seem to handle glob as expected (if it's a file or directory, then it gives this error and never builds r?).  It looks for something called: "outputEval" in binding, but never finds it, so it never writes 'r'.  This is on line 628 here: https://github.com/common-workflow-language/cwltool/blob/master/cwltool/draft2tool.py

Also, I think lines 592 to 616, in the same file, are never used, since r will always be blank when it tries to read "files in r".

I've attached an admittedly pretty mangled output file with combined print statements, stdout, and logging.  The lines of interest are 851 - 857: 
[output_biowardrobe.txt](https://github.com/BD2KGenomics/toil/files/1289153/output_biowardrobe.txt)

I just reran this hopefully with the recently merged #1863 PR, but I am getting the same error as before unfortunately.

@tetron @mr-c @chapmanb

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-6)
┆Issue Number: TOIL-6

chapmanb:
[...]

PR #1867, opened 2017-09-10
CWL: move tmp_outdir_prefix to work directory

chapmanb:
Places the output directories into work directory with other
temporary work files to avoid clutter and improve ability
to easily clean up.

https://github.com/BD2KGenomics/toil/issues/1865#issuecomment-328303879

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -423,7 +423,7 @@ class CWLJob(Job):
         os.environ["TMPDIR"] = os.path.realpath(opts.pop("tmpdir", None) or fileStore.getLocalTempDir())
         outdir = os.path.join(fileStore.getLocalTempDir(), "out")
         os.mkdir(outdir)
-        tmp_outdir_prefix = os.path.join(os.environ["TMPDIR"], "out_tmpdir")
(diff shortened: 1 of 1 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -423,7 +423,7 @@ def run(self, fileStore):
         os.environ["TMPDIR"] = os.path.realpath(opts.pop("tmpdir", None) or fileStore.getLocalTempDir())
         outdir = os.path.join(fileStore.getLocalTempDir(), "out")
         os.mkdir(outdir)
-        tmp_outdir_prefix = os.path.join(os.environ["TMPDIR"], "out_tmpdir")
+        tmp_outdir_prefix = os.path.join(opts.pop("workdir", None) or os.environ["TMPDIR"], "out_tmpdir")
```
tetron: Could write this as `opts.pop("workdir", os.environ["TMPDIR"])` I think.

chapmanb: Peter, thanks you can do but it becomes a problem if `workdir` is set but has a value of None (or False, or an empty string). I do this `or` bit only when I'm not sure the inputs won't be None so it's a little more safe.

ejacox:
Thank you @chapmanb and @tetron

Issue #1851, opened 2017-08-31
Change dockerCall to use the docker-py API.


DailyDreaming:
dockerCall in Toil as it is, is just a wrapper for calling docker via the terminal with minimal documentation.  This issue proposes a move to implement the docker-py API, a well-documented, standardized, and more natural interface for python to interact with docker.

PR #1876, opened 2017-09-13
dockerCall changed to use docker's API.

DailyDreaming:
Addresses #1851

Not ready to pull yet.  PyUnit tests need changes (as well as a couple of other places) and I need to make sure this is backwards compatible.

This PR makes use of docker's powerful python API (including an exception that detects if a docker image exists) and the documentation they've cheerfully provided here: http://docker-py.readthedocs.io/en/stable/containers.html

This PR is also a necessary step towards increasing our docker functionality in preparation for running WDL workflows using docker.


DailyDreaming:
Jenkins skip this please.

DailyDreaming:
Jenkins, skip this please.

src/toil/lib/docker.py
```diff
@@ -114,146 +114,67 @@ def _docker(job,
            RM (2) will stop the container and then forcefully remove it from the system
            using `docker rm -f`. This is the default behavior if defer is set to None.
     """
+
+    # default is 1000, which are standard user permissions
+    # set this to 0 to become root
+    # however all output files will have only root permissions
+    defaultUser = 1000
```
joelarmstrong: This might end up running the container as the "wrong" user. If you have only one user on your system, that user has UID 1000 (on Ubuntu, at least). But that isn't the case if you have multiple users on a system, or if you are running something other than Ubuntu. I have UID 501 on my laptop and 1621 on the CBSE systems, for example.

I'd suggest using os.getuid() to get the UID of the current user, rather than hardcoding 1000 as the default.

DailyDreaming: Yes, good call.  I'll include it in the next commit!

DailyDreaming:
Jenkins, skip this please.

PR #1881, opened 2017-09-18
Fix bugs in the Parasol batch system

joelarmstrong:
This should hopefully make Parasol usable on the master branch, so that we don't have to keep up our toil fork on juggernaut/kolossus. I've tested it out, and the memory requirements seem to be enforced correctly, which I know was a worry before.

- Didn't parse jobStore path correctly: it now does, and fails early
if using anything but the FileJobStore
- The previous version attempted to remove jobs that it didn't launch.

### DIFF
--- a/src/toil/batchSystems/parasol.py
+++ b/src/toil/batchSystems/parasol.py
@@ -35,6 +35,7 @@ from bd2k.util.processes import which
 
 from toil.batchSystems.abstractBatchSystem import BatchSystemSupport
 from toil.lib.bioio import getTempFile
+from toil.common import Toil
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/toil/batchSystems/parasol.py
```diff
@@ -222,15 +228,11 @@ def getJobIDsForResultsFile(self, resultsFile):
         Get all queued and running jobs for a results file.
         """
         jobIDs = []
-        for line in self._runParasol(['-results=' + resultsFile, 'pstat2'])[1]:
-            runningJobMatch = self.runningPattern.match(line)
-            queuedJobMatch = self.queuePattern.match(line)
```
ejacox: queuePattern was only used here and can be deleted

src/toil/batchSystems/parasol.py
```diff
@@ -190,7 +196,7 @@ def setEnv(self, name, value=None):
         return super(ParasolBatchSystem, self).setEnv(name, value)
 
     def __environment(self):
-        return (k + '=' + (os.environ[k] if v is None else v) for k, v in list(self.environment.items()))
+        return (k + '=' + (os.environ[k] if v is None else v) for k, v in self.environment.items())
```
ejacox: How about:
`from future.utils import listitems`
`itemlist = listitems(self.environment)`
to maintain python 3 compatibility?

joelarmstrong:
Good points, thanks.

ejacox:
@adderan, do you have time to review this today?

adderan:
Looks good. The test output is a little confusing though because the paraHub and paraNode threads are exiting with -9, which causes a traceback. I'm trying to see if there's a cleaner way to terminate the threads. Also it seems like we could allow running with other jobstores by just using a fixed place like /tmp to store the parasol results files, instead of the jobstore directory. I guess parasol will probably never be used outside of FileJobStore though.

PR #1892, opened 2017-09-20
Batch: cache instead of block for statePollingWait

chapmanb:
Improves support for configuring interval to query schedulers. This
builds off #1769 which blocked to avoid extra queries. Unfortunately
this also blocks submissions of jobs, making it hard to scale up
analyses. This change uses caching instead of blocking, so Toil can
continue to submit and monitor jobs and will refresh queries to the
scheduler only within the `statePollingWait` period. The behavior when
`statePollingWait` is unset does not change.

### DIFF
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -15,6 +15,7 @@
 from __future__ import absolute_import
 
 from builtins import str
+from datetime import datetime
 import os
 import shutil
 import logging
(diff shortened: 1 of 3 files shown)


brainstorm:
Totally endorse this PR, fixes/improves greatly on top of https://github.com/BD2KGenomics/toil/pull/1686 and https://github.com/BD2KGenomics/toil/pull/1617, thanks a lot @chapmanb!

ohofmann:
Yes please. Almost there to have Toil run on PBS.

ejacox:
Thank you @chapmanb

PR #1896, opened 2017-09-21
Refactor binpacking algorithm

joelarmstrong:
This PR refactors the notoriously unreadable bin-packing code into something a bit more sensible (though not at all perfect), adds actual unit tests for the bin-packing, and fixes a few bugs exposed by those tests. None of these bugs were introduced by the multiple-node-types PR, so I figured it wasn't worth holding up those necessary changes any longer. Specifically, this fixes:

- a bug where new node reservations (reservations are sort of like "time-slices" of resource usage at a node, so you can find the projected resource usage at each point in time) weren't split, meaning that a 5-minute job that uses 5 cores would be projected to use 5 cores for the entire 1-hour "wallTime" of the instance. This means in certain situations, more nodes would be created than would be optimal.
- a bug where a long-running job, which "extends" a node reservation for a very long time, can end up causing other jobs to be packed into its extended reservations, even though they may be projected to start several hours in the future. This ends up underprovisioning nodes if you have a few long-running jobs and many more short jobs. (I've noticed the old pre-multiple-node-types scaler underestimating the number of needed nodes with Cactus, though I don't know whether that's caused by this particular bug or not.)

I also want to use this to discuss a few other autoscaler parameters/behaviors that aren't directly related to the bin-packer. Basically, I think it's worth having a discussion on a few topics:
- Do we still need to pack based on recent jobs as well as queued jobs? I'd argue the queued jobs are enough, and packing recent jobs is a bit confusing.
- The beta parameter has always annoyed me a bit: I like the basic idea, but I think the current behavior is less useful than it could be. Right now it is basically used to ignore small fluctuations, which is fine. But the problem I usually have with the scaler isn't with small fluctuations, it's with fluctuations that are too large, too fast. Usually this happens when thousands of small jobs are being dumped all at once, right after a series of long jobs, causing the autoscaler to freak out and hit its max -- then it rapidly figures out the jobs are pretty short and adjusts to something smaller than the max within 5 minutes, wasting some money. So I'd prefer the inertia parameter to be something much more like a moving average, which could give the autoscaler a chance to adjust, while still being able to hit the max in a reasonable time. (The moving average could still have some minor fluctuations, so maybe the existing beta parameter behavior would still be needed.)
- How should we order the "preference" of nodes to launch? Currently a sort is used which will do the "right" thing in the vast, vast majority of cases--but not all. Should we use the ordering provided by the user, and trust them to do it right? Or should we sort to prioritize the cheapest instances first, which is probably what the user wants?

### DIFF
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -93,8 +93,8 @@ class BinPackedFit(object):
     binPack method on a list of job shapes, and use getRequiredNodes
     to get how many nodes it thinks are required.
 
-    :param Shape nodeShapes: A list of possible types of nodes that can be launched.
-
+    :param list nodeShapes: A list of possible types of nodes that can be launched (as "Shape"s).
+    :param targetTime: The time before which all jobs should at least be started.
     """
     def __init__(self, nodeShapes, targetTime=3600):
         self.nodeShapes = nodeShapes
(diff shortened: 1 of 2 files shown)


src/toil/provisioners/clusterScaler.py
```diff
@@ -84,150 +85,180 @@ def get(self):
...
+        for nodeShape, nodeReservations in iteritems(self.nodeReservations):
+            for nodeReservation in nodeReservations:
+                if nodeReservation.attemptToAddJob(jobShape, nodeShape, self.targetTime):
+                    # We succeeded adding the job to this node reservation. Now we're done.
+                    return
+
+        # If we get here, the job didn't fit into any existing node
+        # reservations. Assign to the smallest node shape that will
+        # fit this job, prioritizing preemptable nodes
+        consideredNodes = [nodeShape for nodeShape in self.nodeShapes if nodeShape.cores >= jobShape.cores and nodeShape.memory >= jobShape.memory and nodeShape.disk >= jobShape.disk and (jobShape.preemptable or not nodeShape.preemptable)]
```
adderan: This is essentially duplicating the NodeReservation.fits() function. Not sure if it's worth the effort to avoid duplicated code, but maybe NodeReservation should be a subclass of Shape, and Shape should define the fits function?

src/toil/provisioners/clusterScaler.py
```diff
@@ -84,150 +85,180 @@ def get(self):
...
+                    # Packed the job.
+                    return True
+
+                # If the job would fit, but is longer than the total node allocation
+                # extend the node allocation
+                elif endingReservation.nReservation == None and startingReservation == self:
+                    # Extend the node reservation to accommodate jobShape
+                    endingReservation.nReservation = NodeReservation(nodeShape)
+            else:
+                if startingReservationTime + jobTimeSoFar <= targetTime:
```
adderan: I'm a little worried that this depends on the wall times for queued jobs being correct, when they are really just guesses. If we're assuming the wall times are correct, this solution shouldn't be necessary in the first place because only the correct number of jobs will be stacked up serially in the long reservation, and the others will result in new nodes being provisioned.

src/toil/provisioners/clusterScaler.py
joelarmstrong: yes! excellent point.

src/toil/provisioners/clusterScaler.py
joelarmstrong: I totally agree that the correct number of jobs will be put into the long reservation, but I think the whole point of a scaler is that it tries to schedule your jobs "soon" (for some, totally arbitrary, definition of "soon"). This tries to define "soon" as 1 hour, which seems OK to me :). But, then again, it won't necessarily speed up the workflow. The time that the workflow finishes is, of course, the time that the last job finishes. So it's not necessarily best to schedule jobs earlier if you can fit them in later.

But I think it's best to schedule jobs as early as possible, because they may spawn *more* jobs--which again may spawn more jobs, etc. But you bring up a good point in that we may be wasting money by scheduling jobs than is strictly necessary.

I don't think there's a great, 100% solution to this.

DailyDreaming:
@joelarmstrong What's the status on this PR?

joelarmstrong:
Invalidated by later work, which I swear I'll merge at some point :)

PR #1904, opened 2017-09-22
Changed file imports to symlinks instead of hard links (resolves #1903)

### LINKED ISSUE #1903: Symlink imported files instead of hard linking in the file jobStore.

ejacox:


ejacox:
Thank you for being a pain @joelarmstrong. The comments are very helpful. I removed the permission setting since this should not be applicable with symlinks. Perhaps that is why hard links were used so that the link could be made read only.

joelarmstrong:
LGTM, thanks for making the changes. Hmm, that failing test is annoying--that should clearly result in a retry, but it's absurd that boto is raising a ValueError. At some point we will just be adding every possible exception to the list of retriable s3 exceptions. (Maybe we should just retry on every exception, no matter what? It seems like total overkill, but at least we wouldn't be playing whack-a-mole with exceptions).

ejacox:
Jenkins, test this please.

ejacox:
Addressed in #1914

Issue #1883, opened 2017-09-18
File names should be preserved in the fileStore.


DailyDreaming:
Currently this is implemented with:

sample2_preserveThisFilename = "GVCF_rawLikelihoods.g.vcf"
sample2_filepath = os.path.join(tempDir, sample2_preserveThisFilename)
sample2 = job.fileStore.readGlobalFile(sample2, userPath=sample2_filepath)

It would make things more streamlined if this name preservation was on by default, prepended with a unique ID.  Nearly all bioinformatics tools require certain extensions to be present on the filenames, which makes this something that is always necessary to implement anyway.  This would make implementation smoother for the user and the code less unnecessarily verbose.

PR #1914, opened 2017-10-06
Define fileStore semantics (resolves #1887, #1903)

### LINKED ISSUE #1887: Clearly define and implement the semantics of global file store
The semantics should be documented clearly in the API.

The motivation for this is to make global file store behavior clear and as consistent and efficient as possible across different files stores and with or without caching.

The suggest public semantics are:

- With readGlobalFile mutable=False
  - File maybe either copied, hard-linked or symbolic-linked to the jobStore.  The exact method depends on the jobStore implementation.
  - The jobStore file may be set to write-protected.  This behavior is not guaranteed and depends on the jobStore implementation.  For instance, on a shared file system jobStore, symbolic links will be created to the global file location, however the file permissions will not be changed.
  - The behavior of modifying a file marked as immutable is undefined.  It may result if crashes, incorrect results or data corruption.  Do not do this.

- With readGlobalFile mutable=True
  - The file will be copied to the jobStore
  - The file permissions will be set to writable.

The above semantics are maintained with cached enabled or disabled.

TO DO:
  - Defined semantics of file naming, including user name, along with rationale.
  - Support directories of files as first class file store objects.
  - readGlobalFile default for mutable should be `False', not `None.
  - Do not reference to internal modules, e.g. cache (bool) – Described in toil.fileStore.CachingFileStore.readGlobalFile()

ejacox:
- [x] Define the semantics of of readGlobalFile.
- [x] Disable caching by default.
- [x] Mutable false by default. -- Oct 6th
- [x] Symlink instead of hard link files when possible (because sym links are more versatile and work across different filesystems)
- [ ] Ensure that caching still correctly tracks links.
- [x] Preserve file names (optional in this PR), see #1883.
- [ ] Better document role of fileStore vs jobStore.

RULE: If Toil did not create the file, Toil will not change permissions.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -183,13 +183,7 @@ class FileJobStore(AbstractJobStore):
         # linking is not done be default because of issue #1755
         srcPath = self._extractPathFromUrl(srcURL)
         if self.linkImports:
-            try:
-                os.link(os.path.realpath(srcPath), destPath)
-            except OSError:
-                shutil.copyfile(srcPath, destPath)
-            else:
-                # make imported files read-only if they're linked for protection
-                os.chmod(destPath, 0o444)
+            os.symlink(os.path.realpath(srcPath), destPath)
         else:
             shutil.copyfile(srcPath, destPath)
 
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -190,12 +190,7 @@ class FileJobStore(AbstractJobStore):
     def _importFile(self, otherCls, url, sharedFileName=None):
         if issubclass(otherCls, FileJobStore):
             if sharedFileName is None:
-                fd, absPath = self._getTempFile()  # use this to get a valid path to write to in job store
-                os.close(fd)
-                os.unlink(absPath)
-                # remove the .tmp extension and add the original file name
-                (noExt,ext) = os.path.splitext(absPath)
-                absPath = noExt + '-' + os.path.basename(url.path)
+                absPath = self._getUniqueName(url.path)  # use this to get a valid path to write to in job store
                 self._copyOrLink(url, absPath)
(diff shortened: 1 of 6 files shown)


ejacox:
Jenkins, skip this test.

ejacox:
I changed the fileJobStore to use the original file name when importing files. The Azure and AWS file stores use UUIDs. I wasn't sure if there was a reason for this and didn't change it.

I had hoped to use the sharedFileName functions to do this. However, those commands have a different behavior, i.e. writeFileStream returns a path to use. This seems overly complicated. I don't see reason to have readFileStream and readSharedFileStream, which are incompatible. Unfortunately, these are part of the API.

src/toil/jobStores/fileJobStore.py
```diff
@@ -193,6 +193,9 @@ def _importFile(self, otherCls, url, sharedFileName=None):
                 fd, absPath = self._getTempFile()  # use this to get a valid path to write to in job store
                 os.close(fd)
                 os.unlink(absPath)
+                # remove the .tmp extension and add the original file name
+                (noExt,ext) = os.path.splitext(absPath)
+                absPath = noExt + '-' + os.path.basename(url.path)
```
jessebrennan: This only fixes files that enter the jobStore from `importFile()`. The underlying filename is set from [`fileJobStore._getTempFile()`](https://github.com/BD2KGenomics/toil/blob/5819ff372f08f146ff17e7053bfe34f1347624a8/src/toil/jobStores/fileJobStore.py#L445)

Which is used to create the file name any time a file enters the jobStore, be it from importFile, writeGlobalFile, or writeGlobalFileStream. The fix should happen there instead.

jessebrennan: To add to this, it may be worth while to add a test that ensures that this behaves as expected and that the file still has the correct name (or at least extension) if it was brought into the jobStore from importFile, writeGlobalFile, or writeGlobalFileStream.

PR #1923, opened 2017-10-13
Move and copy all cgcloud deps to toil/lib (resolves #1813)

### LINKED ISSUE #1813: Remove cgcloud lib dependencies from the AWS provisioner.
At a minimum, the code from cgcloud should be moved into Toil. The instance types should not be hard-coded. Consider keeping them in a separate file. Also, is there way to use instance types that are not pre-defined?

jessebrennan:
Hopefully I didn't miss anything...


setup.py
```diff
@@ -40,8 +40,8 @@ def runSetup():
                 'psutil==3.0.1'],
             'aws': [
                 botoRequirement,
-                'cgcloud-lib==' + version.cgcloudVersion,
-                'futures==3.0.5'],
+                'futures==3.0.5',
+                'pycrypto==2.6.1'],
```
psafont: Hi!

pycrypto hasn't been maintained in 3 years. It also doesn't support Python 3. (I think this last point is a work in progress for toil)

 Consider using [pycryptodome](https://pypi.python.org/pypi/pycryptodome) (drop-in replacement) or [cryptography](https://pypi.python.org/pypi/cryptography) as a dependency.

ejacox: I created a separate issue for this. Thanks @psafont.

Issue #1926, opened 2017-10-17
SystemTest.testAtomicityOfNonEmptyDirectoryRenames fails on some systems (namely Travis-CI)


jessebrennan:
The trace looked like:
```
______________ SystemTest.testAtomicityOfNonEmptyDirectoryRenames ______________
Traceback (most recent call last):
  File "/home/travis/build/jessebrennan/toil/src/toil/test/src/systemTest.py", line 26, in testAtomicityOfNonEmptyDirectoryRenames
    grandChildIds = grandChildIds.get()
  File "/opt/python/2.7.13/lib/python2.7/multiprocessing/pool.py", line 567, in get
    raise self._value
OSError: [Errno 17] File exists
----------------------------- Captured stdout call -----------------------------
2017-10-16 19:10:37,112 - toil.test - INFO - Setting up toil.test.src.systemTest.SystemTest.testAtomicityOfNonEmptyDirectoryRenames ...
2017-10-16 19:10:37,475 - toil.test - INFO - Tore down toil.test.src.systemTest.SystemTest.testAtomicityOfNonEmptyDirectoryRenames
----------------------------- Captured stderr call -----------------------------
INFO:toil.test:Setting up toil.test.src.systemTest.SystemTest.testAtomicityOfNonEmptyDirectoryRenames ...
INFO:toil.test:Tore down toil.test.src.systemTest.SystemTest.testAtomicityOfNonEmptyDirectoryRenames
```

Fortunately I ran into @hannes-ucsc and he explained this all away:

> It turns out that ENOTEMPTY and EEXIST are equivalent and that the test should accept both as expected failure in the processes that lose the race.
> 
> http://man7.org/linux/man-pages/man2/rename.2.html
> 
> Also keep in mind that multiprocessing.pool.AsyncResult.get() blocks until all tasks are done and A) returns and iterable of the task return values or B) re-raises the first exceptions occurring in a task. That's why the stack trace looks funny.
> 
> https://docs.python.org/2/library/multiprocessing.html#multiprocessing.pool.AsyncResult

PR #1930, opened 2017-10-19
Quick integration testing with Travis-CI (resolves #1771, #1926)

jessebrennan:
This gives us a basic Travis-CI setup.

It only runs our simplest and quickest unit tests. This ends up being ~100 tests (not a lot).

To finish setting this up, Travis will need to be activated for the repo. [Super easy instructions](https://docs.travis-ci.com/user/getting-started/) to do this are available on Travis's website, but require access to the BD2KGenomics github account AFAIK.

The default Travis configuration is to run on every commit on every remote branch. I think this makes sense and fits the development model where work is saved on a remote branch. Once Travis tests pass a PR is made in which case all Jenkins and Travis tests run.

Successful Travis tests results are available https://travis-ci.org/jessebrennan/toil/

Things that can be improved upon:
- Adding more extras (and thus more tests) to the install. Probably this would be no problem, but I wanted to be sure the most basic setup would work first. This can also be parallelized so that Travis will simultaneous test Toil with different levels of installation. That may or may not be desired..
- Adding Python 3 testing. This is super simple, but I left it out b/c I figured it would be best to start with something that worked.
- Further parallelizing the testing. Not really an issue now since the current tests only take ~9 minutes.
- Some tests (like `testServiceDeadlock`, and clusterScaler tests) just refused to run on Travis. My only explanation was resource deadlock since Toil expects a semi-powerful machine to be running tests (instead of a t2.micro). These tests should probably be fixed to run on any machine.


jessebrennan:
Jenkins test this please

jessebrennan:
@ejacox this is ready for review. Really the next step is to activate Travis-CI for the repository. Instructions are in the PR message.

jessebrennan:
@ejacox Travis still has to be configured to run on the repository. I think you should be able to do this from this link: https://www.travis-ci.org/BD2KGenomics/toil

PR #1944, opened 2017-10-27
Add convenience Job attributes (resolves #1940)

### LINKED ISSUE #1940: Convenience job attributes
I was talking with a few other pipeline developers and a few additions to the `job` instance, like `job.cores`,  would cut down on boilerplate and make the code more readable.

- [ ] `job.tempDir` —  At the top of every job function I have to write `work_dir = job.fileStore.getLocalTempDir()` to create and get the localTempDir path. Instead, have `job.tempDir` return the file path from `job.fileStore.getLocalTempDir()`, but don't require me to invoke it to create the local temp dir, i.e. create the directory when invoked for the first time in a job function or just always create it if that overhead isn't large. 

- [ ] `job.log` — We could just point this to `job.fileStore.logToMaster` or w/e the preferred logging  system is.

jessebrennan:
This PR adds a job.tempDir attribute that contains the path to a tempDir following @jvivian's suggestion to create the tempDir the first time the attribute is accessed. 

Also it makes job.log an alias for fileStore.LogToMaster.

Docs are updated to make use of the new interface.

Tests are added to ensure things work as expected.


jessebrennan:
Jenkins test this please

ejacox:
Jenkins, test this please.

jessebrennan:
There is more to fix here. I'm still working on it.

jessebrennan:
Seems like failures actually are unrelated.. first only `src.toil.test.src.fileStoreTest.CachingFileStoreTestWithAwsJobStore.testAsyncWriteWithCaching`. second, `src.toil.test.src.promisedRequirementTest.MesosPromisedRequirementsTest.testConcurrencyStatic` failed while the other passed.

Hopefully everything is actually fine. Jenkins, test this please.

jessebrennan:
@ejacox The last commit is a minor spelling fix. I figured it would be a waste of time to test. It looks like the original test that I pushed now "passed" according to Jenkins while before it was a failure (as documented in John's comment above).

We can run them again if you want, but otherwise I think this is ready to merge.

ejacox:
Looks good. Thank you @jessebrennan

Issue #1900, opened 2017-09-21
Monitoring and logging


ejacox:
This is a general PR to track needed improvements in monitoring and logging.

- [ ] Add more fine-grained logging, e.g. run time of individual jobs
- [ ] Add more information on job status, e.g. where things are stuck
- [ ] Make jobStore more accessible. 
- [ ] Remove extraneous logging messages.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-207)
┆Issue Number: TOIL-207

PR #1977, opened 2017-11-29
Logging edits (#1900).

DailyDreaming:
I've edited some of the logging messages, particularly the ones I found most egregious.  Message by message, these are the ones I've changed and why:

**leader.py**
*logger.debug("Job %s services have completed or totally failed, adding to updated jobs", predecessorJob)*
Saying that jobs have either "completed or totally failed" has never seemed that useful to me.  I can deduce that it's completed or failed when I'm looking at the log because the log is no longer generating new output.

**singleMachine.py**
*log.debug('Exiting worker thread normally.')*
*log.debug('Received queue sentinel.')*
These repeat ad nauseam for hundreds of repeated lines and seem to say nothing useful.  If they are insanely useful for some common edge case, someone please let me know.

**toilClean.py**
*logger.info("Attempting to delete the job store")*
**common.py**
*logger.info("Attempting to delete the job store")*
Edited this to be a try/except that only logs on failure or success.

**statsAndLogging.py**
*2017-11-14 16:02:33,780 - toil.statsAndLogging - DEBUG - 'EncapsulatedJob' v/g/jobFA66ni    WARNING:toil.resource:Can't find resource for leader path '/home/lifeisaboutfishtacos/Desktop'*
*2017-11-14 16:02:33,781 - toil.statsAndLogging - DEBUG - 'EncapsulatedJob' v/g/jobFA66ni    WARNING:toil.resource:Can't localize module ModuleDescriptor(dirPath='/home/lifeisaboutfishtacos/Desktop', name='new_compiled', fromVirtualEnv=False)*
These two messages trigger together.  One prints when returning a None value, and the other prints when receiving a None value from that function.  I also had no idea what these meant when first reading them.  I've edited them to just one message, and something slightly more in line with how I think they run.  If this could be written in a better, clearer way, please discuss.

**job.py**
*logger.debug('Getting %s from user module __main__ (%s).', class_name, userModule)*
*logger.debug('Getting %s from module %s.', class_name, module_name)*
They print the modules that are imported at the beginning of each file which I don't see as particularly useful (you can find that information in the stack trace generally if something fails to load).  I've made this a try/except statement just in case, so that it will only log if things fail.


src/toil/common.py
```diff
@@ -691,9 +691,11 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
                             exc_type is None and self.config.clean == "onSuccess" or
                         self.config.clean == "always"):
-                logger.info("Attempting to delete the job store")
-                self._jobStore.destroy()
-                logger.info("Successfully deleted the job store")
+                try:
+                    self._jobStore.destroy()
+                    logger.info("Successfully deleted the job store: %s" % str(self._jobStore))
+                except:
+                    logger.info("Failed to delete the job store: %s" % str(self._jobStore))
```
jessebrennan: Exception should be re-raised here after logging. As it is this will swallow the exception and just log without crashing the program, which i don't think is what we want

src/toil/job.py
```diff
@@ -915,12 +915,16 @@ def _unpickle(cls, userModule, fileHandle, config):
...
+            try:
+                if module_name == '__main__':
+                    return getattr(userModule, class_name)
+                else:
+                    return getattr(importlib.import_module(module_name), class_name)
+            except:
+                if module_name == '__main__':
+                    logger.debug('Failed getting %s from module %s.', class_name, userModule)
+                else:
+                    logger.debug('Failed getting %s from module %s.', class_name, module_name)
```
jessebrennan: again, just add 
`raise`
as the next line so that this doesn't eat up the exception

src/toil/utils/toilClean.py
```diff
@@ -31,7 +31,9 @@ def main():
...
-    logger.info("Attempting to delete the job store")
-    jobStore = Toil.getJobStore(config.jobStore)
-    jobStore.destroy()
-    logger.info("Successfully deleted the job store")
+    try:
+        jobStore = Toil.getJobStore(config.jobStore)
+        jobStore.destroy()
+        logger.info("Successfully deleted the job store: %s" % str(jobStore))
+    except:
+        logger.info("Failed to delete the job store: %s" % str(jobStore))
```
jessebrennan: raise raise raaaaaays

src/toil/common.py
DailyDreaming: Fixed.

src/toil/job.py
DailyDreaming: Fixed.

src/toil/utils/toilClean.py
DailyDreaming: Fixed.

PR #1978, opened 2017-11-30
Update quickStart.rst

ionox0:
pip is picky about installing libraries with extra requirements. 

pip 9.0.1 works with quotes

https://stackoverflow.com/questions/19096155/setuptools-and-pip-choice-of-minimal-and-complete-install


ionox0:
As a followup, I should have actually referenced this issue related to zsh as the source of this error, quotes are still a good solution though

https://stackoverflow.com/questions/30539798/zsh-no-matches-found-requestssecurity/30539963

PR #2007, opened 2018-01-08
expose cwltool's --tmpdir-prefix

mr-c:
Needed for running CWL workflows on OS X


mr-c:
Needs merging after https://github.com/BD2KGenomics/toil/pull/2008

Issue #2017, opened 2018-01-12
Disk usage calculation is hard-coded to Unix.


ejacox:
getDirSizeRecursively in common.py uses a hard-coded, Unix specific block size. The disk usage will not be correct on machines using different block sizes. This causes the test testGetSizeOfDirectoryWorks in miscTests to fail on MacOs, which can use a different block size.

DailyDreaming:
Does this fix it?  #2019

PR #2019, opened 2018-01-13
Change block size dynamically.  #2017

DailyDreaming:
@ejacox I have no way of testing this, but does this fix the issue?

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -54,8 +54,10 @@ from toil.version import dockerRegistry, dockerTag
 
 logger = logging.getLogger(__name__)
 
-# This constant is set to the default block size of the system using os.stat('/').st_blksize.
-systemBlockSize = int(os.stat('/').st_blksize / 8)
+# This constant is set to the default value used on unix for block size (in bytes) when
+# os.stat(<file>).st_blocks is called.
+unixBlockSize = 512
+
 
 class Config(object):
     """
@@ -541,7 +543,7 @@ def _addOptions(addGroupFn, config):
     #Debug options
     #
     addOptionFn = addGroupFn("toil debug options", "Debug options")
-    addOptionFn("--debug-worker", default=False, action="store_true", dest="debugWorker",
+    addOptionFn("--debugWorker", default=False, action="store_true",
             help="Experimental no forking mode for local debugging."
                  " Specifically, workers are not forked and"
                  " stderr/stdout are not redirected to the log.")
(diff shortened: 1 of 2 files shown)


ejacox:
@DailyDreaming, this didn't fix the problem.

joelarmstrong:
This is actually a problem with the test, which is incorrectly calculating the filesystem's block size:

https://github.com/BD2KGenomics/toil/blob/f462225e5ddf2848ede783f953f18fe18ccc8ff3/src/toil/test/src/miscTests.py#L57

The actual getDirSizeRecursively code was probably fine the way it was. But `st_blksize` is *not* guaranteed to be the filesystem-level blocksize (i.e. the minimum allocation size). It happens to be that way on ext2/3/4 on Linux, as far as I can tell, but according to the documentation it is just the "preferred" I/O size. On macOS it's 4MiB.

If we really, really want the filesystem-level blocksize, we could think about using the `f_frsize` entry of [statvfs](http://man7.org/linux/man-pages/man3/statvfs.3.html). But why not just use du in the test to get the actual size on disk, and call it a day? I'm not sure we need to do the allocation-size guessing that the test currently does.

DailyDreaming:
Alright, changed it to use du!

src/toil/common.py
```diff
@@ -543,7 +543,7 @@ def _addOptions(addGroupFn, config):
     #Debug options
     #
     addOptionFn = addGroupFn("toil debug options", "Debug options")
-    addOptionFn("--debug-worker", default=False, action="store_true", dest="debugWorker",
+    addOptionFn("--debugWorker", default=False, action="store_true",
```
joelarmstrong: Why this change? is this because of a bad merge or something? This line got changed 3 times in 4 commits! And the change has already been applied on master. IMO this PR should be squashed or otherwise rebased so this line won't be changed.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
-    :rtype: int
+    This method will return a 'subprocess.CalledProcessError' if it is unable to
+    access a folder or file because of permissions.  Therefore this method should
+    only be called on the jobStore, and will alert the user if some portion is
+    inaccessible.  Everything in the jobStore should have appropriate permissions
+    as there is no way to read the filesize without permissions.
+
+    du is often faster than using os.lstat(), sometimes significantly so.
+
+    The call: 'du -c /some/path' should give the number of 512-byte blocks
```
joelarmstrong: this still reads `du -c` instead of `du -s`

diekhans:
seems like it mostly comments that need fixed ;-)

src/toil/common.py
DailyDreaming: Yes, it was a bad merge.  I'll just squash this and make a new one.

PR #2024, opened 2018-01-17
LSF: Separately determine memory reservation and memory limit values.

tmooney:
In the default LSF configuration `mem` in `-R` and `-M` are in different units, so we can't reuse the same value for both.


ejacox:
Jenkins, test this please.

ejacox:
Thank you @tmooney and @mr-c.

PR #2026, opened 2018-01-19
Add HTCondor batch system

jasoncpatton:
Current version works with HTCondor pools that use a shared file system (this includes, for example, a personal HTCondor pool on a single machine). Future versions will work with distributed HTCondor pools.

Requires the HTCondor Python bindings, which are either installed by the system package manager as part of a HTCondor install or can be installed separately via `pip install htcondor`.

Passes tests in `src/toil/test/batchSystems/batchSystemTest.py` when tested on a machine with a personal HTCondor pool.

To do to make this plugin more complete (will open issue): 
1. Transfer local file stores to the execute nodes.
2. Check pool resources.
3. Support running in the Open Science Grid.

### DIFF
--- a/src/toil/batchSystems/htcondor.py
+++ b/src/toil/batchSystems/htcondor.py
@@ -1,0 +1,306 @@
+# Copyright (C) 2018, HTCondor Team, Computer Sciences Department,
+# University of Wisconsin-Madison, WI.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you
+# may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+from builtins import str
+
+import sys
+import os
+import logging
+import time
+import math
+
+from toil.batchSystems.abstractGridEngineBatchSystem import AbstractGridEngineBatchSystem
+
+import htcondor
+import classad
+
+logger = logging.getLogger(__name__)
+
+class HTCondorBatchSystem(AbstractGridEngineBatchSystem):
+    # When using HTCondor, the Schedd handles scheduling
+
+    class Worker(AbstractGridEngineBatchSystem.Worker):
+
+        # Override the createJobs method so that we can use htcondor.Submit objects
+        # and so that we can get disk allocation requests and ceil the CPU request.
+        def createJobs(self, newJob):
+            activity = False
+
+            if newJob is not None:
+                self.waitingJobs.append(newJob)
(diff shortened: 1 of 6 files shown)


ejacox:
Jenkins, test this please

benedictpaten:
This is great to see. @jasoncpatton Can you take a look at the failures Jenkins is reporting? If there are failures you don't understand we can try to find some developer time to help out to get this merged!

jasoncpatton:
I see what the problem is, but I'm not sure how to address it.

The `import htcondor` statement won't (and shouldn't) work on the test system since htcondor is not a required package for Toil as a whole, though it is required for using the HTCondor batch system plugin. How can this be handled gracefully? Should htcondor be another install option (like mesos)?

mr-c:
@jasoncpatton Yes, it should be an additional install option https://github.com/BD2KGenomics/toil/blob/master/setup.py#L19

ejacox:
Jenkins, test this please

jasoncpatton:
@ejacox Should I add htcondor to the extras in jenkins.sh? (Or is that "cheating"?)

ejacox:
@jasoncpatton, That's exactly right. Add it to the "make develop ..." line so that Jenkins will install it before running a test.

ejacox:
Jenkins, test this please

jasoncpatton:
That should cover all the types of exceptions that our python bindings can raise when they can't find the condor scheduler process or config.

ejacox:
Jenkins, test this please.

ejacox:
Jenkins, add to whitelist

jasoncpatton:
I could use some pointers in debugging these tests, it's not clear to me how (if) they are related to the new batch system (and batch system test) code.

ejacox:
We have been getting more transient errors lately in our AWS tests. We will track down the problems when we have time. In the mean time, we rerun the tests or ignore them.

ejacox:
Jenkins, test this please.

ejacox:
@jasoncpatton Is it ready to merge? We can ignore the AWS failures.

src/toil/batchSystems/htcondor.py
```diff
@@ -0,0 +1,306 @@
...
+        localID = self.handleLocalJob(jobNode)
+        if localID:
+            return localID
+        else:
+            self.checkResourceRequest(jobNode.memory, jobNode.cores, jobNode.disk)
+            jobID = self.getNextJobID()
+            self.currentJobs.add(jobID)
+
+            # Add the jobNode.disk and jobNode.jobName to the job tuple
+            self.newJobsQueue.put((jobID, jobNode.cores, jobNode.memory, jobNode.disk, jobNode.jobName, jobNode.command))
```
mr-c: General question for Toil devs, why don't we put the `jobNode` itself in the `newJobsQueue`?

jasoncpatton:
Yes

src/toil/batchSystems/htcondor.py
ejacox: Toil only needs the jobId. It seems the queue was only jobIds originally and later changed to a tuple to include extra information used by the batch system.

ejacox:
Thank you @jasoncpatton

PR #2029, opened 2018-01-22
Add [all] extra.

DailyDreaming:
Allows toil to be built from source using just: `make develop extras=[all]` so that I don't have to copy-paste all of the time when building from source.  Also relocates a global variable.

### DIFF
--- a/docs/gettingStarted/install.rst
+++ b/docs/gettingStarted/install.rst
@@ -67,11 +67,17 @@ install any necessary headers and libraries (`python-dev`_, `libffi-dev`_). Then
 
     $ pip install toil[aws,mesos,azure,google,encryption,cwl]
 
+or::
+
+    $ pip install toil[all]
+
 Here's what each extra provides:
 
 +----------------+------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


natanlao:
We should update the [documentation](https://toil.readthedocs.io/en/3.12.0/gettingStarted/install.html?highlight=extras#installing-toil-with-extra-features) to reflect this as well

ejacox:
Can it be written so that the dependencies don't need to be in two places? For example, can 'all' invoke installing 'aws', 'azure', ... Having to update two lists would be a source of errors.

DailyDreaming:
Updated docs and consolidated the list.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Some new flaky tests?

src.toil.test.sort.sortTest.SortTest.testAwsMesos
src.toil.test.sort.sortTest.SortTest.testGoogleSingle

DailyDreaming:
Jenkins, test this please.

ejacox:
Thank you @DailyDreaming

PR #2031, opened 2018-01-27
Changed getDirSizeRecursively to use du.

DailyDreaming:
Addresses #2017.

Should be faster (sometimes much faster) and won't skip 'root' protected directories, erroneously telling you it's 0 size.  Also, doesn't break on broken symlinks.  Should work on Apple things too.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1257,26 +1257,29 @@ def cacheDirName(workflowID):
 
 def getDirSizeRecursively(dirPath):
     """
-    This method will return the cumulative filesize in bytes of all of the files
-    in the directory and its subdirectories using 'du'.
+    This method will return the cumulative number of bytes occupied by the files 
+    on disk in the directory and its subdirectories.
 
-    This method will return a 'subprocess.CalledProcessError' if it is unable to
-    access a folder or file because of permissions.  Therefore this method should
-    only be called on the jobStore, and will alert the user if some portion is
-    inaccessible.  Everything in the jobStore should have appropriate permissions
-    as there is no way to read the filesize without permissions.
-
-    du is often faster than using os.lstat(), sometimes significantly so.
-
-    The call: 'du -s /some/path' should give the number of 512-byte blocks
-    allocated with the environment variable: BLOCKSIZE='512' set, and we
-    multiply this by 512 to return the filesize in bytes.
+    This method will raise a 'subprocess.CalledProcessError' if it is unable to
+    access a folder or file because of insufficient permissions.  Therefore this 
+    method should only be called on the jobStore, and will alert the user if some 
+    portion is inaccessible.  Everything in the jobStore should have appropriate 
+    permissions as there is no way to read the filesize without permissions.
+    
+    The environment variable 'BLOCKSIZE'='512' is set instead of the much cleaner
+    --block-size=1 because Apple can't handle it.
 
     :param str dirPath: A valid path to a directory or file.
-    :return: Size, in bytes, of dirPath, and recursively everything within it.
+    :return: Total size, in bytes, of the file or directory at dirPath.
     """
+    
+    # du is often faster than using os.lstat(), sometimes significantly so.
+    
+    # The call: 'du -s /some/path' should give the number of 512-byte blocks
+    # allocated with the environment variable: BLOCKSIZE='512' set, and we
+    # multiply this by 512 to return the filesize in bytes.
     return int(subprocess.check_output(['du', '-s', dirPath],
-               env=dict(os.environ, BLOCKSIZE='512')).split()[0].decode('utf-8')) * 512
+               env=dict(os.environ, BLOCKSIZE='512')).split()[0].decode('ascii')) * 512
 
 
 def getFileSystemSize(dirPath):
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
 
 def getDirSizeRecursively(dirPath):
     """
-    This method will walk through a directory and return the cumulative filesize in bytes of all
-    the files in the directory and its subdirectories.
+    This method will return the cumulative filesize in bytes of all of the files
+    in the directory and its subdirectories using 'du'.
 
-    :param dirPath: Path to a directory.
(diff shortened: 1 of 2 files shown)


src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
 
 def getDirSizeRecursively(dirPath):
     """
-    This method will walk through a directory and return the cumulative filesize in bytes of all
-    the files in the directory and its subdirectories.
+    This method will return the cumulative filesize in bytes of all of the files
+    in the directory and its subdirectories using 'du'.
```
hannes-ucsc: Don't mention implementation details.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
 
-    :param dirPath: Path to a directory.
-    :return: cumulative size in bytes of all files in the directory.
-    :rtype: int
+    This method will return a 'subprocess.CalledProcessError' if it is unable to
```
hannes-ucsc: s/return/raise

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
+    access a folder or file because of permissions.  Therefore this method should
+    only be called on the jobStore, and will alert the user if some portion is
+    inaccessible.  Everything in the jobStore should have appropriate permissions
+    as there is no way to read the filesize without permissions.
+
+    du is often faster than using os.lstat(), sometimes significantly so.
```
hannes-ucsc: Implementation detail. Make this a comment in the body of the method.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
-    This method will walk through a directory and return the cumulative filesize in bytes of all
-    the files in the directory and its subdirectories.
+    This method will return the cumulative filesize in bytes of all of the files
+    in the directory and its subdirectories using 'du'.
 
-    :param dirPath: Path to a directory.
-    :return: cumulative size in bytes of all files in the directory.
-    :rtype: int
+    This method will return a 'subprocess.CalledProcessError' if it is unable to
+    access a folder or file because of permissions.  Therefore this method should
```
hannes-ucsc: insufficient permissions

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
+    only be called on the jobStore, and will alert the user if some portion is
+    inaccessible.  Everything in the jobStore should have appropriate permissions
+    as there is no way to read the filesize without permissions.
+
+    du is often faster than using os.lstat(), sometimes significantly so.
+
+    The call: 'du -s /some/path' should give the number of 512-byte blocks
```
hannes-ucsc: Don't explain implementation details in the docstring. Make this a comment, too.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
 
 def getDirSizeRecursively(dirPath):
     """
-    This method will walk through a directory and return the cumulative filesize in bytes of all
-    the files in the directory and its subdirectories.
+    This method will return the cumulative filesize in bytes of all of the files
```
hannes-ucsc: This is ambiguous: "file size" usually means the number of bytes in each file. Be sure to explain that this is the number of bytes occupied by the files on disk.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
+    access a folder or file because of permissions.  Therefore this method should
+    only be called on the jobStore, and will alert the user if some portion is
+    inaccessible.  Everything in the jobStore should have appropriate permissions
+    as there is no way to read the filesize without permissions.
+
+    du is often faster than using os.lstat(), sometimes significantly so.
+
+    The call: 'du -s /some/path' should give the number of 512-byte blocks
+    allocated with the environment variable: BLOCKSIZE='512' set, and we
+    multiply this by 512 to return the filesize in bytes.
```
hannes-ucsc: Mention that the obvious BLOCKSIZE=1 does not work on OS X and why.

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
+
+    :param str dirPath: A valid path to a directory or file.
+    :return: Size, in bytes, of dirPath, and recursively everything within it.
```
hannes-ucsc: … of the file or directory at dirPath …

src/toil/common.py
```diff
@@ -1257,33 +1257,26 @@ def cacheDirName(workflowID):
...
-                    folderSize += fileStats.st_blocks * unixBlockSize
-                    seenInodes.add(fileStats.st_ino)
-                else:
-                    continue
-            else:
-                folderSize += fileStats.st_blocks * unixBlockSize
-        totalSize += folderSize
-    return totalSize
+    return int(subprocess.check_output(['du', '-s', dirPath],
+               env=dict(os.environ, BLOCKSIZE='512')).split()[0].decode('utf-8')) * 512
```
hannes-ucsc: s/utf-8/ascii

src/toil/test/src/miscTests.py
```diff
@@ -36,25 +36,15 @@ def setUp(self):
 
     @slow
     def testGetSizeOfDirectoryWorks(self):
+        '''A test to make sure toil.common.getDirSizeRecursively does not
+        underestimate the amount of disk space needed.
+
+        Disk space allocation varies from system to system.  The computed value
+        should always be equal to or slightly greater than the creation value.
+        This test generates a number of random directories and randomly sized
+        files to test this using getDirSizeRecursively, which calls 'du'.
```
hannes-ucsc: Great description, except for the mention of the implementation detail.

hannes-ucsc: https://en.wikipedia.org/wiki/Information_hiding

DailyDreaming:
@hannes-ucsc Should all be fixed!  Thanks!

DailyDreaming:
Thanks for testing and the review Joel!

Issue #2039, opened 2018-02-02
Toil uses subprocess in inconsistent, risky, and sometimes buggy ways


diekhans:
After Joel discovered that the gridengine does not check subprocess status, leaving zombies around and possibly incorrectly interpreting results.  This is also seen in lsf. Several places use shell=True, which is full of landmines and is explicitly warned against in the python documentation. 

A review of the use of subprocess is warranted.  Creating common, higher level functions to interface to process would help ensure consistent and robust process creation, consistent logging, and remove redundant code.

Process creation is much trickier than it appears.

joelarmstrong:
We should also consider using [subprocess32](https://github.com/google/python-subprocess32), which is a backport of the Py3 subprocess module. It fixes a lot of race conditions that are (to my knowledge) still present in the 2.7 branch.

DailyDreaming:
I'll make a PR for subprocess32 because that sounds like a really good idea and a couple of `shell=True` statements that I refactored in `template_version.py`.  Overall, after talking with Mark, tearing out the shell=True stuff seems like a lower priority and something we should handle as we go, unless any in particular seem to stand out as risky (many don't interpret user input and so don't seem at risk from that particular vulnerability).

ejacox:
Fixed in #2072.

PR #2041, opened 2018-02-03
Amend subprocess imports.  (#2039)

DailyDreaming:
Switch from subprocess to subprocess32 as per Joel's suggestion and partially addressing #2039.

From their site ( https://github.com/google/python-subprocess32 ):

"This is a backport of the Python 3 subprocess module for use on Python 2. This code has not been tested on Windows or other non-POSIX platforms.

subprocess32 includes many important reliability bug fixes relevant on POSIX platforms. The most important of which is a C extension module used internally to handle the code path between fork() and exec(). This module is reliable when an application is using threads."

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -2,8 +2,14 @@ import glob
 import itertools
 import logging
 import os
-import subprocess
 import sys
+
+# subprocess32 is a backport of python3's subprocess module for use on Python2,
+# and includes many reliability bug fixes relevant on POSIX platforms.
+if os.name == 'posix' and sys.version_info[0] < 3:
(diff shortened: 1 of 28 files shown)


joelarmstrong:
Yeah, definitely a +1 from me as well, though you may need to include it as a dependency in setup.py to get the tests to pass.

DailyDreaming:
@mr-c @joelarmstrong I just needed to make some changes.  I didn't want jenkins running tests on the new commits as I got it to run.  It's back up!  #2045

diekhans:
this should have removed the need for a lock in one of the modules we were looking at (don't remember which one).

PR #2045, opened 2018-02-04
Import updates.  #2039

DailyDreaming:
Originally switched from subprocess to subprocess32 as per Joel's suggestion and partially addressing #2039.  Amended this to import from `__init__.py` to save lines of code.  Added subprocess32 as a dependency.

Did the same for our pickle/cPickle import.

Fixed two subprocess calls in template_version.py (which use the original subprocess module since they are called prior to `make develop`) so that they do not use `shell=True`.  Also partially addressing #2039.

Added some copyright material that was missing (and some that was cut off mid-sentence for some reason).

Now in the process of making these separate PRs.  subprocess32 was the only critical one, so I've made it its own PR ( #2072 ) and the other changes will likely be on the backburner as they are pretty minor.


src/toil/common.py
```diff
@@ -946,20 +943,20 @@ def _setBatchSystemEnvVars(self):
...
+        Puts the environment in a globally accessible json file.
         """
-        # Dump out the environment of this process in the environment pickle file.
-        with self._jobStore.writeSharedFileStream("environment.pickle") as fileHandle:
-            pickle.dump(os.environ, fileHandle, pickle.HIGHEST_PROTOCOL)
-        logger.info("Written the environment for the jobs to the environment file")
+        # Dump out the environment of this process in the environment json file.
+        with self._jobStore.writeSharedFileStream("environment.json") as fileHandle:
+            json.dump(dict(os.environ), fileHandle)
+        logger.debug("Written the environment for the jobs to the environment file")
```
ejacox: @DailyDreaming, This seems like a major enough change to have its own issue.

DailyDreaming: Alright, I'll revert this segment.

version_template.py
```diff
@@ -107,14 +102,29 @@ def buildNumber():
...
+    import os
+
+    with open(os.devnull, 'w')  as FNULL:
+        # return code is 1 if there are still changes relative to the index
+        # otherwise return code is 0
+        retcode1 = call(['git', 'diff', '--exit-code'], stdout=FNULL)
+    # return code is 1 if there are still changes staged for the next commit relative to HEAD
+    # otherwise return code is 0
+    retcode2 = call(['git', 'diff', '--cached', '--exit-code'])
+    return 0 != (retcode1 + retcode2)
```
hannes-ucsc: retcode can be negative.

version_template.py
```diff
@@ -107,14 +102,29 @@ def buildNumber():
...
-                     '&& git diff --cached --exit-code) > /dev/null', shell=True)
+    import os
+
+    with open(os.devnull, 'w')  as FNULL:
+        # return code is 1 if there are still changes relative to the index
+        # otherwise return code is 0
+        retcode1 = call(['git', 'diff', '--exit-code'], stdout=FNULL)
+    # return code is 1 if there are still changes staged for the next commit relative to HEAD
+    # otherwise return code is 0
+    retcode2 = call(['git', 'diff', '--cached', '--exit-code'])
```
hannes-ucsc: This is not equivalent to the shell's `&&`, which does boolean short-circuit.

version_template.py
```diff
@@ -107,14 +102,29 @@ def buildNumber():
...
+    Returns True if the code has uncommitted diffs.
+
+    :return: bool
+    """
     from subprocess import call
-    return 0 != call('(git diff --exit-code '
-                     '&& git diff --cached --exit-code) > /dev/null', shell=True)
+    import os
+
+    with open(os.devnull, 'w')  as FNULL:
```
hannes-ucsc: Did you see https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL

version_template.py
diekhans: I always find these 'backwards' expressions harder to read

src/toil/common.py
DailyDreaming: Should be reverted now.

version_template.py
DailyDreaming: I did, but I didn't see it on the python 2 page.  I agree it's much more elegant.

version_template.py
DailyDreaming: Whoa, thanks for catching this.

version_template.py
DailyDreaming: I was thinking it would produce the same effect (excepting that I missed that it could negative values), but I've nested the two now so that it should be equivalent.

src/toil/worker.py
```diff
@@ -130,8 +133,8 @@ def badWorker():
     ##########################################
     
     #First load the environment for the jobGraph.
-    with jobStore.readSharedFileStream("environment.pickle") as fileHandle:
-        environment = pickle.load(fileHandle)
+    with jobStore.readSharedFileStream("environment.json") as fileHandle:
+        environment = json.loads(fileHandle.read().decode('utf8'))
```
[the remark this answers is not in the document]
DailyDreaming: That's a good point.  I will use it in the future.  Because of Edwin's comment though, I'm reverting back to using pickles.

version_template.py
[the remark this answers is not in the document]
DailyDreaming: It was, but only because I noticed the second command did not produce stdout on my terminal.  I've redirected it now though in the event that the command changes or produces stdout on Apple unbeknownst to me.

version_template.py
hannes-ucsc: You're not using the Python 2 sub process. You're using the backport from Python 3 which has DEVNULL. So why not use it?

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
     from subprocess import call
-    return 0 != call('(git diff --exit-code '
-                     '&& git diff --cached --exit-code) > /dev/null', shell=True)
+    import os
+
+    # pipe stdout to devnull since git diff can generate unnecessary output, and we only want return codes
+    with open(os.devnull, 'w')  as FNULL:
+        # return code is non-zero if there are still changes relative to the index
+        # otherwise return code is 0
+        retcode1 = call(['git', 'diff', '--exit-code'], stdout=FNULL)
```
hannes-ucsc: You should be able to use `subprocess.DEVNULL` and drop the `with` statement.

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
+        if (0 == retcode1):
+            # return code is non-zero if there are still changes staged for the next commit relative to HEAD
+            # otherwise return code is 0
+            retcode2 = call(['git', 'diff', '--cached', '--exit-code'], stdout=FNULL)
```
hannes-ucsc: Ditto.

version_template.py
hannes-ucsc: Consider [inlining](https://refactoring.com/catalog/inlineTemp.html) retcode1 and recode2. It will reduce the # of lines *and* the line length in this case.

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
-    return 0 != call('(git diff --exit-code '
-                     '&& git diff --cached --exit-code) > /dev/null', shell=True)
+    import os
+
+    # pipe stdout to devnull since git diff can generate unnecessary output, and we only want return codes
+    with open(os.devnull, 'w')  as FNULL:
+        # return code is non-zero if there are still changes relative to the index
+        # otherwise return code is 0
+        retcode1 = call(['git', 'diff', '--exit-code'], stdout=FNULL)
+        if (0 == retcode1):
```
hannes-ucsc: This isn't C. No parentheses.

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
+            # return code is non-zero if there are still changes staged for the next commit relative to HEAD
+            # otherwise return code is 0
+            retcode2 = call(['git', 'diff', '--cached', '--exit-code'], stdout=FNULL)
+            if (0 == retcode2):
```
hannes-ucsc: Same here.

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
+    import os
     from subprocess import check_output
-    return check_output('git log --pretty=oneline -n 1 -- $(pwd)', shell=True).split()[0]
+
+    return check_output(['git', 'log', '--pretty=oneline', '-n', '1', os.getcwd()]).split()[0]
 
 
 def dirty():
+    """
+    Returns True if the code has uncommitted diffs.
```
hannes-ucsc: Consider replacing `code` with `project source`

version_template.py
```diff
@@ -107,14 +102,34 @@ def buildNumber():
...
 
 def currentCommit():
+    import os
     from subprocess import check_output
-    return check_output('git log --pretty=oneline -n 1 -- $(pwd)', shell=True).split()[0]
+
+    return check_output(['git', 'log', '--pretty=oneline', '-n', '1', os.getcwd()]).split()[0]
 
 
 def dirty():
```
hannes-ucsc: A purist would commit the change to this method in a separate PR.

src/toil/batchSystems/gridengine.py
```diff
@@ -21,9 +21,16 @@
 import logging
 import os
 from pipes import quote
-import subprocess
 import time
 import math
+import sys
+
+# subprocess32 is a backport of python3's subprocess module for use on Python2,
```
hannes-ucsc: This is rather verbose, especially considering that this import occurs all over the place.

You can make this a one-liner by using a technique called re-exporting:

In a nutshell, this means placing this code in `src/toil/__init__.py` and then using `from toil import subprocess` wherever `import subprocess` occurs.

src/toil/batchSystems/gridengine.py
```diff
@@ -21,9 +21,16 @@
 import logging
 import os
 from pipes import quote
-import subprocess
```
hannes-ucsc: Make sure you catch imports like `from subprocess import Popen`.

version_template.py
DailyDreaming: Done!

version_template.py
DailyDreaming: I will do this in the future then.

version_template.py
DailyDreaming: Done!  I agree, it looks much better.

DailyDreaming:
Added the cPickle incantation to `__init__.py` too, as it seemed like a similar use case.  Also added absolute import to some functions that needed it, along with our copyright header.  Reworked `dirty()` to within the recommendations that Hannes gave (thanks Hannes!), though `subprocess.DEVNULL` does not seem to be available in python 2.7.

DailyDreaming:
Putting a hold on this until after the next release.  Will break this up into separate PRs.

jsmedmar:
Does subprocess32 work on python3 as well?

joelarmstrong:
subprocess32 just acts as a harmless pointer back to subprocess in Python 3; subprocess32 is basically just a straight backport of Python 3's more modern subprocess module to the 2.7 series.

Issue #1530, opened 2017-02-27
fileJobStore chokes on large files


natanlao:
Trying to import a local file to an Azure job store raises a MemoryError in fileJobStore when the size of the file is larger than the amount of available memory. This can be solved by modifying `_readFromUrl()` to parse the file in chunks.

Issue #1983, opened 2017-12-07
exporting a file to local disk (from another jobStore) reads the entire file into RAM


joelarmstrong:
https://github.com/BD2KGenomics/toil/blob/890480506f977ba5faadab4b1cdb90696a7dd2da/src/toil/jobStores/fileJobStore.py#L223

This means that you run out of memory (raising MemoryError) when exporting large files.

DailyDreaming:
I think it might actually be two methods.  I made a PR: #2047.

PR #2047, opened 2018-02-05
Add buffering for large files in fileJobStore.  Addresses #1983.

DailyDreaming:
I set a rather large buffer size (1Gb) after doing some speed tests.  I also made it so that it uses the original method (loads the entire file into memory) if the file is smaller than 1Gb.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -58,10 +58,8 @@ class FileJobStore(AbstractJobStore):
     # Depth of temporary subdirectories created
     levels = 2
 
-    # 1Gb RAM chunks when reading/writing files to avoid overloading RAM capacity
-    BUFFER_SIZE = 1073741824 # 1Gb
-    # Files under SMALL_FILE_SIZE_THRESHOLD will be fully read into memory, else use a buffer
-    SMALL_FILE_SIZE_THRESHOLD = 1073741824 # 1Gb
+    # 10Mb RAM chunks when reading/writing files
+    BUFFER_SIZE = 10485760 # 10Mb
 
     def __init__(self, path):
         """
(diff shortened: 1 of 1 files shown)


src/toil/jobStores/fileJobStore.py
```diff
@@ -215,13 +215,41 @@ def getSize(cls, url):
...
+        """Writes the contents of a file to a source (writes url to writable).
+
+        Uses a buffer of 1Gb if file is larger than 1Gb, else
+        read the entire file into memory for writing.
+        """
+        file_path = cls._extractPathFromUrl(url)
+        with open(file_path, 'r') as f:
+            if os.path.getsize(file_path) > 1073741824:
+                while True:
+                    buffer_size = 1073741824 # 1Gb RAM buffer
```
diekhans: use a single constant here, rather than two copies.  Assigning in the loop doesn't accomplish anything.

DailyDreaming:
@diekhans That looks great!  I'll play with it tomorrow and fix the constant declaration you suggested.

src/toil/jobStores/fileJobStore.py
ejacox: Generally avoid constants in the code. This should be a configuration variable (e.g. environment variable). At a minimum, the constant should be in a variable that is used in the code.

diekhans: This is the kind of configuration variable that is very hard for a user to set right or know when to set.  Unless there is a really good reason to expose it,  it would be better not too.

I think it would be better to have a single largish buffer size that is just used for all files and not try to tune it on a per-file basis.   Maybe 256M?

but yes, one constant in the code.

natanlao:
This will also close #1530, which is a duplicate of #1983.

src/toil/jobStores/fileJobStore.py
```diff
@@ -215,13 +221,55 @@ def getSize(cls, url):
...
+
+        Uses a buffer of 1Gb if file is larger than 1Gb, else
+        read the entire file into memory for writing.
+
+        :param str url: A path as a string of the file to be written to.
+        :param object readable: An open file object to read from.
+        """
+        file_path = cls._extractPathFromUrl(url)
+        with open(file_path, 'wb') as f:
+            if os.fstat(readable.fileno()).st_size > cls.SMALL_FILE_SIZE_THRESHOLD:
```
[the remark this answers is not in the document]
DailyDreaming: Thanks for catching this!  Alright, I've implemented only using the buffer code.  Hannes also pointed out that shutil has a copyfileobj() function that uses buffering.  I did some speed tests that ran overnight and decided to set their buffer a bit higher to 10Mb, rather than their default of 16Kb.  Their source code is exactly the same as the buffering portion of the function I originally had, so calling it should be equivalent.  Thanks again Joel!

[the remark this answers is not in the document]
joelarmstrong: Nice! I forgot all about `shutil.copyfileobj`. That's perfect for our use case.

DailyDreaming:
Two flaky tests?

    

- src.toil.test.provisioners.aws.awsProvisionerTest.AWSStaticAutoscaleTest.testSpotAutoScale

DailyDreaming:
Jenkins, test this please.

PR #2056, opened 2018-02-09
Doc changes.

DailyDreaming:
We sat down with Brian to "kick the tires" on WDL and he had some suggestions for the docs.  This PR addresses some of those issues:

- Add wdl to extras with comments (no dependencies as of now).
- Add a wdl helloworld example to the quickstart.
- Warn the user to use destroy-cluster at the start of the autoscaling scripts in the quickstart rather than relying on them to see it in the final step.
- Change the CWL AWS example to use a t2.medium instead of a t2.micro.
- Change the AWS sort.py example to use a t2.medium instead of a t2.micro.

### DIFF
--- a/docs/gettingStarted/quickStart.rst
+++ b/docs/gettingStarted/quickStart.rst
@@ -30,7 +30,7 @@ A Toil workflow can be run with just three steps:
           print output
 
 
-3. Specify the name of the :ref:`job store <>` and run the workflow::
+3. Specify the name of the job store and run the workflow::
 
        (venv) $ python helloWorld.py file:my-job-store
 
@@ -69,8 +69,7 @@ Running CWL workflows using Toil is easy.
 
        (venv) $ pip install 'toil[cwl]'
 
-   This installs the ``toil-cwl-runner`` and ``cwl-runner`` executables. These are identical -
-   ``cwl-runner`` is the portable name for the default system CWL runner.
+   This installs the ``toil-cwl-runner`` and ``cwltoil`` executables.
 
 #. Copy and paste the following code block into ``example.cwl``:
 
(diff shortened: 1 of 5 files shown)


docs/gettingStarted/quickStart.rst
```diff
@@ -381,12 +428,14 @@ on running Toil workflows on a cluster, see :ref:`runningAWS`.
 Running a CWL Workflow on AWS
 -----------------------------
 After having installed the ``aws`` and ``cwl`` extras for Toil during the :ref:`installation-ref` and set up AWS
-(see :ref:`prepare_aws-ref`),
-the user can run a CWL workflow with Toil on AWS.
+(see :ref:`prepare_aws-ref`), the user can run a CWL workflow with Toil on AWS.
+
+Also!  Remember to use the :ref:`destroyCluster` command when finished to destroy the cluster!  Otherwise things may not be cleaned up properly.
```
bencvdb: Id change this to "Also, remember to... to destroy the cluster!"  I would do the same for the changes on line 468.

docs/gettingStarted/quickStart.rst
```diff
@@ -113,6 +112,52 @@ For information on using CWL with Toil see the section :ref:`cwl`
...
+        workflow write_simple_file {
+          call write_file
+        }
+        task write_file {
+          String message
+          command { echo ${message} > wdl-helloworld-output.txt }
+          output { File test = "wdl-helloworld-output.txt" }
+        }
+
+   and this code into ``wdl-helloworld.json``::
```
bencvdb: Take this out of the code block.

docs/gettingStarted/quickStart.rst
```diff
@@ -113,6 +112,52 @@ For information on using CWL with Toil see the section :ref:`cwl`
...
+
+        {
+          "write_simple_file.write_file.message": "Hello world!"
+        }
+
+#. To run the workflow simply enter ::
```
bencvdb: This number appears as a 1 when it should be a 3

docs/gettingStarted/quickStart.rst
```diff
@@ -113,6 +112,52 @@ For information on using CWL with Toil see the section :ref:`cwl`
...
+
+        (venv) $ toil-wdl-runner wdl-helloworld.wdl wdl-helloworld.json
+
+   Your output will be in ``wdl-helloworld-output.txt`` ::
+
+        (venv) $ cat output.txt
```
bencvdb: the output file name here should be wdl-helloworld-output.txt as specified above.

docs/gettingStarted/quickStart.rst
DailyDreaming: The original has the link to the command so that the user specifically has a reference to what we're referring to.  There are many ways to destroy a cluster and we're trying to avoid having them just delete through the browser.

PR #2059, opened 2018-02-09
Support NVMe ephemeral drives (e.g. the i3 series) (resolves #2057)

### LINKED ISSUE #2057: Toil not putting enough jobs on i3.8xlarge nodes
I'm seeing fewer jobs than I expect on each of my i3.8xlarge nodes.  From the log, my nodes look like 

`Running bin packing for node shapes [_Shape(wallTime=3600, memory=261885630873.6, cores=32, disk=8160437862400, preemptable=True)] and 357 job(s).`

And my jobs:

`Issued job 'run_construct_region_graph' ee16af3b85fa-41b3-a876-8146e640f3d9 with job batch system ID: 96 and cores: 1, disk: 32.0 G, and memory: 32.0 G`

From @joelarmstrong 
>yeah, it looks like the i3 series drives are at a different location than usual, because they're NVMe and not SSD they are at /dev/nvme0n* rather than /dev/sd* so toil doesn't mount them properly right now

So it seems there's an issue mounting these nodes, though the log above does show Toil reading 8TB for it.

joelarmstrong:
NB: this requires a Container Linux version upgrade as well. The (very) old version currently used by toil has a kernel that fails to recognize more than one NVMe drive.

I haven't tested this Container Linux version extensively with this version of toil. On my fork (which may use a different docker client version), the stable branch works great, though, so this shouldn't introduce major problems.

### DIFF
--- a/src/toil/provisioners/aws/__init__.py
+++ b/src/toil/provisioners/aws/__init__.py
@@ -213,7 +213,7 @@ write_files:
         ephemeral_count=0
         drives=""
         directories="toil mesos docker"
-        for drive in /dev/xvd{{b..z}}; do
+        for drive in /dev/xvd{{b..z}} /dev/nvme*n*; do
             echo checking for $drive
             if [ -b $drive ]; then
                 echo found it
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -383,7 +383,7 @@ class AWSProvisioner(AbstractProvisioner):
     @memoize
     def _discoverAMI(cls, ctx):
         def descriptionMatches(ami):
-            return ami.description is not None and 'stable 1235.4.0' in ami.description
+            return ami.description is not None and 'stable 1632.2.1' in ami.description
         coreOSAMI = os.environ.get('TOIL_AWS_AMI')
         if coreOSAMI is not None:
             return coreOSAMI


DailyDreaming:
Possible flaky tests @ http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1961 :
- src.toil.test.provisioners.aws.awsProvisionerTest.AWSStaticAutoscaleTest.testAutoScale

DailyDreaming:
Jenkins, test this please.

joelarmstrong:
The first flaky test I can deal with in an upcoming PR.

The second is because the t2.small is too small to comfortably fit those test jobs. The instance supposedly has 2 GiB of memory, but it only offers slightly less than 1 GiB. There is literally *1 megabyte* of headroom on the t2.small offer [even when that test succeeds](http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/lastSuccessfulBuild/testReport/src.toil.test.provisioners.aws.awsProvisionerTest/AWSAutoscaleTestMultipleNodeTypes/testAutoScale/):

```
2018-02-09 20:22:16,399 - toil.batchSystems.mesos.batchSystem - DEBUG - Got offer 5e4ef370-b8b9-422c-8814-37c94ac81918-O0 for a non-preemptable slave with 1001.00 MiB memory, 1.00 core(s) and 42121.00 MiB of disk.
```
, so I'm not surprised this randomly pushed it over the edge. 

Does anyone know wtf is going on with that? Why are we taking up a full gigabyte of overhead? On my instances I only see around 200 MiB of overhead tops. Still something to account for, probably, but I don't understand why these Mesos offers are so low.

ejacox:
@joelarmstrong, is this ready to merge?

joelarmstrong:
Yep, this should be good to go. It still seems to work well with the toil container's docker client version.

PR #2061, opened 2018-02-11
Fix a few flaky aws tests

joelarmstrong:
This (hopefully) handles the EC2 400 errors on instance creation, and another rare error when creating a leader. Both are related to consistency issues with EC2: the results of instance creation aren't immediately propagated to all the servers that handle our requests.

There are still a few flaky AWS tests remaining, because the t2.small instances used in some of the tests have only a few MB of headroom to be able to run the jobs we issue. We are somehow using nearly a GB (and sometimes more) of RAM as overhead on worker nodes.

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -439,9 +439,18 @@ class AWSProvisioner(AbstractProvisioner):
 
     @classmethod
     def _addTags(cls, instances, tags):
+        def tagThrottlePredicate(e):
+            """Check for common retriable issues when creating tags."""
+            if isinstance(e, BotoServerError) and e.status == 400 \
+               and 'NotFound' in e.body:
+                # EC2 can take a while to propagate instance IDs to all servers.
+                return True
+            else:
+                return cls._throttlePredicate
+
         for instance in instances:
             for key, value in iteritems(tags):
-                for attempt in retry(predicate=AWSProvisioner._throttlePredicate):
+                for attempt in retry(predicate=tagThrottlePredicate):
                     with attempt:
                         instance.add_tag(key, value)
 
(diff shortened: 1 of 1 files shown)


src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -136,18 +136,19 @@ def launchCluster(self, leaderNodeType, leaderSpotBid, nodeTypes, preemptableNod
...
         else:
             logger.info('Launching preemptable leader')
             # force generator to evaluate
-            list(create_spot_instances(ec2=ctx.ec2,
-                                       price=leaderSpotBid,
-                                       image_id=self._discoverAMI(ctx),
-                                       tags={'clusterName': clusterName},
-                                       spec=kwargs,
-                                       num_instances=1))
-        leader = self._getLeader(clusterName=clusterName, wait=True, zone=zone)
```
ejacox: This waits for the leader. The tests are failing because the container is not yet ready when the test code tries to ssh to the leader.

joelarmstrong: d'oh! thank you! I should be able to fix that while still avoiding another DescribeInstances API call.

joelarmstrong:
@jessebrennan I think that looks super cool! I like how clean the decorator approach is.

One thing I just thought of though, is that we might have to be careful that all functions wrapped in a decorator are idempotent. The _addTags method, for example, may not be. If we are adding 3 tags and hit a retriable exception on the 3rd, then we would try to add all 3 tags over again--which might raise some other error from AWS complaining that we already added it. (Or maybe they don't care about "creating" the same tag twice--in which case it doesn't really matter.)

I'd be cool with going with your retry changes in a new PR and closing this one, as long as the changes from `1aa662f` are still included somehow.

jessebrennan:
@joelarmstrong I rebased off of your branch. Feel free to close this, but I'll ping you for a final review.

Good point about idempotentcy. I wonder if it would be worth breaking some of those larger functions into smaller, idempotent ones and then wrapping those.

joelarmstrong:
Yeah I think that's the safest possible approach: methods that do only 1 request which are wrapped in your decorator, then composed into larger, more useful functions. Thanks!

PR #2062, opened 2018-02-12
Set log level to error when jobs fail

jsmedmar:
I'd like to add this small change to this PR. I think that if a LSF job fails, whe should log `error` not `debug`. What do you think?

### DIFF
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -106,8 +106,8 @@ class LSFBatchSystem(AbstractGridEngineBatchSystem):
                                  "{}".format(job))
                     return None
                 elif "Exited with exit code" in line:
-                    exit = int(line[line.find("Exited with exit code "):]
-                               .split()[0])
+                    exit = int(line[line.find("Exited with exit code ")+22:]
+                               .split('.')[0])
                     logger.debug("bjobs detected job exit code: "
                                  "{}".format(exit))
                     return exit
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -25,273 +25,158 @@ from past.utils import old_div
 import logging
 import subprocess
 import time
-from threading import Thread
 from datetime import date
 import os
 
-# Python 3 compatibility imports
-from six.moves.queue import Empty, Queue
-
 from toil.batchSystems import MemoryString
-from toil.batchSystems.abstractBatchSystem import BatchSystemLocalSupport
-from toil.batchSystems.lsfHelper import parse_memory_resource, parse_memory_limit, per_core_reservation
+from toil.batchSystems.abstractGridEngineBatchSystem import \
+        AbstractGridEngineBatchSystem
+from toil.batchSystems.lsfHelper import (parse_memory_resource,
+                                         parse_memory_limit,
+                                         per_core_reservation)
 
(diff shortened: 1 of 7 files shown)


ejacox:
Jenkins, test this please

src/toil/batchSystems/lsf.py
```diff
@@ -124,7 +124,7 @@ def getJobExitCode(self, lsfJobID):
                 return None
 
             # if not found in bjobs, then try bacct (slower than bjobs)
-            logger.debug("bjobs failed to detect job - trying bacct: "
+            logger.error("bjobs failed to detect job - trying bacct: "
```
mr-c: This should remain a `debug` level as we can still try `bacct`

src/toil/batchSystems/lsf.py
```diff
@@ -136,7 +136,7 @@ def getJobExitCode(self, lsfJobID):
                                  "{}".format(job))
                     return 0
                 elif line.find("Completed <exit>") > -1:
-                    logger.debug("Detected job failed for job: "
+                    logger.error("Detected job failed for job: "
```
mr-c: Since this backed extends `AbstractGridEngineBatchSystem` then all implementers of `getJobExitCode` should throw an exception instead of using `1` as the return in case the job itself failed to differentiate between job failure and the underlying tool having an exit code of `1`. I'm the most recent person to work on this code, so I can fix this tomorrow if you don't want to do it today.

Once the exception is thrown then we can have a unified `logger.error` in `AbstractGridEngineBatchSystem`.

jsmedmar: As long as we expose the `LSF` job id to the user through logging we are good. It doesn't really matter *when* it happens as long as you are reported with the failed LSF job id even if your logging level is not `debug` (so you can explore with `bsub -L` for instance).

src/toil/batchSystems/lsf.py
jsmedmar: Yes sorry, this was a mistake :/

src/toil/batchSystems/lsf.py
jsmedmar: But logging to `error` the failed LSF job ID shouldn't affect the return value of the function... I'm not following here 

>Since this backed extends AbstractGridEngineBatchSystem then all implementers of getJobExitCode should throw an exception instead of using 1 as the return in case the job itself failed to differentiate between job failure and the underlying tool having an exit code of 1.

mr-c: Since my proposed change will touch all of the `AbstractGridEngineBatchSystem`s I'm going to do it in a separate PR. In the meantime I've accepted your suggestion to increase the logging level.

jsmedmar:
hey @mr-c I think we never did this. The following lines should log to `error` instead of `debug`:

https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/lsf.py#L113
https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/lsf.py#L117
https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/lsf.py#L141

This [line](https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/lsf.py#L141) is actually the only one logging to `error`. What do you think? these changes are very useful and quite inoffensive, should I submit a PR?

PR #2063, opened 2018-02-13
Fix some flaky AWS tests (resolves #1840)

### LINKED ISSUE #1840: Fix flakey AWS tests
Boto calls fail sometimes when Amazon get's finicky and throttles us. We aren't prepared to handle these exceptions.

Also sometimes with newly created instances, a 400 error is returned and we don't catch it. [example](http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1950/testReport/junit/src.toil.test.provisioners.aws.awsProvisionerTest/AWSAutoscaleTestMultipleNodeTypes/testAutoScale/)

One solution is to make a wrapper for all boto calls. 

Another would be to take the mechanism used in #1837, stick it inside a context manager or decorator or something and then apply the decorator to functions that make these boto calls.

jessebrennan:
This is a continuation of #2061. 

It should fix a few different kinds of AWS errors.

- Adding tags would fail with `EC2ResponseError: EC2ResponseError: 400 Bad Request` [example](http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1950/testReport/junit/src.toil.test.provisioners.aws.awsProvisionerTest/AWSAutoscaleTestMultipleNodeTypes/testAutoScale/)
This is fixed in 6682bac then refactored in 7659c0b

- Counting workers would fail with assertion error. [example](http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1955/testReport/junit/src.toil.test.provisioners.aws.awsProvisionerTest/AWSStaticAutoscaleTest/testSpotAutoScale/)
This is fixed in e01298a

This PR also contains a refactor of the AWS retry mechanism 7659c0b. Now there is a decorator `@awsRetry` which retries the function if any regular AWS exceptions are thrown.

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -618,7 +618,7 @@ class AWSProvisioner(AbstractProvisioner):
                     raise
 
     @awsRetry
-    def _propagateToNode(self, node):
+    def _propagateKeyToNode(self, node):
         # since we're going to be rsyncing into the appliance we need the appliance to be running first
         ipAddress = self._waitForNode(node, 'toil_worker')
         self._coreRsync(ipAddress, [self.config.sseKey, ':' + self.config.sseKey], applianceName='toil_worker')
@@ -627,7 +627,7 @@ class AWSProvisioner(AbstractProvisioner):
         if not self.config or not self.config.sseKey:
             return
         for node in instances:
-            self._propagateToNode(node)
+            self._propagateKeyToNode(node)
 
     @classmethod
     def _getBlockDeviceMapping(cls, instanceType, rootVolSize=50):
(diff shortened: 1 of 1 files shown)


jessebrennan:
Ironically a flaky aws test is failing here that I didn't fix, but I think this is ready for review @ejacox @joelarmstrong.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -589,15 +617,17 @@ def _deleteIAMProfiles(cls, instances, ctx):
                 else:
                     raise
 
+    @awsRetry
+    def _propagateToNode(self, node):
+        # since we're going to be rsyncing into the appliance we need the appliance to be running first
+        ipAddress = self._waitForNode(node, 'toil_worker')
+        self._coreRsync(ipAddress, [self.config.sseKey, ':' + self.config.sseKey], applianceName='toil_worker')
+
     def _propagateKey(self, instances):
```
ejacox: Would a better name be _propagateKeyToNode and _propagateKey for the above function?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -618,24 +648,28 @@ def _getBlockDeviceMapping(cls, instanceType, rootVolSize=50):
 
     @classmethod
     def _getNodesInCluster(cls, ctx, clusterName, nodeType=None, preemptable=False, both=False):
-        for attempt in retry(predicate=AWSProvisioner._throttlePredicate):
+        for attempt in retry(predicate=awsRetryPredicate):
```
ejacox: The decorator only works for a single retry in the function? Would it be cleaner to break up these functions and use the decorator?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -618,24 +648,28 @@ def _getBlockDeviceMapping(cls, instanceType, rootVolSize=50):
...
             with attempt:
                 runningInstances = ctx.ec2.get_only_instances(filters={'instance.group-name': clusterName,
                                                                        'instance-state-name': 'running'})
         if nodeType:
             pendingInstances = [instance for instance in pendingInstances if instance.instance_type == nodeType]
             runningInstances = [instance for instance in runningInstances if instance.instance_type == nodeType]
-        instances = set(pendingInstances)
+        # combine lists of instances by checking id for uniqueness to prevent a lucky
```
ejacox: How about making a single call to get_only_instances without the instance-state-name filter, then separate the lists by state?

src/toil/provisioners/aws/awsProvisioner.py
jessebrennan: I considered it, but it seemed sort of unnecessary since this is still pretty concise as is. I'm happy to change it if you want me to.

src/toil/provisioners/aws/awsProvisioner.py
jessebrennan: That sounds much cleaner!

PR #2066, opened 2018-02-15
Add --disableChaining option (resolves #1698)

### LINKED ISSUE #1698: Job chaining breaks service termination timing
The docs say:

> The services of a job are started before any of its successors have been run and stopped after all the successors of the job have completed successfully.

But that's not the case if a chained job starts a service. The chaining works by pushing the predecessor's "stack" onto the next job's "stack", basically making the predecessor's successors into successors of the next job. So the job graph as designed by the user is no longer the same graph that's in the job store: the chained job has extra successors. This breaks the timing of service termination.

Consider this example (follow-on edges in red, service edges dotted):
toil-chaining-issue dot

`chainEnd` can be chained to `chainStart` (assuming it has the same CPU, memory, etc requirements). If it's chained, then `service` will stick around not just through `job1` and `job2`, but also `job3` and `job4`. This isn't great, because if the service is, say, a very large database, it will hang around on a non-preemptable server costing lots of money for no reason.

I fixed this on my fork by just disabling chaining entirely--I've never been too convinced of its value and I'm not sure there aren't other edge cases like this still remaining to be found. I don't think it's feasible to shoehorn a fix into the chaining code as written, because it's undecidable to figure out whether the next job will run a service or not.

joelarmstrong:
Also refactors worker logic slightly.

This *shouldn't* break the resumability of jobstores that were started before these changes, but I haven't tested that. I have tested that resuming a workflow with/without --disableChaining seems to work as expected (it flips on/off properly, unlike options like --writeLogs).

### DIFF
--- a/src/toil/worker.py
+++ b/src/toil/worker.py
@@ -18,7 +18,6 @@ standard_library.install_aliases()
 from builtins import str
 from builtins import map
 from builtins import filter
-from builtins import object
 import os
 import sys
 import copy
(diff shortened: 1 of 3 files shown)


jessebrennan:
Jenkins ran out of space and stopped building. Manually restarting: Jenkins, test this please.

jessebrennan:
From what I can tell, the unittest failure is boto timing out. The integration test failures are from known flaky tests. Let's run again to see if this clears up. Jenkins test this please

src/toil/worker.py
```diff
@@ -539,18 +490,28 @@ def badWorker():
...
     jobStoreLocator = argv[2]
     jobStoreID = argv[3]
-    
+
     ##########################################
     #Load the jobStore/config file
     ##########################################
-    
+
+    # Try to monkey-patch boto early so that credentials are cached.
```
jessebrennan: why did you move this to the end of the file?

joelarmstrong: I should have been clearer about this change. IMO it's best if this monkey-patch is run *before* any boto calls are made, so that we're using 1 single monkey-patched version of the boto code the entire time. Before this change boto was getting monkey-patched only after `Toil.resumeJobStore` was already called, which I believe contacts the jobStore. Obviously the code worked just fine before, so it doesn't actually matter in practice, but this seems a tiny bit cleaner to me.

(I could have put it at the top of the file as well, but that would monkey-patch any random process that [for some reason] imported stuff from the worker file.)

DailyDreaming:
Flaky tests?
- src.toil.test.src.fileStoreTest.CachingFileStoreTestWithGoogleJobStore.testAsyncWriteWithCaching
- http://jenkins.cgcloud.info/job/toil-pull-requests/3693/
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1989/

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
2 flaky tests:
- src.toil.test.src.fileStoreTest.CachingFileStoreTestWithGoogleJobStore.testCacheEvictionFailCase
- http://jenkins.cgcloud.info/job/toil-pull-requests/3698
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/1994

DailyDreaming:
@joelarmstrong Is this ready to merge?

DailyDreaming:
Jenkins, test this please.

Issue #2081, opened 2018-02-23
Toil Installation via pip -> toil command fails


kmpf:
I installed toil via:
`$ pip install toil[cwl,wdl]`

Running toil by:
`$ toil`
Or
`$ toil --help`
result in:

> Traceback (most recent call last):
>   File "/usr/bin/toil", line 11, in <module>
>     load_entry_point('toil==3.14.0', 'console_scripts', 'toil')()
>   File "/usr/lib/python2.7/site-packages/toil/utils/toilMain.py", line 16, in main
>     printHelp(modules)
>   File "/usr/lib/python2.7/site-packages/toil/utils/toilMain.py", line 62, in printHelp
>     descriptions='\n'.join("%s - %s" % (n, m.__doc__.strip()) for n, m in iteritems(modules))))
>   File "/usr/lib/python2.7/site-packages/toil/utils/toilMain.py", line 62, in <genexpr>
>     descriptions='\n'.join("%s - %s" % (n, m.__doc__.strip()) for n, m in iteritems(modules))))
> AttributeError: 'NoneType' object has no attribute 'strip'

Can this be fixed in the PyPI package?

DailyDreaming:
I made a fix, but it'll likely be in the next release: #2082 

Thanks for catching this!

PR #2082, opened 2018-02-23
Fixes broken help menu.  Addresses #2081.

DailyDreaming:
Addresses #2081.

Currently if one pip installs toil and runs the command `toil --help`, toil breaks.  This PR fixes this.

### DIFF
--- a/src/toil/utils/toilMain.py
+++ b/src/toil/utils/toilMain.py
@@ -59,4 +59,4 @@ def printHelp(modules):
     print(usage.format(
         name=os.path.basename(sys.argv[0]),
         commands='|'.join(iterkeys(modules)),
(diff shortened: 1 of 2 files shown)


DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Flaky test:
- http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2010

DailyDreaming:
m.__doc__ searches through all of the files in the directory for their docstrings.  It returns a string or None.  The method always uses `.strip()` so if it returns None, you get a type error.  This fix added the missing docstring, and also forced m.__doc__ as a string, even if it is None, so that the help menu will show up even if there's no docstring in place for that file (just showing the string 'None' instead on the help menu).

PR #2085, opened 2018-02-24
Main loop refactor

### LINKED ISSUE #2084: Leader.innerloop is very complex
It requires huge amounts of work to understand the monolithic and heavily nested innerloop function, with difficult to follow flow, implicit gotos (ie. continues), and cases were the rationale is not clear.

Very 1970s

diekhans:
Not ready to merge.  Just for discussion.

fixes #2084

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ requires the TOIL_DOCKER_REGISTRY variable to be set to a value other than the d
 accidentally pushing to the official Docker registry for Toil.
 
 The TOIL_DOCKER_NAME environment variable can be set to customize the appliance image name that
-is created by the 'docker' target and pushed by the 'push_docker' target. The Toil team's
+is created by the 'docker' target and pushed by the 'push_docker' target. The Toil team\'s
 continuous integration system overrides this variable to avoid conflicts between concurrently
 executing builds for the same revision, e.g. toil-pr and toil-it.
 
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -53,6 +53,23 @@ from toil.common import ToilMetrics
 
 logger = logging.getLogger( __name__ )
 
+###############################################################################
+# Implementation Notes
+#
+# Multiple predecessors:
+#   There is special-case handling for jobs with multiple predecessors as a
+#   performance optimization. This minimize number of expensive loads of
+#   jobGraphs from jobStores.  However, this special case could be unnecessary.
+#   The jobGraph is loaded to update predecessorsFinished, in
+#   _checkSuccssorReadyToRunMultiplePredecessors, however it doesn't appear to
+#   write the jobGraph back the jobStore.  Thus predecessorsFinished may really
+#   be leader state and could moved out of the jobGraph.  This would make this
+#   special-cases handling unnecessary and simplify the leader.
+#   Issue #2136
+###############################################################################
+
+
+
 ####################################################
 # Exception thrown by the Leader class when one or more jobs fails
 ####################################################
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -74,7 +74,7 @@ requires the TOIL_DOCKER_REGISTRY variable to be set to a value other than the d
 accidentally pushing to the official Docker registry for Toil.
 
 The TOIL_DOCKER_NAME environment variable can be set to customize the appliance image name that
-is created by the 'docker' target and pushed by the 'push_docker' target. The Toil team's
```
joelarmstrong: why escape this one and not the rest? to satisfy syntax highlighting?

src/toil/leader.py
```diff
@@ -136,14 +136,17 @@ def __init__(self, config, batchSystem, provisioner, jobStore, rootJob, jobCache
         self.preemptableServiceJobsIssued = 0
         self.preemptableServiceJobsToBeIssued = []
 
+        # Timing of the jobGraph rescuing method
+        timeSinceJobsLastRescued = None
```
joelarmstrong: presumably you want self. here

Makefile
diekhans: Because it is not matched, so emacs thinks the rest of the makefile is a quote.  Didn't really care about syntax highlighting in the help message.   I have never seen define used this way, so it is not surprising it doesn't understand it.

src/toil/leader.py
diekhans: Doh!  Yes

src/toil/leader.py
```diff
@@ -228,262 +231,293 @@ def run(self):
...
+            return False
+
+
+    def _checkSuccssorReadyToRunMultplePredecessors(self, jobGraph, jobNode, successorJobStoreID):
+        """Handle the special cases of checking if a successor job is
+        ready to run when there are multiple predecessors"""
+        # FIXME: it appears that the reason for special handling is to avoid
+        # recomputing the jobgraph, unclear is this is actually the case or
+        # how important it is.  If possible, it would be clearer without this
+        # special cases.
```
joelarmstrong: I'm fairly sure we don't actually need to load the jobGraph here. We write to the `predecessorsFinished` attribute, but since the jobGraph is likely never written *back* to the job store, there are two possibilities:

- the logic is wrong, or
- `predecessorsFinished` is leader state.

It looks like it's the latter, because that attribute is only ever touched by the leader. Revising predecessorsFinished to be leader state and not job state would make this a little bit simpler.

joelarmstrong: As for whether we could get rid of the special case: yeah, probably. Right now special-casing this is avoiding a ton of leader-blocking jobGraph loads, which is a good thing. But if those go away, I don't really see a reason to have the 1-predecessor case be special.

src/toil/leader.py
```diff
@@ -228,262 +231,293 @@ def run(self):
...
+        or restart it if it has retries left and is a checkpoint job"""
+
+        if jobGraph.jobStoreID in self.toilState.servicesIssued:
+            # The job has services running, signal for them to be killed
+            # once they are killed then the jobGraph will be re-added to
+            # the updatedJobs set and then scheduled to be removed
+            logger.debug("Telling job: %s to terminate its services due to successor failure",
+                         jobGraph.jobStoreID)
+            self.serviceManager.killServices(self.toilState.servicesIssued[jobGraph.jobStoreID],
+                                        error=True)
```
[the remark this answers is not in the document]
diekhans: nit: fake8 is much better at enforcing nits ;-)

joelarmstrong:
Jenkins, test this please.

diekhans:
Jenkins, test this please.

ejacox:
Jenkins, test this please.

DailyDreaming:
Looks good to me.

Issue #2089, opened 2018-02-28
Hot-deployment is completely undocumented.


diekhans:
The only documentation on what `hot-deployment' is

>  Hot-deployment does not work here because Toil cannot hot-deploy itself, the classic "Which came first, chicken or egg?" problem.

jessebrennan:
Actually the whole
http://toil.readthedocs.io/en/latest/running/deploying.html section
explains hot deployment in pretty good detail.

diekhans:
I can't find the word `hot' (agreed, it's bad terminology) or
discussion of restarting in the page.   Can you point out specifics
discussion?

joelarmstrong:
I think remote-deployment == hot-deployment here. I like the remote-deployment terminology a bit better.

I don't think there is any mention of the script being re-deployed on restart. We could add that, though if we *really* want to document everything, we should definitely also mention that this only works on Mesos (right?).

DailyDreaming:
Someone changed `hot-deployment` to `remote-deployment`, but only in one place.  The code still has functions saying stuff like `supportsHotDeployment()` and sprinkled throughout the comments.

DailyDreaming:
Also, wtf is `pip install fairydust` and why is it in the `remote-deployment` explanation?

http://toil.readthedocs.io/en/latest/running/deploying.html?highlight=fairydust

DailyDreaming:
I get that it's probably a joke PyPi repo, but I think it's just confusing.

DailyDreaming:
I think it would be a good idea to change `remote-deployment` back to `hot-deployment`.  There's a test file that needs to be renamed, functions need to be refactored, documentation that was missed, and comments that need to be edited.  It looks like `hot-deployment` renaming to `remote-deployment` only occurred in one section, so it'd be much easier and more consistent to change back to `hot-deployment`.

diekhans:
Main immediate concern now is having consistent user doc.  Long
term, a clean conceptual name is better.  Hot deploy means
changing while running; is not what this does. 

Change names in code is easy (well, easy in emacs, YMMV); easier
than documentation, since running tests will find typos.
Documentation has to be understandable to a wide range of users.
That is hard!!

DailyDreaming:
I've converted everything to `remote-deployment`.  It's pretty straight-forward in PyCharm too.  I really didn't want to leave things half-way.

hannes-ucsc:
[...]

PR #2090, opened 2018-02-28
Change "hot" to "remote"-deployment.  Addresses #2089.

DailyDreaming:
`Hot-deployment` was renamed `remote-deployment` for some reason in one part of the docs but nowhere else.  The current explanation of `hot-deployment` is documented here: http://toil.readthedocs.io/en/latest/running/deploying.html?highlight=remote-deployment

However, this was done improperly.  Some of the docs still mention `hot-deployment`, the test file is still called `hotDeploymentTest.py`, there are mentions of hot-deployment in many code comments (which in docstrings carries over to the API docs), there's an option called `--disableHotDeployment` and even functions like `supportsHotDeployment(cls)`.  This PR addresses this and reverts back this small section to `hot-deployment`.

Also took out `pip install fairydust` which is a fake PyPi and can confuse the easily confused like myself.  I used `matplotlib` as an example instead.

### DIFF
--- a/docs/running/amazon.rst
+++ b/docs/running/amazon.rst
@@ -21,6 +21,8 @@ that will dynamically scale depending on the workflow's needs.
 The :ref:`StaticProvisioning` section explains how a static cluster (one that
 won't automatically change in size) can be created and provisioned (grown, shrunk, destroyed, etc.).
 
+Currently, it's recommended that users do not run their scripts from system folders (``/var``, ``/``, or ``/tmp`` for example.).
+
 
 .. _EC2 instance type: https://aws.amazon.com/ec2/instance-types/
 
--- a/docs/running/amazon.rst
+++ b/docs/running/amazon.rst
@@ -130,7 +130,7 @@ look like::
 
 .. note::
 
-    If your toil workflow has dependencies have a look at the :ref:`remoteDeploying`
+    If your toil workflow has dependencies have a look at the :ref:`hotDeploying`
     section for a detailed explanation on how to include them.
 
 
--- a/docs/running/amazon.rst
+++ b/docs/running/amazon.rst
@@ -22,7 +22,7 @@ The :ref:`StaticProvisioning` section explains how a static cluster (one that
 won't automatically change in size) can be created and provisioned (grown, shrunk, destroyed, etc.).
 
 Currently, it's recommended that users do not run their scripts from system folders (``/var``, ``/``, or ``/tmp`` for example.).
-
+It's recommended to create a new folder such as ``/tmp/working`` as the working directory.
 
 .. _EC2 instance type: https://aws.amazon.com/ec2/instance-types/
 
@@ -130,7 +130,7 @@ look like::
 
 .. note::
 
-    If your toil workflow has dependencies have a look at the :ref:`hotDeploying`
+    If your toil workflow has dependencies have a look at the :ref:`remoteDeploying`
     section for a detailed explanation on how to include them.
 
 
@@ -139,7 +139,7 @@ look like::
 Running a Workflow with Autoscaling
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Autoscaling is a feature of running Toil in a cloud whereby additional cloud instances are launched to run the workflow.  Autoscaling leverages Mesos containers to provide an execution environment for these workflows.  
+Autoscaling is a feature of running Toil in a cloud whereby additional cloud instances are launched to run the workflow.  Autoscaling leverages Mesos containers to provide an execution environment for these workflows.
 
 
 
--- a/docs/running/deploying.rst
+++ b/docs/running/deploying.rst
@@ -84,7 +84,7 @@ From here, you can install your project and its dependencies::
        └── main.py
 
    3 directories, 5 files
-   $ pip install fairydust
+   $ pip install matplotlib
    $ cp -R workflow util venv/lib/python2.7/site-packages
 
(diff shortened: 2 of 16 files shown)


DailyDreaming:
Also, wtf is `pip install fairydust` and why is it in the `remote-deployment` explanation?

http://toil.readthedocs.io/en/latest/running/deploying.html?highlight=fairydust

DailyDreaming:
Addresses #2089 .

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Looks like jenkins is down again.

diekhans:
please leave as remote-deployment.  The code can be fixed later.  That is easier than trying to explain to the user that hot deployment is not really hot deployment

  https://www.quora.com/Difference-between-hot-deployment-vs-cold-deployment

DailyDreaming:
I reverted the changes.  Hot-deployment should be completely converted to remote-deployment now.

docs/running/amazon.rst
```diff
@@ -22,7 +22,7 @@ The :ref:`StaticProvisioning` section explains how a static cluster (one that
 won't automatically change in size) can be created and provisioned (grown, shrunk, destroyed, etc.).
 
 Currently, it's recommended that users do not run their scripts from system folders (``/var``, ``/``, or ``/tmp`` for example.).
-
+It's recommended to create a new folder such as ``/tmp/working`` as the working directory.
```
diekhans: Please explain why it is recommended.  Recommendations that don't teach people is for robots ;-)

 Not necessarily a new folder, because it might be a venv or one where your app is being developed.  Mentioning /var is almost weird, as I don't think many people would do this.   The fact that it  /tmp is shared by other processes means attempts to copy other files in /tmp is the real problem.

docs/running/deploying.rst
```diff
@@ -82,7 +84,7 @@ From here, you can install your project and its dependencies::
        └── main.py
 
    3 directories, 5 files
-   $ pip install matplotlib
+   $ pip install fairydust
```
diekhans: how about <requiredpackages>

src/toil/batchSystems/options.py
```diff
@@ -106,10 +106,10 @@ def addOptions(addOptionFn, config):
...
                 help=("The type of batch system to run the job(s) with, currently can be one "
                       "of %s'. default=%s" % (', '.join(uniqueNames()), defaultBatchSystem())))
-    addOptionFn("--disableHotDeployment", dest="disableHotDeployment", action='store_true', default=None,
-                help=("Should hot-deployment of the user script be deactivated? If True, the user "
-                      "script/package should be present at the same location on all workers. "
-                      "default=false"))
+    addOptionFn("--disableRemoteDeployment", "--disableHotDeployment", dest="disableRemoteDeployment",
+                action='store_true', default=None,
+                help=("Remote-deployment=True copies the toil virtualenv and user script to the same location "
+                "on all workers automatically.  The default is False."))
```
diekhans: we should leave in --disableHotDeployment as a deprecated option (it can have dest=disableRemoteDeployment) and comment that it is deprecated. 

For API functions, we should put in deprecated ones that call the new ones (change the comments to fooHotDeploy is deprecated, use fooRemoveDeploy

src/toil/batchSystems/parasol.py
```diff
@@ -51,7 +51,7 @@ def supportsWorkerCleanup(cls):
         return False
 
     @classmethod
-    def supportsHotDeployment(cls):
+    def supportsRemoteDeployment(cls):
```
diekhans: see comment about deprecating above.,

src/toil/common.py
```diff
@@ -875,10 +875,10 @@ def createBatchSystem(config):
 
         return batchSystemClass(**kwargs)
 
-    def _setupHotDeployment(self, userScript=None):
+    def _setupRemoteDeployment(self, userScript=None):
```
diekhans: by definition, _* names are fine to change

src/toil/batchSystems/parasol.py
diekhans: Yes, if you no it is only use by the toil code, then fine.
No, don't change the names.  Rest of code doesn't follow that and that gets complicated.  
Best to just use _ to mean non-public to class or module.

Instead, a guide would be `is it part of the API?

src/toil/batchSystems/options.py
diekhans: The old option should not be left in as-is.  Change the help to just say `Deprecated, use --disableRemoteDeployment instead' and have dest="disableRemoteDeployment"

src/toil/batchSystems/options.py
```diff
@@ -106,10 +106,14 @@ def addOptions(addOptionFn, config):
     addOptionFn("--batchSystem", dest="batchSystem", default=defaultBatchSystem(),
                 help=("The type of batch system to run the job(s) with, currently can be one "
                       "of %s'. default=%s" % (', '.join(uniqueNames()), defaultBatchSystem())))
-    addOptionFn("--disableHotDeployment", dest="disableHotDeployment", action='store_true', default=None,
-                help=("Should hot-deployment of the user script be deactivated? If True, the user "
-                      "script/package should be present at the same location on all workers. "
-                      "default=false"))
+    addOptionFn("--disableHotDeployment", dest="disableRemoteDeployment",
```
ejacox: Why is this left in? Does it fail if this option is selected?

DailyDreaming:
Just created a new branch because it seemed easier: #2137 .

PR #2112, opened 2018-03-07
torque: fix resource allocation

### LINKED ISSUE #2111: Resource allocation in torque
Hi,

I've noticed when using the Torque BatchSystem API, the resource requirements are not properly relayed to the cluster.

In the code only the `ncpus` flag is used to allocate cores, but this results in reservations of one core  on a node with ncpu number of cores available. To properly allocate the number of requested processes, the `ppn` flag should also be used.

More explanation on this [here](https://research.computing.yale.edu/hpc/faq/multiple-cpus-under-pbs).

I'll look into a quick fix for this.

Cheers
M

matthdsm:
fixes #2111


matthdsm:
Tested, can confirm this works.

Cheers
M

ejacox:
Jenkins, test this please.

PR #2114, opened 2018-03-08
Autoscaler revamp

joelarmstrong:
This is a slight refactoring and retouching of the autoscaler. This fixes several bugs, some of which are blockers for certain niche use-cases. Notably:
- Services weren't being treated specially by the autoscaler. Services are one case where the degree of parallelization really, really matters. If you have multiple services, they may all need to be running simultaneously before any real work can be done. Previously, if your services had a fairly short runtime, the autoscaler would believe that it could pack them serially, one after the other. This would lead to service deadlocks in some situations.
- New node reservations (reservations are sort of like "time-slices" of resource usage at a node, so you can find the projected resource usage at each point in time) weren't split, meaning that a 5-minute job that uses 5 cores would be projected to use 5 cores for the entire 1-hour "wallTime" of the instance. This means in certain situations, more nodes would be created than would be optimal.
- A long-running job, which "extends" a node reservation for a very long time, could end up causing other jobs to be packed into its extended reservations, even though they may be projected to start several hours in the future. This ends up underprovisioning nodes if you have a few long-running jobs and many more short jobs. (I've noticed the old pre-multiple-node-types scaler underestimating the number of needed nodes with Cactus, though I don't know whether that's caused by this particular bug or not.)
- The ignored nodes functionality didn't work at all, or when it did, only rarely, because the ignoredNodes set was based on IP addresses, and the Mesos batch system was checking *hostnames* to see if they were ignored.
- When running services or other long-running jobs, the ignored nodes functionality would rarely lead to deadlocks, where all N nodes were ignored despite wanting to scale up to N nodes. Now the scaler will "unignore" nodes if it changes its mind about scaling up or down.

For context, the original scaler worked like this: take the resource requirements of the last 10000 completed jobs, attempt to figure out how many nodes would be able to run those so that all jobs started within an hour. Then multiply that by the length of the current queue times a constant "alpha packing" factor. This is nice from a performance point of view (because it only needs to pack at most 10000 jobs), but leads to some *really* weird results, because it used the past to predict the future. It also isn't ideal when you absolutely need parallelism (services), because you *never* want to under-provision like you would with an alphaPacking < 1.0. The current scaler is a blend of using "recent" jobs vs. queued jobs, with alphaPacking determining the mix between the two.

This alters the scaler so that it works entirely on the basis of queued jobs, and the "alpha" parameter is a time. The packing attempts to spawn enough nodes that all the queued jobs will be started by that time. Intuitively this is just a parameter which controls the degree of parallelization, without (hopefully) causing deadlocks if it can help it.

Anyway, this has been mangled a bit since I last tested it thoroughly, because it's been through a rebase and now selectively cherry-picked off my fork. So it's not ready to merge just yet, but it should be relatively OK.

Some considerations:
- The scaler basically attempts to set the rate of job completion (assuming homogeneous jobs for simplicity) to n/a, n being the number of jobs and a being the alpha factor. But the scaler will run again in (by default) 60 seconds, and assuming some jobs have been completed, it will try to set the rate of job completion lower (because n is lower). The total number of jobs remaining, n, will decrease from the starting point N as roughly n = N/(t/a + 1). So the end result isn't the local result, that all jobs are started within alpha seconds--quite the contrary, they are almost guaranteed not to unless they can all fit into one node. This isn't a new problem: the old scalers had the exact same behavior, but ironically it didn't really matter because they couldn't scale down very well. We should solve this properly at some point, but for now I recommend increasing the scaler interval.
- The beta parameter has always annoyed me a bit: I like the basic idea, but I think the current behavior is less useful than it could be, and can sometimes lead to problematic under-provisioning. Right now it is basically used to ignore small fluctuations, which is fine. But the problem I usually have with the scaler isn't with small fluctuations, it's with fluctuations that are too large, too fast. Usually this happens when thousands of small jobs are being dumped all at once, right after a series of long jobs, causing the autoscaler to freak out and hit its max -- then it rapidly figures out the jobs are pretty short and adjusts to something smaller than the max within 5 minutes, wasting some money. So I'd prefer the inertia parameter to be something much more like a moving average, which could give the autoscaler a chance to adjust, while still being able to hit the max in a reasonable time. (The moving average could still have some minor fluctuations, so maybe the existing beta parameter behavior would still be needed.)
- How should we order the "preference" of nodes to launch? Currently a sort is used which will do the "right" thing in the vast, vast majority of cases--but not all. Should we use the ordering provided by the user, and trust them to do it right? Or should we sort to prioritize the cheapest instances first, which is probably what the user wants?

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -39,7 +39,7 @@ except ImportError:
 # Python 3 compatibility imports
 from six import iteritems
 
-from bd2k.util.humanize import bytes2human
+from toil.lib.humanize import bytes2human
 from bd2k.util.retry import retry
 
 from toil import logProcessContext
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2018 Regents of the University of California
+# Copyright (C) 2015-2016 Regents of the University of California
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -53,7 +53,7 @@ from toil import lookupEnvVar
 from toil.version import dockerRegistry, dockerTag
 
 # aim to pack autoscaling jobs within a 20 minute block before provisioning a new node
-defaultTargetTime = 1200
+defaultTargetTime = 1800
 logger = logging.getLogger(__name__)
 
 
@@ -103,7 +103,7 @@ class Config(object):
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
-        self.deadlockWait = 80  # Wait 80 seconds before declaring a deadlock
+        self.deadlockWait = 60  # Wait 80 seconds before declaring a deadlock
         self.statePollingWait = 1  # Wait 1 seconds before querying job state
 
         # Resource requirements
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -49,11 +49,11 @@ from toil.batchSystems.options import addOptions as addBatchOptions
 from toil.batchSystems.options import setDefaultOptions as setDefaultBatchOptions
 from toil.batchSystems.options import setOptions as setBatchOptions
 from toil.provisioners import clusterFactory
-from toil.provisioners.clusterScaler import defaultTargetTime
-
 from toil import lookupEnvVar
 from toil.version import dockerRegistry, dockerTag
 
+
+defaultTargetTime = 1800
 logger = logging.getLogger(__name__)
 
 
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -231,6 +231,8 @@ class Config(object):
         setOption("minNodes", parseIntList)
         setOption("maxNodes", parseIntList)
         setOption("alphaTime", int)
+        if self.alphaTime <= 0:
+            raise RuntimeError('alphaTime must be a positive integer (was set to %s)' % self.alphaTime)
         setOption("betaInertia", float)
         setOption("scaleInterval", float)
         setOption("metrics")
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2016 Regents of the University of California
+# Copyright (C) 2015-2018 Regents of the University of California
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -52,8 +52,8 @@ from toil.provisioners import clusterFactory
 from toil import lookupEnvVar
 from toil.version import dockerRegistry, dockerTag
 
-
-defaultTargetTime = 1800
+# aim to pack autoscaling jobs within a 20 minute block before provisioning a new node
+defaultTargetTime = 1200
 logger = logging.getLogger(__name__)
 
 
@@ -95,7 +95,7 @@ class Config(object):
         self.maxNodes = [10]
         self.targetTime = defaultTargetTime
         self.betaInertia = 0.1
-        self.scaleInterval = 30
+        self.scaleInterval = 60
         self.preemptableCompensation = 0.0
         self.nodeStorage = 50
         self.metrics = False
@@ -103,7 +103,7 @@ class Config(object):
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
-        self.deadlockWait = 60  # Wait one minute before declaring a deadlock
+        self.deadlockWait = 80  # Wait 80 seconds before declaring a deadlock
         self.statePollingWait = 1  # Wait 1 seconds before querying job state
 
         # Resource requirements
@@ -408,16 +408,17 @@ def _addOptions(addGroupFn, config):
                      "default=%s" % config.maxNodes[0])
 
     addOptionFn("--targetTime", dest="targetTime", default=None,
-                help=("Sets how rapidly you aim to complete jobs. Shorter times mean more "
-                      "aggressive parallelization. The autoscaler attempts to scale up/down "
+                help=("Sets how rapidly you aim to complete jobs in seconds. Shorter times mean "
+                      "more aggressive parallelization. The autoscaler attempts to scale up/down "
                       "so that it expects all queued jobs will complete within targetTime "
                       "seconds. default=%s" % config.targetTime))
     addOptionFn("--betaInertia", dest="betaInertia", default=None,
                 help=("A smoothing parameter to prevent unnecessary oscillations in the "
                       "number of provisioned nodes. This controls an exponentially weighted "
                       "moving average of the estimated number of nodes. A value of 0.0 "
-                      "disables any smoothing, and a value of 1.0 will smooth so much that "
-                      "no change will ever be made. default=%s" % config.betaInertia))
+                      "disables any smoothing, and a value of 0.9 will smooth so much that "
+                      "few changes will ever be made.  Must be between 0.0 and 0.9.  "
+                      "default=%s" % config.betaInertia))
     addOptionFn("--scaleInterval", dest="scaleInterval", default=None,
                 help=("The interval (seconds) between assessing if the scale of"
                       " the cluster needs to change. default=%s" % config.scaleInterval))
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -52,7 +52,7 @@ from toil.provisioners import clusterFactory
 from toil import lookupEnvVar
 from toil.version import dockerRegistry, dockerTag
 
-# aim to pack autoscaling jobs within a 20 minute block before provisioning a new node
+# aim to pack autoscaling jobs within a 30 minute block before provisioning a new node
 defaultTargetTime = 1800
 logger = logging.getLogger(__name__)
 
@@ -103,8 +103,8 @@ class Config(object):
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
-        self.deadlockWait = 60  # Wait 80 seconds before declaring a deadlock
-        self.statePollingWait = 1  # Wait 1 seconds before querying job state
+        self.deadlockWait = 60  # Number of seconds to wait before declaring a deadlock
+        self.statePollingWait = 1  # Number of seconds to wait before querying job state
 
         # Resource requirements
         self.defaultMemory = 2147483648
@@ -240,7 +240,7 @@ class Config(object):
         setOption("metrics")
         setOption("preemptableCompensation", float)
         if not 0.0 <= self.preemptableCompensation <= 1.0:
-            raise RuntimeError('preemptableCompensation (%f) must be between 0.1 and 0.9!'
+            raise RuntimeError('preemptableCompensation (%f) must be between 0.0 and 1.0!'
                                '' % self.preemptableCompensation)
         setOption("nodeStorage", int)
 
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -209,8 +209,6 @@ class Config(object):
         elif self.clean is None:
             self.clean = "onSuccess"
         setOption('clusterStats')
-
-        # Restarting the workflow options
         setOption("restart")
 
         # Batch system options
@@ -222,7 +220,6 @@ class Config(object):
         setOption("parasolCommand")
         setOption("parasolMaxBatches", int, iC(1))
         setOption("linkImports")
-
         setOption("environment", parseSetEnv)
 
         # Autoscaling options
@@ -233,17 +230,18 @@ class Config(object):
         setOption("maxNodes", parseIntList)
         setOption("targetTime", int)
         if self.targetTime <= 0:
-            raise RuntimeError(
-                'targetTime must be a positive integer (was set to %s)' % self.targetTime)
+            raise RuntimeError('targetTime (%s) must be a positive integer!'
+                               '' % self.targetTime)
         setOption("betaInertia", float)
-        if not 0.0 <= self.betaInertia <= 1.0:
-            raise RuntimeError('--betaInertia (%f) must be >= 0.0 and <= 1.0.' % self.betaInertia)
+        if not 0.1 <= self.betaInertia <= 0.9:
+            raise RuntimeError('betaInertia (%f) must be between 0.1 and 0.9!'
+                               '' % self.betaInertia)
         setOption("scaleInterval", float)
         setOption("metrics")
         setOption("preemptableCompensation", float)
         if not 0.0 <= self.preemptableCompensation <= 1.0:
-            raise Exception(
-                '--preemptableCompensation (%f) must be >= 0.0 and <= 1.0.' % self.preemptableCompensation)
+            raise RuntimeError('preemptableCompensation (%f) must be between 0.1 and 0.9!'
+                               '' % self.preemptableCompensation)
         setOption("nodeStorage", int)
 
         # Parameters to limit service jobs / detect deadlocks
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -94,7 +94,7 @@ class Config(object):
         self.nodeOptions = None
         self.minNodes = None
         self.maxNodes = [10]
-        self.alphaPacking = 0.0
+        self.alphaTime = 1800
         self.betaInertia = 1.2
         self.scaleInterval = 30
         self.preemptableCompensation = 0.0
@@ -230,7 +230,7 @@ class Config(object):
         setOption("nodeOptions")
         setOption("minNodes", parseIntList)
         setOption("maxNodes", parseIntList)
-        setOption("alphaPacking", float)
+        setOption("alphaTime", int)
         setOption("betaInertia", float)
         setOption("scaleInterval", float)
         setOption("metrics")
@@ -398,12 +398,11 @@ def _addOptions(addGroupFn, config):
                 "as a default if the list length is less than the number of nodeTypes. "
                 "default=%s" % config.maxNodes[0])
 
-    # TODO: DESCRIBE THE FOLLOWING TWO PARAMETERS
-    addOptionFn("--alphaPacking", dest="alphaPacking", default=None,
-                help=("The total number of nodes estimated to be required to compute the issued "
-                      "jobs is multiplied by the alpha packing parameter to produce the actual "
-                      "number of nodes requested. Values of this coefficient greater than one will "
-                      "tend to over provision and values less than one will under provision. default=%s" % config.alphaPacking))
+    addOptionFn("--alphaTime", dest="alphaTime", default=None,
+                help=("Sets how rapidly you aim to complete jobs. Shorter times mean more "
+                      "aggressive parallelization. The autoscaler attempts to scale up/down "
+                      "so that it expects all queued jobs will complete within alphaTime "
+                      "seconds. default=%s" % config.alphaTime))
     addOptionFn("--betaInertia", dest="betaInertia", default=None,
                 help=("A smoothing parameter to prevent unnecessary oscillations in the "
                       "number of provisioned nodes. If the number of nodes is within the beta "
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -15,6 +15,7 @@
 from __future__ import absolute_import
 
 from future import standard_library
+
 standard_library.install_aliases()
 from builtins import str
 from builtins import range
@@ -63,6 +64,7 @@ class Config(object):
     """
     Class to represent configuration operations for a toil workflow run.
     """
+
     def __init__(self):
         # Core options
         self.workflowID = None
@@ -82,32 +84,32 @@ class Config(object):
         self.cleanWorkDir = None
         self.clusterStats = None
 
-        #Restarting the workflow options
+        # Restarting the workflow options
         self.restart = False
 
-        #Batch system options
+        # Batch system options
         setDefaultBatchOptions(self)
 
-        #Autoscaling options
+        # Autoscaling options
         self.provisioner = None
         self.nodeTypes = []
         self.nodeOptions = None
         self.minNodes = None
         self.maxNodes = [10]
-        self.alphaTime = 1800
+        self.targetTime = 1800
         self.betaInertia = 0.1
         self.scaleInterval = 30
         self.preemptableCompensation = 0.0
         self.nodeStorage = 50
         self.metrics = False
-        
+
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
-        self.deadlockWait = 60 # Wait one minute before declaring a deadlock
-        self.statePollingWait = 1 # Wait 1 seconds before querying job state
+        self.deadlockWait = 60  # Wait one minute before declaring a deadlock
+        self.statePollingWait = 1  # Wait 1 seconds before querying job state
 
-        #Resource requirements
+        # Resource requirements
         self.defaultMemory = 2147483648
         self.defaultCores = 1
         self.defaultDisk = 2147483648
@@ -117,12 +119,12 @@ class Config(object):
         self.maxMemory = sys.maxsize
         self.maxDisk = sys.maxsize
 
-        #Retrying/rescuing jobs
+        # Retrying/rescuing jobs
         self.retryCount = 1
         self.maxJobDuration = sys.maxsize
         self.rescueJobsFrequency = 3600
 
-        #Misc
+        # Misc
         self.disableCaching = True
         self.disableChaining = False
         self.maxLogFileSize = 64000
@@ -134,7 +136,7 @@ class Config(object):
         self.useAsync = True
         self.forceDockerAppliance = False
 
-        #Debug options
+        # Debug options
         self.debugWorker = False
         self.badWorker = 0.0
         self.badWorkerFailInterval = 0.01
@@ -146,11 +148,11 @@ class Config(object):
         """
         Creates a config object from the options object.
         """
-        from bd2k.util.humanize import human2bytes #This import is used to convert
-        #from human readable quantites to integers
+        from bd2k.util.humanize import human2bytes  # This import is used to convert
+        # from human readable quantites to integers
         def setOption(varName, parsingFn=None, checkFn=None, default=None):
-            #If options object has the option "varName" specified
-            #then set the "varName" attrib to this value in the config object
+            # If options object has the option "varName" specified
+            # then set the "varName" attrib to this value in the config object
             x = getattr(options, varName, None)
             if x is None:
                 x = default
@@ -167,7 +169,7 @@ class Config(object):
                 setattr(self, varName, x)
 
         # Function to parse integer from string expressed in different formats
-        h2b = lambda x : human2bytes(str(x))
+        h2b = lambda x: human2bytes(str(x))
 
         def parseJobStore(s):
             name, rest = Toil.parseLocator(s)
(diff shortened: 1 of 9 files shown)


DailyDreaming:
This is amazing @joelarmstrong , thank you!  Will give a full review tomorrow.

diekhans:
I feel like some of this should be captured in the autoscaler doc instead of lost in a
pR>

src/toil/provisioners/clusterScaler.py
```diff
@@ -399,108 +400,81 @@ def __init__(self, scaler):
...
+                                self.preemptableNodeDeficit[nodeType] = 0
+
+                    #Attempt to terminate any nodes that we previously designated for 
+                    #termination, but which still had workers running.
+                    self._terminateIgnoredNodes()
+
+                    if self.stats:
+                        self.stats.checkStats()
+            except:
+                logger.exception("Exception encountered in scaler thread. Making a "
```
DailyDreaming: Maybe a comment about why we'd want to keep soldiering on?

src/toil/provisioners/clusterScaler.py
```diff
@@ -542,7 +530,7 @@ def setNodeCount(self, nodeType, numNodes, preemptable=False, force=False):
                                                                    preemptable=preemptable,
                                                                    force=force)
                 else:
-                    logger.info('Cluster already at desired size of %i. Nothing to do.', numNodes)
+                    logger.info('Cluster (minus ignored nodes) already at desired size of %i. Nothing to do.', numNodes)
```
DailyDreaming: Should this have a different log message if `force` is set?

src/toil/common.py
```diff
@@ -392,12 +392,11 @@ def _addOptions(addGroupFn, config):
                 "as a default if the list length is less than the number of nodeTypes. "
                 "default=%s" % config.maxNodes[0])
 
-    # TODO: DESCRIBE THE FOLLOWING TWO PARAMETERS
-    addOptionFn("--alphaPacking", dest="alphaPacking", default=None,
-                help=("The total number of nodes estimated to be required to compute the issued "
-                      "jobs is multiplied by the alpha packing parameter to produce the actual "
-                      "number of nodes requested. Values of this coefficient greater than one will "
-                      "tend to over provision and values less than one will under provision. default=%s" % config.alphaPacking))
+    addOptionFn("--alphaTime", dest="alphaTime", default=None,
```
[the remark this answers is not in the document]
joelarmstrong: Yes, I'd love to change it! (though I don't have any great ideas ATM.) I just didn't want to keep it as alphaPacking.

src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -170,7 +208,7 @@ def testClusterScaling(self):
         config.maxNodes = [10]
 
         # Algorithm parameters
-        config.alphaPacking = 0.0
+        config.alphaTime = 3600
```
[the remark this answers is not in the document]
joelarmstrong: I kept these tests at 3600 because they were written I had 3600 as the default. They probably still pass with 1800, but I haven't tried.

(It wouldn't be a bad thing if they didn't pass with 1800. Some of these tests are testing specific behavior of the bin-packing algorithm with a fixed value of alpha. I think for those tests I made sure that any remotely reasonable value of alpha [less than a day or so] would do, but I'm not 100% sure.)

src/toil/provisioners/clusterScaler.py
```diff
@@ -313,21 +325,14 @@ def addCompletedJob(self, job, wallTime):
...
-    completed jobs.
-    The scaling calculation is essentially as follows: Use the RecentJobShapes instance to
-    calculate how many nodes, n, can be used to productively compute the last N completed
-    jobs. Let M be the number of jobs issued to the batch system. The number of nodes
-    required is then estimated to be alpha * n * M/N, where alpha is a scaling factor used to
-    adjust the balance between under- and over- provisioning the cluster.
+    nodes according to the resource requirements of the queued jobs.
+    The scaling calculation is essentially as follows: start with 0 estimated worker nodes. For
+    each queued job, check if we expect it can be scheduled into a worker node before a certain time
+    (currently one hour). Otherwise, attempt to add a single new node of the smallest type that
```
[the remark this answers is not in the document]
joelarmstrong: d'oh, i knew I'd forget to change something in the docstrings! Yeah, probably better to leave this out than to have to change it everytime the default changes.

src/toil/provisioners/clusterScaler.py
```diff
@@ -399,108 +400,81 @@ def __init__(self, scaler):
...
+                        logger.info("Nodes of type %s to run queued jobs = %s" % (nodeType, nodesToRunQueuedJobs[nodeShape]))
+                        # Actual calculation of the estimated number of nodes required
+                        estimatedNodes = 0 if nodesToRunQueuedJobs[nodeShape] == 0 else max(1, int(round(nodesToRunQueuedJobs[nodeShape])))
+                        logger.info("Estimating %i nodes of shape %s" % (estimatedNodes, nodeShape))
+
+
+                        # If we're scaling a non-preemptable node type, we need to see if we have a 
+                        # deficit of preemptable nodes of this type that we should compensate for.
+                        if not nodeShape.preemptable:
+                            compensation = self.scaler.config.preemptableCompensation
```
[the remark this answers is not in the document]
joelarmstrong: Honestly, I have no idea if preemptableCompensation even works. I've never used it, barely understand its purpose, and there aren't any tests for it that I know of (go figure!).

src/toil/provisioners/clusterScaler.py
```diff
@@ -47,193 +46,196 @@
...
-                                                    cores=config.defaultCores,
-                                                    disk=config.defaultDisk,
-                                                    preemptable=True)])
-        # Calls to add and getLastNJobShapes may be concurrent
-        self.lock = Lock()
-        # Number of jobs to average over
-        self.N = N
-
-    def add(self, jobShape):
+    def __init__(self, nodeShapes, targetTime=3600):
```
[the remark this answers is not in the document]
joelarmstrong: It actually is set to alphaTime (hopefully), through the `binPacking` function. Yes, maybe targetTime is what alphaTime should be called.

src/toil/provisioners/clusterScaler.py
```diff
@@ -278,8 +281,17 @@ def shutdown(self):
         """
         self.stop = True
         self.scaler.join()
-                
-    def getAverageRuntime(self, jobName):
+
+    def getAverageRuntime(self, jobName, service=False):
```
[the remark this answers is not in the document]
joelarmstrong: good point! i think, for simplicity, we should enforce it to be a positive integer when parsing arguments.

src/toil/provisioners/clusterScaler.py
joelarmstrong: I wouldn't be against it, but I'm not sure what it'd look like. What kind of message were you thinking?

DailyDreaming:
So I really like a lot of the changes made.  Some points I'd like to address:

1. I'd like more testing for cases where we have a known desired outcome.  Particularly, the new way that AWS bills (per minute) makes the old way of prioritizing wallTime to maximize the hour that was paid for doesn't make as much sense anymore.  My preference would be launching jobs in parallel as aggressively as possible and almost ignoring the wallTime blocks.  I'm not sure if 60 seconds has been tested as a default wallTime, but I think that would make more sense now (to me at least).
2. Increasing the 60 second intervals for the scaler interval seems sensible to me.  Should we double it?
3. Should we add a constant to the service job time rather than simply `alphaTime * 24`?  I'm just worried the user will set `alphaTime` to zero, and then have unexpected deadlocks.  I wouldn't mind having a unittest for this particular case.
4. Should the `--betaInertia` parameter be a user option?  I agree with your change, though it sounds more like a feature that should be behind the scenes rather than the "try it and see if it works" tweaking that each new user might have to go through.  I'd like to implement the better compensation for rapid swings you're suggesting though I feel like it might be easier if the algorithm in general were reworked to be simpler and just launch jobs as finishing jobs initiate the queue without so much importance on packing based on these time intervals (again, on a pay-by-the-hour system, this made more sense, but I feel like it's less important now).
5. I agree that it would be pretty great to have an option that would sort by price, and have the workflow use the cheapest instances (especially for, I assume rare, cases where a non-preemptable is cheaper than a preemptable instance), but since these fluctuate, I feel like implementing this would be more of a pain than it's worth.  Allowing the user to provide a price list (which takes the burden off of toil accidentally mispricing something) and prioritize based on this might be a feature worth having (especially for long running, expensive runs like cactus), though I'm not sure it's a priority feature right now.  I like the general idea that jobs should run on the instance with the fewest resources meeting their requirements, but you're right that while generally that's cheaper, it's not always the case, especially when certain instances aim to be cpu "cheap" or memory "cheap".

I like a lot of the commenting, and it's much clearer than it once was.  The NodeReservation object augmentation was pretty great as well.  I agree with Mark that some of these questions and the design choices made should be included in some of the comments as well.  This is really awesome overall and thank you for these changes!

src/toil/common.py
```diff
@@ -93,7 +93,7 @@ def __init__(self):
         self.nodeOptions = None
         self.minNodes = None
         self.maxNodes = [10]
-        self.alphaPacking = 0.0
+        self.alphaTime = 1800
```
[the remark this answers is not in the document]
joelarmstrong: I think lower would be fine, but how much lower I'm not sure. My gut tells me 60 would be too low, because the per-node overhead becomes a larger proportion of cost as they run for less time. Basically, because it can take a decent amount of time (IIRC over a minute) just to download and launch the toil container, lower alphaTime means higher cost. I'm not sure there is really an "optimal" value of alphaTime, but intuitively I think the cost will increase massively as the value goes below a few minutes.

As a totally contrived example, imagine that you are running 1000 jobs that take 1 minute to run and take up exactly one instance each, which costs (say) 1 cent per minute. Also assume the toil container takes one minute to initialize after the instance boots (IIRC this is in the right ballpark). Then running this with different values of alphaTime would yield different costs*:

- alphaTime = 60: 1000 instances launched for 2 minutes each, $20.00
- alphaTime = 120: 500 instances launched for 3 minutes each, $15.00
- alphaTime = 600: 100 instances launched for 11 minutes each, $11.00

[*] This is under the simplifying assumption that there is only one scaler decision made, at the start: in reality the last two costs will be different (lower, but with noticably longer runtime; see the first point under 'considerations' up above). Unfortunately calculating the costs with extra scaler decisions is kind of a pain.

src/toil/provisioners/clusterScaler.py
DailyDreaming: Just the original message, since `delta = numNodes - (numCurrentNodes - numIgnoredNodes)` (and thus `(minus ignored nodes)`) is only True if != force.

src/toil/provisioners/clusterScaler.py
```diff
@@ -47,193 +46,196 @@
...
-            self.jobShapes.append(jobShape)
+        chosenNodeShape = None
+        for nodeShape in self.nodeShapes:
+            if NodeReservation(nodeShape).fits(jobShape):
+                # This node shape is the first that fits this jobShape
+                chosenNodeShape = nodeShape
+                break
+
+        if chosenNodeShape is None:
+            logger.warning("Couldn't fit job with requirements %r into any nodes in the nodeTypes list." % jobShape)
```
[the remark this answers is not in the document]
joelarmstrong: I'd vote no, I think, because at least progress could be made via other jobs (or the job may fit onto a previously statically provisioned worker node).

src/toil/provisioners/clusterScaler.py
joelarmstrong: ah OK, that makes sense.

src/toil/provisioners/clusterScaler.py
```diff
@@ -47,193 +46,196 @@
 ch.setFormatter(formatter)
 logging.getLogger().addHandler(ch)
 
-
-class RecentJobShapes(object):
+class BinPackedFit(object):
     """
-    Used to track the 'shapes' of the last N jobs run (see Shape).
+    Use a first fit decreasing (FFD) bin packing like algorithm to
```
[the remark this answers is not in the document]
joelarmstrong: Yes! definitely! we should mention that BinPackingFit takes the input nodeShapes order as the "node preference" order. It assumes any sorting of nodeShapes is done outside of BinPackingFit. That is, when virtually "creating" nodes, the first node within the list that fits the job is the one that's added.

joelarmstrong:
OK, after an additional refactor I'm handing this off to Lon now, who graciously volunteered to shepherd this the rest of the way. It's totally possible that I've broken things a bit with these additional changes (though there are a few more tests that actually test things now). Adding more tests revealed a few bugs, which are fixed now, but I assume are around on master branch as well.

I forgot that in the original PR, I had actually removed betaInertia (because the way it's formulated caused unrecoverable service-deadlock problems by under-provisioning nodes). I added this back in and converted it to an exponentially weighted moving average. I have no idea if having betaInertia actually a good idea or not, but IMO it's a better solution than the old betaInertia regardless--I'll try a run over the weekend and see how it feels.

Lon, I didn't get around to everything we agreed was a good idea yesterday (alphaTime is still called alphaTime, for one). Feel free to change/rename whatever you feel like, I don't have any particular preference. I think for sanity's sake the core algorithm should remain the same, and that alphaTime should remain (gut feeling) above 10min, but otherwise go nuts. We can discuss the alpha parameter and the core algorithm more widely to get other people's input--I think it's a useful discussion to have, because the current algorithm is obviously sub-optimal. But better to get the incremental improvements in even if in the future the core deserves a rethink. I think it's easily possible that betaInertia should be set to 0.0 by default, or just plain removed, I'm not too attached to it. Please let me know if you have questions, sorry for dumping a giant PR on you!

src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -58,6 +56,47 @@
...
+        """Pack one shape and check that the resulting reservations look sane."""
+        self.bpf.nodeReservations[self.c4_8xlarge] = [NodeReservation(self.c4_8xlarge)]
+        self.bpf.addJobShape(Shape(wallTime=1000, cores=2, memory=1000000000, disk=2000000000, preemptable=True))
+        self.assertEqual(self.bpf.nodeReservations[self.r3_8xlarge], [])
+        self.assertEqual([x.shapes() for x in self.bpf.nodeReservations[self.c4_8xlarge]],
+                         [[Shape(wallTime=1000, memory=59000000000, cores=34, disk=98000000000, preemptable=True),
+                           Shape(wallTime=2600, memory=60000000000, cores=36, disk=100000000000, preemptable=True)]])
+
+    def testAddingInitialNode(self):
+        """Pack one shape when no nodes are available and confirm that we fit one node properly."""
```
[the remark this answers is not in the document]
joelarmstrong: It's sort of subtle, but the previous one already puts a "blank" node reservation in *before* exercising the method. This one is expecting the node to be added and reserved properly *during* the method.

More importantly, why do both? These are testing two different code paths: the above reservation modification happens within attemptToAddJob, the other one happens in the latter half of addJobShape. And I forget exactly, but I have a hazy memory of newly "created" nodes not actually reserving space for the job that triggered their creation in the original code: that is, one of these code paths was broken but not the other. (One could argue that these two code paths should probably be the same code path--which is probably true. I only refactored up to the point where I was convinced everything worked.)

DailyDreaming:
Thanks Joel!  Still in progress and trying to make the tests better.  Fixed one bug at least.

joelarmstrong:
d'oh! thank you! I don't use the clusterStats stuff myself so that totally slipped by me.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -41,10 +41,17 @@ def __eq__(self, other):
...
-            self.memory < other.memory or \
-            self.cores < other.cores or \
-            self.disk < other.disk or \
-            self.wallTime < other.wallTime
+               self.memory < other.memory or \
+               self.cores < other.cores or \
+               self.disk < other.disk or \
+               self.wallTime < other.wallTime
+
+    def __gt__(self, other):
```
joelarmstrong: You shouldn't need `__gt__` here, at least if this weird multiple-inheritance crap does its job. The `functools.total_ordering` decorator defines all seven comparison methods for you consistently, as long as you define `__eq__` and one of <=, <, >, >=. The point of this weird Ordering stuff is to avoid defining all seven, because this is super error-prone and a flipped sign can lead to bugs. Did something come up which meant `__gt__` needed to be defined?

diekhans: also, pep8 recommends putting multiline expressions like this in parenthesis rather than backslash escaping each line.  just an FYI, not a requested change, but it does improve readability IMUO.

DailyDreaming: You're right, I misunderstood originally.  I used it temporarily, and saw the `@total_ordering` decorator later but don't really have a good reason for leaving it in.  Will remove it when I push the next few commits.  It's taking a bit for me to absorb all of the inheritance throughout.

DailyDreaming:
I think this is ready to be reviewed and merged.
- Added some additional tests.
- Added a constant to the service job handling so that it expects service jobs to always take at least an hour.
- Scaled down the 1800 self.targetTime default to 1200.  
- Bumped the self.scaleInterval default up to 60 from 30 as per earlier comments (though should we go above 60?  Maybe I misunderstood this since the comment was that the current default was 60).  
- A betaInertia of 1.0 smoothed too much and the clusterScaler would never actually change the scaling, allowing things to seemingly run forever, so I set it so that betaInertia must be a value between 0 and 0.9.
- Changed the algorithm slightly with Joel to account for tests not provisioning t2.micro tests as expected.
- Saved some minutes on jenkins by reducing a time.sleep in one of the tests.
- Added the humanize stuff in from bd2k since I was using the human2bytes function.
- General commenting.

Overall, thanks for letting me add a little to your PR Joel!  The refactoring you did looks great, and I'm really glad to have these fixes in toil master!

One other thing, some of the comments here are valuable but perhaps not useful to have in our user docs.  I'm going to try to capture it in a design doc on the toil github wiki if that's appropriate.

src/toil/common.py
```diff
@@ -51,7 +51,7 @@
 from toil import lookupEnvVar
 from toil.version import dockerRegistry, dockerTag
 
-
+# aim to pack autoscaling jobs within a 20 minute block before provisioning a new node
```
joelarmstrong: Sorry, this is such a nitpick, but I gotta point out that this is 30 minutes, not 20.

joelarmstrong:
I think wiki documentation would be great. I agree it doesn't really fit into the user docs well, at least not within the ReadTheDocs framework.

Oops, my bad about the scaleInterval. I thought the default was 60 for some reason, but I never actually checked to see if that was accurate.

src/toil/common.py
```diff
@@ -94,15 +94,15 @@ def __init__(self):
...
+        self.scaleInterval = 60
         self.preemptableCompensation = 0.0
         self.nodeStorage = 50
         self.metrics = False
 
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
-        self.deadlockWait = 60  # Wait one minute before declaring a deadlock
+        self.deadlockWait = 60  # Wait 80 seconds before declaring a deadlock
```
joelarmstrong: Again, comment seems to be referring to a different value. (Maybe we should just rewrite it to "number of seconds to wait before declaring a deadlock"?)

src/toil/common.py
```diff
@@ -230,17 +227,18 @@ def parseIntList(s):
...
+        if not 0.0 <= self.betaInertia <= 0.9:
+            raise RuntimeError('betaInertia (%f) must be between 0.0 and 0.9!'
+                               '' % self.betaInertia)
         setOption("scaleInterval", float)
         setOption("metrics")
         setOption("preemptableCompensation", float)
         if not 0.0 <= self.preemptableCompensation <= 1.0:
-            raise Exception(
-                '--preemptableCompensation (%f) must be >= 0.0 and <= 1.0.' % self.preemptableCompensation)
+            raise RuntimeError('preemptableCompensation (%f) must be between 0.1 and 0.9!'
```
joelarmstrong: Values are wrong here

src/toil/common.py
DailyDreaming: Sorry, I had it at 1200, but the autoscaling tests failed (possibly still flaky, but I think this may have actually caused them to fail).  Anyway, I forgot to set it back, thanks for catching this!

src/toil/common.py
DailyDreaming: Ack, sorry about that.

src/toil/common.py
DailyDreaming: Thanks for catching this too.

DailyDreaming:
Problems should be addressed.  Thanks for catching those and for letting me contribute!  I saw this passed all tests on the last one, and I haven't seen that in a while.  I think this PR is much needed and really appreciate the time you put into submitting it!  Hopefully the scale test goes well.

joelarmstrong:
Thanks! I have a large dataset to run now anyway, so I'll try it with this and see if I can catch anything.

joelarmstrong:
Hey, sorry for the radio silence on this front. I struggled a bit to avoid deadlocks when running from this branch. The good news is that the autoscaler works just fine now, it seems (thanks for the bugfixes!).

But I'm having two issues that cause most runs to stall within ten hours or so:
1. a cPickle deadlock bug, where occasionally a `_toil_worker` process will completely stall trying to unpickle the Job object (likely only if the pickle is larger than 4Kb).
2. Docker daemon crashes causing all Docker `run`s / `exec`s to stall. (This may be something you only run into with usage of stdin through an attached docker run process.)

I haven't run from upstream toil in a long time so I didn't really notice these before. Have you guys run into any of this during your scale testing? Any unexplained stalls where nothing is happening? I especially think you may have run into 1) before, it seems to happen once in every few thousand jobs for me. But since you are running Cactus every so often, have you noticed 2) too?

DailyDreaming:
Sorry for the silence as well.  I was at the hackathon Mon-Wed and forgot to check this.

Ah, are you using the previous commit?  I'm guessing this is fixed in your fork, but there appears to be an undefined variable currently in the PR called `memoryOverhead` that I assume should be some percentage of the total memory available: https://github.com/BD2KGenomics/toil/blob/f32fd565f324e74980dc163af953b5334df02e0f/src/toil/provisioners/aws/awsProvisioner.py#L307

I've run 1000 jobs for integration testing on master only, but these were simple, `time.sleep(#)` jobs that I created for this explicit purpose, and we've only started doing this in the past month or so.  Because of how the current autoscaler functions, I had been hoping to have this PR implemented before continuing to develop more complex runs.  Running 1000+ jobs is pretty new, so I can't say we've run into either problem before.

As for our small pestis cactus runs, they are run pretty infrequently since they're not part of the main unit tests, and take far less than 10 hours.  I don't think we currently test anything right now that takes 10 hours.  As for this particular branch, the only integration test has been your cactus run.

joelarmstrong:
Oops, my bad about the memoryOverhead thing. That is from a different piece of work, trying to solve the problem of the autoscaler and Mesos disagreeing about available RAM. (Long story. Mesos reserves a hardcoded 1GB of ram for some reason, which is a really big problem on small instances.)

Ah, OK. That makes sense. I can push a fix for the pickle deadlock. I've tracked it down and unfortunately it's a bug in cPickle. There isn't much we can do to avoid it aside from never using pickle.load from large, remote jobStore streams.

The docker issue is a bit weird. I can't tell exactly what the deal is, but it doesn't seem like it's really toil's problem. I can work around it on the Cactus side, it may be that we are abusing too many weird, uncommon docker use cases, and that some docker bug was introduced recently.

DailyDreaming:
Yes, I remember seeing that.  I think you might have even been the one to show me.  Seems like a necessary thing to correct for.

Is [this link](https://bugs.python.org/issue13555) the cPickle error?  It seems as though normal pickle works if this is the case, and it's only cPickle that breaks.  Is that the plan of attack?  :)

I'm unclear about the docker error.  Are these the worker mesos images themselves or docker images called from within the mesos images or everything?  I wouldn't mind having a look because now you have me curious.

joelarmstrong:
I'm not sure it's a bug that anyone else but toil could run into. It has to do with cPickle occasionally not releasing the GIL when reading. (C extensions must release the GIL before a potentially blocking operation, for exactly the reason we are having problems with: the operation may be relying on a separate Python thread to run and provide the data.) Switching to HIGHEST_PROTOCOL means that it releases the GIL the vast majority of the time, but sadly I discovered there are still sometimes a few  read_global opcodes that get dumped, and the data for those is read without releasing the GIL. So we are inevitably stuck if we try to load from a stream.

(We could switch to pickle rather than cPickle to avoid this. But that has performance implications, and Python 3 aliases pickle to cPickle anyway.)

Yeah, these are issues with docker images called from within the toil_worker container. It's a bit difficult to track down, but I did get a few kernel errors about dockerd filesystem sync syscalls getting stuck for several minutes. So it may be a kernel issue, or just some weird coincidental interaction with EBS.

DailyDreaming:
I see.  So is the fix to stream the pipe to a file before running `cPickle.load`?

(Makes sense, and I think we can't move to py3 soon enough.)

Huh, that's pretty interesting.  If you have the errors, could you send them my way?

joelarmstrong:
Here's the hang. I got quite a few of these, which roughly corresponded to the time where docker run commands seemed to start hanging, though I'm not positive about that.

```[31089.543761] INFO: task dockerd:1720 blocked for more than 120 seconds.
[31089.547240]       Not tainted 4.14.16-coreos #1
[31089.549550] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[31089.553487] dockerd         D    0  1720      1 0x00000080
[31089.556215] Call Trace:
[31089.557493]  ? __schedule+0x2e3/0x890
[31089.559328]  schedule+0x32/0x80
[31089.560800]  jbd2_log_wait_commit+0x93/0x110 [jbd2]
[31089.563393]  ? wait_woken+0x80/0x80
[31089.565115]  ext4_resize_fs+0x8f50/0xe390 [ext4]
[31089.567393]  0xffffffffc03d8233
[31089.568972]  sync_filesystem+0x6a/0x90
[31089.570847]  generic_shutdown_super+0x22/0x120
[31089.573127]  kill_anon_super+0xe/0x20
[31089.574782]  deactivate_locked_super+0x39/0x70
[31089.577020]  cleanup_mnt+0x3b/0x70
[31089.578799]  task_work_run+0x80/0xa0
[31089.580565]  exit_to_usermode_loop+0xa4/0xb0
[31089.582786]  do_syscall_64+0xe9/0x1c0
[31089.584690]  entry_SYSCALL64_slow_path+0x25/0x25
[31089.587035] RIP: 0033:0x495e24
[31089.588630] RSP: 002b:000000c4248d0eb0 EFLAGS: 00000246 ORIG_RAX: 00000000000000a6
[31089.592223] RAX: 0000000000000000 RBX: 0000000000000000 RCX: 0000000000495e24
[31089.595714] RDX: 0000000000000000 RSI: 0000000000000002 RDI: 000000c426163420
[31089.599226] RBP: 000000c4248d0f10 R08: 0000000000000000 R09: 0000000000000000
[31089.602722] R10: 0000000000000000 R11: 0000000000000246 R12: 0000000000000000
[31089.606311] R13: 00000000ffffffee R14: 0000000000001420 R15: 0000000000000049
```
Looks like it's something to do with syncing the filesystem, though it would be interesting to know exactly what syscall it's hanging on from Docker. It could be some weird problem caused by mounting /var/lib/docker in the containers, given that it appears to be something to do with mounting/unmounting filesystems. But I'm not sure I totally trust the stacktrace: why would ext4_resize_fs get called?

I also had the journal crash for some reason:
```[39528.688191] systemd-coredump[67557]: MESSAGE=Process 956 (systemd-journal) of user 0 dumped core.
[39528.694227] systemd-coredump[67557]: Coredump diverted to /var/lib/systemd/coredump/core.systemd-journal.0.296fcc15f88b46bab231ebcc5
63f456d.956.1522843727000000.lz4
[39528.695542] systemd-journald[67633]: File /var/log/journal/6ed0fa8a724b40a0b4c57e5d4f7724c9/system.journal corrupted or uncleanly sh
ut down, renaming and replacing.
[39528.730025] systemd-journald[67633]: Received SIGTERM from PID 1 (systemd).```

Yes, reading the stream into a file is the solution I had in mind. I have a patch which seems to fix the immediate problem. I can do a PR tonight or early tomorrow.

benedictpaten:
What's the status for this? @glennhickey @adamnovak are having autoscaler issues with multiple node types that is somewhat blocking and at least slowing down vg testing. @DailyDreaming Is this is the current sprint?

joelarmstrong:
I forget the status of this. I think at one point I was waiting for the provisioner overhaul to get committed. But I ran the 48-way on this, or something close to it, and really the only issues I ran into were docker bugs.

I can rebase and remove the conflicts, give me an hour or two. Hopefully the tests will pass; i don't think there were really any remaining issues with this PR.

glennhickey:
This would be really handy for toil-vg and would resolve #2195.

Issue #2095, opened 2018-03-01
Provisioner Refactor


ejacox:
Various issues to address for provisioners.

- [ ] set and test vpc-subnet
- [ ] zone is hard-coded for AWS
- [ ] environment variable to change image
- [ ] test SSE keys (implement them if not yet added)
- [ ] Add TOIL_<>_ZONE for azure and gce and make a consistent zone call
- [ ] Remove hard-coded 3000 port forward from coreSSH.
- [ ] The current AWS provisioner uses various adhoc methods to check on progress, e.g. 'ssh ps' to see if the docker daemon is running. This is not the most reliable method and often hangs.

┆Issue Number: TOIL-65

DailyDreaming:
Adding to #2203 and closing this.

PR #2124, opened 2018-03-13
Use googleJobstore in provisioner tests.

ejacox:


ejacox:
Jenkins, test this please

src/toil/utils/toilLaunchCluster.py
```diff
@@ -84,14 +84,16 @@ def main():
         try:
             from toil.provisioners.aws.awsProvisioner import AWSProvisioner
         except ImportError:
-            raise RuntimeError('The aws extra must be installed to use this provisioner')
+            print('The aws extra must be installed to use this provisioner')
+            raise
```
jessebrennan: I don't disapprove, Just wondering why this change.

src/toil/test/provisioners/gceProvisionerTest.py
```diff
@@ -287,7 +286,7 @@ def _runScript(self, toilOptions):
         # instances
         runCommand = ['/home/venv/bin/python', '/home/sort.py', '--fileToSort=/home/s3am/bin/asadmin', '--sortMemory=0.6G', '--mergeMemory=3.0G']
         runCommand.extend(toilOptions)
-        runCommand.append('--sseKey=/home/keyFile')
+        #runCommand.append('--sseKey=/home/keyFile')
```
jessebrennan: I guess google job store doesn't work with sse? Is there an issue for this?

src/toil/provisioners/gceProvisioner.py
```diff
@@ -482,56 +487,81 @@ def addNodes(self, nodeType, numNodes, preemptable):
...
-                        else:
-                            logger.debug("Rsync to new nodes failed, trying again")
-                            retries += 1
-                            time.sleep(20*retries)
-            workersCreated += 1
-        logger.info('Launched %s new instance(s)', numNodes)
-        if len(instancesLaunched) != workersCreated:
-            logger.error("Failed to launch %d workers", len(instancesLaunched)-workersCreated)
+        # Try a few times to create the requested number of workers
+        while numNodes-workersCreated > 0 and retries < 3:
```
jessebrennan: This is a nice, robust retry mechanism. I like it 👍

src/toil/provisioners/gceProvisioner.py
```diff
@@ -482,56 +487,81 @@ def addNodes(self, nodeType, numNodes, preemptable):
...
+                self.terminateNodes(failedWorkers)
+            retries += 1
+
+        logger.info('Launched %d new instance(s)', numNodes)
+        if numNodes != workersCreated:
+            logger.error("Failed to launch %d worker(s)", numNodes-workersCreated)
         return workersCreated
 
+
+    def _injectWorkerFiles(self, ip):
```
jessebrennan: These seem like much more robust ways of dealing with this. Definitely want to keep this model in mind for the provisioner refactor.

src/toil/provisioners/gceProvisioner.py
```diff
@@ -852,12 +876,9 @@ def _waitForAppliance(cls, ip_address, role, keyName='core'):
                     logger.info('...Still waiting for appliance, trying again in %s sec...' % sleepTime)
                     time.sleep(sleepTime)
             except:
-                if retries == 5:
-                    raise # givup
```
jessebrennan: Why did you get rid of this?

src/toil/provisioners/gceProvisioner.py
```diff
@@ -852,12 +876,9 @@ def _waitForAppliance(cls, ip_address, role, keyName='core'):
                     logger.info('...Still waiting for appliance, trying again in %s sec...' % sleepTime)
                     time.sleep(sleepTime)
             except:
```
jessebrennan: Now it will just eat up exception and keep retrying which is _not_ good behavior. What exception do we want to retry on?

src/toil/test/provisioners/gceProvisionerTest.py
ejacox: SSE keys are not yet supported in Google. Partially because they are not tested. I pushed adding it to the refactor (see #2095).

src/toil/utils/toilLaunchCluster.py
ejacox: It is clearer that way. Hiding exceptions by raising a different error is not a good practice.

src/toil/provisioners/gceProvisioner.py
ejacox: I was thinking a decorator would look cleaner. Something for the refactor.

src/toil/provisioners/gceProvisioner.py
ejacox: Time based failure seems to work better. That is covered by the loop on line 839. This mechanism seemed redundant or not very useful.

src/toil/provisioners/gceProvisioner.py
ejacox: At this point, the connection works and docker is running. I don't know of any specific errors other than transient connection failures. What do you suggest?

jessebrennan: As far as I can tell, the only exception we would expect to encounter is from `_coreSSH` and is a `RuntimeError` so I think you should only retry on that exception or maybe only that exception and with a particular error code.

src/toil/utils/toilLaunchCluster.py
```diff
@@ -84,14 +84,16 @@ def main():
...
         provisioner = AWSProvisioner()
     elif config.provisioner == 'gce':
         logger.info('Using a gce provisioner.')
         try:
             from toil.provisioners.gceProvisioner import GCEProvisioner
         except ImportError:
-            raise RuntimeError('The libCloud extra must be installed to use this provisioner')
+            print('The libCloud extra must be installed to use this provisioner')
```
jessebrennan: Somehow this slipped through the cracks... There is no `libCloud` extra! it should be google, right?

src/toil/provisioners/gceProvisioner.py
```diff
@@ -351,13 +355,14 @@ def launchCluster(self, leaderNodeType, leaderSpotBid, nodeTypes, preemptableNod
...
+            or not self._copySshKeys(leader.public_ips[0], keyName)
+            or not self._injectFile(leader.public_ips[0], self.credentialsPath, GoogleJobStore.nodeServiceAccountJson,
+                                    'toil_leader')
+            or (self.botoPath and
+                not self._injectFile(leader.public_ips[0], self.botoPath, self.nodeBotoPath, 'toil_leader'))):
 
-        if self.botoPath:
-            self._rsyncNode(leader.public_ips[0], [self.botoPath, ':' + nodeBotoPath],
-                            applianceName='toil_leader')
+            raise RuntimeError("Failed to start leader")
```
DailyDreaming: I recall making this comment before, but I think each of these conditions should throw a different error message to relate to the user why it failed.

src/toil/utils/toilLaunchCluster.py
DailyDreaming: I agree.  A plain `raise` is almost always better.

diekhans: Is this a program or a library?  print from libraries is generally not a good idea, they should go through the logger instead.

If this is a program, error messages should always be printed to stderr, not stdout.

I don't completely have the context here.  How are tests reporting error>

ejacox: Program. Stderr is better.

src/toil/utils/toilLaunchCluster.py
DailyDreaming: Commented on these in the previous PR but they were not fixed.

ejacox: Try refreshing your page to get the latest changes.

PR #2137, opened 2018-03-17
Change to "auto-deployment".  Addresses #2089.

DailyDreaming:
Addresses #2089.

There was confusion over hot, remote, and now the chosen one, "auto-deployment".  Hopefully this resolves these mentions throughout the code.


DailyDreaming:
I did!

Issue #1987, opened 2017-12-10
Metrics dashboard broken after prometheus:v2.0.0 container was published


joelarmstrong:
Any toil-prometheus containers pushed after v.2.0.0 was released are non-functional. We can fix this by hard-coding the tag in the FROM statement, or by changing the flags a bit to support v2.0.0. Ideally both.

PR #2139, opened 2018-03-17
Update prometheus to fix metrics.  Addresses #1987.

DailyDreaming:
Addresses #1987.

Now we can use grafana again.


joelarmstrong:
Added another necessary change for node-exporter which I forgot to commit on my fork.

DailyDreaming:
Thanks @joelarmstrong !  In the middle of testing this now.

Issue #2126, opened 2018-03-13
Toil should check if docker image (toil's appliance) exists.


DailyDreaming:
The python docker API has a nice way of handling this.  See the `docker.py` code for an example: https://github.com/BD2KGenomics/toil/blob/master/src/toil/lib/docker.py#L354

Issue #1831, opened 2017-08-15
launch-cluster waits forever for the Toil appliance to start


adamnovak:
To fix #1830, I uploaded my SSH public key to AWS.

```
$ aws ec2 import-key-pair --key-name "$(whoami)" --public-key-material "$(cat ~/.ssh/id_rsa.pub)"
{
    "KeyName": "anovak", 
    "KeyFingerprint": "14:f5:f3:39:a6:cf:c8:27:dd:78:7c:12:cc:22:85:3a"
}
```

Now I can start up an instance, but it looks like Docker never starts the Toil appliance:

```
$ TOIL_APPLIANCE_SELF=quay.io/ucsc_cgl/toil:3.9.1 toil launch-cluster novakcluster --nodeType=t2.micro -z us-west-2a --keyPairName=anovak
INFO:toil.lib.bioio:Root logger is at level 'INFO', 'toil' logger at level 'INFO'.
INFO:toil.utils.toilLaunchCluster:Using aws provisioner.
INFO:toil:Using default docker registry of quay.io/ucsc_cgl as TOIL_DOCKER_REGISTRY is not set.
INFO:toil:Using default docker name of toil as TOIL_DOCKER_NAME is not set.
INFO:toil:Overriding docker appliance of quay.io/ucsc_cgl/toil:3.10.0a1-dd898ab49f8a6605c4c88ef18df8fb384e4392f5-dirty with quay.io/ucsc_cgl/toil:3.9.1 from TOIL_APPLIANCE_SELF.
INFO:toil.provisioners.aws.awsProvisioner:Launching non-preemptable leader
INFO:cgcloud.lib.ec2:Creating t2.micro instance(s) ... 
INFO:toil.provisioners.aws.awsProvisioner:Waiting for toil_leader to enter 'running' state...
INFO:toil.provisioners.aws.awsProvisioner:... toil_leader is running
INFO:toil.provisioners.aws.awsProvisioner:Waiting for ip...
INFO:toil.provisioners.aws.awsProvisioner:...got ip
INFO:toil.provisioners.aws.awsProvisioner:Waiting for ssh port to open...
INFO:toil.provisioners.aws.awsProvisioner:...ssh port open
INFO:toil.provisioners.aws.awsProvisioner:Attempting to establish SSH connection...
Warning: Permanently added 'ec2-54-149-8-63.us-west-2.compute.amazonaws.com,54.149.8.63' (RSA) to the list of known hosts.
[...]

Issue #1707, opened 2017-06-08
Check if docker pull is not working in provisioner when launching a cluster.


ejacox:
I put in the wrong docker address into TOIL_APPLIANCE_SELF and then used "toil ssh-cluster ..."  (from the autoTest/local.sh script). The provisioner hung waiting for the docker to launch.

ejacox:
Related to #1697 and #1688.

In awsProvisioner.py, _waitForAppliance() waits for the docker appliance to start on the leader. It does this by calling 'docker ps' on the leader. However, if the docker image doesn't exist, it will wait forever.  

Before launching an instance or a cluster, Toil should check if the image exists. Unfortunately, there is no standard way to do this.

https://stackoverflow.com/questions/32113330/check-if-imagetag-combination-already-exists-on-docker-hub

Issue #1697, opened 2017-05-31
'create-cluster' fails without hand-specifying coreos and toil appliance images


glennhickey:
Running the example command from the docs fails (error at bottom)

`toil launch-cluster cluster-name --nodeType=t2.micro -z us-west-2a --keyPairName=glenn.hickey@gmail.com`

@joelarmstrong suggested this fix, which seems to help:
prefix with `TOIL_AWS_AMI=ami-f4d4b694` (from https://stable.release.core-os.net/amd64-usr/current/coreos_production_ami_hvm_us-west-2.txt) as apparently wherever toil's getting coreos images from isn't always up to date. 

This causes the command above to loop forever.  This can be fixed by specifying a quay url that exists (more thanks to joel): 

`TOIL_APPLIANCE_SELF="quay.io/ucsc_cgl/toil:3.8.0a1.dev392-14bc15847c2c187431beae2ecb77649ff124b124" TOIL_AWS_AMI=ami-f4d4b694 toil launch-cluster cluster-name6 --nodeType=t2.micro -z us-west-2a --keyPairName=glenn.hickey@gmail.com`

pip install toil
toil --version
3.7.0

boreale 2017-05-31 09:49:58,732 MainThread INFO toil.lib.bioio: Root logger is at level 'INFO', 'toil' logger at level 'INFO'.
boreale 2017-05-31 09:49:58,732 MainThread INFO toil.lib.bioio: Root logger is at level 'INFO', 'toil' logger at level 'INFO'.
boreale 2017-05-31 09:49:58,732 MainThread INFO toil.utils.toilLaunchCluster: Using aws provisioner.
boreale 2017-05-31 09:50:01,416 MainThread INFO toil: Using default docker registry of quay.io/ucsc_cgl as TOIL_DOCKER_REGISTRY is not set.
boreale 2017-05-31 09:50:01,416 MainThread INFO toil: Using default docker name of toil as TOIL_DOCKER_NAME is not set.
boreale 2017-05-31 09:50:01,417 MainThread INFO toil: Using default docker appliance of quay.io/ucsc_cgl/toil:3.7.0-73c62d7698bb6f4154d276994df4353ce4d7826b as TOIL_APPLIANCE_SELF is not set.
boreale 2017-05-31 09:50:01,417 MainThread INFO toil.provisioners.aws.awsProvisioner: Launching non-preemptable leader
Traceback (most recent call last):
[...]

PR #2142, opened 2018-03-20
Check if docker image exists before sallying forth.  Addresses #2126.

DailyDreaming:
Addresses #2126, but also apparently #1707, #1697, #1688 and #1831 after digging a bit.


ejacox:
Is there a better way to do this? This could take a long time if the connection is slow.

joelarmstrong:
This is sorely needed, *but* like Edwin mentioned, I'm a little bit hesitant about using pull to do this. (I don't really like that it has the side effect of messing with the user's existing docker images. Plus, what if someone doesn't have Docker installed?) Sadly it doesn't look like the Docker python library has this functionality directly.

IMO we should use the [HTTP API](https://docs.docker.com/registry/spec/api/), which is fairly well documented and reasonably stable. We already include `requests` in the dependencies, right? So HTTP should be relatively painless. I think we want to hit the endpoint that describes "manifests", though I haven't looked at it in detail yet.

joelarmstrong:
Looks like `HEAD /v2/path/to/image/manifests/tag` should work. You get a 200 on existence, 404 on non-existence:

```
$ http head https://quay.io/v2/ucsc_cgl/toil/manifests/3.14.0
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 17760
Content-Type: application/vnd.docker.distribution.manifest.v1+prettyjws
Date: Tue, 20 Mar 2018 18:31:36 GMT
Docker-Content-Digest: sha256:36d0858e8fea5bc6a42a776e46245572eab2ad0627b414e3a20fe8bc655ce681
Server: nginx/1.13.6
Strict-Transport-Security: max-age=63072000; preload
X-Frame-Options: DENY



$ http head https://quay.io/v2/ucsc_cgl/toil/manifests/foobar
HTTP/1.1 404 NOT FOUND
Connection: keep-alive
Content-Length: 82
Content-Type: application/json
Date: Tue, 20 Mar 2018 18:31:42 GMT
Server: nginx/1.13.6
```

DailyDreaming:
Maybe I have an unfounded bias in the docker API, but I'd like to argue for its virtues.  XD

I've changed it so that it fetches the list of docker images in the registry specified, and then checks against existing tags.  It will throw separate errors now if either does not match and it should all happen without pulling an image (so near instantly on my machine).  Apologies for the naive approach earlier.

DailyDreaming:
Weird, they pass on my machine.  >.<  Will handle in the morning.

DailyDreaming:
Guessing this is Jenkins using an older docker version.

DailyDreaming:
Okay, I think this is done.  Sorry I delved a little too deep with the docker API.  I've torn out the docker API stuff in the interest of portability.  I think it's best just to support `quay.io` and to use Joel's suggestion to use `requests` (definitely seems to be the best solution).  Right now I have it so that it will reject any repo that isn't `quay.io`, but this can be overridden if a power user really wants to.  I included unittests and documentation.

DailyDreaming:
Okay, I think this is done.  Sorry I delved a little too deep with the docker API.  I've torn out the docker API stuff in the interest of portability.  I think it's best just to support `quay.io` and to use Joel's suggestion to use `requests` (definitely seems to be the best solution).  Right now I have it so that it will reject any repo that isn't `quay.io`, but this can be overridden if a power user really wants to.  I included unittests and documentation.

DailyDreaming:
Alright, done.  Checks `quay.io` and `docker.io` using the v2 schema with requests.head.  `docker.io` needs an OAuth token, but it seems to work fine.

Makefile
```diff
@@ -99,19 +99,13 @@ sdist_name:=toil-$(dist_version).tar.gz
 docker_tag:=$(shell $(python) version_template.py dockerTag)
 default_docker_registry:=$(shell $(python) version_template.py dockerRegistry)
 docker_path:=$(strip $(shell which docker))
-ifdef docker_path
```
ejacox: Why was this check removed? It seems necessary.

Makefile
```diff
@@ -99,19 +99,13 @@ sdist_name:=toil-$(dist_version).tar.gz
...
+
+ifdef docker_registry
+    export TOIL_DOCKER_REGISTRY?=$(docker_registry)
 else
-    $(warning Cannot find 'docker' executable. Docker-related targets will be skipped.)
-    export TOIL_DOCKER_REGISTRY:=
+    export TOIL_DOCKER_REGISTRY?=$(default_docker_registry)
 endif
 export TOIL_DOCKER_NAME?=$(shell $(python) version_template.py dockerName)
-# Note that setting TOIL_DOCKER_REGISTRY to an empty string yields an invalid TOIL_APPLIANCE_SELF
```
ejacox: This seems like a useful comment. Why remove it?

docs/contributing/contributing.rst
```diff
@@ -186,6 +186,16 @@ This is described in the next section.
 Making Your Own Toil Docker Image
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+**Note!**  Toil only supports checking if quay.io and docker.io images exist.  To use 
+an image not hosted on docker.io or quay.io, you will have to specifically override this 
+by specifying ``TOIL_APPLIANCE_SELF`` with the ``[override]`` modifier (though be warned, 
+Toil may loop forever when launching a cluster if this docker image does not, in fact,
```
ejacox: It will no longer loop forever, but will fail to launch.

src/toil/__init__.py
```diff
@@ -145,10 +147,116 @@ def lookupEnvVar(name, envName, defaultValue):
...
+    :return: Raises an exception if the docker image cannot be found or is invalid.  Otherwise, it
+             will return the appliance string.
+    :rtype: str
+    """
+    appliance = appliance.lower()
+    tag = appliance.split(':')[-1]
+    registry_name = appliance[:-(len(':' + tag))] # remove only the tag
+
+    # jenkins tests
+    if ('quay.io/ucsc_cgl/toil-pr' in registry_name) or ('quay.io/ucsc_cgl/toil-it' in registry_name):
```
ejacox: Why don't the tests use this mechanism?

src/toil/__init__.py
```diff
@@ -145,10 +147,116 @@ def lookupEnvVar(name, envName, defaultValue):
...
+        path_name = path_name[:-1]
+    quay_url = 'https://quay.io/v2/{path_name}/manifests/{tag}'.format(path_name=path_name, tag=tag)
+    response = requests.head(quay_url)
+    if not response.ok:
+        raise ImageNotFound("TOIL_APPLIANCE_SELF only supports quay.io and official docker.io "
+                            "hosted images.  The image: %s is unsupported (or malformed).  Please "
+                            "supply a docker image with the format: 'quay.io/<repo_path>:<tag>' or "
+                            "'<repo_path>:<tag>'.  Examples: 'quay.io/ucsc_cgl/toil:latest', "
+                            "'ubuntu:latest', or 'broadinstitute/genomes-in-the-cloud:2.0.0'."
+                            "" % registry_name + ':' + tag)
```
ejacox: Should this message include the override option?

Makefile
DailyDreaming: This checks if docker is installed for the `@needs_appliance` decorator to disable docker tests.  But if you look at the def, it already does that in `def needs_appliance`.  So first of all, it's redundant and unnecessary.

Second, its purpose is to set `TOIL_APPLIANCE_SELF` to `None` if docker is not installed.  Since this checks that `TOIL_APPLIANCE_SELF` is valid, this is painfully unhelpful.

Makefile
DailyDreaming: I tore this out because, as above, it serves no purpose anymore and is setting `TOIL_APPLIANCE_SELF` to an invalid string on purpose as a signal to skip docker tests (redundantly, so it doesn't actually need to do this to skip docker tests).  

Since this checks that `TOIL_APPLIANCE_SELF` is valid, this is painfully unhelpful.

docs/contributing/contributing.rst
DailyDreaming: Alright, I can modify this.

src/toil/__init__.py
DailyDreaming: I don't catch your meaning.  Jenkins sets `TOIL_APPLIANCE_SELF` to `quay.io/ucsc_cgl/toil-pr` and `quay.io/ucsc_cgl/toil-it`, which are both invalid, on purpose.  Since this checks for valid docker images, it needs to skip these or jenkins will have a conniption.

src/toil/__init__.py
DailyDreaming: I'd rather have users make a `quay.io` or `docker.io` account first, rather than try to use `[override]` first, and complain.  They can dig in the docs if they're really bent on some strange docker image, but otherwise they should use the canonical path.  I think it will create fewer complaints for us.  Mostly people should not use `[override]` (or have a reason to).

ejacox:
Jenkins, test this please.

DailyDreaming:
Flaky tests?
- src.toil.test.src.fileStoreTest.CachingFileStoreTestWithGoogleJobStore.testCacheEvictionFailCase
- src.toil.test.jobStores.jobStoreTest.EncryptedAzureJobStoreTest.testImportFtpFile
- http://jenkins.cgcloud.info/job/toil-pull-requests/3884

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
This should be ready.

PR #2152, opened 2018-03-23
Provisioner class reorganization.

ejacox:
This is the first part of a provisioner refactor. The class structure is simplified in this step. The next step will refactor within the classes.


benedictpaten:
This looks like an impressive reduction in the amount of code, did this delete a lot of duplicated code? It's a little hard to tell from the big diffs.

ejacox:
Yes. Some of the methods in the Node class were duplicated.

Issue #2149, opened 2018-03-22
ec2_instance_types does not include new p3 instances


Chasearmer:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-243)
┆Issue Number: TOIL-243

Issue #2135, opened 2018-03-16
nodeType validation


bencvdb:
Toil does not validate that what is specified for --nodeType. Attempting to run Cactus using the GCE provisioner and specifying c3.4xlarge(an AWS instance type) for --nodeType yielded the following error: 

Traceback (most recent call last):
  File "/cact_venv/bin/cactus", line 11, in <module>
    load_entry_point('progressiveCactus==1.0', 'console_scripts', 'cactus')()
  File "/cact_venv/local/lib/python2.7/site-packages/cactus/progressive/cactus_progressive.py", line 485, in main
    halID = toil.start(RunCactusPreprocessorThenProgressiveDown(options, project, memory=configWrapper.getDefaultMemory()))
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 744, in start
    self._setProvisioner()
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 794, in _setProvisioner
    self._provisioner = GCEProvisioner(self.config)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/gceProvisioner.py", line 243, in __init__
    self.nonPreemptableNodeShapes = [self.getNodeShape(nodeType=nodeType, preemptable=False) for nodeType in self.nonPreemptableNodeTypes]
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/gceProvisioner.py", line 376, in getNodeShape
    assert len(sizes) == 1
AssertionError

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-13)
┆Issue Number: TOIL-13

PR #2153, opened 2018-03-25
Dynamic instanceType fetching.  Addresses #2149.

DailyDreaming:
Addresses #2149 and #2135.

Still adding to this, but this should allow us to fetch EC2 instance price/specs/types programatically now.

To Do:
- ~Check user input for poorly specified instance types~
- ~Create better static list and region specific diff lists~
- Unittests


ejacox:
Why is cost in the list? Is it used anywhere? Getting the instance types should be fast. What is this huge file to download?

DailyDreaming:
Cost is there to sort by cost in the clusterScaler, which hasn't been refactored yet. We use resource requirements as a proxy right now but it should be cost.

The API fetches a 220Mb+ json. I agree the API should be better but I don't see any way around it. That's why I made it an option to fetch the latest info rather than pulling it. There are comments in the code explaining it.

ejacox:
I don't see cost in #2114. What refactor?

DailyDreaming:
#2114 doesn't sort by price because it needs this PR's ability to fetch price.  This PR doesn't use cost because changing the clusterScaler would interfere with #2114.

I felt this was separate enough to warrant its own PR.  Otherwise, this would have been a part of #2114.

I'd already eliminated all of the unnecessary fields from cgcloud that we never used to begin with.  I'd like to argue for having cost in there because now that we can access it easily, it seems like the most important field for running workflows cheaply.  #2114 does a priority sort by resources (the instance with the lowest memory that can run the job is the one that's used), but this is kind of a proxy, because we assume low memory == low cost.  That's not always true though.  With this new field, that would be relatively straight forward to change.

DailyDreaming:
I'd like to argue for keeping cost in and adding the sort by cost after these are both merged.  I think it's worth doing.

ejacox:
Can you please put the cost in a different PR. I need this PR without the cost.

DailyDreaming:
Alright, no problem.

src/toil/common.py
```diff
@@ -575,6 +582,38 @@ def addGroup(headingString, bodyString):
         raise RuntimeError("Unanticipated class passed to addOptions(), %s. Expecting "
                            "argparse.ArgumentParser" % parser.__class__)
 
+def checkValidNodeTypes(provisioner, nodeTypes):
```
ejacox: Azure and GCE already do this. This method should be in the AWSProvisioner.

src/toil/common.py
```diff
@@ -378,10 +380,15 @@ def _addOptions(addGroupFn, config):
                       "new nodes once the maximum number of preemptable-nodes has been"
                       "reached.")
 
+    addOptionFn('--useLatestNodeTypes', default=False,
```
ejacox: Instance types should always be loaded dynamically.

src/toil/lib/generatedEC2Lists.py
```diff
@@ -0,0 +1,1329 @@
+# Copyright (C) 2015-2018 UCSC Computational Genomics Lab
+#
```
ejacox: Avoid checking in generated code.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -297,7 +313,15 @@ def terminateNodes(self, nodes):
         self._terminateNodes(nodes, self.ctx)
 
     def addNodes(self, nodeType, numNodes, preemptable):
-        instanceType = ec2_instance_types[nodeType]
+        if self.ec2_instance_types:
```
ejacox: This code is duplicated three times. It should go in a function, e.g. getInstanceType.

src/toil/common.py
DailyDreaming: It's required to fetch a 220MiB+ JSON file.  On campus ethernet, that's a couple minute delay, which isn't bad.  But on a slower internet connection, that could be much longer.  If you're very opposed to this, then maybe two other ideas?

1. Just stick with the static lists, and update every release as a PR.  Provide a user command if the user wishes to update Toil's lists internally (this function already exists).

or

2. Serve our own, much smaller, json from a server.  AWS has some kind of notification system when changes occur so that we'd know when to update, though I haven't looked into it much.

src/toil/provisioners/aws/awsProvisioner.py
DailyDreaming: Twice, but good point!  Will fix.

src/toil/common.py
DailyDreaming: GCE doesn't seem to: #2135

PR #2155, opened 2018-03-26
Add option not to generate a master public key (resolves #2154)

### LINKED ISSUE #2154: Add option not to generate a master public key for the AWS provisioner
Add an option to Toil to not create a master public key for the provisioner when using autoscaling and running the pipeline from Dockstore (which calls cwl-runner) where the pipeline is running as the leader node, because cwl-runner locks the file system on the leader and creation of an ssh key is not possible. If the provisioner attempts to create the ssh key an exception will be thrown.

wshands:
Add an option to Toil to not create a master public key for the
provisioner when using autoscaling and running the pipeline from
Dockstore (which calls cwl-runner) where the pipeline is running
as the leader node, because cwl-runner locks the file system on
the leader and creation of an ssh key is not possible. If the
provisioner attempts to create the ssh key an exception will be thrown.
Resolves: #2154

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -99,7 +99,8 @@ class Config(object):
         self.preemptableCompensation = 0.0
         self.nodeStorage = 50
         self.metrics = False
-        
+        self.noMasterPublicKey = False
+
         # Parameters to limit service jobs, so preventing deadlock scheduling scenarios
         self.maxPreemptableServiceJobs = sys.maxsize
         self.maxServiceJobs = sys.maxsize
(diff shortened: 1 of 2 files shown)


ejacox:
The masterPublicKey is needed when the leader needs to send files to the workers. On AWS, this is currently needed to enable server side encryption of file streams. We are currently passing credentials for Azure and Google to the workers, so it can't be disabled.

Would it be easier to modify the cwl-runner?

DailyDreaming:
Jenkins, test this please.

wshands:
I was able to make the root folder writable by using a VOLUME command in the pipeline Dockerfile so I am closing this pull request

PR #2156, opened 2018-03-26
Added encryption test.

ejacox:
Added a test for shared file encryption. AWS and Google use server side and Azure uses client side.


ejacox:
Ready for review.

Issue #1503, opened 2017-02-10
deadlock due to cPickle not releasing GIL (again!)


joelarmstrong:
A long time ago I tracked down a toil deadlock caused by attempting to unpickle a file as it is streamed from the job store. If a pickle file isn't created with cPickle.HIGHEST_PROTOCOL, cPickle (incorrectly) fails to release the GIL before starting a read. If the filestream buffer runs out during that read, the process will block forever, because the Python thread on the other end can't wake up to fill the buffer. We fixed this by always dumping with HIGHEST_PROTOCOL.

But just now, I found a deadlocked worker, and guess what it's doing?

```
gdb) bt
#0  0x00007fcdc5af26cd in read () at ../sysdeps/unix/syscall-template.S:81
#1  0x00007fcdc5a815a0 in _IO_new_file_underflow (fp=0x2791f80) at fileops.c:613
#2  0x00007fcdc5a8252e in __GI__IO_default_uflow (fp=0x2791f80) at genops.c:435
#3  0x00007fcdc5a78ace in _IO_getc (fp=0x2791f80) at getc.c:39
#4  0x000000000054177c in readline_file.57371 (self=0x7fcdc066ac50, s=0x7ffc94a13220) at ../Modules/cPickle.c:591
#5  0x000000000047a014 in load_global (self=0x7fcdc066ac50) at ../Modules/cPickle.c:4034
#6  load (self=0x7fcdc066ac50) at ../Modules/cPickle.c:4860
#7  0x0000000000524cdc in call_function (oparg=<optimized out>, pp_stack=0x7ffc94a13360) at ../Python/ceval.c:4004
```
```
(gdb) py-bt
[...]

PR #2157, opened 2018-03-26
Add a more robust deleter to the file job store

adamnovak:
This will hopefully fix #1927, or at least make it less likely to happen. If Toil actually does have a file in the job store open while the job store is being deleted (and it's not just hitting a race condition against the NFS silly rename file cleanup) then it's not really possible for Toil to remove job stores on NFS, and we'll have to find where/why Toil has files open and make it close them.

PR #2174, opened 2018-04-12
Fix cPickle deadlock when unpickling Job objects

joelarmstrong:
This is relevant to earlier issues like #1503. The root cause is that cPickle sometimes doesn't release the GIL when reading. When reading using the `read_file` function, cPickle correctly releases the GIL before performing a potentially-blocking operation:

https://github.com/python/cpython/blob/a61f5da54772b0ea6a7eccf21df08e61585ef712/Modules/cPickle.c#L552

But in the `readline_file` function, it doesn't:

https://github.com/python/cpython/blob/a61f5da54772b0ea6a7eccf21df08e61585ef712/Modules/cPickle.c#L591

This means that anytime the "file-like-object" that cPickle is reading from is a pipe filled from a different Python thread (i.e. in remote jobStores), we have to avoid using `readline_file` at all costs. Otherwise, the unpickling thread will stall waiting for data, but the thread that fills the other end of the pipe will never run, because the GIL is still held by the reading thread. We originally thought we worked around this by enforcing `cPickle.HIGHEST_PROTOCOL` on all pickle dumps. Pickle's protocol version 2 dumps most things in a binary format and not a line-based one, which means `readline_file` is almost never used (we thought it was never used). But unfortunately, `load_global` opcodes always cause `readline_file` to be used, even in the higher protocol versions:

https://github.com/python/cpython/blob/a61f5da54772b0ea6a7eccf21df08e61585ef712/Modules/cPickle.c#L4039

So sometimes, if you're unlucky, a load_global opcode will happen to be right on the end of the pipe-buffer (I believe that's 4KB on Linux), and there's a race between one thread reading past the end of the pipe, and the other writing more data into it. In my tests, this race gets lost roughly once in every 10K to 100K jobs for me, but it may depend heavily on the size of your Job objects and even the length of the names of the files/classes you use.

Basically, this PR solves the problem by never actually streaming the file when unpickling. Instead, it's fed into a special function which just reads all the data and calls `pickle.loads` on it, which can't stall because all the data is already in memory. (Why not avoid using readFileStream entirely and just use readFile?: Because this way we don't have to worry about cleaning up temporary files if there's an unpickling error.)

Unpickling the job objects is a special case, because it uses a custom unpickler, which doesn't have the `loads` method: in that case, we just read it into a temporary file and delete it after we're done or an exception is raised.

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -932,7 +932,7 @@ class Toil(object):
             from toil.jobStores.abstractJobStore import NoSuchFileException
             try:
                 with self._jobStore.readSharedFileStream('userScript') as f:
-                    userScript = pickle.load(f)
+                    userScript = safeUnpickleFromStream(f)
             except NoSuchFileException:
                 logger.info('User script neither set explicitly nor present in the job store.')
                 userScript = None
@@ -1323,3 +1323,7 @@ def getFileSystemSize(dirPath):
     freeSpace = diskStats.f_frsize * diskStats.f_bavail
     diskSize = diskStats.f_frsize * diskStats.f_blocks
     return freeSpace, diskSize
+
+def safeUnpickleFromStream(stream):
+    string = stream.read()
+    return pickle.loads(string)
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -25,9 +25,7 @@ import importlib
 import inspect
 import logging
 import os
-import sys
 import time
-import uuid
 import dill
 import tempfile
 
@@ -47,7 +45,7 @@ from six import iteritems, string_types
 from bd2k.util.expando import Expando
 from bd2k.util.humanize import human2bytes
 
-from toil.common import Toil, addOptions
+from toil.common import Toil, addOptions, safeUnpickleFromStream
 from toil.fileStore import DeferredFunction
 from toil.lib.bioio import (setLoggingFromOptions,
                             getTotalCpuTimeAndMemoryUsage,
@@ -1750,7 +1748,7 @@ class Promise(object):
         with cls._jobstore.readFileStream(jobStoreFileID) as fileHandle:
             # If this doesn't work then the file containing the promise may not exist or be
             # corrupted
-            value = pickle.load(fileHandle)
+            value = safeUnpickleFromStream(fileHandle)
             return value
 
 
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -29,6 +29,7 @@ import sys
 import time
 import uuid
 import dill
+import tempfile
 
 try:
     import cPickle as pickle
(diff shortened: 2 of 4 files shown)


DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Thanks so much for getting this in @joelarmstrong !  It looks pretty clean.  The failing test is a bit odd (unable to find the s3 bucket?) and I haven't seen it before.  Not sure if it's a random fail or not.  Assuming tests pass, this looks ready to merge.

Also, thanks for the clear explanation of the problem with links.  Was hoping to browse python 3 and find that it properly released the GIL, but it seems pretty similar to the python 2.7 code: https://github.com/python/cpython/blob/master/Modules/_pickle.c#L1352 .  This looks like a great solution, thanks for submitting it!

joelarmstrong:
Are these remaining failing tests known to be flaky? It looks like the static Azure test may have failed because the container took too long to start. (You guys might be getting throttled by Quay, I've run into similar issues). Not sure what's going on with the CWL conformance test.

Yeah, sorry for the verbosity. I just wanted to dump out all that information so other people know what to watch out for: I think no one on the current toil team was around when we first hit this bug. Yeah, sadly it was never fixed in any version of Python, as far as I can tell. (Honestly, I don't think anyone ever ran into it except us.)

DailyDreaming:
The azure autoscaling test may be flaky.  Azure autoscaling was a little flaky for a time, but hasn't been failing in a while.  The cwl conformance tests have never been flaky since I've been here.

No, it was really valuable.  I didn't realize this had a history to it.  I'm glad you explained it and gave pointers to how the GIL release is controlled in the first function, and the second function lacking that control.

Not sure what's going on with the tests.  Will run again.  I suspect you're right and there may be a bottleneck somewhere causing some timing problem.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
The cwl conformance test oddly failed on #2179 as well, which should be a PR that has no effect at all on cwl performance.  I'm guessing they may have added new tests to the tarball that we fetch.

DailyDreaming:
Jenkins, test this please.

mr-c:
The CWL conformance tests are fetched via hash, so there is no way for use to sneak additional tests in 😁

DailyDreaming:
@mr-c thanks!  I should have seen that.  It looks like the 3 failing cwl tests are all timing out (they all pass fine running it locally though).  A bunch of cloud tests also appear to be timing out too.  I suspect it might be #2157 that was added recently simply taking longer to delete things in general, in which case we should only apply the fix to nfs, rather than deletions in general.  I'll make a PR to test it.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Flaky tests seem to be addressed and this is passing all but the known cwl test failure.  LGTM.

Issue #2181, opened 2018-04-16
Flaky tests.


DailyDreaming:
CWL conformance tests have been timing out towards the end fairly consistently, but seem to pass locally just fine.

Cloud tests have been randomly failing among all three clouds it seems (especially azure though?), also (most commonly) with timeouts.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-16)
┆Issue Number: TOIL-16

Issue #2162, opened 2018-03-29
Toil leaks open files that prevent job store deletion


adamnovak:
With #2157 in I have better visibility into a problem I have been having where the job store can't be cleaned up:

```
leader.py                  485 INFO     Job ended successfully: 'run_map_eval_align' 9/S/jobLrGOF2
leader.py                  546 INFO     Finished the main loop: no jobs left to run
serviceManager.py          156 INFO     Waiting for service manager thread to finish ...
serviceManager.py          163 INFO     ... finished shutting down the service manager. Took 0.198056936264 seconds
statsAndLogging.py         156 INFO     Waiting for stats and logging collator thread to finish ...
statsAndLogging.py         160 INFO     ... finished collating stats and logs. Took 0.173548221588 seconds
leader.py                  240 INFO     Finished toil run successfully
realtimeLogger.py          175 INFO     Stopping real-time logging server.
realtimeLogger.py          179 INFO     Joining real-time logging server thread.
fileJobStore.py            159 WARNING  Cannot delete /cluster/home/anovak/build/vg/vgci-work/jobstoresim-MHC-cactus/tmp/l/T; retrying in 1 seconds
fileJobStore.py            139 WARNING  Open NFS files like /cluster/home/anovak/build/vg/vgci-work/jobstoresim-MHC-cactus/tmp/l/T/.nfs0000000000b626210007d20b may prevent job or job store deletion
fileJobStore.py            159 WARNING  Cannot delete /cluster/home/anovak/build/vg/vgci-work/jobstoresim-MHC-cactus/tmp/l/T; retrying in 2 seconds
fileJobStore.py            159 WARNING  Cannot delete /cluster/home/anovak/build/vg/vgci-work/jobstoresim-MHC-cactus/tmp/l; retrying in 1 seconds
fileJobStore.py            159 WARNING  Cannot delete /cluster/home/anovak/build/vg/vgci-work/jobstoresim-MHC-cactus/tmp/l/T; retrying in 1 seconds
...
common.py                  706 INFO     Failed to delete the job store: <toil.jobStores.fileJobStore.FileJobStore object at 0x7f3ce16a8610>
```

The "..." elides a bunch of looping and retrying the delete (it seems that each level of the recursion gets its own retries, which I didn't think about when I wrote #2157 and which may cause delays).

[...]

PR #2182, opened 2018-04-16
Fix failing flaky tests.  Addresses #2181.

DailyDreaming:
Addresses #2181 .

CWL conformance tests have been timing out towards the end, and seem to work locally.  This is likely due to recently updating these tests, and so I've increased the timeout by another 5 minutes.

Cloud tests have been randomly failing among all three clouds it seems, also (most commonly) with timeouts, and this is likely due to #2157.  I've reverted back to the standard method for removing directories before trying the potentially time consuming custom deletion function written to address mostly nfs specific issues.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -35,7 +35,6 @@ except ImportError:
     import pickle
 
 # toil and bd2k dependencies
-from bd2k.util.files import rm_f
 from toil.fileStore import FileID
 from toil.lib.bioio import absSymPath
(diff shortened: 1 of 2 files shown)


adamnovak:
For referenc,e the real problem we have is https://github.com/BD2KGenomics/toil/issues/2162, and #2157 didn't really do anything other than slow stuff down and produce more descriptive errors.

adamnovak:
Mashing rmtree a few times is probably just as good as the manual recursive delete, and easier to reason about the runtime of.

DailyDreaming:
Jenkins, test this please.

PR #2194, opened 2018-04-26
Require zone arg if no default set (fixes #2191)

### LINKED ISSUE #2191: Azure provisioner launch-cluster fails unclearly when `--zone` is not specified
When the `--zone` argument isn't provided to `toil launch-cluster`, Toil exits with an unclear traceback:

```
$ toil launch-cluster -p azure --leaderNodeType Standard_A0 --keyPairName natanlao@users.noreply.github.com natanlao-test-cluster
...
  File "/home/natan/toil/toil/bin/toil", line 11, in <module>
    load_entry_point('toil', 'console_scripts', 'toil')()
  File "/home/natan/toil/src/toil/utils/toilMain.py", line 34, in main
    module.main()
  File "/home/natan/toil/src/toil/utils/toilLaunchCluster.py", line 133, in main
    azureStorageCredentials=config.azureStorageCredentials)
  File "/home/natan/toil/src/toil/provisioners/azure/azureProvisioner.py", line 139, in launchCluster
    self.callPlaybook(self.playbook['create-cluster'], ansibleArgs, wait=True)
  File "/home/natan/toil/src/toil/provisioners/ansibleDriver.py", line 43, in callPlaybook
    command.append(" ".join(["=".join(i) for i in ansibleArgs.items()]))  # Arguments being passed to playbook
TypeError: sequence item 1: expected string, NoneType found
```

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-14)
┆Issue Number: TOIL-14

natanlao:


natanlao:
On second thought, this doesn't really fix the problem (or do anything of use) and this kind of problem should be fixed at argument-parsing time - something that I think would be better addressed with the provisioner refactor.

PR #2204, opened 2018-05-10
Investigate cause of CWL conformance timeout

joelarmstrong:
The CWL conformance tests keep failing due to timeouts (though they should be very quick), and in the back of my mind I've sort of been wondering why. This PR is just meant to get Jenkins to run and output *actual* debug information. Please don't merge this as-is, because right now it breaks the cwltoil `--quiet` flag.

I suspect it's something to do with the singleMachine batch system running out of resources (disk space?) and deadlocking. But we'll see: these commits will show the toil debug output so we can see what's happening in the timed-out tests.


joelarmstrong:
Trying to get Jenkins to run this properly. (maybe I'm not on the whitelist anymore?)

Jenkins, test this please.

PR #2226, opened 2018-05-22
Update subprocess32 to 3.5.1.

DailyDreaming:


DailyDreaming:
True.  cwltool now requires it though, so it has the additional benefit of reestablishing compatibility with the latest version of cwltool.

Issue #2230, opened 2018-06-01
Version-pinned, outdated boto3 dependency causes problems when installing toil alongside awscli


adamnovak:
The fundamental problem here is that pip can't solve dependency constraints and select a pair of toil and awscli versions that agree on what version of botocore needs to be installed (see https://github.com/pypa/pip/issues/988).

You ought to be able to do:
```
pip install toil[aws,mesos] awscli
```

But this doesn't work, because it tries to install the current version of both packages. Toil depends on `boto3==1.4.7`:
https://github.com/BD2KGenomics/toil/blob/e14b41e7552e2aa7c161e01f93b8260299590e2d/setup.py#L24

Which in turn depends on `botocore>=1.7.0,<1.8.0`:
https://github.com/boto/boto3/blob/9d5fac2a36a31b9665a8d80c7dfe8df9fac4ceb3/setup.py#L17

But the current awscli depends on `botocore==1.10.30`:
https://github.com/aws/aws-cli/blob/9b781a5796fc950a295cc2b000f640dcdc8cd6ab/setup.py#L26

Which in turn can only be used by Boto3 1.7.0 or higher, but which will stop working when Boto3 1.8.0 comes out (since they keep the minor versions in sync):
https://github.com/boto/boto3/blob/6374e82f1d5879c4abef3873606812cd838b68e3/setup.py#L17

So there's no version of botocore that can be installed to make both Toil and awscli happy. And having one tool or the other trying to use the wrong version has caused me mysterious problems in the past; botocore is an internal module for boto/AWS things and is not very cross-compatible.

If we can't let the Toil boto3 version just float (`boto3>=1.4.7` or maybe `boto3>=1.4.7,<2.0`), Toil needs to keep up to date with Boto3 releases, so that it can be installed alongside whatever the current awscli is without problems.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-261)
┆Issue Number: TOIL-261

adamnovak:
This is particularly troublesome because if you get this configuration:

```
awscli                       1.14.70
boto                         2.38.0   
boto3                        1.6.7    
botocore                     1.9.23   
toil                         3.16.0
```
Then you can be in a situation where toil can create clusters but it cannot destroy them:

```
toil destroy-cluster 'wgcluster' -z us-west-2a
INFO:toil.lib.bioio:Root logger is at level 'INFO', 'toil' logger at level 'INFO'.
[...]

PR #2233, opened 2018-06-06
Update boto3.  Addresses #2230.

DailyDreaming:
Addresses #2230.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ def runSetup():
     functionality. The `version` module is imported dynamically by importVersion() below.
     """
     boto = 'boto==2.38.0'
-    boto3 = 'boto3==1.4.7'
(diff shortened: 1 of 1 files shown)


DailyDreaming:
@adamnovak Does this work for you?

DailyDreaming:
Hmmm... it looks like boto3 is only used in one function ( https://github.com/BD2KGenomics/toil/blob/45de1029ae190b8e0bb3a4addcc7ebc7bec69173/src/toil/jobStores/aws/utils.py#L258 ).  I think this is safe to merge as all the tests are passing except for two known flakes (http://jenkins.cgcloud.info/job/toil-pull-requests/4103/).  I'll make an issue to write a test for it explicitly later.

PR #2247, opened 2018-06-12
Report why a Docker manifest might be unavailable

adamnovak:
Right now, if you give a non-working Docker container specifier as TOIL_APPLIANCE_SELF, you get no feedback. This adds code to show you the manifest URL that didn't work and the HTTP status code that describes the problem (for example, 401 if you forgot to make your Quay repo public).

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -184,29 +184,29 @@ def checkDockerImageExists(appliance):
         return requestCheckDockerIo(registryName=registryName, tag=tag)
 
 
-def makeApplianceImageNotFound(name, tag, url, statusCode):
+class ApplianceImageNotFound(ImageNotFound):
     """
-    Compose an ImageNotFound error complaining that the given name and tag
-    for TOIL_APPLIANCE_SELF specify an image manifest which could not be
+    Compose an ApplianceImageNotFound error complaining that the given name and
+    tag for TOIL_APPLIANCE_SELF specify an image manifest which could not be
     retrieved from the given URL, because it produced the given HTTP error
     code.
     
-    :param str registryName: The url of a docker image's registry.  e.g. "quay.io/ucsc_cgl/toil"
+    :param str name: The name of a docker image.  e.g. "quay.io/ucsc_cgl/toil"
     :param str tag: The tag used at that docker image's registry.  e.g. "latest"
-    :param str url: The URL at which the image's manifewst is supposed to appear
+    :param str url: The URL at which the image's manifest is supposed to appear
     :param int statusCode: the failing HTTP status code returned by the URL
-    :return: Return an ImageNotFound error containing an informative message.
     """
-    
-    return ImageNotFound("The docker image that TOIL_APPLIANCE_SELF specifies (%s) produced "
-                         "a nonfunctional manifest URL (%s). The HTTP status returned was %s. "
-                         "The specifier is most likely unsupported or malformed.  "
-                         "Please supply a docker image with the "
-                         "format: '<websitehost>.io/<repo_path>:<tag>' or '<repo_path>:<tag>' "
-                         "(for official docker.io images).  Examples: "
-                         "'quay.io/ucsc_cgl/toil:latest', 'ubuntu:latest', or "
-                         "'broadinstitute/genomes-in-the-cloud:2.0.0'."
-                         "" % (name + ':' + tag, url, str(statusCode)))
+    def __init__(self, name, tag, url, statusCode):
+        msg = ("The docker image that TOIL_APPLIANCE_SELF specifies (%s) produced "
+               "a nonfunctional manifest URL (%s). The HTTP status returned was %s. "
+               "The specifier is most likely unsupported or malformed.  "
+               "Please supply a docker image with the format: "
+               "'<websitehost>.io/<repo_path>:<tag>' or '<repo_path>:<tag>' "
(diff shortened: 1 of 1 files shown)


DailyDreaming:
This should be done for the other checking function as well (which is separate because the official docker.io site for some reason needs oauth). https://github.com/BD2KGenomics/toil/blob/45de1029ae190b8e0bb3a4addcc7ebc7bec69173/src/toil/__init__.py#L227

adamnovak:
The tests are failing because they couldn't download Docker itself during the build, which I suspect is Not My Fault:

```
Step 10/21 : RUN curl https://get.docker.com/builds/Linux/x86_64/docker-1.13.1.tgz          | tar -xvzf - --transform='s,[^/]*/,,g' -C /usr/local/bin/          && chmod u+x /usr/local/bin/docker
 ---> Running in 87ac75de6b93
[91m  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0[0m[91m
100   454  100   454    0     0   3562      0 --:--:-- --:--:-- --:--:--  3574
[0m[91m
gzip: stdin: not in gzip format
[0m[91mtar: Child returned status 1
tar: Error is not recoverable: exiting now
[0mThe command '/bin/sh -c curl https://get.docker.com/builds/Linux/x86_64/docker-1.13.1.tgz          | tar -xvzf - --transform='s,[^/]*/,,g' -C /usr/local/bin/          && chmod u+x /usr/local/bin/docker' returned a non-zero code: 2
make: *** [docker] Error 2
```

Someone should say "Jenkins test this please" again I think? That URL works for me right now.

src/toil/__init__.py
```diff
@@ -184,6 +184,30 @@ def checkDockerImageExists(appliance):
         return requestCheckDockerIo(registryName=registryName, tag=tag)
 
 
+def makeApplianceImageNotFound(name, tag, url, statusCode):
```
DailyDreaming: This seems to be a custom error, I'd like to argue that it should inherit `ImageNotFound` following the conventions of other custom errors in Toil.  Something roughly like the following:

```
class ApplianceNotFoundException(ImageNotFound):
    def __init__(self, name, tag, url, statusCode):
        msg = ("The docker image that TOIL_APPLIANCE_SELF specifies (%s) produced "
               "a nonfunctional manifest URL (%s). The HTTP status returned was %s. "
               "The specifier is most likely unsupported or malformed.  "
               "Please supply a docker image with the format: "
               "'<websitehost>.io/<repo_path>:<tag>' or '<repo_path>:<tag>' "
               "(for official docker.io images).  Examples: "
               "'quay.io/ucsc_cgl/toil:latest', 'ubuntu:latest', or "
               "'broadinstitute/genomes-in-the-cloud:2.0.0'."
               "" % (name + ':' + tag, url, str(statusCode)))
        super(ApplianceNotFoundException, self).__init__(msg)
```

DailyDreaming:
Yes, not your fault, just a hiccup with the url it seems.  I made a comment about the format.  Once that's addressed, this should be fine to merge.

src/toil/__init__.py
adamnovak: This isn't a custom error type; it's a function that returns ImageNotFound errors. I just needed one place to put the template string logic that was common to both image checking functions.

adamnovak: Unless you think we really do want a custom error type for exactly this particular case, which always has a message that looks like that? It seems a bit overkill for something that would be thrown in pretty much one situation ever.

DailyDreaming: A class that is raised to return an error is equivalent in function to a custom error.  Two is not "pretty much one", and I don't see having a more explicitly named error as necessarily a bad thing.

DailyDreaming: Something that is raised to return an error seems equivalent in function to a custom error.

Two may be "pretty much one" sometimes, but isn't the fact that it's used in more than one situation the impetus behind writing the function in the first place?

Also, having a more explicitly named error in general isn't necessarily a bad thing.

adamnovak: The type of the error being raised in this case is still ImageNotFound.

The syntax at the point of use is the same between a function call and a new type. If you really want the class version I can do that instead.

DailyDreaming: It's true that it's an ImageNotFound error.  It's also a RuntimeError.  These are varying degrees of specificity.  I think it's fine to keep it as an ImageNotFound error, rather than making it more specific.  A more specific error name was just a suggestion.

The error class, however, is a convention in Toil (and in python in general).  See PEP8 guidelines for creating exceptions:

https://www.python.org/dev/peps/pep-0008/#exception-names

I'm trying to argue for consistency in toil.  I can link all of the exception classes that are consistently written throughout Toil if it will convince you that it's a good idea to follow convention, most of which simply hold a custom string error message just as your function is doing.

adamnovak:
OK, how's this version with the new error type?

DailyDreaming:
Looks good to me.  Thanks for writing it and making the error message more useful.  Will merge when tests pass.

DailyDreaming:
The two failures are known errors ( http://jenkins.cgcloud.info/job/toil-pull-requests/4110 ).  Looks good to merge.

Issue #2237, opened 2018-06-07
Provisioner can't fit jobs it could fit before, and can't explain itself.


adamnovak:
I woke up to an 802 megabyte Toil log this morning. It seems that the provisioner can't fit jobs that used to work with node types of "i3.8xlarge,i3.8xlarge:0.90" into any of "i3.8xlarge,r4.8xlarge:0.60".

The log looks like this.

```
DEBUG:toil.provisioners.clusterScaler:Running bin packing for node shapes [<toil.provisioners.abstractProvisioner.Shape object at 0x7fa0441a3dd0>, <toil.provisioners.abstractProvisioner.Shape object at 0x7fa0441a3f50>] and 35 job(s).
2018-06-06 23:22:07,570 - toil.provisioners.clusterScaler - DEBUG - Running bin packing for node shapes [<toil.provisioners.abstractProvisioner.Shape object at 0x7fa0441a3dd0>, <toil.provisioners.abstractProvisioner.Shape object at 0x7fa0441a3f50>] and 35 job(s).
WARNING:toil.provisioners.clusterScaler:Couldn't fit job with requirements <toil.provisioners.abstractProvisioner.Shape object at 0x7fa044094fd0> into any nodes in the nodeTypes list.
2018-06-06 23:22:07,571 - toil.provisioners.clusterScaler - WARNING - Couldn't fit job with requirements <toil.provisioners.abstractProvisioner.Shape object at 0x7fa044094fd0> into any nodes in the nodeTypes list.
ERROR:toil.provisioners.clusterScaler:Exception encountered in scaler thread. Making a best-effort attempt to keep going, but things may go wrong from now on.
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/clusterScaler.py", line 857, in tryRun
    currentNodeCounts)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/clusterScaler.py", line 460, in getEstimatedNodeCounts
    goalTime=self.targetTime)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/clusterScaler.py", line 288, in binPacking
    bpf.binPack(jobShapes)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/clusterScaler.py", line 85, in binPack
    self.addJobShape(jS)
  File "/usr/local/lib/python2.7/dist-packages/toil/provisioners/clusterScaler.py", line 104, in addJobShape
[...]

PR #2248, opened 2018-06-15
Update instance downloader and redownload instances

adamnovak:
This should solve the last part of #2237. It looks like Amazon previously described i3 instance disk sizes with numbers in TB, which the human-readable-description-parsing code interpreted as GB, leading to tiny estimates of available disk space on i3 nodes.

Amazon has since fixed the issue on their end, so I have rerun the instance type downloader `toil/utils/toilUpdateEC2Instances.py`. Now we have an instance list that properly has all sizes in GB.

I also had to fix the downloader to filter on the "operation" field, because now the same instance type with the same OS ("Linux") can appear multiple times with different operations, depending on whether you want your instance to come with e.g. Microsoft SQL Server.


DailyDreaming:
AWS must have updated their json.  Thanks for updating the downloader.

LGTM.

Issue #2245, opened 2018-06-12
Toil only supports Docker registries hosted in the .io TLD


adamnovak:
https://github.com/BD2KGenomics/toil/blob/45de1029ae190b8e0bb3a4addcc7ebc7bec69173/src/toil/__init__.py#L204-L212

This seems... overly specialized. Just because docker.io and quay.io are in .io doesn't mean *every* registry is going to be in .io.

The spec url of https://docs.docker.com/registry/spec/manifest-v2-2/ given to support this British-Indian-Ocean-Territory-centric restriction doesn't seem to have anything to say about registry domain names; it's all about image manifest files.

There exist real registries in other TLDs; for example, [Gitlab operates a registry](https://docs.gitlab.com/ee/user/project/container_registry.html) at `registry.gitlab.com`.

The relevant code should be rewritten to support registries at arbitrary domain names.

┆Issue Number: TOIL-283

DailyDreaming:
I included the wrong link in the docstring: `https://docs.docker.com/registry/spec/api`.  Fair point though, `gcr.io`, `docker.io`, and `quay.io` were the only ones I saw at the time and I think I conflated the convention with actually being part of the schema.  I'll amend that.

DailyDreaming:
I've rewritten the functions to accommodate more exotic repos, and will create a PR once #2247 is merged.  Specifically, if you have an image in mind, I'd like to test it.

PR #2249, opened 2018-06-16
Check non-io docker images.  Addresses #2245.

DailyDreaming:
Addresses #2245.

This should cover some of the more exotic repositories out there.


DailyDreaming:
Thanks!

PR #2251, opened 2018-06-19
More Python 3 fixes

adamnovak:
@DailyDreaming Here are some fixes I have done on top of your python23compat branch. I "fixed" some problems I ran into getting the tests up and running (such as the Makefile trying to always use 2.7 versions of things), and I've fixed a couple bugs that made tests fail.

I've now hit the `find_global` problem that was mentioned this morning. I can try fixing that, but I wanted to PR this first, into your python23compat pr.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 __pycache__
 .eggs/
 .cache/
+.pytest_cache/
 .idea/
 /test-report.xml
 /test-report-*.xml
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ The 'prepare' target installs Toil's build requirements into the current virtual
 The 'develop' target creates an editable install of Toil and its runtime requirements in the
 current virtualenv. The install is called 'editable' because changes to the source code
 immediately affect the virtualenv. Set the 'extras' variable to ensure that the 'develop' target
-installs support for extras. Consult setup.py for the list of supported extras. To install Toil
-in develop mode with all extras, run
+installs support for extras; some tests require extras to be installed. Consult setup.py for the
+list of supported extras. To install Toil in develop mode with all extras, run
 
 	make develop extras=[all]
 
--- a/Makefile
+++ b/Makefile
@@ -90,8 +90,8 @@ help:
 
 # This Makefile uses bash features like printf and <()
 SHELL=bash
-python=python
-pip=pip
+python=python2.7
+pip=pip2.7
 tests=src
 tests_local=src/toil/test
 # do slightly less than travis timeout of 10 min.
--- a/Makefile
+++ b/Makefile
@@ -267,7 +267,7 @@ check_build_reqs:
 
 
 prepare: check_venv
-	$(pip) install sphinx==1.5.5 mock==1.0.1 pytest==2.8.3 stubserver==1.0.1 \
+	$(pip) install sphinx==1.5.5 mock==1.0.1 pytest==3.6.2 stubserver==1.0.1 \
 		pytest-timeout==1.2.0 cwltest
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
-[pytest]
+[tool:pytest]
 # Look for any python file, the default of test_*.py wouldn't work for us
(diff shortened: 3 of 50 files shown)


DailyDreaming:
I left the Makefile 2.7 requirements in purposely just for the time being, since that meant being able to merge incrementally while keeping it python2.7 consistent.  I took out the other 2.7 gatekeepers that mostly seemed redundant though.

If you're curious, using the python3 Unpickler's `find_class` method as a first naive drop in replacement:

```
        if sys.version_info[0] == 2:
            unpickler.find_global = filter_main
        else:
            unpickler.find_class = filter_main
```

Yields the error:

```
  File "/home/quokka/git/deltoil23/toil3/toil/src/toil/test/src/workerTest.py", line 58, in testNextChainableJobGraph
    self.assertEquals(jobGraph2, nextChainableJobGraph(jobGraph1, self.jobStore))
  File "/home/quokka/git/deltoil23/toil3/toil/src/toil/worker.py", line 98, in nextChainableJobGraph
    successorJob = Job._loadJob(successorJobGraph.command, jobStore)
  File "/home/quokka/git/deltoil23/toil3/toil/src/toil/job.py", line 900, in _loadJob
    return cls._unpickle(userModule, fileHandle, jobStore.config)
  File "/home/quokka/git/deltoil23/toil3/toil/src/toil/job.py", line 930, in _unpickle
    unpickler.find_class = filter_main
AttributeError: '_pickle.Unpickler' object attribute 'find_class' is read-only
```

DailyDreaming:
Also, jenkins is down since we disabled Edwin's account on AWS. I've asked Erich to re-enable it, but expect a lot of tests to fail before that's addressed.

adamnovak:
OK, I think I've come up with a solution for the unpickler: if monkey-patching the Python 2 method name fails, define a class and override the Python 3 method name.

Now I'm on to finding all the places we try to pickle to/from text-mode files; I've caught enough of them that I can get `src/toil/test/batchSystems/batchSystemTest.py::MaxCoresSingleMachineBatchSystemTest::testServices` to pass, which seems to exercise some of the fileStore and jobStore code.

DailyDreaming:
Cool.  I like the unpickler solution.

adamnovak:
It looks like we've settled on file store streams (for e.g. writeSharedFileStream) being binary-mode, so you have to encode/decode text going to/from them in Python 3. Does that make sense? Do we want to think about enhancing the API to let you specify what kind of stream you want, to make text IO easier?

DailyDreaming:
`It looks like we've settled on file store streams (for e.g. writeSharedFileStream) being binary-mode, so you have to encode/decode text going to/from them in Python 3. Does that make sense?`

I chose binary-mode because it seemed the closest to the old python2.7 code, since the old strings were bytestrings, but I don't have any strong reason to keep it that way.  The decision was semi-arbitrary, but honestly, having not coded a lot of python 3, I am not familiar with what the typical convention there is.  If it looks odd to you having it be 'wb' and 'rb', then I am not opposed at all to going in the other direction so long as we try to be consistent.

`Do we want to think about enhancing the API to let you specify what kind of stream you want, to make text IO easier?`

If that's a use-case you have, and feel others would use, then sure.  I'm not opposed at all.  I do have a vague suspicion that that might create complications, but no concrete reason that it would.

diekhans:
If the data can be either text or binary, then the best approach is
it to have the same parameters as open (encoding and error) and behaive
consistently.

adamnovak:
I think we want to put off a change to the stream API until we get Python 3 actually working and passing the tests. Then we can add support for Pythonic specification of what kind of stream we want, and remove a bunch of encode/decode calls.

@DailyDreaming I'm getting to the end of the tests I can run locally that are failing. Is there a way to run e.g. the Azure job store tests outside of Jenkins?

DailyDreaming:
@adamnovak To the first, I agree with deferring other changes until this is done.  To the second, run `make docker` and set:

TOIL_TEST_INTEGRATIVE=True;
TOIL_AZURE_KEYNAME=<yourstorageaccountname>;
TOIL_AZURE_ZONE=westus;

Issue #2252, opened 2018-06-19
Update initscript to handle NVM root devices (c5.19xlarge)


joelarmstrong:
In keeping with the tradition of AWS changing what devices need to get formatted every time they release a new instance type, the c5 types (at least c5.19xlarge) use the first NVM device as the root filesystem. Sadly, in previous instances (e.g. the i2 and i3 types), NVM devices were always the ephemeral drives, so we reformat them in our initscript.

This is obviously not a great thing, because it means our initscript corrupts the root filesystem on these instance types. The fix should be changing the initscript to go through all of sd* and nvme*, filter out anything already mounted, and reformat the rest.

┆Issue Number: TOIL-284

PR #2253, opened 2018-06-20
Avoid overwriting NVMe root volumes on current-gen AWS instances (addresses #2252)

joelarmstrong:
c5, i3.metal, c5d, m5, and m5d instances all have their EBS volumes
show up as NVMe devices. Previously toil assumed that all NVMe devices
were ephemeral, so it formed a RAID device from them. On these
instance types, though, the root volume would be reformatted,
corrupted, and then remounted as readonly, causing failures.

### DIFF
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -269,13 +269,17 @@ write_files:
         ephemeral_count=0
         drives=""
         directories="toil mesos docker cwl"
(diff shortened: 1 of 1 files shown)


DailyDreaming:
http://jenkins.cgcloud.info/job/toil-pull-requests/4130/
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2420/

DailyDreaming:
Jenkins, test this please.

PR #2254, opened 2018-06-21
Add CWL destBucket runtime option (resolves #2241)

### LINKED ISSUE #2241: Upload final output files to AWS S3/Azure Blob/Google Cloud Storage
Currently (and please correct me if I'm wrong) Toil will spit out its
output files at the head node where you run a workflow. This is fine for
many cases, but sometimes it's more convenient to have all of the output
files in cloud storage.

One way of facilitating this is to add a runtime argument which lets us
specify an output file endpoint.

An implementation of this, written by @Irrationone, which works for
(only) CWL and Azure can be seen here:

https://github.com/BD2KGenomics/toil/compare/master...Irrationone:304682394b5687a88f1f48415e2e462138e83e76

This changeset makes a new object, a "final job store", which is another
job store which essentially represents the final storage destination for
the output files.

The downside of this is that it conflates a final storage destination
with a job store, which may not be desired; and that the final job store
contains much more functionality than is necessary for its purpose.

The upside of making the final storage destination another job store is
that it only requires adding in minimal code.

Anyway, any discussion on how to implement will be much appreciated.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-164)
┆Issue Number: TOIL-164

mwiens91:
Adds an option to specify cloud bucket endpoints for the final outputs
of a CWL workflow. This works (and has been tested) for the following
cases:

AWS S3:
```toil-cwl-runner --destBucket s3://my-bucket-name workflow.cwl config.yaml```

Azure Blob Storage
```toil-cwl-runner --destBucket wasb://container_name@account_name.blob.core.windows.net workflow.cwl config.yaml```

Google Storage:
```toil-cwl-runner --destBucket gs://my-bucket-name workflow.cwl config.yaml```

Make sure you have the right authentication for all three bucket types,
as is detailed in the docs.

Note that the JSON dump at the end still reports that the files are
stored locally, which is something that should be changed eventually
when using the --destBucket option.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -357,7 +357,7 @@ class ResolveIndirect(Job):
         return resolve_indirect(self.cwljob)
 
 
-def toilStageFiles(fileStore, cwljob, outdir, index, existing, export):
+def toilStageFiles(fileStore, cwljob, outdir, index, existing, export, destBucket=None):
     """Copy input files out of the global file store and update location and
     path."""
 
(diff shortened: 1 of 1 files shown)


DailyDreaming:
Jenkins, add to the whitelist please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
The cwl test is failing because subprocess32 accidentally got reverted to version: `3.5.0rc1`.  Once #2255 is merged, I'll test this again.  Otherwise it looks good.

DailyDreaming:
Jenkins, test this please.

Issue #2222, opened 2018-05-21
CWL conformance tests are timing out.


DailyDreaming:
┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-94)
┆Issue Number: TOIL-94

PR #2255, opened 2018-06-21
More Python 3 fixes, against master

adamnovak:
This is just #2251 against master, since the branch it was against has been merged in.

On the other PR, the only test it failed was the CWL conformance test. @DailyDreaming Is that expected to fail?

Also I believe the Jenkins tests are only testing Python 2, right? They still need to be upgraded to also do Python 3.

Closes #2251.

### DIFF
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -333,7 +333,7 @@ class AzureJobStore(AbstractJobStore):
         return jobStoreFileID
 
     def updateFile(self, jobStoreFileID, localFilePath):
-        with open(localFilePath) as read_fd:
+        with open(localFilePath, 'rb') as read_fd:
             with self._uploadStream(jobStoreFileID, self.files) as write_fd:
                 while True:
                     buf = read_fd.read(self._maxAzureBlockBytes)
@@ -344,7 +344,7 @@ class AzureJobStore(AbstractJobStore):
     def readFile(self, jobStoreFileID, localFilePath, symlink=False):
         try:
             with self._downloadStream(jobStoreFileID, self.files) as read_fd:
-                with open(localFilePath, 'w') as write_fd:
+                with open(localFilePath, 'wb') as write_fd:
                     while True:
                         buf = read_fd.read(self._maxAzureBlockBytes)
                         write_fd.write(buf)
--- a/src/toil/jobStores/azureJobStore.py
+++ b/src/toil/jobStores/azureJobStore.py
@@ -247,12 +247,12 @@ class AzureJobStore(AbstractJobStore):
         return job
 
     def exists(self, jobStoreID):
-        if self.jobItems.get_entity(row_key=bytes(jobStoreID)) is None:
+        if self.jobItems.get_entity(row_key=jobStoreID.encode('utf-8')) is None:
             return False
         return True
 
     def load(self, jobStoreID):
-        jobEntity = self.jobItems.get_entity(row_key=bytes(jobStoreID))
+        jobEntity = self.jobItems.get_entity(row_key=jobStoreID.encode('utf-8'))
         if jobEntity is None:
             raise NoSuchJobException(jobStoreID)
         return AzureJob.fromEntity(jobEntity)
@@ -262,7 +262,7 @@ class AzureJobStore(AbstractJobStore):
 
     def delete(self, jobStoreID):
         try:
-            self.jobItems.delete_entity(row_key=bytes(jobStoreID))
+            self.jobItems.delete_entity(row_key=jobStoreID.encode('utf-8'))
         except AzureMissingResourceHttpError:
             # Job deletion is idempotent, and this job has been deleted already
             return
@@ -356,7 +356,7 @@ class AzureJobStore(AbstractJobStore):
 
     def deleteFile(self, jobStoreFileID):
         try:
-            self.files.delete_blob(blob_name=bytes(jobStoreFileID))
+            self.files.delete_blob(blob_name=jobStoreFileID.encode('utf-8'))
             self._dissociateFileFromJob(jobStoreFileID)
         except AzureMissingResourceHttpError:
             pass
@@ -366,7 +366,7 @@ class AzureJobStore(AbstractJobStore):
         # python API) we just try to download the metadata, and hope
         # the metadata is small so the call will be fast.
         try:
-            self.files.get_blob_metadata(blob_name=bytes(jobStoreFileID))
+            self.files.get_blob_metadata(blob_name=jobStoreFileID.encode('utf-8'))
             return True
         except AzureMissingResourceHttpError:
             return False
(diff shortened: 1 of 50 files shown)


DailyDreaming:
@adamnovak CWL conformance tests have been timing out since they added a 15 minute timeout on the cwltool side.  The tests pass locally and otherwise.  Issue is: #2222 .  It basically just needs to be extended.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Also, I want to add tox to jenkins for this, but I leave tomorrow for Portland for a week.  This and the toil move to `github.com/DataSphere` I felt would be better to attempt when I got back.

src/toil/jobStores/azureJobStore.py
```diff
@@ -247,12 +246,12 @@ def create(self, jobNode):
         return job
 
     def exists(self, jobStoreID):
-        if self.jobItems.get_entity(row_key=bytes(jobStoreID)) is None:
+        if self.jobItems.get_entity(row_key=str(jobStoreID)) is None:
```
joelarmstrong: I just want to check that my understanding is correct here. jobStoreID is likely a `builtins.str` (i.e. some weird hybrid in py2, and the native `str` in python 2). So, with the removal of the `builtins.str` import from above, we're switching back to the native `str` of each version here and in all similar cases below, right? That is, `row_key` will be a `bytes` (aka Py2 `str`) in Python 2, and a `str` (aka Py2 `unicode`) in Python 3? Is that the intention? (It sounds like a reasonable thing to do if that's what Azure expects, but just want to make sure.)

adamnovak: The thing I am trying to deal with is that the Azure API code has type checks in it, so only a 100% genuine `str` will do. No `newstr` or `newbytes` objects, which come from the builtins module in Python 2, will be accepted.

I think that's why we were using `bytes` here before; we were replacing `str` from builtins, but we still had the original `bytes`, which in Python 2 is equal to `str` anyway:
```
>>> type(bytes('abc'))
<type 'str'>
```

We can't do this in Python 3, I don't think, since you can't just go from a string to bytes without encoding, and even if we could I think the Azure API still wants strings.

So I got rid of the str import, and now I am using native str, which converts `newstr` to `str` on Python 2 and does no-op conversions of `str` to `str` on Python 3.

adamnovak: I don't think I've actually managed to *run* this test under Python 3; I only have it going on Jenkins under Python 2. I still have to actually set up my local Azure setup.

joelarmstrong: OK, cool. I just wanted to make sure that was intentional, because it seemed a bit weird to me. What a pain that Azure is doing these type checks. (surely they should be checking against `basestring` anyway...)

src/toil/provisioners/clusterScaler.py
```diff
@@ -370,6 +370,34 @@ def __init__(self, provisioner, leader, config):
 
         if not sum(config.maxNodes) > 0:
             raise RuntimeError('Not configured to create nodes of any type.')
+            
+    def _round(self, number):
```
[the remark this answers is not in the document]
adamnovak: I've added some unit tests for the rounding.

adamnovak:
I've fixed up a bunch of tests that were still failing in Python 3; I don't know why I didn't catch them before. Some stuff still fails on me when I run it locally under Python 3, and I think some of that is due to me trying to run things on Courtyard with the Docker daemon on a different machine (so bind mounts don't work at all).

I'm running another pass through all the tests now, to make sure that, of the ones I can run properly, none are broken.

I think we still need to enable Python 3 in the CI before we can declare that Python 3 support exists, but I think before we do that we might want to do #2258 so that the appliance can run under Python 3.

DailyDreaming:
Also, #2258 is a good point and I agree definitely needs to be addressed after this.  Once it is, running something a bit beefer like cactus on python 3 would be a good test.

src/toil/provisioners/clusterScaler.py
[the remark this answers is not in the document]
DailyDreaming: Thanks for that, looks great.

joelarmstrong:
Cactus won't work on python 3 in the near future (though toil supporting it would give some reason to clean it up). We (ab)use a lot of py2-only features right now.

DailyDreaming:
@joelarmstrong Alright, I have some workflows for TOPMed that I can try out.

adamnovak:
This is blocked on one caching test with the Google job store failing on Jenkins, which I'm a bit befuddled as to how to debug:

http://jenkins.cgcloud.info/job/toil-pull-requests/4145/testReport/src.toil.test.src.fileStoreTest/CachingFileStoreTestWithGoogleJobStore/testCacheEvictionFailCase/

It looks like we're getting more CacheUnbalancedErrors than the test is expecting, and 3 jobs are failing instead of the expected 1. Is this something to do with chaining and how that affects how failing jobs are counted? Do we just want to remove that check on the number of failed jobs?

adamnovak:
Looks like the Ubuntu keyserver was down.

Jenkins, test this please.

adamnovak:
OK, now these tests failed, which didn't fail last time the tests ran:

```
src.toil.test.provisioners.aws.awsProvisionerTest.AWSRestartTest.testAutoScaledCluster
src.toil.test.provisioners.gceProvisionerTest.GCERestartTest.testAutoScaledCluster
```

It looks like they timed out? @DailyDreaming @joelarmstrong Is this expected behavior?

DailyDreaming:
@adamnovak testCacheEvictionFailCase has been flaky for a while. It's not Google specific.

The others seem like something with Jenkins or one of the make checks. Will test again and check the Jenkins logs:

http://jenkins.cgcloud.info/job/toil-pull-requests/4149/
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2439/

```ERROR:__main__:2 out 11 child processes failed
make: *** [jenkins_test_integration] Error 2
$ ssh-agent -k
unset SSH_AUTH_SOCK;
unset SSH_AGENT_PID;
echo Agent pid 2212 killed;
[ssh-agent] Stopped.
Recording test results
Adding one-line test results to commit status...
Setting status of 30296dc0f8248c0d0cb14f9a0d97b6253f1e641a to FAILURE with url http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2439/ and message: 'Build finished. 18 tests run, 0 skipped, 2 failed.'
Using context: jenkins-it
Finished: FAILURE
```

Jenkins, test this please.

DailyDreaming:
I suspect it may be this check and your adding the cached files to the ignored list:

https://github.com/BD2KGenomics/toil/blob/master/Makefile#L278

DailyDreaming:
This looks good to merge.

I think I have a fix for the timeout that makes sense.  I'll post a PR right after this.

Issue #2260, opened 2018-06-26
Add examples of executing CWL and WDL workflows with Toil scripts


bencvdb:
There was a request for an example of executing non-Toil workflows in a Toil script.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-288)
┆Issue Number: TOIL-288

PR #2261, opened 2018-06-26
Add examples of running prexisting workflows with Toil scripts.

bencvdb:
Addresses #2260. Provides examples of executing CWL and WDL workflows from a Toil script.

### DIFF
--- a/docs/running/cwl.rst
+++ b/docs/running/cwl.rst
@@ -32,8 +32,8 @@ To run a CWL workflow in AWS with toil see :ref:`awscwl`.
 Running CWL with Toil Scripts
 ------------------------------------
 
-CWL workflows can be evoked in native Toil scripts. For example, if you wanted to run a CWL workflow, with 3 YML files
-specifying different samples inputs, it could look something like::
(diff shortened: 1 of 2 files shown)


DailyDreaming:
Maybe a mention about Toil running cwl/wdl directly vs. indirectly, as in the example. The wdl script has a Cromwell dependency that should be mentioned as well. Looks great though, thanks for adding this!

bencvdb:
Given that these were on the same page as running CWL/WDL with Toil, I did not feel it was necessary to point them to another location where they can read up more. I did make sure to explicitly state that they were being run indirectly.

DailyDreaming:
Alright, maybe I'd better explain my reasoning.  This is useful, but not quite the standard way that we want people to run cwl.  It might be useful to explain when something like this might be appropriate vs. running with `toil-cwl-runner`.

DailyDreaming:
Due to Toil's move to `github.com/DataBiosphere`, we want to close PRs.  Merging this for now since it's just documentation, and little changes can be addressed in the future.

PR #2263, opened 2018-06-26
Fix AWSJobStore updates not working if values are shrunk

joelarmstrong:
AWSJobStore persists its data as SimpleDB "items", which are basically just dictionaries. It converts arbitrary binary data to uuencoded values, chunked across multiple keys if necessary (when the binary data is larger than the maximum value size). It ends up looking something like this:
```
{'000': ''VQsomeuuencod', '001': 'eddatawhichis', '002': 'split=='}
```

It then uses the `PutAttributes` method in SimpleDB to store the data. There's no problem the first time a fresh item is created. But there are massive problems when an item is *changed* rather than created (through `AWSJobStore.update` or `.updateFile`). `PutAttributes` updates *only* those attributes that are specified in the request. So shrinking the above file to something that encodes to "shorterunpaddedvalue" may end up doing something like this:

```
{'000': "shorterunpaddedvalue", "001": "eddatawhichis", "002": "split=="}
```

That is, the jobStore only attempted to update the key `000`, so the other chunks still exist and represent stale data. This is really, really bad. Thankfully, the bug doesn't actually trigger most of the time, because when the data is compressible it's encoded with bzip2, which silently discards all trailing unexpected data. In that case, you don't get any weird stale data after your new data. But if the data happens not to be compressible, this bug triggers and gives very surprising results.

Separately, a similar bug also affects the AWSJobStore "overlarge jobs". These are jobs that are stored in S3 because they're too large to store in a single SimpleDB item. The "overlargeID" key indicates the presence of an overlarge job. But, again, when shrinking a previously-overlarge job to be smaller than the "overlarge boundary", we don't provide the "overlargeID" key in the SimpleDB request, meaning that the "overlargeID" will still be present despite pointing to old data. The next read will give stale data.

Reading stale data for jobs / files affects the correctness of your Toil workflow. Workflows run without `--disableChaining` are especially vulnerable since their jobGraphs are updated more often. This can result in jobs being marked as completed even when they haven't yet succeeded.

This PR fixes all of the above (according to the tests.  testing on a large pipeline now), and adds a new regression test to that shrinking a job doesn't yield stale data.


src/toil/jobStores/aws/jobStore.py
```diff
@@ -233,7 +233,7 @@ def _registered(self, value):
                                                        attributes=attributes)
     
     def _awsJobFromItem(self, item):
-        if "overlargeID" in item:
+        if item["overlargeID"]:
```
DailyDreaming: Can this potentially KeyError?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -339,7 +340,7 @@ def delete(self, jobStoreID):
         for attempt in retry_sdb():
             with attempt:
                 item = self.jobsDomain.get_attributes(bytes(jobStoreID), consistent_read=True)
-        if "overlargeID" in item:
+        if item["overlargeID"]:
```
DailyDreaming: Same?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -651,6 +651,24 @@ def testZeroLengthFiles(self):
                 self.assertEquals(f.read(), "")
             self.master.delete(job.jobStoreID)
 
+        def testGrowingAndShrinkingJob(self):
```
DailyDreaming: This is great.

src/toil/jobStores/aws/jobStore.py
joelarmstrong: I wanted this to KeyError because it will fail if overlargeID isn't always present--this makes sure there isn't some sort of logic bug which would cause us to fail to update that key. I could add a catch which raises a more descriptive error, if you'd prefer?

It *will* KeyError when resuming from jobStores created with a previous version, but the other changes throw resumability out the window anyway.

src/toil/test/jobStores/jobStoreTest.py
joelarmstrong: It does waste 50MB of upload/download, though. I could shrink it down to whatever the minimal amount is that still triggers the overlarge job code.

DailyDreaming: I think it's 5MB, though looking at the helper, I think 200kb should trigger it.

joelarmstrong: D'oh, you're totally right. I shifted it down to 500kB, because the Azure jobStore only allows jobs that are less than 1MB. (Which means we eventually need to add overlarge jobs to Azure, too... but one step at a time.)

DailyDreaming: That's a good idea.  And Azure will get there.  I'll file a ticket so that I don't forget.

DailyDreaming:
This looks good and I'm guessing will pass once the conflict in `src/toil/jobStores/aws/utils.py` is resolved.  @joelarmstrong did this need any other changes?

DailyDreaming:
Awesome, thanks @joelarmstrong !  Will merge once the tests pass.

PR #2265, opened 2018-06-27
Extend cwltest (conformance run) timeout to 1800 seconds.  Addresses #2181.

DailyDreaming:
Addresses #2181.

Extends the cwl conformance tests we run from the default 900 seconds to 1800 seconds.  It also reverts the cwl confromance test to the code it contained previously, except for this small addition.

Also, I hope I'm not abusing the `TEST_N` variable here @mr-c .

### DIFF
--- a/src/toil/test/cwl/cwlTest.py
+++ b/src/toil/test/cwl/cwlTest.py
@@ -143,6 +143,7 @@ class CWLTest(ToilTest):
     def test_run_conformance(self, batchSystem=None):
         rootDir = self._projectRootPath()
         cwlSpec = os.path.join(rootDir, 'src/toil/test/cwl/spec')
+        workDir = os.path.join(cwlSpec, 'v1.0')
         # The latest cwl git hash. Update it to get the latest tests.
         testhash = "f96bca6911b6688ff614c02dbefe819bed260a13"
         url = "https://github.com/common-workflow-language/common-workflow-language/archive/%s.zip" % testhash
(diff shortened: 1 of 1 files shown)


DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

src/toil/test/cwl/cwlTest.py
```diff
@@ -153,46 +153,22 @@ def test_run_conformance(self, batchSystem=None):
             shutil.move("common-workflow-language-%s" % testhash, cwlSpec)
             os.remove("spec.zip")
         try:
-            cmd = ["bash", "run_test.sh", "RUNNER=toil-cwl-runner",
-                   "DRAFT=v1.0", "-j4"]
+            cmd = ['bash', 'run_test.sh', 'RUNNER=toil-cwl-runner', 'DRAFT=v1.0',
+                   'TEST_N="--timeout=1800"', '-j4']
```
mr-c: Nice hack (ab)using `TEST_N`, but `--timeout` is for `cwltest`, so just call that directly:

```
cd v1.0
cwltest --tool ${path_to_toil-cwl-runner} --test=conformance_test_v1.0.yaml --timeout=1800 --basedir=${PWD}
```

DailyDreaming: Yes, that's much cleaner, and I'll make the change. I felt odd about it so thanks for replying back.

I was hoping to push back a change to the bash script to include another var to adjust the timeout if you feel like that would be useful, but that would be for the next hash.

DailyDreaming:
@mr-c can you take a look?  I think this is ready.  Thanks in advance!

PR #2268, opened 2018-06-29
Improved wdl support.

DailyDreaming:
Fixes some docker permissions and adds in some extra debugging logs.


DailyDreaming:
Jenkins, test this please.

DailyDreaming:
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2458/
http://jenkins.cgcloud.info/job/toil-pull-requests/4169/ 

Jenkins, test this please.

DailyDreaming:
@joelarmstrong I think this is ready.  Would you mind taking a look?

src/toil/wdl/wdl_synthesis.py
```diff
@@ -706,10 +706,13 @@ def write_function_dockercall(self, job):
...
                                image={docker_image}, 
                                working_dir=tempDir, 
-                               parameters=["/root/{job_task_reference}_script.sh"], 
+                               parameters=[os.path.join(tempDir, "{job_task_reference}_script.sh")], 
                                entrypoint="/bin/bash", 
-                               user={docker_user},
-                               volumes={{tempDir: {{"bind": "/root"}}}})
+                               user={docker_user}, 
+                               stderr=True, 
+                               volumes={{tempDir: {{"bind": tempDir}}}})
```
joelarmstrong: Are you sure this works? I've had no success doing this in the past, but maybe my understanding is incomplete here.

I've never been able to use Docker to bind-mount a directory into a nonexistent directory in the container. It runs the container, but silently fails to actually perform the mount. Which does makes sense: there's nowhere for it to be mounted. But even if you later create the directory, you can never get the bind to actually occur (again, this makes sense).

Is this happening here? It looks to me like you may be bind mounting to an empty dir. `tempDir` seems to be set from `fileStore.getLocalTempDir()`, which (IIRC) is usually under the workDir, which by default gets set to something like `${TMPDIR}/toil-${machine-id}/randomstring`.  That directory surely doesn't exist in the container.

src/toil/wdl/wdl_synthesis.py
```diff
@@ -706,10 +706,13 @@ def write_function_dockercall(self, job):
...
+        with open(os.path.join(asldijoiu23r8u34q89fho934t8u34fcurrentworkingdir, '{job_task_reference}.log'), 'w') as f:
```
joelarmstrong: haha, what's the story behind this? was this to avoid an accidental name collision?

src/toil/wdl/wdl_synthesis.py
```diff
@@ -730,7 +733,7 @@ def write_function_cmdline(self, job):
         if 'raw_commandline' in self.tasks_dictionary[job]:
             for cmd in self.tasks_dictionary[job]['raw_commandline']:
                 if not cmd.startswith("r'''"):
-                    cmd = 'str({}).strip("\\n")'.format(cmd)
+                    cmd = 'str({i} if not isinstance({i}, tuple) else process_and_read_file({i}, tempDir, fileStore)).strip("\\n")'.format(i=cmd)
```
joelarmstrong: Oops, I didn't catch this the first time, and maybe I'm mistaken about your intention here, but I don't think this `.strip("\\n")` is doing what you want.

It looks to me like you are trying to get rid of trailing and leading `\\n`s: that is, not newline characters, but the actual characters `\` and `n`. This does that, but it *also* gets rid of other characters if they happen to be either `\`s or `n`s! The strip method works on the basis of characters, not substrings.

OK, that is confusing to try to explain, but an example is worth a thousand words:
```
>>> "\\nnn\\nno correction\\n".strip("\\n")
'o correctio'
```
Couldn't that cause major issues if a command starts/ends with an `n`?

src/toil/wdl/wdl_synthesis.py
DailyDreaming: I remember trying this in the past and having the same experience!  I saw someone try it recently though and thought it wouldn't work but it does now?  For example, if you want to create a mythical:
`/home/lifeisaboutfishtacos/Desktop/dragon`
folder with some files, and you run:
`docker run -it -v /home/lifeisaboutfishtacos/Desktop/dragon:/home/lifeisaboutfishtacos/Desktop/dragon ubuntu`
and run:
`ls /home/lifeisaboutfishtacos/Desktop/dragon`
You will see the contents of your folder.  It also runs the unittests fine, but yes, I was surprised too!  I wish I'd always been able to do this.

src/toil/wdl/wdl_synthesis.py
DailyDreaming: Yes, if you know of a better way, please advise.  XD

src/toil/wdl/wdl_synthesis.py
joelarmstrong: Wow, I just tried it on my laptop and it works! I literally tried it just last week on my desktop and it didn't work, which is why I was so sure.

This is probably a feature that was released with a new version of Docker. I think my desktop is running a slightly old version, but my laptop is running the latest.

Well, this is good to know. It certainly simplifies things quite a bit.

src/toil/wdl/wdl_synthesis.py
DailyDreaming: Yes, thanks for the catch!  I tested this and yes, misunderstood the logic behind strip().  I did still want to strip normal newlines as well, so I've added a function to perform this and a short test.

I need to test with Cromwell at some point to make sure that Toil's parsing and Cromwell's is one to one, but this seems to work in the meantime.

joelarmstrong: Wait, maybe the double-escaping (because of the indirection through 2 layers of Python) threw me off here.

If I do this:
```
print 'str(cmd if not isinstance(cmd, tuple) else process_and_read_file(cmd, tempDir, fileStore)).strip("\\n")'
```
I get this:
```
str(cmd if not isinstance(cmd, tuple) else process_and_read_file(cmd, tempDir, fileStore)).strip("\n")
```
Which probably means that the original code *is* what you want. Sorry if I led you on a wild goose chase with that one.

PR #2272, opened 2018-07-02
Issues/2122 add azure environ var

bencvdb:
This is a redo of PR #2264 which was closed to a force-push related incident.

There was no support for setting a default Azure zone via an environment variable. This was addressed as well as updating documentation to reflect the change. Incidentally, an the helpfulness of an error message I encountered was increased.

### DIFF
--- a/docs/appendices/environment_vars.rst
+++ b/docs/appendices/environment_vars.rst
@@ -52,6 +52,8 @@ There are several environment variables that affect the way Toil runs.
 |                        | deleted until all associated nodes have been       |
 |                        | terminated.                                        |
 +------------------------+----------------------------------------------------+
+| TOIL_AZURE_ZONE        | The default region for provisioning instances.     |
++------------------------+----------------------------------------------------+
 | TOIL_SLURM_ARGS        | Arguments for sbatch for the slurm batch system.   |
 |                        | Do not pass CPU or memory specifications here.     |
 |                        | Instead, define resource requirements for the job. |
--- a/docs/gettingStarted/quickStart.rst
+++ b/docs/gettingStarted/quickStart.rst
@@ -525,8 +525,8 @@ Also!  Remember to use the :ref:`destroyCluster` command when finished to destro
 
         **A Helpful Tip**
 
-        When using AWS, setting the environment variable eliminates having to specify the ``--zone`` option for each command.
-        This will be supported for GCE and Azure in the future. ::
+        When using AWS or Azure, setting the environment variable eliminates having to specify the ``--zone`` option
+        for each command. This will be supported for GCEin the future. ::
 
             (venv) $ export TOIL_AWS_ZONE=us-west-2c
 
--- a/src/toil/lib/bioio.py
+++ b/src/toil/lib/bioio.py
@@ -287,9 +287,6 @@ def parseBasicOptions(parser):
     if options.provisioner == 'azure':
         if os.environ.get('TOIL_AZURE_ZONE'):
             options.zone = os.environ.get('TOIL_AZURE_ZONE')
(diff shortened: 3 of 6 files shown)


bencvdb:
Jenkins, test this.

docs/appendices/environment_vars.rst
```diff
@@ -52,6 +52,8 @@ There are several environment variables that affect the way Toil runs.
 |                        | deleted until all associated nodes have been       |
 |                        | terminated.                                        |
 +------------------------+----------------------------------------------------+
+| TOIL_AZURE_ZONE        | The default region for provisioning instances.     |
```
DailyDreaming: Default may be the wrong word here?  It can certainly change to a non-default value.

docs/gettingStarted/quickStart.rst
```diff
@@ -525,8 +525,8 @@ Also!  Remember to use the :ref:`destroyCluster` command when finished to destro
 
         **A Helpful Tip**
 
-        When using AWS, setting the environment variable eliminates having to specify the ``--zone`` option for each command.
-        This will be supported for GCE and Azure in the future. ::
+        When using AWS or Azure, setting the environment variable eliminates having to specify the ``--zone`` option
+        for each command. This will be supported for GCEin the future. ::
```
DailyDreaming: `GCEin`

docs/appendices/environment_vars.rst
DailyDreaming: "Default" may be the wrong word here? It can change to a non-default region.

src/toil/lib/bioio.py
```diff
@@ -283,6 +283,11 @@ def parseBasicOptions(parser):
 
     setLoggingFromOptions(options)
 
+    # Look for a default zone.
```
DailyDreaming: This works, but it should be fetched via a function that checks both the environment variable and the user options.

In this case, if the `TOIL_AZURE_ZONE` has a default value (now or added in the future) it will overwrite what the user specifies via `--zone uswest`, which may not be the behavior we want since then the user can't specify a zone via the commandline anymore once this is set.

There should be a function that checks all places (options and env variable) and returns the zone.  See here for how we fetch the AWS zone:
https://github.com/BD2KGenomics/toil/blob/9455dd4c2662979da9ff18eef8bf5bf7b5f10344/src/toil/provisioners/aws/__init__.py#L37

There needs to be a function like that for Azure.  This file might be a good place to put it:
https://github.com/BD2KGenomics/toil/blob/9455dd4c2662979da9ff18eef8bf5bf7b5f10344/src/toil/provisioners/azure/__init__.py  (include the toil copyright header there if you do).

src/toil/provisioners/azure/azureProvisioner.py
```diff
@@ -211,8 +211,7 @@ def _checkIfClusterExists(self):
         try:
             self.callPlaybook(self.playbook['check-cluster'], ansibleArgs, wait=True)
         except RuntimeError:
-            logger.info("The cluster could not be created. Try deleting the cluster if it already exits.")
-            raise
+            raise RuntimeError("The cluster could not be created. Try deleting the cluster if it already exits.")
```
DailyDreaming: Looks good.

docs/appendices/environment_vars.rst
bencvdb: "A specified region for provisioning instances."?

DailyDreaming:
We'll be moving toil soon so I'm closing this for now.  New changes can be made and submitted to the new `github.com/DataBiophere` location of Toil.

bencvdb:
Jenkins, test this please.

PR #2302, opened 2018-07-13
Issues/2266 break up test job store test

bencvdb:
Breaks up a unit test that was too long and had too ambiguous of a name into smaller ones.


DailyDreaming:
My previous review of this: https://github.com/DataBiosphere/toil/pull/2281

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -544,7 +742,7 @@ def testFileDeletion(self):
             Intended to cover the batch deletion of items in the AWSJobStore, but it doesn't hurt
             running it on the other job stores.
             """
-            master = self.master
+            master = self.jobstore_initialized
```
jessebrennan: It seems like this 'master' variable serves no purpose. It should probably be deleted and it's use replaced with 'self.jobstore_initialized'

PR #2314, opened 2018-07-19
Remove log message.

DailyDreaming:
Really small change but this log message spams the terminal if one's importing a lot of public s3 files and it's not particularly helpful.  Also removed an import since I was actually writing a PR for s3, most of which is no longer needed, but might as well keep the import change.


src/toil/lib/ec2nodes.py
```diff
@@ -246,8 +246,7 @@ def updateStaticEC2Instances():
         # See the License for the specific language governing permissions and
         # limitations under the License.
         from six import iteritems
-        from toil.lib.ec2nodes import InstanceType\n\n\n''',
-                            dictionary={'year': datetime.date.today().strftime("%Y")}))
+        from toil.lib.ec2nodes import InstanceType\n\n\n''').format(year=datetime.date.today().strftime("%Y"))[1:])
```
joelarmstrong: Not really important,  but why `[1:]`?

DailyDreaming: The string starts with a newline char so this just makes it neater when it writes the file.

PR #2321, opened 2018-07-25
Fix mixing up S3 with Google Cloud Storage in docs

mwiens91:
Also, haven't tested this, but is

```
gce:us-west2-2a:my-google-jobstore-name
```

still the correct format for specifying Google Cloud Storage buckets?


bencvdb:
@mwiens91 Thanks for catching this! And no `gce:us-west2-2a:my-google-jobstore-name` should be `gce:us-west2-a:my-google-jobstore-name`.

mwiens91:
@bencvdb Thanks for letting me know. I've added the correct bucket syntax to my PR.

PR #2324, opened 2018-07-26
Improve documentation around fileID (resolves #2278)

### LINKED ISSUE #2278: No documentation for the FileID.size attribute appears on ReadTheDocs
I have a job that needs to use some files from the file store. I initially tried with with just 1 GB of disk, but that's not enough to download the files that it wants to work with.

I'm trying to rewrite the job to measure how big the files it wants to use are, and to re-queue itself with the appropriate amount of disk required, so it can actually download them. But the file store doesn't seem to expose a file size query. All you can do with a file in the file store is read it---and I think even if you read it streaming, your job is liable to fail if the file is bigger than the amount of disk available on the node you are running on.

The file store needs to expose file sizes, so that jobs can know how much disk they need to ask for in order to use particular files.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-296)
┆Issue Number: TOIL-296

bencvdb:
It was pointed out that there was no documentation around accessing the size of a file stored in a jobstore. There was a little blurb about FileID's in general in the "Developing Workflows" section but I can see how it could be improved. This PR adds the FileID class to the "Filestore API" section as well as a link between this and the previously mentioned blurb. @adamnovak @joelarmstrong did you miss the blurb or not feel it was a sufficient description? Let me know if you have any feedback that I can use to improve this documentation.

### DIFF
--- a/docs/developingWorkflows/developing.rst
+++ b/docs/developingWorkflows/developing.rst
@@ -569,7 +569,7 @@ multiple jobs' output values::
 FileID
 ------
 
-This object is a small wrapper around Python's builtin string class. It is used to
+The :class:`toil.fileStore.FileID` class is a small wrapper around Python's builtin string class. It is used to
 represent a file's ID in the file store, and has a ``size`` attribute that is the
 file's size in bytes. This object is returned by ``importFile`` and ``writeGlobalFile``.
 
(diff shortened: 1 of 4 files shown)


adamnovak:
I think I must have missed the blurb; I was looking at the file store and file ID documentation, since that's where I assumed the size would have to come from. I didn't look at "Developing Workflows".

PR #2332, opened 2018-08-03
WDL Changes.

DailyDreaming:
In line with supporting WES for toil, I've made a few changes to the WDL functions and options.

Also, the filestore was reporting filesizes that were too small with `fileID.size()` and weren't taking blocksize into account so I've also changed how `fileID.size()` determines size.

### DIFF
--- a/src/toil/wdl/wdl_synthesis.py
+++ b/src/toil/wdl/wdl_synthesis.py
@@ -43,9 +43,17 @@ class SynthesizeWDL:
     then write the main and all of its subsections.
     '''
 
-    def __init__(self, tasks_dictionary, workflows_dictionary, output_directory, json_dict, docker_user):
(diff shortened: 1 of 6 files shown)


src/toil/wdl/wdl_synthesis.py
```diff
@@ -43,10 +43,18 @@ class SynthesizeWDL:
...
+                 tasks_dictionary,
+                 workflows_dictionary,
+                 output_directory,
+                 json_dict,
+                 docker_user,
+                 jobstore=None):
         self.output_directory = output_directory
         self.output_file = os.path.join(self.output_directory, 'toilwdl_compiled.py')
 
+        self.jobstore = jobstore if jobstore else './toilWorkflowRun'
```
bencvdb: I like this!

PR #2337, opened 2018-08-09
Fix 2162 (at least partially) by refactoring stream writes in the fileJobStore

adamnovak:
I've backed out the code that I wrote to diagnose what is holding a file open and try to kill it or close the file descriptor, but this at least solves the immediate problem I was having in that issue and somehow stops the file descriptor from leaking in the first place.

It's not a general solution to misbehaving user code, which can still do bad things like start a subprocess attached to a file stream in the job store and leave it running when the Toil job exits.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -364,6 +364,15 @@ class FileJobStore(AbstractJobStore):
 
     @contextmanager
     def writeFileStream(self, jobStoreID=None):
+        fd, absPath = self._getTempFile(jobStoreID)
+        relPath = self._getRelativePath(absPath)
+        with open(absPath, 'wb') as f:
+            logger.debug('Write file stream {}'.format(relPath))
+            yield f, relPath
+        os.close(fd)  # Close the os level file descriptor
+        return
+                
+    
         # Record the name of the job/function writing the file in the file name
         try:
             # It ought to be third-to-last on the stack, above us and the context manager stuff
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -284,43 +284,38 @@ class FileJobStore(AbstractJobStore):
     def _supportsUrl(cls, url, export=False):
         return url.scheme.lower() == 'file'
 
-    def writeFile(self, localFilePath, jobStoreID=None):
+    def _getUserCodeFunctionName(self):
+        """
+        Get the name of the function 4 levels up the stack (above this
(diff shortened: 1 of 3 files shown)


adamnovak:
I'm duplicating some of the stack-walking code form writeFile, but I'm reluctant to pull it out into its own function because of the aformentioned stack walking it has to do, and because the rationale for why the stack walking ought to work is different in each case.

src/toil/jobStores/fileJobStore.py
```diff
@@ -280,27 +285,47 @@ def _supportsUrl(cls, url, export=False):
...
+        # Record the name of the job/function writing the file in the file name
         try:
-            sourceFunctionName = traceback.extract_stack()[0][3].split("(")[0]
+            # It ought to be third-to-last on the stack, above us and the FileStore. Probably.
+            sourceFunctionName = traceback.extract_stack()[-3][2]
         except:
-            sourceFunctionName = "x"
+            sourceFunctionName = "UNKNOWNJOB"
             # make sure the function name fetched has no spaces or oddities
+        finally:
```
DailyDreaming: This looks like a no-op.

src/toil/jobStores/fileJobStore.py
```diff
@@ -280,27 +285,47 @@ def _supportsUrl(cls, url, export=False):
...
+        relPath = self._getRelativePath(absPath)
         shutil.copyfile(localFilePath, absPath)
-        return self._getRelativePath(absPath)
+        return relPath
 
     @contextmanager
     def writeFileStream(self, jobStoreID=None):
-        fd, absPath = self._getTempFile(jobStoreID)
+        # Record the name of the job/function writing the file in the file name
+        try:
```
DailyDreaming: This looks like duplicated code from the above and should probably be its own function.

adamnovak:
OK @DailyDreaming, this should address your comments.

I'm not sure what went wrong with that WDL test. I'll see if I can get it to misbehave for me and maybe say more than that it is launching a job and then that it has timed out.

adamnovak:
I wrangled a Docker machine, and that test passed in 350.27 seconds when I ran it with:
```
make test tests=src/toil/test/wdl/toilwdlTest.py::ToilWdlIntegrationTest::testENCODE
```

Maybe something I changed to address the review comments also fixed the test? Or maybe something important is different between Jenkins and my environment.

DailyDreaming:
@adamnovak Strange.  Well, tests are passing now and it looks good to me.  Merging.

PR #2338, opened 2018-08-13
WIP Mesos update.

DailyDreaming:
Edit: The code in this PR is now largely Adam's quality work, which has been marged in from #2366.

I've basically only converted toil's mesos executor at this point (probably poorly) based on the `pymesos` example here: https://github.com/douban/pymesos/blob/master/examples/executor.py

The next step is converting our scheduler: https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/mesos/batchSystem.py
To conform to their scheduler example: https://github.com/douban/pymesos/blob/master/examples/scheduler.py

Feel free to add to this.


benedictpaten:
@adamnovak When you have time, perhaps pick this up to get us to Python3.0

DailyDreaming:
@adamnovak Travis failing is nothing that you did: https://github.com/boto/boto/issues/3717  It started failing right before I changed the AMI for no reason that I can discern.

Still needs to be fixed, but most tests on travis run on Jenkins as well (was going to have our new undergraduate student worker apply the `@needs_travis` decorator around) so this should be fine if it passes the jenkins tests.

DailyDreaming:
And this one before?

<Code>NumberDomainsExceeded</Code><Message>Number of domains limit exceeded.</Message>

```
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/test/jobStores/jobStoreTest.py", line 1033, in setUp
    super(AbstractEncryptedJobStoreTest.Test, self).setUp()
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/test/jobStores/jobStoreTest.py", line 128, in setUp
    self.jobstore_initialized.initialize(self.config)
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/jobStores/aws/jobStore.py", line 135, in initialize
    self.destroy()
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/lib/exceptions.py", line 55, in __exit__
    raise_(exc_type, exc_value, traceback)
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/jobStores/aws/jobStore.py", line 132, in initialize
    self._bind(create=True)
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/jobStores/aws/jobStore.py", line 160, in _bind
    self.jobsDomain = self._bindDomain(qualify('jobs'), create=create, block=block)
  File "/mnt/ephemeral/workspace/toil-pull-requests/src/toil/jobStores/aws/jobStore.py", line 773, in _bindDomain
    return self.db.create_domain(domain_name)
  File "/mnt/ephemeral/workspace/toil-pull-requests/venv/local/lib/python2.7/site-packages/boto/sdb/connection.py", line 321, in create_domain
    d = self.get_object('CreateDomain', params, Domain)
  File "/mnt/ephemeral/workspace/toil-pull-requests/venv/local/lib/python2.7/site-packages/boto/connection.py", line 1208, in get_object
    raise self.ResponseError(response.status, response.reason, body)
SDBResponseError: SDBResponseError: 409 Conflict
<?xml version="1.0"?>
<Response><Errors><Error><Code>NumberDomainsExceeded</Code><Message>Number of domains limit exceeded.</Message><BoxUsage>0.0055590278</BoxUsage></Error></Errors><RequestID>54c58285-9961-665d-8a95-8a7e027ecd51</RequestID></Response>
```
http://jenkins.cgcloud.info/job/toil-pull-requests/4539/

adamnovak:
I don't have an account in the relevant Amazon account, I don't think. @DailyDreaming are you going to go through and clean up excess buckets/domains?

DailyDreaming:
@adamnovak Yes, I'll go through and delete the relevant buckets, though it has to be done one-by-one unfortunately with AWS.  I deleted half a dozen after the first time this happened, though I guess it wasn't enough.

The buckets left behind all seem to be empty.  I have a feeling this PR may not be deleting some and instead just emptying them?  I'll keep track of how many there are before you run the next tests, and then what's left after the next run.

adamnovak:
OK, let me know when you have deleted some and are ready to rerun the tests.

Maybe we aren't cleaning up the buckets when the tests fail or time
out. That hopefully shouldn't be happening so much anymore, but we had
a lot of timeouts happen when I was debugging the tests.

DailyDreaming:
@adamnovak All tests passed!  I don't see any stray buckets left behind either.  I'm going to do a `git pull` from `master` just to launch another round of tests, but if those pass, I'll do a code review and probably ask @joelarmstrong to weigh in as well (if he doesn't mind!  :) ).

Thanks for the PR, this is awesome and I'm excited to see this merged!

src/toil/batchSystems/mesos/executor.py
```diff
@@ -212,11 +233,75 @@ def frameworkMessage(self, driver, message):
...
-def main(executorClass=MesosExecutor):
+def main():
     logging.basicConfig(level=logging.DEBUG)
     log.debug("Starting executor")
-    executor = executorClass()
-    driver = mesos.native.MesosExecutorDriver(executor)
-    exit_value = 0 if driver.run() == mesos_pb2.DRIVER_STOPPED else 1
+
+    if not os.environ.has_key("MESOS_AGENT_ENDPOINT"):
+        # Some Mesos setups in our tests somehow lack this variable. Provide a
```
DailyDreaming: That does seem strange.  Did you get any further in solving this one?

src/toil/batchSystems/mesos/executor.py
```diff
@@ -212,11 +233,75 @@ def frameworkMessage(self, driver, message):
...
+    try:
+        urlopen("http://%s/logging/toggle?level=1&duration=15mins" % os.environ["MESOS_AGENT_ENDPOINT"]).read()
+        log.debug("Toggled agent log level")
+    except Exception as e:
+        log.debug("Failed to toggle agent log level")
+        
+    # Parse the agent state
+    agent_state = json.loads(urlopen("http://%s/state" % os.environ["MESOS_AGENT_ENDPOINT"]).read())
+    if agent_state.has_key('completed_frameworks'):
+        # Drop the completed frameworks whichg grow over time
```
DailyDreaming: nit: `whichg`

src/toil/batchSystems/mesos/executor.py
```diff
@@ -212,11 +233,75 @@ def frameworkMessage(self, driver, message):
...
+                if 'limit_in_bytes' not in filename:
+                    continue
+                log.debug('cgroup memory info from %s:' % os.path.join(dirpath, filename))
+                try:
+                    for line in open(os.path.join(dirpath, filename)):
+                        log.debug(line.rstrip())
+                except Exception as e:
+                    log.debug("Failed to read file")
+                    
+    # Mesos can also impose rlimit limits, including on things that really
```
DailyDreaming: Sorry if it's a dumb question, but do you know why it does this, or the limiting by cgroup?

src/toil/batchSystems/mesos/test/__init__.py
```diff
@@ -52,18 +85,31 @@ def tryRun(self):
             self.popen.wait()
             log.info('Exiting %s', self.__class__.__name__)
 
-        def findMesosBinary(self, name):
-            try:
-                return next(which(name))
-            except StopIteration:
+        def findMesosBinary(self, names):
+            if isinstance(names, basestring):
```
DailyDreaming: More py2/3, I think `basestring` needs to imported for py3.

src/toil/jobStores/googleJobStore.py
```diff
@@ -24,6 +24,7 @@
 import logging
 import time
 import os
+import os.path
```
DailyDreaming: Nit: `import os.path` and `import os`.

src/toil/batchSystems/mesos/executor.py
adamnovak: Not really. I'm not sure it happens in 1.01, either. I added this before we switched.

src/toil/batchSystems/mesos/executor.py
```diff
@@ -212,11 +233,75 @@ def frameworkMessage(self, driver, message):
...
-    exit_value = 0 if driver.run() == mesos_pb2.DRIVER_STOPPED else 1
+
+    if not os.environ.has_key("MESOS_AGENT_ENDPOINT"):
+        # Some Mesos setups in our tests somehow lack this variable. Provide a
+        # fake one to maybe convince the executor driver to work.
+        os.environ["MESOS_AGENT_ENDPOINT"] = os.environ.get("MESOS_SLAVE_ENDPOINT", "127.0.0.1:5051")
+        log.warning("Had to fake MESOS_AGENT_ENDPOINT as %s" % os.environ["MESOS_AGENT_ENDPOINT"])
+        
+    try:
+        urlopen("http://%s/logging/toggle?level=1&duration=15mins" % os.environ["MESOS_AGENT_ENDPOINT"]).read()
```
adamnovak: We might want to ditch this log level toggle. I had it in there to try and make debugging failing executors easier, but I don't think I ever got much useful out of it, and we probably don't want to be tinkering with people's settings in production.

src/toil/batchSystems/mesos/executor.py
adamnovak: Mesos applies rlimits to jobs to keep them from using more memory than they asked for and interfering with other scheduled jobs on the same node. It's one of the containerization/isolation strategies that you can have it use instead of cgroups, which I don't think you can do without root, at least on a lot of deployed kernels.

src/toil/jobStores/googleJobStore.py
adamnovak: I totally thought you needed both. Apparently you don't.

adamnovak:
@joelarmstrong Ccan have a look at this and hit the big merge button if you approve?

src/toil/batchSystems/mesos/test/__init__.py
DailyDreaming: Resolved this.

src/toil/batchSystems/mesos/executor.py
DailyDreaming: I put a manual switch in.  Not sure if it's the best method, but if it ever comes up again someone reading the code might find it useful.

DailyDreaming:
@adamnovak Hope you don't mind, I made some of the changes just so that I don't forget about them.

DailyDreaming:
@adamnovak Sounds good, I'll add in the changes from the comments here in a newPR.  Accidentally committed and relaunched the tests, but it's the same code and they've passed 2 already so I imagine they will again a 3rd time.

DailyDreaming:
LGTM!  Will make another PR to address the above issues.

DailyDreaming:
Thanks for reviewing this @joelarmstrong !

adamnovak:
I don't think this is a race with 256. This line adds items to the
set. The code around line 256 finds items in the set with that
intersection, and then removes the items it found. It's possible a new
item could get in between the intersection and the removal, or during
either process, but the GIL ensures that we can't corrupt the sets.
The worst thing that can happen, it seems to me, is that an item gets
added but we don't pick it up, and then we just pick it up on the next
pass through the loop.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -533,67 +589,92 @@ def _newMesosTask(self, job, offer):
...
+            try:
+                self.killJobIds.remove(jobID)
+            except KeyError:
+                pass
+            else:
+                # We were asked to kill this job, so say that we have done so.
+                # We do this LAST, after all status updates for the job have
+                # been handled, to ensure a consistent view of the scheduler
+                # state from other threads.
+                self.killedJobIds.add(jobID)
```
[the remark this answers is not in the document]
adamnovak: I don't think this is a race with 256. This line adds items to the
set. The code around line 256 finds items in the set with that
intersection, and then removes the items it found. It's possible a new
item could get in between the intersection and the removal, or during
either process, but the GIL ensures that we can't corrupt the sets.
The worst thing that can happen, it seems to me, is that an item gets
added but we don't pick it up, and then we just pick it up on the next
pass through the loop.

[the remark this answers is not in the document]
joelarmstrong: My worry is that at a slightly lower level, it seems possible that line 256 really looks something more like this:
```
1. create a new set with the items removed
2. assign self.killedJobIds to that new set
```

1 probably compiles down to a bunch of opcodes, and 2 might be only one -- but Python can easily switch to the other (batch system) thread after 1, add a killed job to self.killedJobIds, which promptly gets overwritten by 2. So the fact that the job's been killed may never be seen by the leader thread (if the intersection/assignment isn't atomic).

Anyway, not 100% sure on whether this is indeed not atomic, I'm just a bit paranoid about races in the Mesos code. We've had this exact type of 1-opcode race cause a crash before (though it only triggered after a few days!).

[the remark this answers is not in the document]
adamnovak: I don't think that can happen unless the set is internally using some data structure that makes that replacement and the GIL doesn't ensure a consistent view of it. For sets, `|=` and `-=` operate on the same `set` object, as evidenced by the fact that we can see their results through other references:

```
>>> a = set()
>>> b = a
>>> a |= set([1, 2, 3])
>>> a
set([1, 2, 3])
>>> b
set([1, 2, 3])
>>> a -= set([2])
>>> b
set([1, 3])
>>> 
```

So there's no new `set` at the Python level.

But it's never that nice to rely on the GIL; we might want to rely on some kind of synchronized queue to pass these things around between threads, or just add some locks around the set accesses.

PR #2347, opened 2018-08-21
Grid engine job creation fix

bencvdb:
This Pr will change how the abstract grid engine batch system will decide whether or not to continue making jobs. Currently it the --maxLocalJobs option is not being respected. This fixes the issue by checking if the number of running jobs is less than the value set by the user. When it exceeds that number it stops.


DailyDreaming:
LGTM!

PR #2351, opened 2018-08-22
Retry HTTP connections on BadStatusLine errors

adamnovak:
Fix https://github.com/DataBiosphere/toil/issues/2350

### DIFF
--- a/src/toil/lib/retry.py
+++ b/src/toil/lib/retry.py
@@ -176,6 +176,7 @@ def retry_http( delays=default_delays, timeout=default_timeout, predicate=retrya
     ...         raise BadStatusLine('sad-cloud.gif')
     Traceback (most recent call last):
     ...
+    BadStatusLine: sad-cloud.gif
     >>> i > 1
     True
     """
--- a/src/toil/lib/retry.py
+++ b/src/toil/lib/retry.py
@@ -22,6 +22,11 @@ from builtins import next
 from contextlib import contextmanager
 import time
 import urllib.request, urllib.error, urllib.parse
+try:
+    from httplib import BadStatusLine
+except ImportError:
+    # This has moved in Python 3
+    from http.client import BadStatusLine
 import logging
 
 log = logging.getLogger( __name__ )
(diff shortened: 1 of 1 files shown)


adamnovak:
Would it make sense to require the message to be empty? It looks like when this happens randomly we usually get an empty status lime (maybe because something timed out), while any other data might indicate a non-HTTP server that isn't likely to start working correctly. Or it could be a partial status line from a working but slow server.

DailyDreaming:
The test isn't catching the error properly: http://jenkins.cgcloud.info/job/toil-pull-requests/4340/testReport/junit/src.toil.lib/retry/toil_lib_retry_retry_http/

My opinion is that leaving the message seems useful based on your diagnosis of different reasons it might omit or create partial messages.  Ultimately though, you've been dealing with this and know better whether it's helpful or not.

Looks good otherwise!

DailyDreaming:
Failed tests: http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2627/

These look like stochastic failures.  PR looks good!  Thanks @adamnovak .

adamnovak:
@DailyDreaming What's with Jenkins not reporting in and also failing the build when all the tests pass? I am seeing things in the log like this:

```
Checking if your working copy is clean ...
You have untracked files:
failed.log
fileToSort.txt
output.txt
sortedFile.txt
succeeded.log
tmp/
make: *** [check_clean_working_copy] Error 1
```

And also this:
```
Recording test results
Adding one-line test results to commit status...
Setting status of db26a25079876e458b5da77318099092d925f4cf to FAILURE with url http://jenkins.cgcloud.info/job/toil-pull-requests/4341/ and message: 'Build finished. 885 tests run, 39 skipped, 0 failed.'
Using context: jenkins
FileNotFoundException means that the credentials Jenkins is using is probably wrong. Or the user account does not have write access to the repo.
org.kohsuke.github.GHFileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/statuses/#create-a-status"}
```

adamnovak:
I think the failure to post back is due to @cgcloud-jenkins not being in DataBiosphere. I'm adding the bot manually to Toil with write access, but we probably want to put the account into DataBiosphere's bots team, which means we need to find somebody who can log in as it to turn on 2-factor.

PR #2353, opened 2018-08-23
Fix #2352 by quoting all Docker arguments with an appropriate quote()

### LINKED ISSUE #2352: Non-pipe Docker calls turn parameter lists into strings without escaping, allowing arguments to wreak havok
https://github.com/DataBiosphere/toil/blob/27ec2ff4d357fe80957c70d97c48c52214deea2f/src/toil/lib/docker.py#L327-L335

This code turns nicely separated lists of arguments like `['rm', '-f', 'Not Important.txt']` into single strings like `'rm -f Not Important.txt'`, which can then be mis-parsed as something like `['rm', '-f', 'Not', 'Important.txt']`, with potentially disastrous results.

We need to use `pipes.quote()` (or `shlex.quote()` for Python 3) to escape each argument separately if we're going to gom them together into a single string, as we do for the pipe case:

https://github.com/DataBiosphere/toil/blob/27ec2ff4d357fe80957c70d97c48c52214deea2f/src/toil/lib/docker.py#L320

I noticed this when I tried to pass a nice, quoted, human-readble plot title with spaces in it to an R script running via Docker, and got only the first word of the title in the resulting plot.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-335)
┆Issue Number: TOIL-335

adamnovak:


DailyDreaming:
Failing tests:
http://jenkins.cgcloud.info/job/toil-pull-requests/4343/
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2629/

- `testConcurrencyDynamic` fails occasionally because it's just setting a sleep duration and hoping that 8 concurrent processes write to a file before that timer ends.  It could probably be written better.
- The other is a cloud test, which seem to fail randomly sometimes, this one just waiting too long for the appliance to start.

This looks good to me, and I'll merge.

PR #2367, opened 2018-09-08
Set the default duration for batch systems with getWaitDuration().  Resolves #2362.  Resolves #2358.

### LINKED ISSUE #2358: Wrong environment variable in log message
https://github.com/DataBiosphere/toil/blob/f658acce644998ba3a565150b44e762a372d2e3b/src/toil/batchSystems/slurm.py#L190

The message says `TOIL_SLURM_RESOURCES` but really it reads from `TOIL_SLURM_ARGS`.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-340)
┆Issue Number: TOIL-340

### LINKED ISSUE #2362: getWaitDuration is defined in all batch systems but remains unused
Method [`getWaitDuration()`](https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L394) is defined in all batch systems but is never called in the codebase. The difference between [`getWaitDuration()`](https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L394) and [`sleepSeconds`](https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L397) is yet not clear to me. However, I think if we don't want to overload batch systems, `statePollingWait` should somehow honor `getWaitDuration()`  [here](https://github.com/DataBiosphere/toil/blob/619fbc3e012c0081911b266ec1e5ad4e2c031142/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L184).  For example, only use `statePollingWait` if greater than `getWaitDuration()`:

```python
        def checkOnJobs(self):
            ....
            polling_wait = self.getWaitDuration()
            current_wait = (datetime.now() - self._checkOnJobsTimestamp).total_seconds()

            if self.boss.config.statePollingWait > polling_wait:
                 polling_wait = self.boss.config.statePollingWait

            if self._checkOnJobsTimestamp and current_wait < polling_wait:
                return self._checkOnJobsCache
           .... 
```

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-342)
┆Issue Number: TOIL-342

DailyDreaming:
Addresses #2362 and #2358.

### DIFF
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -183,9 +183,8 @@ class LSFBatchSystem(AbstractGridEngineBatchSystem):
             return bsubline
 
     def getWaitDuration(self):
-        """We give LSF a second to catch its breath (in seconds)
-        """
-        return 15
+        """We give LSF a second to catch its breath (in seconds)"""
+        return 8
 
     @classmethod
     def obtainSystemConstants(cls):
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -54,6 +54,7 @@ class AbstractGridEngineBatchSystem(BatchSystemLocalSupport):
 
             """
             Thread.__init__(self)
+            self.boss = boss
             self.boss.config.statePollingWait = self.boss.config.statePollingWait or self.boss.getWaitDuration()
             self.newJobsQueue = newJobsQueue
             self.updatedJobsQueue = updatedJobsQueue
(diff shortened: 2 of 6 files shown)


src/toil/batchSystems/lsf.py
```diff
@@ -183,9 +183,8 @@ def prepareBsub(cpu, mem, jobID):
             return bsubline
 
     def getWaitDuration(self):
-        """We give LSF a second to catch its breath (in seconds)
-        """
-        return 15
+        """We give LSF a second to catch its breath (in seconds)"""
+        return 8
```
mr-c: Why was this lowered?

PR #2339, opened 2018-08-13
Issues/2335 add get status func

bencvdb:
This PR will add getStatus and getPIDStatus methods to the ToilStatus class, as well as corresponding unit tests.

PR #2376, opened 2018-09-23
Update CoreOS's AMI to the latest stable version.

DailyDreaming:


DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Old failure because of AMI change (apt update && apt upgrade): http://jenkins.cgcloud.info/job/toil-pull-requests/4446/

DailyDreaming:
Failures:
http://jenkins.cgcloud.info/job/toil-pull-request-integration-tests/2741/
http://jenkins.cgcloud.info/job/toil-pull-requests/4448/

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Jenkins, test this please.

DailyDreaming:
Okay, some notes.

I originally just changed the CoreOS AMI to the latest stable version.  I quickly learned that tests were failing because of the previous status PR https://github.com/DataBiosphere/toil/pull/2339 (the current failure on jenkins is still because of this).  And possibly some failed because of the CoreOS version.

The #2339 status tests written left behind some debris, so I added some content to the tearDown.  There were more problems though so this PR has been reverting the status PR as minimally as I could since it should be redone.  Also a couple of things in sortTest that at first I thought had been generating cruft but thought might be good to leave in.

However, the CoreOS AMI update here seems to fix about 5-6 AWS integration tests that recently, randomly started failing.

src/toil/test/utils/utilsTest.py
```diff
@@ -68,8 +68,17 @@ def setUp(self):
             self.correctSort.sort()
 
     def tearDown(self):
+        cruft = ['failed.log', 'fileToSort.txt',
+                 'output.txt', 'sortedFile.txt',
+                 'succeeded.log']
+        if os.path.exists(self.tempDir):
+            shutil.rmtree(self.tempDir)
+        if os.path.exists('tmp/'):
```
joelarmstrong: I like using rmtree rather than `rm -fr` above, but this line makes me a bit nervous. What if I (stupidly) had created my own folder in the toil directory called "tmp" where I (stupidly) kept something important? Or worse, what if I were running these tests from my home directory, or the root folder? A similar argument could go for the other special-cased cruft files, though that's stretching it.

I totally get the desire to clean up the cruft created by some of these tests, but surely it's better to clean up the cruft where it's created: making sure tempDir is always used for all test temp files. A stopgap may be to `chdir` to the self.tempDir directory in setUp. That might solve having to special case these stray files without having to change any other code.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -341,7 +341,7 @@ def descriptionMatches(ami):
         for attempt in retry(predicate= lambda e : isinstance(e, SSLError)):
             # SSLError is thrown when get_all_images times out
             with attempt:
-                amis = self._ctx.ec2.get_all_images(owners=['679593333241'])
+                amis = self._ctx.ec2.get_all_images(image_ids=['ami-70d5b608'])
```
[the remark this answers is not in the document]
DailyDreaming: Really good point, I'll test it!  Thanks!

src/toil/test/utils/utilsTest.py
DailyDreaming: That's a good point.  I'll see if I can't improve it.  Thanks again.

DailyDreaming: So the cruft generated was removed when the status PR code was removed. I've removed these steps and I'll create a new PR for status after this including this clean up step and ask for your review there if that's alright!

src/toil/provisioners/aws/awsProvisioner.py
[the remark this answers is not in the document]
DailyDreaming: You were right, it was only passing because it was in the correct region.  I've reverted it to the old code, which is the reference to the aws marketplace.  The latest version I tested this with I thought had errored on that "user" but now I suspect it may have been something different.

Anyway, should be fixed.  Thanks for catching my error!

src/toil/test/utils/utilsTest.py
joelarmstrong: Sounds great, thanks!

src/toil/provisioners/aws/awsProvisioner.py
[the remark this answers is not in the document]
joelarmstrong: No worries--I only know this because I made this mistake before!

joelarmstrong:
LGTM when tests pass!

DailyDreaming:
Okay, I think I figured out what happened with AWS breaking and which subsequently updating the CoreOS AMI fixed:

1. Toil fetches all of the images in the AWS marketplace.
2. Toil retries SSL on this fetching for 300 seconds.
3. The AWS marketplace finally became large enough that fetching took longer than 300 seconds.

I used a filter to reduce this time so that's why tests appeared to pass at first when I only fetched the AMI explicitly instead of the list.

DailyDreaming:
@joelarmstrong Awesome, thanks!

joelarmstrong:
Haha, nice. So will instance launches get faster as a bonus?

DailyDreaming:
@joelarmstrong Seems that way.  I ran this test code:
```
from boto.ec2.connection import EC2Connection

conn = EC2Connection('name', 'secret name')
inst = conn.get_all_images(owners=['679593333241'], filters={'name': 'CoreOS-stable-1688.5.3-hvm-0d1e0bd0-eaea-4397-9a3a-c56f861d2a14-ami-9e2685e3.4'})
for i in inst:
    print(i)
    print(i.id)
    print(i.description)
    print('\n\n')
```
Which ran in less than a second.  Versus this code:
```
from boto.ec2.connection import EC2Connection

conn = EC2Connection('name', 'secret name')
inst = conn.get_all_images(owners=['679593333241'])
for i in inst:
    print(i)
    print(i.id)
    print(i.description)
    print('\n\n')
```
Which takes a few minutes.

Hopefully things don't break.  XD  I'll try other versions too to see how far we can upgrade.

DailyDreaming:
Some tests seem to have failed for newer and more exciting reasons:
`The maximum number of security groups for VPC 'vpc-be2148d9' has been reached.`

This should be addressed now.

PR #2380, opened 2018-09-26
Stop logging every message twice in certain cases

joelarmstrong:
Occassionally toil will get in a state where every message is logged
twice. The root cause of these extra messages was a stray logging
handler set up in clusterScaler.py.


DailyDreaming:
Did a `git pull` to add fixes from the recently merged #2376.

DailyDreaming:
Jenkins, test this please.

PR #2382, opened 2018-09-26
first attempt at adding retries for batch commands

### LINKED ISSUE #2381: Please consider to add retry limit for SLURM
When using SLURM **sbatch** in busy cluster, sometime you will see error like this: 
```
sbatch:error:Batch job submission failed: Socket timed out on send/recv operation
```
This will happen for command **sacct**, **squeue** and **scontrol** as well.

When we read the code, we noticed that "src/toil/batchSystems/slurm.py" has caught the exception, but there is no code to retry submiting, it just print out some error messages. This will cause the workflow fail:
```
def submitJob(self, subLine):
    try:
        output = subprocess.check_output(subLine, stderr=subprocess.STDOUT)
        # sbatch prints a line like 'Submitted batch job 2954103'
        result = int(output.strip().split()[-1])
        logger.debug("sbatch submitted job %d", result)
        return result
    except subprocess.CalledProcessError as e:
        logger.error("sbatch command failed with code %d: %s", e.returncode, e.output)
        raise e
    except OSError as e:
        logger.error("sbatch command failed")
        raise e
```

Would you please consider to implement the retry function, and let user choose how many times to retry submitting a job? 
Note, there might be a same issue for SGE and Torque.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-349)
┆Issue Number: TOIL-349

mr-c:
Will fix #2381 for all abstractGridEngineBatchSystems

### DIFF
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -33,6 +33,21 @@ from toil.batchSystems.abstractBatchSystem import BatchSystemLocalSupport
 logger = logging.getLogger(__name__)
 
 
+def with_retries(operation, *args, **kwargs):
+    retries = 3
+    latest_err = None
+    while retries:
+        retries -= 1
+        try:
+            return operation(*args, **kwargs)
+        except subprocess.CalledProcessError as err:
+            latest_err = err
+            logger.error(
+                "Operation %s failed with code %d: %s",
+                operation, err.returncode, err.output)
+    raise latest_err
+
+
 class AbstractGridEngineBatchSystem(BatchSystemLocalSupport):
     """
     A partial implementation of BatchSystemSupport for batch systems run on a
@@ -116,21 +131,6 @@ class AbstractGridEngineBatchSystem(BatchSystemLocalSupport):
                 subLine = self.prepareSubmission(cpu, memory, jobID, command)
                 logger.debug("Running %r", subLine)
 
-                # submit job and get batch system ID
-                batchJobID = None
-                retries = 3
-                latest_err = None
-                while not batchJobID and retries:
-                    retries -= 1
-                    try:
-                        batchJobID = self.submitJob(subLine)
-                    except subprocess.CalledProcessError as err:
-                        latest_err = err
-                        logger.error(
-                            "submitJob failed with code %d: %s",
-                            err.returncode, err.output)
-                if not batchJobID and latest_err:
-                    raise latest_err
                 logger.debug("Submitted job %s", str(batchJobID))
 
                 # Store dict for mapping Toil job ID to batch job ID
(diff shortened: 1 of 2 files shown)


DailyDreaming:
@mr-c There are some tests failing for known reasons related to #2339 and also an outdated CoreOS AMI which #2376 should resolve.  Once that's addressed and merged, tests should pass as usual.

mr-c:
Thanks @DailyDreaming, once @xzhub tests this I'll merge it

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -116,7 +117,22 @@ def createJobs(self, newJob):
...
+                        batchJobID = self.submitJob(subLine)
+                        success = True
+                    except subprocess.CalledProcessError as err:
+                        latest_err = err
+                        logger.error(
+                            "submitJob failed with code %d: %s",
+                            err.returncode, err.output)
+                    finally:
+                        retries -= 1
+                if latest_err:
```
DailyDreaming: I think you may want to add `and not success` here.  Otherwise, if you have a failure, then retry and it succeeds, it looks like it will still have the error saved and raise.

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -275,6 +291,30 @@ def getJobExitCode(self, batchJobID):
...
+                try:
+                    exit_code = self.getJobExitCode(batchJobID)
+                except subprocess.CalledProcessError as err:
+                    latest_err = err
+                    logger.error(
+                        "getJobExitCode failed with code %d: %s",
+                        err.returncode, err.output)
+                finally:
+                    retries -= 1
+            if latest_err:
```
DailyDreaming: Same here, except with `and not exit_code==0`.

How would you feel about reducing the total code and trying to make one retry function for all of these cases?

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -346,16 +386,33 @@ def getIssuedBatchJobIDs(self):
...
+                try:
+                    batchIds = self.worker.getRunningJobIDs()
+                except subprocess.CalledProcessError as err:
+                    latest_err = err
+                    logger.error(
+                        "getRunningJobIDs failed with code %d: %s",
+                        err.returncode, err.output)
+                finally:
+                    retries -= 1
+                if latest_err:
```
DailyDreaming: Ditto.

DailyDreaming:
Rebased add fixes from #2376.

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -346,16 +366,19 @@ def getIssuedBatchJobIDs(self):
...
-        if (self._getRunningBatchJobIDsTimestamp and
-             (datetime.now() - self._getRunningBatchJobIDsTimestamp).total_seconds() < self.config.statePollingWait):
+        if (self._getRunningBatchJobIDsTimestamp and (
+                datetime.now() -
+                self._getRunningBatchJobIDsTimestamp).total_seconds() <
+                self.config.statePollingWait):
             batchIds = self._getRunningBatchJobIDsCache
         else:
-            batchIds = self.worker.getRunningJobIDs()
+            batchIds = with_retries(self.workfer.getRunningJobIDs)
```
DailyDreaming: `workfer` ;D

PR #2425, opened 2018-10-20
resource.getrusage returns KB already, so dividing by 1024 yields MB

ionox0:
not KB

Still would like confirmation that this memory for the report does indeed come from `resource.getrusage()` (I see that being used in the `worker` script, but can't exactly tie it to the reference used in `toilStats`


DailyDreaming:
Jenkins, add to the whitelist please.

DailyDreaming:
Jenkins, test this please.

PR #2430, opened 2018-10-30
Update the CoreOS AMI.

DailyDreaming:


DailyDreaming:
The two failures are because of a deleted bucket from our Production account: http://jenkins.cgcloud.info/job/toil-pull-requests/4642/#showFailuresLink  Merging this.

Issue #2434, opened 2018-11-01
`screen` on Toil AWS leader nodes starts /bin/sh instead of Bash


adamnovak:
When you run `screen` on a Toil-created cluster's head node, to run your workflow under it, it starts /bin/sh instead of your user's shell, `bash`, as set in /etc/passwd. This means things like the arrow keys to move the cursor do not work inside of `screen`. This in turn annoys our rotation students who are trying to run long-running Toil workflows on AWS from laptops.

If you `export SHELL` before running `screen`, it picks up that you are supposed to be using `bash`.

Toil's images ought to set themselves up so that `bash` is run in `screen`.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-384)
┆Issue Number: TOIL-384

DailyDreaming:
Just tested this and it works for me.  Let me know if it works alright for you: https://github.com/DataBiosphere/toil/pull/2435

DailyDreaming:
Closed by #2435 .

PR #2435, opened 2018-11-02
Default to bash on the leader node rather than sh.  Addresses #2434.

DailyDreaming:
Addresses #2434.


DailyDreaming:
@adamnovak I tried it briefly and echo'ed `$SHELL` inside of screen to demonstrate that it routed to bash.  Will merge.  Let me know if there are any issues!

Issue #2436, opened 2018-11-02
Metrics dashboard doesn't work if not on AWS / us-west-2


joelarmstrong:
The us-west-2 region got hardcoded in the prometheus config we use:

https://github.com/DataBiosphere/toil/blob/master/dashboard/prometheus/prometheus.yml#L36

The correct region (and maybe the correct cloud) should be detected and supplied as a parameter to the Docker wrapper, which could generate the right config YAML.

DailyDreaming:
Looks like the intended route is using `sed`: https://github.com/DataBiosphere/toil/blob/master/dashboard/prometheus/run.sh#L6

Thanks for the issue!  Will try to put something together that makes sense soon!

joelarmstrong:
Nice catch! I didn't realize it was already parameterized.

DailyDreaming:
I made a quick preliminary PR here: #2437.  Not sure when I'll get a chance to test it, but if you get to it before me, let me know!  :)

PR #2437, opened 2018-11-05
Specify region in prometheus.  Addresses #2436.

DailyDreaming:
Addresses #2436.

### DIFF
--- a/src/toil/utils/toilStatus.py
+++ b/src/toil/utils/toilStatus.py
@@ -165,7 +165,11 @@ class ToilStatus():
         :return: A string indicating the status of the PID of the workflow as stored in the jobstore.
         :rtype: str
         """
-        jobstore = Toil.resumeJobStore(jobStoreName)
+        try:
+            jobstore = Toil.resumeJobStore(jobStoreName)
+        except NoSuchJobStoreException:
+            return 'QUEUED'
+
         try:
             with jobstore.readSharedFileStream('pid.log') as pidFile:
                 pid = int(pidFile.read())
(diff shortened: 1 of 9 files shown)


joelarmstrong:
Thanks a ton! I should have a chance to test it within the next day or so. I'll let you know if it works or not!

DailyDreaming:
@joelarmstrong Awesome, thank you!

DailyDreaming:
@joelarmstrong I know you're busy with workflow runs atm, but if you get a chance this week to review this, I'd appreciate it!

I implemented what I was thinking of.  If it doesn't make sense to you, please let me know and I'll try the sentinel file idea (the status API used natively makes a sentinel file though and that's what it checks for the existence of indirectly, though I could try to do it more directly if that's what you were suggesting)!

src/toil/utils/toilStatus.py
```diff
@@ -184,21 +188,29 @@ def getStatus(jobStoreName):
...
         try:
             with jobstore.readSharedFileStream('succeeded.log') as successful:
-                return 'COMPLETED'
+                pass
+            return 'COMPLETED'
         except NoSuchFileException:
             try:
-                with jobstore.readSharedFileStream('failed.log') as failedLog:
-                    return 'ERROR'
+                with jobstore.readSharedFileStream('failed.log') as failed:
```
joelarmstrong: Sorry, I know this isn't relevant to this PR, but I just noticed: is `failed.log` ever actually deleted on resume? I can't see anywhere that the deletion happens. If not, a failed but successfully restarted workflow will forever get the FAILED status (until it completes).

DailyDreaming: That's a really good point.  I'll create an issue and a corresponding PR.  Thanks for catching this!

PR #2439, opened 2018-11-06
Fix massive slowdown when many jobs are running

joelarmstrong:
`checkForDeadlocks` currently gets run every leader iteration. That wasn't really the original intention -- it was meant to be run only when "nothing happened", i.e. there were no jobs completed or issued that iteration. I think this check may have disappeared in the leader refactor a while back.

`checkForDeadlocks` is fairly expensive because it needs to traverse all running jobs. That means it gets slower as more jobs are running; running the check every iteration makes you hit a hard cap of how many jobs you can actually keep running at any given time. How high that cap is depends on how long your jobs take to finish. In my case it seemed to be at about 300-400 simultaneous jobs. The autoscaler is totally unaware of this, meaning it provisions a *ton* of unusable capacity -- in my (extreme) case I had several thousand idle cores as the leader loop struggled to keep up.

Some aspects of the new leader code prevent us from reinstating the "nothing is happening" check properly, so I just put in an extra guard that makes it only check once every `deadlockWait` seconds. That should be sufficient to prevent this from slowing down the leader loop.

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -37,6 +37,7 @@ except ImportError:
     CWL_INTERNAL_JOBS = ()
 from toil.jobStores.abstractJobStore import NoSuchJobException
 from toil.jobStores.fileJobStore import FileJobStore
+from toil.lib.throttle import LocalThrottle
 from toil.provisioners.clusterScaler import ScalerThread
 from toil.serviceManager import ServiceManager
 from toil.statsAndLogging import StatsAndLogging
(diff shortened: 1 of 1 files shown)


DailyDreaming:
Failed tests: http://jenkins.cgcloud.info/job/toil-pull-requests/4693/

Jenkins, test this please.

DailyDreaming:
LGTM!

PR #2441, opened 2018-11-06
Make ServiceManager stop blocking while (most) services are established

joelarmstrong:
This is an old bit I never merged upstream--probably only useful for Cactus on very large trees.

The ServiceManager thread previously launched only one "batch" of
services at a time. This caused massive bottlenecks when many services
needed to be started at once. This causes all services to be issued
simultaneously, *unless* they have a special hierarchy (i.e. parent
and child services), in which case the old behavior is kept to avoid
problems where e.g., Spark master services would crowd out Spark
worker services.

### DIFF
--- a/src/toil/serviceManager.py
+++ b/src/toil/serviceManager.py
@@ -21,7 +21,7 @@ import time
 
 from threading import Thread, Event
 
-from bd2k.util.throttle import throttle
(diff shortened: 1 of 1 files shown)


DailyDreaming:
Thanks for submitting this!

PR #2469, opened 2018-11-16
Update cwltool version to the latest

psafont:
Right the referenced cwltool version has not been yet released


DailyDreaming:
@psafont Pip error is: `No matching distribution found for cwltool==1.0.20181115172452`

The latest pip shows is `1.0.20181116032456`.

It's also necessary to change the test commit hash when updating cwl (the git hash associated with the commit for that release):
https://github.com/DataBiosphere/toil/blob/master/src/toil/test/cwl/cwlTest.py#L50

This ensures the latest cwl tests are run with the corresponding version.

PR #2473, opened 2018-11-19
Clean up.

DailyDreaming:
Shouldn't be anything functional.  Mostly general clean up.  Deleted some unused functions.  Also noticed that `ResourceRequirement()` in `mesos.__init__.py` looked nearly identical to `Shape()`, so I attempted to consolidate those two.


DailyDreaming:
Waiting on #2437 to be merged, which might fix the failing tests.

PR #2479, opened 2018-11-25
WIP Translate cloud-config to ignition.

DailyDreaming:
The `cloud-config` format should be accurately translated over into `Container Linux Config` format as per documentation here: https://coreos.com/os/docs/latest/migrating-to-clcs.html

This requires a `ct` dependency to be set up as the new format using ignition is now a two-step process: https://coreos.com/os/docs/latest/overview-of-ct.html

This basically means converting the yaml format `Container Linux Config` into an ignition json using `ct` as needed.  This is kind of silly in my mind, and I want to discuss just using the json directly and cutting out the `ct` requirement, though this sacrifices some readibility.

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -14,7 +14,7 @@
 
 from __future__ import print_function
 import os
-import textwrap
+from toil.lib.misc import heredoc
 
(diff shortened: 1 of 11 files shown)


DailyDreaming:
Closing in favor of: #3488

PR #2480, opened 2018-11-25
Upgrade dockerfile to Ubuntu 16.04.

DailyDreaming:

PR #2485, opened 2018-11-26
Adjust docs.

DailyDreaming:
Doc builds were failing and now I know why.  Should have caught this earlier.


DailyDreaming:
Docker bug seems to have been the reason for one failure: https://status.docker.com/pages/533c6539221ae15e3f000031

The other should be fixed by: #2480.

Issue #2408, opened 2018-10-14
Toil passing tests with python 3.6.


cricketsloan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-285)
┆Issue Number: TOIL-285

PR #2494, opened 2018-11-30
fixes for python 3.x and slurm try 2

mr-c:
Sorry @markasbach, I screwed up your PR https://github.com/DataBiosphere/toil/pull/2492, here is a re-do

> This is a set of a few small commits to enhance python 3.x compatibility, especially when using slurm. Unfortunately, I'm not a git pro and didn't find an easy way so squash them to a single commit as requested by https://toil.readthedocs.io/en/latest/contributing/contributing.html#naming-conventions.

> Is related to #2408 but doesn't fix the issue (over 60 offline tests still fail).

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -349,7 +349,7 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
     >>> with open(path,'w') as f: 
     ...     _ = f.write('from toil.resource import ModuleDescriptor\\n'
     ...                 'print(ModuleDescriptor.forModule(__name__))')
-    >>> subprocess.check_output([ sys.executable, path ]).decode('utf-8') # doctest: +ELLIPSIS +ALLOW_BYTES
+    >>> subprocess.check_output([ sys.executable, path ]) # doctest: +ELLIPSIS +ALLOW_BYTES
     b"ModuleDescriptor(dirPath='...', name='foo', fromVirtualEnv=False)\\n"
 
     >>> from shutil import rmtree
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -349,7 +349,7 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
     >>> with open(path,'w') as f: 
     ...     _ = f.write('from toil.resource import ModuleDescriptor\\n'
     ...                 'print(ModuleDescriptor.forModule(__name__))')
-    >>> subprocess.check_output([ sys.executable, path ]) # doctest: +ELLIPSIS +ALLOW_BYTES
+    >>> subprocess.check_output([ sys.executable, path ]).decode('utf-8') # doctest: +ELLIPSIS +ALLOW_BYTES
     b"ModuleDescriptor(dirPath='...', name='foo', fromVirtualEnv=False)\\n"
 
     >>> from shutil import rmtree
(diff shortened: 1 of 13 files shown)


src/toil/resource.py
```diff
@@ -349,7 +349,7 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
     >>> with open(path,'w') as f: 
     ...     _ = f.write('from toil.resource import ModuleDescriptor\\n'
     ...                 'print(ModuleDescriptor.forModule(__name__))')
-    >>> subprocess.check_output([ sys.executable, path ]) # doctest: +ELLIPSIS +ALLOW_BYTES
+    >>> subprocess.check_output([ sys.executable, path ]).decode('utf-8') # doctest: +ELLIPSIS +ALLOW_BYTES
```
mr-c: Close!
```
343     >>> ModuleDescriptor.forModule('toil.resource') # doctest: +ELLIPSIS
344     ModuleDescriptor(dirPath='/.../src', name='toil.resource', fromVirtualEnv=False)
345 
346     >>> import subprocess, tempfile, os
347     >>> dirPath = tempfile.mkdtemp()
348     >>> path = os.path.join( dirPath, 'foo.py' )
349     >>> with open(path,'w') as f: 
350     ...     _ = f.write('from toil.resource import ModuleDescriptor\n'
351     ...                 'print(ModuleDescriptor.forModule(__name__))')
352     >>> subprocess.check_output([ sys.executable, path ]).decode('utf-8') # doctest: +ELLIPSIS +ALLOW_BYTES
Expected:
    b"ModuleDescriptor(dirPath='...', name='foo', fromVirtualEnv=False)\n"
Got:
    u"ModuleDescriptor(dirPath='/mnt/ephemeral/tmp/tmpuybOtA', name='foo', fromVirtualEnv=False)\n"

/mnt/ephemeral/workspace/toil-pull-requests/src/toil/resource.py:352: DocTestFailure
```
http://jenkins.cgcloud.info/job/toil-pull-requests/4812/testReport/junit/src.toil/resource/toil_resource_ModuleDescriptor/

markasbach: I've converted *all* calls to `check_output()` to return strings (as it was in Python 2) instead of bytes as most code in toil expects to work with the output using string operations. In this specific case, nothing is done with the result, so we could in principle leave it in its original form (without the `decode()`), but then the code here would be different to most other calls to `check_output()` within the toil codebase.

My suggestion would be to change the test to expect a string instead of bytes. But this is just a personal opinion.

src/toil/resource.py
```diff
@@ -349,7 +349,7 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
...
     b"ModuleDescriptor(dirPath='...', name='foo', fromVirtualEnv=False)\\n"
```
mr-c:
```suggestion
    u"ModuleDescriptor(dirPath='...', name='foo', fromVirtualEnv=False)\\n"
```
Sure, go for it

src/toil/resource.py
markasbach: Oh, I see now that it seems to be impossible to solve it as proposed - it will either fail on Python 2 or on Python 3. I'll revert to the original doctest setup in a second and update the PR.

markasbach: As I don't have commit rights to your repo, I've now added the markasbach-master branch to my own fork. Could you pull the last commit from https://github.com/markasbach/toil/tree/markasbach-master and update this PR please? Thanks!

mr-c: Done!

DailyDreaming:
Looks good to me!  Thank you @markasbach and @mr-c .

PR #2497, opened 2018-12-05
WIP: Try to fix #2488 with Google Analytics reporting from setup.py

### LINKED ISSUE #2488: TOIL Analytics
Benedict has asked that we figure out a way to see who is using toil besides us.

┆Issue is synchronized with this [JIRA Epic](https://ucsc-cgl.atlassian.net/browse/TOIL-419)
┆Issue Number: TOIL-419

adamnovak:
This (probably; I haven't been able to check on the receiving end) sends Google Analytics events (with random, non-persistent user IDs) when Toil's setup.py does things like install Toil.

Due to https://github.com/pypa/packaging-problems/issues/64 this won't work with installs from wheels, because wheels don't have a chance to run any code on installation. If we're serious about counting all the installations we'd need to not distribute wheels. We could also just instrument the library on load instead and count usage.

Before we merge this, someone ought to check if the events are actually being received, or if some more setup on the GA side is needed, or if I'm just sending them wrong.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,9 +12,118 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from setuptools import find_packages, setup
-import sys
+from setuptools import find_packages, setup, command
 
+try:
+    # On Python 3 urlopen is here
+    from urllib.request import urlopen
+    from urllib.error import URLError
+except ImportError:
+    # On Python 2 it is here
+    from urllib2 import urlopen, URLError
+
+import sys
+import uuid
+import inspect
+import importlib
+
+# We have an analytics system to keep track of installs. The next few functions
+# support that. TODO: We can't hook a wheel being installed, only made. See
+# <https://github.com/pypa/packaging-problems/issues/64>
+
+def phone_home(event):
+    """
+    Report setup.py activity events and Toil installation/packaging to the UCSC
+    Genomics Institute.
+    
+    Reports high bits of the IP address, Toil version, and the setuptools
+    command being used.
+    
+    Does not report any personally identifiable information. Uses a random
+    ephemeral user ID for every action, even if an install takes multiple
+    actions, or if the same machine installs multiple times.
+    """
+    
+    print("reporting {} to UCSC Genomics Institute".format(event))
+        
+    # Compose the GA report
+    report_data = '&'.join([
+        'v=1',
+        'aip=1',
+        'tid=UA-103168923-3',
+        'cid=' + str(uuid.uuid4()),
+        't=event',
+        'ec=lifecycle',
+        'ea=' + event,
+        'an=toil',
+        'av=' + version.distVersion,
+        'aid=edu.ucsc.gi.toil'
+    ]).encode('utf-8')
+    
+    # We assume that Toil version numbers don't need to be URL encoded.
+    # That means they can't contain +
+    
+    print(report_data)
+    
+    try:
+        # Send the data and open the reply with a 1 second timeout
+        handle = urlopen('https://www.google-analytics.com/collect', report_data, 1.0)
+        print("report sent")
+    except URLError:
+        print("could not report event; continuing")
+        
+def wrap_command(imported_command, command_name):
+    """
+    Wrap a single setuptools command class.
+    Uses its local scope so we can find the base class.
+    """
+    
+    class WrappedCommand(imported_command):
+        """
+        Setuptools install command wrapper that phones home.
+        """
+        
+        def run(self):
+            # First do the base work
+            # Problem: we need to get ahold of the original imported command but our enclosing scope vars may have changed.
+            imported_command.run(self)
+            
+            # Then report in
+            phone_home(command_name)
+            
+    return WrappedCommand
+    
+    
+def wrap_commands():
+    """
+    Wrap all commands from setuptools.command and make them phone home.
+    
+    Returns a dict from Setuptools command name to wrapped implementing class.
+    """
+    
+    to_return = {}
+    
+    for command_name in command.__all__:
+        # For each (basic) command defined by setuptools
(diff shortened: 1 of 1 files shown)


adamnovak:
I'm having trouble finding any sort of download count on libraries.io, and BigQuery doesn't seem like the most convenient place to get download count information, compared to something like Google Analytics which has a nice dashboard. Although right now my UCSC account isn't set up for either, so I suppose they're equally inconvenient to me at the moment.

#2488 came out of a meeting I was in with @DailyDreaming, @joelarmstrong, @benedictpaten, and @diekhans, where we were trying to work out The Future Of Toil. @benedictpaten seemed to think that we would have a lot of trouble getting grants to do Toil development specifically, as opposed to cramming Toil into other grants that use software that relies on it, unless we had some impressive usage statistics. Working out how to query the pip BigQuery download logs might be able to fill that need. But [as the documentation says](https://packaging.python.org/guides/analyzing-pypi-package-downloads/#background) those download counts are inflated by mirroring.

But if the install counter scares off all the users, that's also not going to help us fund Toil.

I'm not really happy with this PR's approach, either. There's no way to turn it off, and it's not going to be any more accurate than the BigTable numbers if it can't count wheels.

What would probably get us the most impressive grant-application-supporting figures would be growth-hacker-style telemetry from Toil itself that reported every workflow run and all the workers and leaders started up. But, taking off my analytics collection hat, people probably don't actually want to report that information, not because it actually reveals anything secret, but because nobody wants to hand out information for no reason. We could make it opt-in, but there's no good place to store an opt-in setting when you're constantly spinning up and tearing down clusters and nodes, and that also means we'd have a difficult extrapolation problem to interpret the data from the people who bothered to opt in.

Maybe the Right Way to do this is to make it less of an analytics ping and more of a service that actually does something for users. For example, when a Toil workflow starts, it could ask a UCSC server what the latest version of Toil is, and display a helpful message to the user to announce the new release if it's newer than the running Toil. Then we could also count the check-ins and get an idea of how many workflows are running in the world, while not just sucking up statistics about everyone for no benefit to them. That's the sort of feature that makes sense to be opt-out instead of opt-in, IMHO, which means we wouldn't have a terrible self-report bias problem trying to tally the results.

Another approach would be to set up an online user survey, and then have Toil or the installer encourage people to take it. That could get us useful information on what users want, but I don't think it will really be effective for counting them.

@mr-c How would you approach this? Just get someone on our team set up with the BigQuery data and declare the potentially inflated download counts to be a good enough metric?

adamnovak:
I think we also wanted the anonymized IP addresses from Google Analytics, because they could tell us who is using Toil at an institutional level, by coming from university or corporate-owned IP ranges. It looks like BigQuery only has country code resolution.

Here's the last 30 days of stats from BigQuery, broken out by country.

Row | num_downloads | place |  
-- | -- | -- | --
1 | 12282 | US |  
2 | 9293 | CN |  
3 | 3330 | KR |  
4 | 1897 | null |  
5 | 1704 | JP |  
6 | 1303 | FR |  
7 | 852 | CA |  
8 | 850 | NL |  
9 | 850 | GB |  
10 | 843 | RU |  
11 | 841 | IL |  
12 | 833 | ES |  
13 | 833 | DE |  
14 | 832 | IE |  
15 | 831 | HK |  
...

BigQuery says we have 39,679 Toil downloads in the past month, which seems a bit high. It also runs through ~30 GB of data, every time you want the download count, and I think there's a 1 TB free tier limit beyond which Google wants to charge for access. So if we really want to use this or to consult it often we're going to have to try to pull the data out for just Toil, so we're not scanning over every PyPI download of any project all the time, and process it ourselves.

@benedictpaten Is this level of detail and accuracy sufficient? Or do we really need to find a way to get uninflated download counts and WHOIS data on downloaders?

adamnovak:
I think we're getting lots of hits on old versions from bots that drive up our overall counts. Here's the top Toil versions by download count in the last 30 days:

Row | num_downloads | version |  
-- | -- | -- | --
1 | 1131 | 3.18.0 |  
2 | 380 | 3.17.0 |  
3 | 188 | 3.16.0 |  
4 | 163 | 3.15.0 |  
5 | 139 | 3.13.0 |  
6 | 128 | 3.5.0a1.dev251 |  
7 | 105 | 3.14.0 |  
8 | 101 | 3.16.0a1.dev12345 |  
9 | 100 | 3.12.0 |  
10 | 95 | 3.11.0 |  
11 | 95 | 3.7.0a1.dev365 |  
12 | 95 | 3.9.0a1.dev410 |  
13 | 95 | 3.8.0 |  
14 | 95 | 3.9.0a1.dev402 |  
15 | 95 | 3.7.0a1.dev368 |  

We have 416 versions up with any hits in the last 30 days, with the least downloaded being hit 47 times.

We can probably suss out a believable "real" download count from this, by subtracting out the hits on the versions that nobody has a real reason to want. I'd say that number is about 1.5K for Toil. We could even work out what countries the real downloads came from. But that's all we could get, and that's assuming nobody mirrors only the past one or two versions.

DailyDreaming:
@adamnovak I agree with @mr-c , this probably isn't the right approach.  For now, I think it's probably fine to stick to Google BigQuery and user stories.

That aside, I like your suggestion to fetch something useful, possibly as an enabled API which might be, as you say, The Right Way.  One suggestion I have...

Our AWS instance types are updated to a new static version every release (weeks/months), because the json is 100's of Mbs and takes forever to download if your internet is slow, so we do it for you.  I'd suggest maintaining a t2.micro server that would fetch this large amount of data regularly, shave it down, and then toil could fetch a small (~1 kb?) file dynamically with live AWS instance information.  This saves us from maintaining a static copy in toil itself, adds value, and would give us aws numbers at least.

@mr-c Do you feel like that's a reasonable approach?

PR #2495, opened 2018-11-30
WDL changes to enable resource usage.

DailyDreaming:

PR #2498, opened 2018-12-06
Remove linear-time bottleneck from Mesos batch system

joelarmstrong:
This fixes an annoying bug where Toil slows to a crawl when dealing with many jobs.

Previously the system had essentially 0 throughput when there were
more than ~100-300k queued jobs. The existing check took linear time
in the number of issued jobs, but was actually redundant for determining
the number of queued jobs, because `jobTypes` contains only queued
jobs. Checking the length of `jobTypes` is sufficient by itself to determine
whether there are jobs to issue to Mesos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,8 @@ install:
   - make develop extras=[aws,google] # adding extras to avoid import errors
 script:
   - TOIL_TEST_QUICK=True make test_offline
+env:
+  # Necessary to get boto to work in Travis's Ubuntu Precise
+  # environment (see #2498). Consider removing this if/when we
+  # transition to the Xenial environment.
+  - BOTO_CONFIG=/dev/null
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - "3.6"
   - "2.7"
-sudo: false
 install:
   - make prepare
   - make develop extras=[aws,google] # adding extras to avoid import errors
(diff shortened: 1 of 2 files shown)


joelarmstrong:
OK, awesome, I was trying to debug the travis error with absolutely no luck. That helps a lot.

The Jenkins failures have: `Failed to read cached auth token from file "/home/jenkins/.google_libcloud_auth.toil-dev": [Errno 2] No such file or directory: u'/home/jenkins/.google_libcloud_auth.toil-dev'`. Not sure if that's relevant; maybe an auth file was deleted?

PS: the reason `sudo:false` doesn't work anymore is because the `sudo: false` used to transition you to a different (Docker-based) infrastructure. [That's gone now.](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures) Which is probably good for us in the long run, but now means we have to figure out the root cause of this.

joelarmstrong:
Alright, the `sudo:  false` hack is gone, replaced with a different hack that seems to fix the problem. Would you prefer me to squash the Travis debugging commits or not?

DailyDreaming:
@joelarmstrong Nice, will merge this when tests pass.  Thanks for the debugging efforts!

joelarmstrong:
OK, thanks. I squashed, I want code archaeologists 1000 years from now to think I never make any mistakes.

DailyDreaming:
Just ran tests on another unrelated PR and Google is failing there too: #2495 .  This looks good to me and the failures have nothing to do with this PR.  Merging.

DailyDreaming:
The code archaeologists can blame me.  I don't mind my image in their textbooks as a cautionary tale.

PR #2500, opened 2018-12-13
Return stream FileID to fix #1020

### LINKED ISSUE #1020: writeGlobalFileStream should return FileID object with .size
Issue #1014 woks on adding attributes to the return value of writeGlobalFile. This issue concerns setting up the same for writeGlobalFileStream.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/TOIL-106)
┆Issue Number: TOIL-106

adamnovak:
This should solve #1020 by making `writeGlobalFileStream` yield a stream that, when written to, updates the size of a real genuine FileID.

This should solve the issue where `str`s were masquerading as FileIDs after being introduced to workflows by `writeGlobalFileStream`.

### DIFF
--- a/src/toil/fileStore.py
+++ b/src/toil/fileStore.py
@@ -180,6 +180,7 @@ class FileStore(with_metaclass(ABCMeta, object)):
         """
         raise NotImplementedError()
 
+    @contextmanager
     def writeGlobalFileStream(self, cleanup=False):
         """
         Similar to writeGlobalFile, but allows the writing of a stream to the job store.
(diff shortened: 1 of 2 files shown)


adamnovak:
@DailyDreaming Are all those GCE Jenkins tests supposed to fail? It looks like the nodes just never showed up.

DailyDreaming:
@adamnovak I think I know what's causing the Google tests to fail, it looks an expired key on AWS.  Will fix soon.

WDL tests should be working now.  There was a problem with one of the test files, not the code.

Since those are the only tests that failed, I think this is good to merge.

PR #2507, opened 2019-01-10
remove a few mentions of bd2k-python-lib

mr-c:


DailyDreaming:
@mr-c Testing is down but I see no problem with merging this.

PR #2508, opened 2019-01-10
Don't force all users to install pytest

mr-c:
Move the pytest-cov install to `make prepare` to join the existing
pytest installation command there.

This is a partial revert of beeb56c8b3c16e814e392987a327e904076a4cce


DailyDreaming:
Merging and testing this with the gitlab runner.

PR #2515, opened 2019-01-25
Removed HTML elements from amazon.rst

nikhil:


DailyDreaming:
Looks good to me.  Thanks for the fix!

PR #2521, opened 2019-01-31
Small typo fix: Replace 'insure' with 'ensure'.

arostamianfar:
While trying out Toil, I got the RuntimeError in awsProvisioner.py, and figured to send a small fix to change the rest of the code as well :)


DailyDreaming:
Looks good to me.  We're currently transitioning to new testing on gitlab so some tests are run but don't appear here at the moment.  Since this is a doc PR, I see no problem merging this.

Thank you @arostamianfar .

PR #2527, opened 2019-02-27
Add support for assuming roles

adamnovak:
To get Toil to work with our new gi-gateway AWS account setup, we need it to be able to assume roles based on `~/.aws/config`, at least when interacting with a cluster.

This PR tracks my progress towards that goal.


adamnovak:
OK, I'm now monkey patching Boto as soon as anything in the Toil module is loaded, if boto and boto3 are around. And instead of just cacheing credentials like we used to, I'm also querying boto3 to actually get the credentials. This seems to be working fine for me.

I had to patch some parts of the create-cluster/destroy-cluster flow to work under Python 3; they didn't seem to when I started. I guess we haven't tested them until now.

adamnovak:
@DailyDreaming Since this passes tests now, can you please go ahead and review it? Especially with regard to the wisdom of my decision to replace classes inside of Boto when Toil is imported.

benedictpaten:
@DailyDreaming Are you reviewing this, is important for CGL usage of Toil.

adamnovak:
@DailyDreaming @joelarmstrong Could one of you please review this PR?

PR #2531, opened 2019-03-09
Add --no-read-only option to the CWL runner.

arostamianfar:
This option is available in the cwltool, but not exposed in Toil.

Tested:
Built a docker image and ran the tool.


DailyDreaming:
Looks good to me too!

adamnovak:
This shouldn't have broken the CWL conformance tests, but they have broken: https://ci.commonwl.org/blue/organizations/jenkins/toil-conformance/detail/toil-conformance/685/pipeline

@mr-c It looks like the test script didn't have your bash-isms fix when it ran on Toil, and our `RUNNER=` was ignored. Do you have any idea how to get that pulled in and actually used?

mr-c:
The error with the runner script has been fixed. Sorry for the false alert!

PR #2549, opened 2019-03-18
Fix boto3 credentials for permanent credentials.

arostamianfar:
An [empty credentials file is created](https://github.com/DataBiosphere/toil/blob/044970558318d28654d2f79d1c9ccf31c4dfab81/src/toil/__init__.py#L613) in case of permanent credentials. The check [here](https://github.com/DataBiosphere/toil/blob/044970558318d28654d2f79d1c9ccf31c4dfab81/src/toil/__init__.py#L565) simply returns if an empty file exists, but does not set the credentials. As a result, credentials are not found and it shows an error.

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -14,6 +14,7 @@
 
 from __future__ import absolute_import
 
+import errno
 import logging
 import os
 import sys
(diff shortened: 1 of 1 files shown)


arostamianfar:
@adamnovak thanks for the quick review!
RE `errno`: I actually got an import error when it tried to run [`if e.errno == errno.ENOENT`](https://github.com/DataBiosphere/toil/blob/044970558318d28654d2f79d1c9ccf31c4dfab81/src/toil/__init__.py#L569). Seemed like a missing import issue, which is fixed now.

PR #2554, opened 2019-03-19
latest version more-itertools required python3.4+

aaronluoq:


adamnovak:
Does this PR have everything that was intended to be included? As is it appears unnecessary to me.

It looks like Toil never imports `more_itertools` or anything from it, whether it's running on Python 3.4+ or not, so why should Toil need to have a dependency on it?

If one of the modules we depend on tries to use but does not properly depend on more-itertools, then that module, rather than Toil, ought to be patched, or at least we ought to reference the bug in that module's install process that we are working around by depending on more-itertools ourselves.

aaronluoq:
> Does this PR have everything that was intended to be included? As is it appears unnecessary to me.
> 
> It looks like Toil never imports `more_itertools` or anything from it, whether it's running on Python 3.4+ or not, so why should Toil need to have a dependency on it?
> 
> If one of the modules we depend on tries to use but does not properly depend on more-itertools, then that module, rather than Toil, ought to be patched, or at least we ought to reference the bug in that module's install process that we are working around by depending on more-itertools ourselves.



> Does this PR have everything that was intended to be included? As is it appears unnecessary to me.
> 
> It looks like Toil never imports `more_itertools` or anything from it, whether it's running on Python 3.4+ or not, so why should Toil need to have a dependency on it?
> 
> If one of the modules we depend on tries to use but does not properly depend on more-itertools, then that module, rather than Toil, ought to be patched, or at least we ought to reference the bug in that module's install process that we are working around by depending on more-itertools ourselves.

If you use the command 
`pip install toil` 
to  install toil,you will get an error like the following:
`
more-itertools requires Python '>=3.4' but the running Python is 2.7.10
`
This is because toil 3.18.x depends on pytest 3.7.4 and pytest 3.7.4 needs to rely on more-itertools>=4.0.0，
Or better yet, toil3.18.x depends on the latest version of pytest

adamnovak:
OK, so more-itertools released 6.0.0, which matches `>=4.0.0` and drops support for Python 2. Pip finds it, because it satisfies Pytest, and installs it without checking to see whether it is actually compatible with the current Python. This PR forces a particular more-itertools version to work around Pytest's now-incorrect dependency specifier.

Probably the right solution here *is* to update to the latest pytest, assuming they've revised their dependency specifiers to retain Python 2 compatibility.

adamnovak:
@aaronluoq The issue with picking the wrong `more-itertools` seems to be fixed now in master, by upgrading `pytest`. Thanks for bringing this to our attention.

I'm going to close this PR, since it shouldn't be needed anymore.

PR #2557, opened 2019-03-20
Add missing imports

adamnovak:
I forgot a couple of the imports that the credential caching code uses when I moved it over.

I've now checked `src/toil/__init__.py` with `pylint`, so there shouldn't be any more missing imports in it.


adamnovak:
Actually #2558 is a better version of this.

PR #2558, opened 2019-03-20
Add missing imports to toil/__init__.py.

arostamianfar:
There were two other missing import errors when running on a clean EC2 instance. Would be great if there were tests covering these paths (perhaps there are integration tests already?).

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -17,12 +17,14 @@ from __future__ import absolute_import
 import errno
 import logging
 import os
-import sys
 import requests
+import sys
+import time
 from datetime import datetime
 from pytz import timezone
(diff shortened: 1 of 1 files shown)


adamnovak:
I'm not quite sure how to convince the tests to cover these codepaths, or why they aren't covered when we run our tests, which are also running on EC2. 

I did manage to turn up these particular imports with `pylint`. Maybe we should add pylint-based static analysis for missing imports to our tests.

arostamianfar:
Thanks, @adamnovak! I agree that covering these environment-dependent paths is difficult. +1 to adding pylint checks! We caught a lot of errors both ways in previous projects (missing imports and unused ones).

Issue #2166, opened 2018-04-03
Tool can't authenticate to Docker registry


brucehoff:
I understand that a cwl tool descriptor can tell Toil to run code in a Docker container.  However, we have a case in which the `docker run` command has custom parameters, so we wish to invoke `docker run` directly.  Before running a container the repository must be pulled from a registry (unless it's already resident on the worker node.)  We have found that the worker node is unable to pull the specified image from the Docker registry *if the image is private*.  That is, the worker seems unable to use the Docker registry credentials cached locally.  To reproduce the issue:

### First, the happy case that works:
docker_run_tool.cwl:
```
#!/usr/bin/env cwl-runner
cwlVersion: v1.0
class: CommandLineTool
baseCommand: docker
arguments: [run, --rm, -i, $(inputs.docker_repository)]

stdout: stdout.txt

inputs:
 docker_repository:
  type: string

outputs:
 stdout:
  type: stdout
```
docker_run_job_hello_world.cwl:
```
docker_repository: hello-world
```
(`hello-world` is a public Docker repository hosted on DockerHub.)

```
toil-cwl-runner --retryCount 0 docker_run_tool.cwl docker_run_job_hello_world.cwl 
...
SAGE-C02H51GMDV17.local 2018-04-03 13:45:22,980 MainThread INFO toil.leader: Finished toil run successfully
2018-04-03 13:45:22,980 - toil.leader - INFO - Finished toil run successfully
{
    "stdout": {
        "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709", 
        "basename": "stdout.txt", 
        "nameext": ".txt", 
        "nameroot": "stdout", 
        "http://commonwl.org/cwltool#generation": 0, 
        "location": "file:///Users/bhoff/Documents/toil/stdout.txt", 
        "class": "File", 
        "size": 0
    }
SAGE-C02H51GMDV17.local 2018-04-03 13:45:23,133 MainThread INFO toil.common: Successfully deleted the job store: <toil.jobStores.fileJobStore.FileJobStore object at 0x10a109c50>
}2018-04-03 13:45:23,133 - toil.common - INFO - Successfully deleted the job store: <toil.jobStores.fileJobStore.FileJobStore object at 0x10a109c50>
```

### Now, replace the repository with a private one:

docker_run_job_private_repository.cwl:
```
docker_repository: docker.synapse.org/syn4224222/dm-python-example
```

[...]

PR #2561, opened 2019-03-20
Add support for running custom init commands in docker.

arostamianfar:
This solves the private docker authentication issue #2166 as one can add `TOIL_CUSTOM_DOCKER_INIT_COMMAND='pip install awscli && eval $(aws ecr get-login --no-include-email --region us-east-1)'` to the environment variable, which is then executed on the docker instance in all workers (note: this is easy to do manually on the master, but not possible on the workers when running a large cluster).

Notes:
* I figured to make this more generic to allow any type of custom initialization and not be AWS ECR specific. We can even consider removing `waitForKey.sh` in favor of this.
* Using an environment variable is not ideal, but passing the argument through all provisioners made it complicated.
* This does not work if the Toil image itself is hosted in a private repository, but can be addressed either via using a custom AMI or adding another similar pre-init command that is run prior to `toil-{role}.service`.

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -216,6 +216,22 @@ def applianceSelf(forceDockerAppliance=False):
         return checkDockerImageExists(appliance=appliance)
 
 
+def customDockerInitCmd():
+    """
+    Returns the custom command (if any) provided through the ``TOIL_CUSTOM_DOCKER_INIT_COMMAND``
+    environment variable to run just after the docker leader/worker has been initialized.
+    This can be useful for doing any custom initialization on instances (e.g. authenticating to
+    private docker registries). An empty string is returned if the environment variable is not
+    set.
+
+    :rtype: str
+    """
+    command = lookupEnvVar(name='user-defined custom docker init command',
+                           envName='TOIL_CUSTOM_DOCKER_INIT_COMMAND',
+                           defaultValue='')
+    return command.replace("'", "'\\''")  # Ensure any single quotes are escaped.
+
+
 def lookupEnvVar(name, envName, defaultValue):
     """
     Use this for looking up environment variables that control Toil and are important enough to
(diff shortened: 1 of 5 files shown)


src/toil/__init__.py
```diff
@@ -17,12 +17,14 @@
 import errno
 import logging
 import os
-import sys
 import requests
+import sys
```
arostamianfar: I'll remove these once #2558 is merged.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -402,6 +402,10 @@ def _getCloudConfigUserData(self, role, masterPublicKey=None, keyPath=None, pree
         if keyPath:
             mesosArgs = keyPath + ' ' + mesosArgs
             entryPoint = "waitForKey.sh"
+        customDockerInitCommand = customDockerInitCmd()
+        if customDockerInitCommand:
+            mesosArgs = " ".join(["'" + customDockerInitCommand + "'", entryPoint, mesosArgs])
+            entryPoint = "customDockerInit.sh"
```
adamnovak: Oh, this is chaining on top of `waitForKey.sh` if it is used. It looked to me like it was replacing it.

src/toil/__init__.py
```diff
@@ -214,6 +216,22 @@ def applianceSelf(forceDockerAppliance=False):
         return checkDockerImageExists(appliance=appliance)
 
 
+def customDockerInitCmd():
+    """
+    Returns the custom command (if any) provided through the ``TOIL_CUSTOM_DOCKER_INIT_COMMAND``
```
adamnovak: The new variable needs to be documented in https://github.com/DataBiosphere/toil/blob/master/docs/appendices/environment_vars.rst I think.

arostamianfar: Done!

src/toil/provisioners/abstractProvisioner.py
arostamianfar: Correct. We could actually do the chaining in-place here and remove the need for `waitForKey.sh` altogether (I can do that in another PR if desired).

PR #2569, opened 2019-03-25
Add private IP support to node operations.

### LINKED ISSUE #2568: Private IP support for Toil leader
Toil currently assumes that the leader node has a public IP ([example](https://github.com/DataBiosphere/toil/blob/863e2f0e0a0e7bca028ba7b66a9e6a4201ec9f58/src/toil/provisioners/node.py#L240)). However, there are cluster setups where public IPs are not available.

It would be ideal to have an option to use private IPs.

Note: once the cluster is launched, Mesos works fine with private IPs, so this is just a blocker for launching the cluster and using tools like `toil ssh-cluster`.

arostamianfar:
With this change, Toil can now launch clusters in environments that just have private IPs (tested on AWS only). Other SSH settings (e.g. ProxyCommand) are assumed to be provided through SSH config.

Fixes #2568

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -401,7 +401,7 @@ class AWSProvisioner(AbstractProvisioner):
         while True:
             time.sleep(a_short_time)
             instance.update()
-            if instance.ip_address or instance.public_dns_name:
+            if instance.ip_address or instance.public_dns_name or instance.private_ip_address:
                 logger.debug('...got ip')
                 break
 
--- a/src/toil/provisioners/node.py
+++ b/src/toil/provisioners/node.py
@@ -33,6 +33,7 @@ class Node(object):
     def __init__(self, publicIP, privateIP, name, launchTime, nodeType, preemptable, tags=None):
         self.publicIP = publicIP
         self.privateIP = privateIP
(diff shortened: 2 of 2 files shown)


buddyroo30:
I don't think this is actually fixed. I just tested in our VPC that only assigns private addresses and it fails. Basically line 391 in awsProvisioner.py checks if there is a public IP address and fails if it doesn't find [one ](https://github.com/DataBiosphere/toil/blob/master/src/toil/provisioners/aws/awsProvisioner.py#L391). I commented out the 2 lines of code here and then I was able to start the cluster.

arostamianfar:
@buddyroo30 thanks for flagging this! Looks like a regression introduced in https://github.com/DataBiosphere/toil/pull/3357 ... could you please file a separate issue for this?
P.S. I'm no longer actively working on Toil, but hoping Adam or you can take on fixing this small bug :)

buddyroo30:
Sure, I'm happy to file a new issue for this.

buddyroo30:
Okay, I submitted a [new](https://github.com/DataBiosphere/toil/issues/4278) issue for this, but not able to assign to anyone. TBH, I just started trying out Toil yesterday in our private VPC, so I don't really know the code and would not really be the right person to do a fix for this (although it does seem like a fairly simple fix).

PR #2570, opened 2019-03-26
Optimize filestore links

### LINKED ISSUE #2550: Determine and implement what consistent behavior would be between the CWL workdir and the CWL jobstore.
See #2534.

### LINKED ISSUE #2551: Enable caching on grid engine (optimize hard links and copying to reduce redundancy).
See #2534.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-47)
┆Issue Number: TOIL-47

liinus:
Use --linkImports also when importing files from jobStore to workDir and allow symlinks across devices (resolves #2550, resolves #2551)

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -374,7 +374,7 @@ class FileJobStore(AbstractJobStore):
         # If local file would end up on same file system as the one hosting this job store ...
         if os.stat(jobStoreFilePath).st_dev == os.stat(localDirPath).st_dev:
             # ... we can hard-link the file, ...
-            if symlink:
+            if symlink or self.linkImports:
                 try:
                     os.symlink(jobStoreFilePath, localFilePath)
                 except OSError as e:
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -387,14 +387,14 @@ class FileJobStore(AbstractJobStore):
                         raise
             else:
                 try:
-                    os.link(jobStoreFilePath, localFilePath)
+                    os.link(os.path.realpath(jobStoreFilePath), localFilePath)
                 except OSError as e:
                     if e.errno == errno.EEXIST:
                         # Overwrite existing file, emulating shutil.copyfile().
                         os.unlink(localFilePath)
                         # It would be very unlikely to fail again for same reason but possible
                         # nonetheless in which case we should just give up.
-                        os.link(jobStoreFilePath, localFilePath)
+                        os.link(os.path.realpath(jobStoreFilePath), localFilePath)
                     else:
                         logger.critical('jobStoreFilePath: ' + jobStoreFilePath + ' ' + str(os.path.exists(jobStoreFilePath)))
                         logger.critical('localFilePath: ' + localFilePath + ' ' + str(os.path.exists(localFilePath)))
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -90,6 +90,7 @@ class FileJobStore(AbstractJobStore):
         if not os.path.isdir(self.jobStoreDir):
             raise NoSuchJobStoreException(self.jobStoreDir)
         super(FileJobStore, self).resume()
+        self.linkImports = self.config.linkImports
 
     def robust_rmtree(self, path, max_retries=3):
         """Robustly tries to delete paths.
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -371,35 +371,33 @@ class FileJobStore(AbstractJobStore):
         self._checkJobStoreFileID(jobStoreFileID)
         jobStoreFilePath = self._getAbsPath(jobStoreFileID)
         localDirPath = os.path.dirname(localFilePath)
-        # If local file would end up on same file system as the one hosting this job store ...
-        if os.stat(jobStoreFilePath).st_dev == os.stat(localDirPath).st_dev:
-            # ... we can hard-link the file, ...
-            if symlink or self.linkImports:
-                try:
+        if symlink or self.linkImports:
+            try:
+                os.symlink(jobStoreFilePath, localFilePath)
+            except OSError as e:
+                if e.errno == errno.EEXIST:
+                    # Overwrite existing file, emulating shutil.copyfile().
+                    os.unlink(localFilePath)
+                    # It would be very unlikely to fail again for same reason but possible
+                    # nonetheless in which case we should just give up.
                     os.symlink(jobStoreFilePath, localFilePath)
-                except OSError as e:
-                    if e.errno == errno.EEXIST:
-                        # Overwrite existing file, emulating shutil.copyfile().
-                        os.unlink(localFilePath)
-                        # It would be very unlikely to fail again for same reason but possible
-                        # nonetheless in which case we should just give up.
-                        os.symlink(jobStoreFilePath, localFilePath)
-                    else:
-                        raise
-            else:
-                try:
+                else:
+                    raise
+        # If local file would end up on same file system as the one hosting this job store ...
+        elif os.stat(jobStoreFilePath).st_dev == os.stat(localDirPath).st_dev:
+            try:
+                os.link(os.path.realpath(jobStoreFilePath), localFilePath)
+            except OSError as e:
+                if e.errno == errno.EEXIST:
+                    # Overwrite existing file, emulating shutil.copyfile().
+                    os.unlink(localFilePath)
+                    # It would be very unlikely to fail again for same reason but possible
+                    # nonetheless in which case we should just give up.
                     os.link(os.path.realpath(jobStoreFilePath), localFilePath)
-                except OSError as e:
-                    if e.errno == errno.EEXIST:
-                        # Overwrite existing file, emulating shutil.copyfile().
-                        os.unlink(localFilePath)
-                        # It would be very unlikely to fail again for same reason but possible
-                        # nonetheless in which case we should just give up.
-                        os.link(os.path.realpath(jobStoreFilePath), localFilePath)
-                    else:
-                        logger.critical('jobStoreFilePath: ' + jobStoreFilePath + ' ' + str(os.path.exists(jobStoreFilePath)))
-                        logger.critical('localFilePath: ' + localFilePath + ' ' + str(os.path.exists(localFilePath)))
-                        raise
+                else:
+                    logger.critical('jobStoreFilePath: ' + jobStoreFilePath + ' ' + str(os.path.exists(jobStoreFilePath)))
+                    logger.critical('localFilePath: ' + localFilePath + ' ' + str(os.path.exists(localFilePath)))
+                    raise
         else:
             # ... otherwise we have to copy it.
             shutil.copyfile(jobStoreFilePath, localFilePath)


src/toil/jobStores/fileJobStore.py
```diff
@@ -370,35 +371,33 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
         self._checkJobStoreFileID(jobStoreFileID)
         jobStoreFilePath = self._getAbsPath(jobStoreFileID)
         localDirPath = os.path.dirname(localFilePath)
-        # If local file would end up on same file system as the one hosting this job store ...
-        if os.stat(jobStoreFilePath).st_dev == os.stat(localDirPath).st_dev:
-            # ... we can hard-link the file, ...
-            if symlink:
-                try:
+        if symlink or self.linkImports:
```
adamnovak: The [`readGlobalFile` implementation in the `NonCachingFileStore`](https://github.com/DataBiosphere/toil/blob/37b55569122e7d3f5e260d149bf4d62ca71187f1/src/toil/fileStore.py#L1665-L1675) assumes that if you don't pass `symlink=True` to `readFile` you will always get a fresh, mutable copy, which is how it manages to ignore its own `mutable` parameter. If we change things so that that won't always be the case (i.e. if the user passes `--linkImports`), we need to change the `readGlobalFile` implementation to deal with that.

Given the current code here, which will symlink or hard link even if the `readFile` caller passes `symlink=False`, the only way to get a mutable copy to fulfill a user request for one would be to make a copy ourselves in the  `NonCachingFileStore`.

Also, since this PR touches the semantics of the `symlink` kwarg, which doesn't have a description in any of the docstrings for `readFile` or `readGlobalFile`, this PR should define the semantics for that parameter and document them in the docstrings of those functions. Since the option is now going to result in either a symlink *or* a hardlink, as appropriate, the name should probably be changed to something like `allowLink`. It could also be removed altogether, in favor of bringing the `mutable` option down into `readFile`; maybe @DailyDreaming can comment on how `symlink` and `mutable` need to differ.

src/toil/jobStores/fileJobStore.py
```diff
@@ -90,6 +90,7 @@ def resume(self):
         if not os.path.isdir(self.jobStoreDir):
             raise NoSuchJobStoreException(self.jobStoreDir)
         super(FileJobStore, self).resume()
+        self.linkImports = self.config.linkImports
```
adamnovak: I don't think it makes sense to tie this to `linkImports`. If we can, we should *always* symlink or hardlink from the file store, unless the caller doing the read explicitly needs a mutable copy or a non-symlink file.

If we want to permit links by default, with the current symlink-only-if-allowed-explicityl semantics, we could just touch [here](https://github.com/DataBiosphere/toil/blob/8057b47f3841f177e040def8ef1b2f1968f0cd94/src/toil/cwl/cwltoil.py#L300) and [here](https://github.com/DataBiosphere/toil/blob/8057b47f3841f177e040def8ef1b2f1968f0cd94/src/toil/cwl/cwltoil.py#L309) where Toil's CWL runner reads from the `FileStore`, and have it pass `symlink=True`.

But I think really what we should do is just change the default for `symlink` to True, and make workflows that can't use symlinks demand non-symlinks or mutable copies.

src/toil/jobStores/fileJobStore.py
DailyDreaming: I agree that `symlink` and `mutable` seem to serve similar purposes in `readGlobalFile` and `mutable` alone could be brought in and used instead of `symlink`, though perhaps a clearer name would be nice.

src/toil/jobStores/fileJobStore.py
DailyDreaming: I agree with this.  This seems like adding a `dev` lever to force symlinks in special cases, but it really should be implemented as a broader design decision.  I think defaulting to using symlinks as a default should be tested.

src/toil/jobStores/fileJobStore.py
adamnovak: I think that `mutable` is not currently implemented right. The code as it is now will create a hard link even if you pass `mutable=True`, if it thinks everything is on the same device. And the official docs for `mutable` are missing; the base class refers you to the `CachingFileStore` docstring, which doesn't exist.

adamnovak:
@liinus We recently merged a major revision of the file store code, and changed the way it determines what kind of links to make. I think we fixed #2551. I think those changes conflict with these.

Can you check to see what if anything you would still want to change about the current code? If you still see something to change you can go ahead and update this PR; if not I will go ahead and close it.

liinus:
@adamnovak  This still seem to be a problem in 4.0.0, and I think it was also a problem in 3.23.1 and 3.24.0. I'm using toil-cwl-runner.

If I distribute 16 SGE jobs on a single node, all having the reference genome or other large files on a nfs drive, it will generate 16 copies on the local device (/tmp). This is also the case when running toil-cwl-runner locally on a single host where the input data is on a nfs.

In the most typical cluster environment, you are using a local directory for your filestore (for performance) and nfs for your input data and /jobstore. That means it will always copy the data to the filestore.

adamnovak:
We're still [defaulting to `--disableCaching=True`](https://github.com/DataBiosphere/toil/blob/8b45f1a1296d7ae0e3b5288db8712bf8cba3181f/src/toil/common.py#L122) even in 4.0, and SGE runs can't use caching anyway as it stands because we don't have an opportunity to run cleanup on the node after we're done using it to run jobs.

On `--batchSystem singleMachine`, if you pass `--disableCaching=False`, you should get one copy of the data on the local filesystem, with hard links.

With the default `--disableCaching=True`, on either single machine or grid engine, you just get copies, unless your workflow passes `symlink=True` when reading files. I did some digging and it looks like we could easily handle the everything-on-one-filesystem case better with some hard links, but whyen things are on different filesystems, the copying was by design. I didn't want to change the user-visible behavior and show workflows symlinks when they were expecting files. If, for example, the workflow downloads a file, and then mounts the directory it put it in into a Docker container to process the file with a tool, then if it actually got a symlink to some NFS path, it will be a broken symlink inside the container.

So I think there are a couple ways we can improve this:

1. Add `mutable` as an argument to `readFile()` in the `AbstractJobStore`, pull it through in the `NonCachingFileStore`, and make all the implementations make hard links when hard linking is possible and `mutable` is false. This should definitely happen, but it won't fix the case where the job store is on NFS and the Toil work directories (where the file store temp directories are, and where jobs download files to) are not.

2. Make `symlink` default to `True` instead of `False`. Don't wait around for user workflows to declare that they can accept symlinks; just change all their inputs out from under them and make them pass `symlink=False` whenever they actually need ordinary files. This will make people start seeing symlinks from the toil work directories to their job stores on NFS, which will save a bunch of copies, but it's going to break probably every workflow that uses Dockerized tools and isn't already passing `symlink=False` for its downloads.

3. Bolt the cleanup workaround I wrote for Kubernetes onto the Grid Engine batch systems, so they can use the `CachingFileStore`, with the downside being that as soon as the last job running on a node finishes, it deletes the whole cache. This might thrash the cache more than a proper cleanup-at-the-end system would, and it would still be making one local copy where a symlink might do, but it would save you from 16 downloads of the same file for 16 simultaneous jobs, and it would be transparent to existing workflows.

@liinus As a workaround, you should be able to modify your Toil Python workflows to specify `symlink=True` when downloading files from the file store, and avoid the 16 copies with the current codebase. It already knows how to make symlinks when it is explicitly allowed to. I'm not sure if a similar workaround would work for your CWL workflow; are you passing your files along to CWL-managed Dockerized tools? And if we feed `cwlrunner` symlinks, is it smart enough to make its own local copies or do something else clever when it is told to pass those symlink files to a Dockerized tool that won't have the destination directories mounted?

liinus:
@adamnovak I have been running cwltoil and docker fine when forcing symlinks to be created, see the pull request https://github.com/DataBiosphere/toil/pull/2570. It looks like the docker daemon works with having symlinks in bind mounts. This works cross device, so if jobstore is on a nfs it still works the same.

```
$ mkdir jobstore
$ echo Hello world! > jobstore/file.txt
$ ln -s jobstore filestore
$ docker run --name test -v $PWD/filestore:/mnt alpine:latest cat /mnt/file.txt
Hello world!
$ docker inspect test | jq '.[].Mounts'
[
  {
    "Type": "bind",
    "Source": "/home/dockervm/filestore",
    "Destination": "/mnt",
    "Mode": "",
    "RW": true,
    "Propagation": "rprivate"
  }
]
$ docker --version
Docker version 19.03.5, build 633a0ea838
$ uname -a
Linux dockervm 4.15.0-20-generic #21-Ubuntu SMP Tue Apr 24 06:16:15 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

adamnovak:
That's not a symlink *in* a bind mount, that looks more like a symlink *as* a bind mount. You madfe a symlink `workdir` that points to `jobstore`, and told Docker to mount `workdir` into the container. I think what's happening is the symlink is being followed and what it points to is being mounted, which is sensible behavior.

Here's an example of a symlink *within* the mounted directory not working right:

```
docker --version
# Reports: Docker version 19.03.6, build 369ce74a3c
uname -a
# Reports: Linux octagon 4.15.0-99-generic #100-Ubuntu SMP Wed Apr 22 20:32:56 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
mkdir docktest
cd docktest
mkdir mountme
cd mountme
ln -s /etc/passwd ./passwd
cd ..
md5sum mountme/passwd 
# Reports: f507e88e5ffef9a40b66c2665188c30c  mountme/passwd
docker run --rm -ti -v `pwd`/mountme:/mnt ubuntu:20.04
# In the container:
md5sum /mnt/passwd
# Reports: 9911b793a6ca29ad14ab9cb40671c5d7  /mnt/passwd
```

The symlink points to one file outside the container, but a different file (or potentially no file at all) inside the container. When symlinking files from the file store, we're not going to be symlinking a file or directory and then mounting it, we're probably (at least the way our lab uses Docker in Toil jobs) going to symlink several files into a directory and then mount the directory in the container.

Does this actually work in toil-cwl-runner? Maybe cwltool creates a separate Docker mount for every file, since it has such a good idea of what files exactly a tool is going to use.

liinus:
It was a while ago I used my forked branch but in that example everything worked fine using `cwltoil` (`toil-cwl-runner`) with `--batchSystem gridEngine`.

I don't see why this would be a problem in 4.0 either:
```
$ cat test.cwl
class: CommandLineTool
cwlVersion: v1.0
requirements:
  DockerRequirement:
    dockerPull: alpine:latest
inputs:
  file:
    type: File
  dir:
    type: Directory
baseCommand: echo
outputs: []
$ mkdir workdir
$ mkdir dir
$ touch file.txt dir/file1.txt dir/file2.txt
$ toil-cwl-runner --logDebug --workDir workdir --jobStore jobstore --clean never --cleanWorkDir never --noLinkImports test.cwl --file file.txt --dir dir
[...]
dockervm-docker 2020-05-08 21:52:01,492 MainThread INFO toil: Running Toil version 4.0.0-1c5f5fa1d8811660ee9b1b602b41607b63c6dd05.
[...]
$ docker \
	    run \
	    -i \
	    --mount=type=bind,source=/home/dockervm/workdir/node-86e2ff33-0970-462c-814c-cb10d036e97b-a62b6e1f74fc452a8dff7e5a02a101a7/tmp9gi8_7y1/3e7a9d00-3f30-49d5-9f52-f43e1b1fa297/tg9vht3mp/tmp-outzu6pku7v,target=/hEZweT \
	    --mount=type=bind,source=/home/dockervm/workdir/node-86e2ff33-0970-462c-814c-cb10d036e97b-a62b6e1f74fc452a8dff7e5a02a101a7/tmp9gi8_7y1/3e7a9d00-3f30-49d5-9f52-f43e1b1fa297/t3l2jkcpdg6soc4lp,target=/tmp \
	    --mount=type=bind,source=/home/dockervm/dir,target=/var/lib/cwl/stgba174f94-0bfd-48f4-803d-448f9119ef02/dir,readonly \
	    --mount=type=bind,source=/home/dockervm/workdir/node-86e2ff33-0970-462c-814c-cb10d036e97b-a62b6e1f74fc452a8dff7e5a02a101a7/tmp9gi8_7y1/3e7a9d00-3f30-49d5-9f52-f43e1b1fa297/tmp1ngfprgd.tmp,target=/var/lib/cwl/stg5e487c51-a076-43ba-a471-429c90b9f644/file.txt,readonly \
	    --workdir=/hEZweT \
	    --read-only=true \
	    --user=1116:1005 \
	    --rm \
	    --env=TMPDIR=/tmp \
	    --env=HOME=/hEZweT \
	    --cidfile=/home/dockervm/workdir/node-86e2ff33-0970-462c-814c-cb10d036e97b-a62b6e1f74fc452a8dff7e5a02a101a7/tmp9gi8_7y1/3e7a9d00-3f30-49d5-9f52-f43e1b1fa297/t3l2jkcpdwfg939kj/20200508220234-032190.cid \
	    alpine:latest \
	    echo
[...]
```

The temp file for `file.txt` in workdir (`tmp1ngfprgd.tmp`) is bind mounted directly. If this file is a symlink on the host, the container should see the host file (as in my previous example) and not a symlink (as your /etc/passwd example). For directories, it looks like it's bypassing the workdir completely and mounts the host path directly (`--mount=type=bind,source=/home/dockervm/dir,target=/var/lib/cwl/stgba174f94-0bfd-48f4-803d-448f9119ef02/dir,readonly`). Considering how bind mounts are created, dockerized tools should have no problem if all files and directories in workdir are symlinks to the jobstore.

I also noticed that `--noLinkImports` is not functioning. I still get symlinks in my jobstore:
```
$ tree jobstore/
jobstore/
├── files
│   ├── for-job
│   │   └── kind-file_home_dockervm_test.cwl
│   │       └── instance-0h4pcq3d
│   ├── no-job
│   │   ├── file-f7rk0z_c
│   │   │   └── file2.txt -> /home/dockervm/dir/file2.txt
│   │   ├── file-l_hmiyz_
│   │   │   └── file1.txt -> /home/dockervm/dir/file1.txt
│   │   ├── file-y4qw4aqw
│   │   │   └── file.txt -> /home/dockervm/file.txt
│   │   └── file-yjlt80pn
│   │       └── stream
│   └── shared
│       ├── config.pickle
│       ├── environment.pickle
│       ├── pid.log
│       ├── rootJobReturnValue
│       ├── rootJobStoreID
│       └── succeeded.log -> /home/dockervm/succeeded.log
├── jobs
│   └── kind-file_home_dockervm_test.cwl
└── stats
    ├── statsu33nyo55.new
    └── statsz6rk4ajq

13 directories, 12 files
```

adamnovak:
That's weird that --noLinkImports doesn't work. I thought we didn't
link imports by default. Maybe that's its own bug?

It does sound like Dockerized tools under CWL would not have a problem
with getting symlinks for all their files, because cwltool is mounting
each file individually.

But the way Docker is used in toil-vg would break, because in that
workflow we download a bunch of files into a directory and mount the
directory. If those files are changed to symlinks in a new version of
Toil, that workflow will break and would have to be modified to run
correctly on the new Toil (by explicitly passing symlink=False, I
guess, or else changing the way it does Docker mounts). So I'm
hesitant to change the current default behavior to make everything
symlinks if there's an almost equally good, backward-compatible
solution (getting caching working).

DailyDreaming:
Hmmm... this is a few years old now and I think that this has been addressed properly: https://github.com/DataBiosphere/toil/blob/1f0930b7b9ecc31ca556d15ab07ff836ba85eb23/src/toil/jobStores/fileJobStore.py#L280 ?

If not, please reopen and I will do a more thorough review on how to adapt this to the current codebase to fix the issue.

PR #2584, opened 2019-04-05
Add an option to set disableCaching to false.

### LINKED ISSUE #2586: Caching is disabled by default and no way to turn it on
Caching is currently disabled by default ( from https://github.com/DataBiosphere/toil/commit/2e54371ea316fe2803ce53d1477a51fadcfb5d2e ) and there is no way to turn it on.

Options:
* Change the default value of `disableCaching` back to False
* Add an `enableCaching` option to preserve the current default behavior.

arostamianfar:
Fixes #2586

This was changed in https://github.com/DataBiosphere/toil/commit/2e54371ea316fe2803ce53d1477a51fadcfb5d2e , which effectively disables caching altogether (there is no way to turn caching on). Was that the intention?

This PR changes the existing default behavior though. I can add an `enableCaching` option if we want to keep the existing default behavior. Please let me know!

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -117,7 +117,7 @@ class Config(object):
         self.rescueJobsFrequency = 3600
 
         # Misc
-        self.disableCaching = False
+        self.disableCaching = True
         self.disableChaining = False
         self.maxLogFileSize = 64000
         self.writeLogs = None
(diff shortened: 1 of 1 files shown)


DailyDreaming:
@arostamianfar Caching's had some odd issues and is not recommended as a default.  I left a comment in #2586.

arostamianfar:
@DailyDreaming I updated the code to add a `disableCaching=False` option (unfortunately, the built-in `store_true` or `store_false` don't work with `=False`).

PR #2587, opened 2019-04-05
Make sure to use --tmpdir-prefix to place workdir and jobstore

### LINKED ISSUE #2576: toil writes to /tmp although --tmpdir-prefix is set
Hi. I'm running toil (3.19.0) like:
```
$ /toil-cwl-runner --cleanWorkDir always \               
        --logFile $(CURDIR)/log --outdir $(CURDIR)/results \                
        --jobStore file:///$(CURDIR)/$(RUN)/job_store \               
        --workDir $(CURDIR)/work --tmpdir-prefix=$(CURDIR)/tmp --clean always                  
```
My pipeline creates quite some huge intermediate files, which toil writes to `/tmp`. I can't have that in `/tmp` since no space, but `--tmpdir-prefix` doesn't seem to have much effect.

adamnovak:
I think this will fix #2576, although I still haven't gotten myself set up with CWL enough to actually check that.

I also think we're still making temporary output directories under --tmpdir-prefix and not --tmp-outdir-prefix. I'm not quite sure if that's the expected behavior.

@gijzelaerr Are you able to test this to see if it solves your issue?


arostamianfar:
@adamnovak Thanks for this fix! I also ran into the same issue and can verify that this fixes the problem.
However, I also noticed that the `workDir` as set in line 1137 (i.e. the cwl local job store) does not get cleaned up after the job is done (even with `--clean always --cleanWorkDir always`), which leads to temporary files being left in the workers. I understand that this is a separate issue, but was wondering if there is a simple fix that can be applied here? If not, I can create a separate issue for this.

adamnovak:
@arostamianfar Is that a regression introduced by this PR? If not I think it should be a separate issue; it sounds like the sort of thing I actually need to reproduce to debug.

Issue #2589, opened 2019-04-10
Restarting two workflows caused overlap in outputs files


ionox0:
I intentionally killed two running workflows (using `bkill` command on our lsf cluster), and attempted to restart them using `--restart`. 

I should note that the cause of this may be that the intended way to end a workflow is the `toil kill` command, and perhaps using `bkill` causes some part of the jobstore or a perhaps the jobStoreFileID to be deleted or cached or corrupted in some way. 

The result of this was that one of the runs has been restarted from files of the other jobstore, which resulted in duplicate outputs at the end of both runs. 

The commands were:
```
toil-cwl-runner \
--logFile /home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-2.0.0/log/cwltoil.log \
--jobStore file:///home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-2.0.0/tmp/jobstore-40fb870c-5612-11e9-addb-ec0d9a88a0ee \
--batchSystem lsf \
--workDir /home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-2.0.0/tmp/ \
--outdir /home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-2.0.0 \
--writeLogs /home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-2.0.0/log \
--logLevel DEBUG \
--stats   \
--cleanWorkDir onSuccess  \
--defaultDisk 10G  \
--preserve-environment PATH PYTHONPATH TOIL_LSF_ARGS  \
--retryCount 2  \
--no-container   \
--defaultMem 10G  \
--disableCaching   \
--maxLogFileSize 20000000000  \
--restart  \
/home/johnsoni/pipeline_2.0.0/ACCESS-Pipeline/workflows/ACCESS_pipeline.cwl \
inputs_group_1.yaml
```

and
```
toil-cwl-runner \
--logFile /home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-group_3-2.0.0/log/cwltoil.log \
--jobStore file:///home/johnsoni/Projects/Ian/06302R_06302S/set_4/Project_06302_R/Project-group_3-2.0.0/tmp/jobstore-0e487c9a-561f-11e9-84eb-8cdcd4acd360 \
--batchSystem lsf \
[...]

PR #2593, opened 2019-04-11
Improve NoSuchFileException usage and robustness

adamnovak:
In #2589 there's evidence that NoSuchFileException might be getting constructed with too many arguments somewhere, masking a real error. This fixes that, so the real error might become apparent.

I couldn't find any code that actually seemed to be doing that, but this makes the exception handle extra arguments being dumped into it.


src/toil/jobStores/abstractJobStore.py
```diff
@@ -80,15 +80,24 @@ def __init__(self, jobStoreFileID):
...
+        # most three arguments error reported in
+        # https://github.com/DataBiosphere/toil/issues/2589#issuecomment-481912211
         if customName is None:
             message = "File '%s' does not exist." % jobStoreFileID
         else:
             message = "File '%s' (%s) does not exist." % (customName, jobStoreFileID)
+        
+        if extra:
+            # Append extra data.
+            message += " " + " ".join((str(x) for x in extra))
```
DailyDreaming: Would prepending a string here explaining what this is trying to solve be good to have in front of the mysterious args?

adamnovak: OK I put a generic "Extra info:" message. Do you think that works? Or do we want to complain that the extra info shouldn't be provided?

DailyDreaming: I think that ought to do the trick.  At least you know it's intentional and not something else leaking over.

mr-c:
Huh, why isn't  GitLab testing this?

adamnovak:
Gitlab has decided it doesn't like this branch. Pussing the same commit to another branch kicked off the tests.

Issue #2598, opened 2019-04-12
Problem with helloWorld.py


pb-cdunn:
I think I''ve followed the tutorial faithfully, but I'm stuck. I don't see any indication of what causes the error.

```
% python helloWorld.py --logDebug --retryCount=0 file:my-job-store
ERROR:toil.batchSystems.singleMachine:Got exit code 1 (indicating failure) from job _toil_worker helloWorld file:/scratch/cdunn/repo/ipa/examples/greg200k-sv2/my-job-store k/x/jobm9HHl4.
```
[toil-error.txt](https://github.com/DataBiosphere/toil/files/3073688/toil-error.txt)

pb-cdunn:
And btw, I do not see `./my-job-store/`. I tried creating that as a directory, but that didn't help either. I deleted that directory and tried again; same error.

pb-cdunn:
Ugh! I see what I did wrong. I had to switch to python3 in order to install Toil successfully with `$PYTHONUSERBASE` (`pip3 install --user`), but I forgot to switch to python3 when running the example. It works perfectly now.

PR #2599, opened 2019-04-12
Use current python executable

cdunn2001:
I had to switch to python3 to get Toil working with `$PYTHONUSERBASE` (#2598). But I still have old python in my `$PATH`. This uses the proper python. (Some tests could be updated too, but this is the only actual bug I found.)

### DIFF
--- a/src/toil/wdl/toilwdl.py
+++ b/src/toil/wdl/toilwdl.py
@@ -18,6 +18,7 @@ from __future__ import division
 import argparse
 import os
 import logging
+import sys
 
 from toil import subprocess
 from toil.wdl.wdl_analysis import AnalyzeWDL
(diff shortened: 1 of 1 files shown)


src/toil/wdl/toilwdl.py
```diff
@@ -141,7 +142,8 @@ def main():
         sWDL.write_mappings(aWDL)
     else:
         wdllogger.debug('WDL file compiled to toil script.  Running now.')
-        cmd = ['python', sWDL.output_file]
+        exe = sys.executable
```
DailyDreaming: In some cases, `sys.executable` will be `None`, so this should be:

`exe = sys.executable if sys.executable else 'python'`

Otherwise, lgtm.

cdunn2001: I will update this PR. Glad you spotted that.

pb-cdunn:
ETA on getting this change into PyPI? From PyPI, I am using `3.19.0`.

adamnovak:
We're planning a Toil release on May 24th; this fix should be in PyPI
after that.

PR #2607, opened 2019-04-18
Improve PR import

### LINKED ISSUE #2605: Note in the contributing rules that PRs that change after review need re-review
The Github review-required mechanism dosn't enforce that, if you make changes to fix your PR to pass tests after it was reviewed, it requires a new review for the changes before it can be merged. But that's the workflow we want to use.

This isn't documented in the contributing document. We should update it to say that that's what we want to happen.

### LINKED ISSUE #2606: PR import is hard
There's a block of commands in the contributing docs that you need to run to import a PR.

We should have a script that makes this easy. Ideally it should pick up the issue being fixed from Github when given the PR, but at minimum it should automate the import and not clobber the `$USER` variable in people's shells.

adamnovak:
This will fix #2606.

I had to import a PR for testing, so instead of just doing it I wrote a script.

While I was touching the contributing docs to reference the script, I fixed #2605.

### DIFF
--- a/contrib/admin/test-pr
+++ b/contrib/admin/test-pr
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# test-pr: submit a PR from an external contributor for Gitlab testing
+
+set -e
+
+# Check arguments
+if [[ "${#}" -ne 3 ]] ; then
+    printf >&2 "test-pr: submit a PR from an external contributor for Gitlab testing\n\n"
+    printf >&2 "\tusage: test-pr GH_USER SOURCE_BRANCH DEST_BRANCH\n\n"
+    printf >&2 "\texample: test-pr arostamianfar defaultcaching issues/2586-allow-enabling-caching\n\n"
+    exit 1
+fi
+
+# Parse arguments
+GH_USER="${1}"
+shift
+SOURCE_BRANCH="${1}"
+shift
+DEST_BRANCH="${1}"
+shift
+
+echo "Fetch ${SOURCE_BRANCH} from ${GH_USER}"
+git fetch "git@github.com:${GH_USER}/toil.git" "${SOURCE_BRANCH}"
+echo "Push to Toil ${DEST_BRANCH}"
+git push git@github.com:DataBiosphere/toil.git "FETCH_HEAD:refs/heads/${DEST_BRANCH}"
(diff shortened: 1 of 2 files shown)


docs/contributing/contributing.rst
```diff
@@ -360,22 +360,21 @@ Pull Requests
...
+  repository. This can be accomplished with (from a Toil clone):
   
   .. code-block:: bash
   
-      USER="jqrandom"
-      SOURCE_BRANCH="fix-thingy"
-      DESTINATION_BRANCH="issues/1234-fix-thingy"
-      git fetch git@github.com:${USER}/toil.git ${SOURCE_BRANCH}
-      git push git@github.com:DataBiosphere/toil.git FETCH_HEAD:refs/heads/${DESTINATION_BRANCH}
+      ./contrib/admin/test-pr jqrandom their-branch issues/123-fix-description-here
```
DailyDreaming: Why is the github user named: "jqrandom"?

contrib/admin/test-pr
```diff
@@ -0,0 +1,26 @@
...
+shift
+SOURCE_BRANCH="${1}"
+shift
+DEST_BRANCH="${1}"
+shift
+
+echo "Fetch ${SOURCE_BRANCH} from ${GH_USER}"
+git fetch "git@github.com:${GH_USER}/toil.git" "${SOURCE_BRANCH}"
+echo "Push to Toil ${DEST_BRANCH}"
+git push git@github.com:DataBiosphere/toil.git "FETCH_HEAD:refs/heads/${DEST_BRANCH}"
```
DailyDreaming: I think we could have travis do this?  Travis could be given credentials and when tests run it'll push a new gitlab branch.

adamnovak: I don't want PRs automatically pulled into branches unless they're from people we trust, because running on our CI nodes gives them access into our AWS infrastructure. Especially with the autoscaling setup I'm working on, where the CI nodes will have access to start more machines, which they will eventually need in order to run any launch-cluster tests.

Either someone has to manually check the commit, or we trust the person enough that we can just give them access to push to a branch themselves, is my thinking.

docs/contributing/contributing.rst
adamnovak: This should match better.

contrib/admin/test-pr
DailyDreaming: Alright, that sounds sane.

docs/contributing/contributing.rst
DailyDreaming: Ha, looks good.  Thanks!

PR #2609, opened 2019-04-18
Update cwltool to handle the new conformance tests.

DailyDreaming:
Looks like there were a couple of breaking changes in cwltool recently, but nothing too bad.


mr-c:
Thanks!

Issue #2579, opened 2019-04-02
Dependency versions too strict


gijzelaerr:
hi! It looks like the Toil requirement versions are a bit too strict, which confuses the dependency resolver when Toil is installed with other projects in the same virtual env.

For example, there is no need to set `requests` exactly to `2.18.4`:

https://github.com/DataBiosphere/toil/blob/master/setup.py#L44

Strict requirements are usually set in a `requirements.txt` file, but `setup.py` is generally more flexible.

The python documentation has a page about this:

https://packaging.python.org/discussions/install-requires-vs-requirements/

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-39)
┆Issue Number: TOIL-39

DailyDreaming:
Ties into https://github.com/DataBiosphere/toil/issues/1812 .  Closing this in favor of that ticket and adding a comment there.

PR #2610, opened 2019-04-18
Loosen and update the 'requests' package version.

DailyDreaming:
Partially addresses #2579.


DailyDreaming:
@adamnovak I've added a minimum.  Will test locally to ensure it works.

PR #2641, opened 2019-04-22
StatsAndLogging(): ensure json.load() receives only string input

kwwette:
I've been trying out Toil on a cluster with Python 3.4. It looks like the stats and logging files are being written in binary mode (`wb`) but `json.load()` doesn't handle binary files [until Python 3.6](https://docs.python.org/3.6/library/json.html#json.load). I therefore get errors like this:
```
  File "/usr/lib64/python3.4/threading.py", line 911, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.4/threading.py", line 859, in run
    self._target(*self._args, **self._kwargs)
  File ".../toil/src/toil/statsAndLogging.py", line 142, in statsAndLoggingAggregator
    if jobStore.readStatsAndLogging(callback) == 0:
  File ".../toil/src/toil/jobStores/fileJobStore.py", line 480, in readStatsAndLogging
    callback(fH)
  File ".../toil/src/toil/statsAndLogging.py", line 114, in callback
    stats = json.loads(statsStr, object_hook=Expando)
  File "/usr/lib64/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```
This small patch first reads the file into a variable `statsStr`, then calls `.decode()` if it is not a `str`. `json.loads()` is then called with guaranteed string input.

Fixes https://github.com/DataBiosphere/toil/issues/2643

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - "3.5"
   - "3.6"
   - "2.7"
 install:
--- a/docs/contributing/contributing.rst
+++ b/docs/contributing/contributing.rst
@@ -368,10 +368,10 @@ Pull Requests
   
   .. code-block:: bash
   
-      USER="jqrandom"
+      CONTRIBUTOR="jqrandom"
       SOURCE_BRANCH="fix-thingy"
       DESTINATION_BRANCH="issues/1234-fix-thingy"
-      git fetch git@github.com:${USER}/toil.git ${SOURCE_BRANCH}
+      git fetch git@github.com:${CONTRIBUTOR}/toil.git ${SOURCE_BRANCH}
       git push git@github.com:DataBiosphere/toil.git FETCH_HEAD:refs/heads/${DESTINATION_BRANCH}
      
   These steps must be repeated every time the PR submitter updates their PR,
--- a/src/toil/statsAndLogging.py
+++ b/src/toil/statsAndLogging.py
@@ -109,7 +109,10 @@ class StatsAndLogging( object ):
         startClock = getTotalCpuTime()
 
         def callback(fileHandle):
-            stats = json.load(fileHandle, object_hook=Expando)
+            statsStr = fileHandle.read()
+            if not isinstance(statsStr, str):
+                statsStr = statsStr.decode()
+            stats = json.loads(statsStr, object_hook=Expando)
             try:
                 logs = stats.workers.logsToMaster
             except AttributeError:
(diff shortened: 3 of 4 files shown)


mr-c:
Thank you @kwwette . FYI, Python 3.4 reached end of life last month: https://devguide.python.org/#status-of-python-branches

I've added Python 3.5 to the testing framework as part of this PR. I'll investigate Python 3.7 compatibility in another branch.

kwwette:
Thanks @mr-c !

mr-c:
@kwwette I'm surprised that last commit was necessary, looks like it passed all the tests?

kwwette:
I found another [issue](https://github.com/DataBiosphere/toil/pull/2641/commits/93a8cdb9f86ce878b63faf82b9cff206c72cdb96) in `StatsAndLogging()` to do with byte/string input, so thought I'd tack it on to this MR:

When writing log file output from `jobLogList` it looks like either byte or string input in `l` is possible, but when 'l' is bytes an error is raised:
```
  File ".../statsAndLogging.py", line 97, in <genexpr>
    f.writelines((l + '\n').encode('utf-8') for l in jobLogList)
TypeError: can't concat str to bytes
```
This is with Python 3.6, so may be a more generic Python 2-vs-3 issue. The patch just tries to `.decode()` l if it is bytes, but passes if it is a string.

mr-c:
@kwwette Can you add a test that triggers the bug you fixed in https://github.com/DataBiosphere/toil/pull/2641/commits/93a8cdb9f86ce878b63faf82b9cff206c72cdb96 ?

mr-c:
FYI, a subset of the test are run on Python 2.7, 3.5, and 3.6: https://travis-ci.org/DataBiosphere/toil/builds/523844416

kwwette:
@mr-c : Two additional patches:

`StatsAndLogging(): add newline to 'l' only if required`: it looks like the lines read from the log file may or may not still have newlines; when I ran on my system, the extra newline causing the log files to be written double-spaced. This patch adds newlines only if required.

`jobStoreTest.py: add test for StatsAndLogging.writeLogFiles()`: as requested, adds a new test `testWriteLogFiles()` in `jobStoreTest.py` to test the `StatsAndLogging.writeLogFiles()` function with string/byte input and with/without newlines.

mr-c:
Thank you @kwwette This looks great to me! Evidently the ucsc-ci gitlab runner is having some issues, but I don't think that is due to your PR

adamnovak:
It looks like a job on the Gitlab runner was canceled, and didn't do its cleanup. It left root-owned stuff laying around that Gitlab insists on trying to delete, as non-root, before running the next job.

I've manually cleaned it up, and I'm hoping that tomorrow I can get the Docker-based runner working and we won't have to deal with this problem anymore.

PR #2646, opened 2019-04-24
SlurmBatchSystem(): include both exit code and terminating signal in job return status

kwwette:
- Out-of-memory-killed jobs will show zero exit code but non-zero terminating signal
  code, so only returning the exit code will tell Toil these jobs have succeeded.
- If terminating signal code N > 0, instead return 128+N (following Bash behaviour).

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -113,9 +113,12 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                 state, exitcode = values
                 logger.debug("sacct job state is %s", state)
                 # If Job is in a running state, return None to indicate we don't have an update
-                status, _ = exitcode.split(':')
-                logger.debug("sacct exit code is %s, returning status %s", exitcode, status)
-                return (state, int(status))
+                status, signal = [int(n) for n in exitcode.split(':')]
+                if signal > 0:
+                    # A non-zero signal may indicate e.g. an out-of-memory killed job
+                    status = 128 + signal
+                logger.debug("sacct exit code is %s, returning status %d", exitcode, status)
+                return (state, status)
             logger.debug("Did not find exit code for job in sacct output")
             return None
 
@@ -148,9 +151,12 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
             try:
                 exitcode = job['ExitCode']
                 if exitcode is not None:
-                    status, _ = exitcode.split(':')
-                    logger.debug("scontrol exit code is %s, returning status %s", exitcode, status)
-                    rc = int(status)
+                    status, signal = [int(n) for n in exitcode.split(':')]
+                    if signal > 0:
+                        # A non-zero signal may indicate e.g. an out-of-memory killed job
+                        status = 128 + signal
+                    logger.debug("scontrol exit code is %s, returning status %d", exitcode, status)
+                    rc = status
                 else:
                     rc = None
             except KeyError:


src/toil/batchSystems/slurm.py
```diff
@@ -113,9 +113,12 @@ def _getJobDetailsFromSacct(self, slurmJobID):
                 state, exitcode = values
                 logger.debug("sacct job state is %s", state)
                 # If Job is in a running state, return None to indicate we don't have an update
-                status, _ = exitcode.split(':')
-                logger.debug("sacct exit code is %s, returning status %s", exitcode, status)
-                return (state, int(status))
+                status, signal = [int(n) for n in exitcode.split(':')]
+                if signal > 0:
+                    # A non-zero signal may indicate e.g. an out-of-memory killed job
+                    status += 128 + signal
```
DailyDreaming: Should this be `status = 128 + signal`?

src/toil/batchSystems/slurm.py
```diff
@@ -148,9 +151,12 @@ def _getJobDetailsFromScontrol(self, slurmJobID):
             try:
                 exitcode = job['ExitCode']
                 if exitcode is not None:
-                    status, _ = exitcode.split(':')
-                    logger.debug("scontrol exit code is %s, returning status %s", exitcode, status)
-                    rc = int(status)
+                    status, signal = [int(n) for n in exitcode.split(':')]
+                    if signal > 0:
+                        # A non-zero signal may indicate e.g. an out-of-memory killed job
+                        status += 128 + signal
```
DailyDreaming: And this?

src/toil/batchSystems/slurm.py
kwwette: @DailyDreaming The idea was to preserve both the exit code of the Slurm job and the code of any terminating signal, just for the sake of completeness. There are probably cases where a program might trap a particular signal and then return a specific exit code, which Slurm may then register as both exit code and signal code being non-zero. I'm not encountered this possibility, but if it exists it would seem prudent to preserve as much information as possible. I'm not sure whether the combined value of `status` is accessible to a user of Toil, but it is recorded in the logs, and so preserving maximum information may help someone debug such a problem with their jobs, should it arise.

src/toil/batchSystems/slurm.py
kwwette: See above

src/toil/batchSystems/slurm.py
DailyDreaming: @kwwette Would that make it difficult to distinguish?  For the sake of argument `1 + 128 + 3` could look a lot like `2 + 128 + 2`.  This isn't strictly bash behavior and might confuse users.

I think the most important case is to return a non-zero exit code when appropriate for the purpose of retrying, and I think that this is really good for that.  I'd like to argue for more strictly conforming to expected return code structure though.

kwwette: @DailyDreaming I got the Bash behaviour idea from [here](https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html), but I don't feel strongly about this though, so we could think of an alternative.

What about instead `status = exit_code + 256 * signal`? This would avoid conflicting with the [0-255 range of valid exit codes](https://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Exit-Status.html), and would be unambiguous: you could always distinguish e.g. `1 + 256 * 3` from `2 + 256 * 2` by unpacking the bits of `status`, i.e. `exit_code = status mod 256`, `signal = (status - exit_code) / 256`.

adamnovak: The actual interface for getting at exit status and signal numbers is documented [here](http://man7.org/linux/man-pages/man2/waitpid.2.html).

```
WIFEXITED(wstatus)
              returns true if the child terminated normally, that is, by
              calling exit(3) or _exit(2), or by returning from main().

       WEXITSTATUS(wstatus)
              returns the exit status of the child.  This consists of the
              least significant 8 bits of the status argument that the child
              specified in a call to exit(3) or _exit(2) or as the argument
              for a return statement in main().  This macro should be
              employed only if WIFEXITED returned true.

       WIFSIGNALED(wstatus)
              returns true if the child process was terminated by a signal.

       WTERMSIG(wstatus)
              returns the number of the signal that caused the child process
              to terminate.  This macro should be employed only if WIFSIG‐
              NALED returned true.
```

There's an exit status *only* if the child process terminated normally, and a signal number *only* if the child process was terminated by a signal. These cases are mutually exclusive: if the child calls exit() from a signal handler, you'll see a normal termination and not a signal termination. That's how Bash gets away with packing these both into one integer.

I'm not sure if Slurm will report something else for the case where e.g. it sent the child a signal, but the child managed to exit normally before receiving it.

I think if the exit code is nonzero we should report only that, and if the exit code is zero and the signal is nonzero we should report `128 + signal` as Bash does.

kwwette: @adamnovak Thanks for the info. I've modified the PR to set `status = 128 + signal` if `signal > 0`, as @DailyDreaming originally suggested. Since a non-zero exit and a non-zero signal are mutually exclusive, this should not lose any information. In any case, a failing job will be sure to return a non-zero status to Toil.

PR #2653, opened 2019-04-25
Deprecate subprocessDockerCall().

DailyDreaming:
As part of the effort to clean up tests, I'm increasing the deprecation of subprocessDockerCall down from warning the user to switch, to raising an error and telling the user to switch.

Testing is being taken out for the old subprocessDockerCall.  Only apiDockerCall should be used and will be the only supported/tested method after this.


DailyDreaming:
@adamnovak All of these work in py2.7, but apparently a bunch of these fail in py3.6.  We didn't notice because of the `@needsappliance` decorator, which for some reason wasn't registering.  Fixing now.

PR #2658, opened 2019-05-02
Add a ToilNodeType tag to distinguish leader vs. workers.

### LINKED ISSUE #2657: Add a tag to distinguish leader vs. worker nodes in AWS
Currently, there is no easy way to distinguish Toil leader(s) from workers in AWS as they all have the same tags (the leader is assumed to be the first instance that was launched).
It would be easier to locate and filter them in the AWS console if we add a special tag (e.g `ToilNodeType`) that equals `leader` or `worker`.

We can optionally make the [`getLeader`](https://github.com/DataBiosphere/toil/blob/a8252277ff814e7bee0971139c2344f88e44b644/src/toil/provisioners/aws/awsProvisioner.py#L367) function more robust by verifying that the leader does indeed have the `leader` tag (e.g. in case the leader has terminated, so we don't accidentally think that a worker is a leader).

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-354)
┆Issue Number: TOIL-354

arostamianfar:
Resolves #2657

### DIFF
--- a/docs/running/cloud/amazon.rst
+++ b/docs/running/cloud/amazon.rst
@@ -181,9 +181,10 @@ to launch a cluster named "my-cluster" with a t2.medium leader in the us-west-2a
     (venv) $ toil launch-cluster my-cluster --leaderNodeType t2.medium --zone us-west-2a --keyPairName <your-AWS-key-pair-name>
 
 The cluster name is used to uniquely identify your cluster and will be used to
-populate the instance's ``Name`` tag. In addition, the Toil provisioner will
+populate the instance's ``Name`` tag. Also, the Toil provisioner will
 automatically tag your cluster with an ``Owner`` tag that corresponds to your
-keypair name to facilitate cost tracking.
+keypair name to facilitate cost tracking. In addition, the ``ToilNodeType`` tag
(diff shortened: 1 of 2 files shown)


arostamianfar:
@DailyDreaming the primary motivation is to make it more robust and easier to find the leader (e.g. searching for `ToilNodeType : leader` in the AWS console). Also, the `_getLeader` method can be made more robust to also verify that the `leader` tag exists rather than just assuming that the first launched instance is the leader (e.g. it's possible that the leader's instance is terminated and some workers are left over, and we'd end up choosing one of the workers as the leader). I didn't add the check to the `_getLeader` method in this PR, but can do that if you like.

DailyDreaming:
@arostamianfar Understandable.  I'm only worried that without any documentation or code around the tags, their purpose from someone digging in the code could be ambiguous.  Either the check or a short example added to the docs of your current use in the browser would take care of that I think.

Thanks for contributing this, I think it's definitely useful in general for the community.

arostamianfar:
@DailyDreaming excellent point! I added documentation and comments in the code as you suggested. PTAL.

PR #2662, opened 2019-05-04
Capture and log standard output/error from batch systems

kwwette:
When diagnosing job failures, it's always useful to have as much info to hand as possible, in particular any standard output/error, or other log files (e.g. HTCondor job event log files) from the batch system. (Toil has its own logging system for capturing information from jobs, but some messages e.g. from an OOM killer will probably still go to the standard output/error attached to the batch job.)

Currently Toil has no systematic way of capturing standard output/error from batch jobs. Some batch systems rely on the default behaviour, which has different effects: for Slurm, by default standard output is saved to the current directory, whether you want it or not; for HTCondor, by default standard output and error are discarded. For some other batch systems, e.g. LSF, standard output and error are hard-coded directed to `/dev/null`.

The aim of this PR is to 1) capture standard output/error and other files from batch systems in a systematic way, and 2) log standard output/error and other batch system output from failing jobs.

- `BatchSystemSupport(): add function formatStdOutErrPath()`: this commit adds a function which formats paths for standard output/error and other files. These files are written to the Toil workflow directory, with file names `toil_job_<Toil job ID>_batch_<name of batch system>_<job ID from batch system>_<file description>.log`, where `<file description>` is `std_output` for standard output, and `std_error` for standard error.

- `Redirect batch system standard output/error ...`: these commits use the new `formatStdOutErrPath()` function to redirect standard output and error to files for Slurm, HTCondor, PBS/Torque, Sun Grid Engine, and LSF. (HTCondor will also write job event log files with `<file description> = job_events`.) I couldn't figure out how to redirect standard output and error for the Parasol or Mesos batch systems (or whether it even makes sense for those systems).

- `Leader.processFinishedJob(): log any output files created by the batch system`: in this commit, if `resultStatus != 0` (i.e. the batch system job failed), Toil will look for any files created in the Toil workflow directory named `toil_job_<Toil job ID>_batch_*.log`. This will pick up e.g. any standard output and/or error from the failed job. The contents of any files found, if non-zero, is logged with `StatsAndLogging.logWithFormatting()`, and if `--writeLogs` is given, written to files with `StatsAndLogging.writeLogFiles()`.

- `Document handling of standard output and error from batch systems`: documents location of standard output/error files in `--workDir`, and output of those files with `--writeLogs`. Also documents handling of standard output/error and other files in `docs/running/hpcEnvironments.rst`.


kwwette:
Note: I have tested this PR with clusters running Slurm and HTCondor only, as I do not have access to clusters running PBS/Torque, Sun Grid Engine, and/or LSF.

adamnovak:
This looks like it could be pretty useful.

It looks like you had to do something special for HTCondor to make sure that the output files got transferred back to the leader node to be read, given that `workDir` is not necessarily on a shared filesystem. But I don't see anything like that for the other batch systems that you have implemented this for.

Do you have any evidence that they do the right thing by default? If they don't, is there any way we could ever get the files back for them, or will this way of doing things (writing the output to workDir) only ever work for some of them?

Do these log files ever get cleaned up? I imagine that standard output files from successful jobs might eventually fill up all the space. We could also end up leaving these log files laying around when a shared filesystem is not in use, on nodes we can't get back to to delete them.

One possible enhancement would be to have a way to separate the `workDir` from the logging directory. You could leave the `workDir` under `/tmp`, while making sure the logs get written to an actual shared filesystem.

Finally, how does this interact with Toil's existing log collection? If a job fails, will we get its logs printed out twice?

kwwette:
@adamnovak :

> It looks like you had to do something special for HTCondor to make sure that the output files got transferred back to the leader node to be read, given that `workDir` is not necessarily on a shared filesystem. But I don't see anything like that for the other batch systems that you have implemented this for.
> 
> Do you have any evidence that they do the right thing by default? If they don't, is there any way we could ever get the files back for them, or will this way of doing things (writing the output to workDir) only ever work for some of them?

It could well be that for standard output/error to be captured in general, `workDir` must be on a shared filesystem. That seems to already be required, e.g. in `src/toil/worker.py`, the workers appear to write their logs directory to `workDir`, so `workDir` must be accessible from the nodes of an HPC cluster. Is that correct?

I don't have enough knowledge of HPC batch systems to know whether they all require standard output/error locations to be on a shared filesystem. I know from testing that, without the `transfer_output_files` option, HTCondor will complain if the standard output/error locations are not on a shared filesystem, presumably because by default it writes them directly from the nodes. On the Slurm system I have access to, a job will not write standard output into a file in `/tmp` (not shared), but will do so into my home directory (shared) - but I've no idea whether this is Slurm behaviour or due to local cluster configuration.

If this is a problematic restriction, then perhaps a command-line flag will need to be added to turn off standard output/error capture if `workDir` is not on a shared filesystem - or even disable standard output/error capture by default. Another option could be to disable standard output/error capture of `workDir` is in `/tmp`, on the assumption that `/tmp` is unlikely to ever be shared.
 
> Do these log files ever get cleaned up? I imagine that standard output files from successful jobs might eventually fill up all the space. We could also end up leaving these log files laying around when a shared filesystem is not in use, on nodes we can't get back to to delete them.
>
> One possible enhancement would be to have a way to separate the `workDir` from the logging directory. You could leave the `workDir` under `/tmp`, while making sure the logs get written to an actual shared filesystem.

Toil removes `workDir` by default, which was the attraction of putting then in that directory. The user can user `--cleanWorkDir` to control what is cleaned up. I think it makes sense to keep standard output/error files here, since

> Finally, how does this interact with Toil's existing log collection? If a job fails, will we get its logs printed out twice?

Once a worker starts, I believe all standard output/error is redirected to the `worker_log.txt` file. So output won't be logged twice. The only output the original standard output/error will capture will be messages sent to the batch system controlling process, e.g. OOM-killer errors.

Another alternative could be to, where possible, redirect the batch system standard output/error directly to the `worker_log.txt`, rather than having the worker do the redirects itself. That seems a much more complicated change, however, and may not work with all batch systems.

adamnovak:
> It could well be that for standard output/error to be captured in general, workDir must be on a shared filesystem. That seems to already be required, e.g. in src/toil/worker.py, the workers appear to write their logs directory to workDir, so workDir must be accessible from the nodes of an HPC cluster. Is that correct?

I don't think we in general guarantee that a shared filesystem exists at all, even in HPC environments. `worker_log.txt` is collected and returned by the worker via the job store if the worker runs into problems, but that only happens if the worker doesn't run into really severe problems like being OOM-killed.

Since the worker is dropping that file in the work dir, though, and doesn't seem to bother cleaning it up, and since these log files being added are just going to have the extra worker logs that the worker doesn't manage to capture itself, I think we probably want to adopt this change. As long as the leader code doesn't get too upset when the work dir isn't actually on a shared filesystem and it can't read the logs, and as long as we have a workaround in case some of the batch systems we can't test refuse to run jobs when told to write logs to non-shared directories.

kwwette:
@adamnovak :

I've pushed an additional patch, `Add command-line option --noStdOutErr`, which adds an option to disable capture of standard output and error (i.e. send to `/dev/null`). This option exists so that, in case capturing standard output and error into `workDir` causes any problems, or is undesirable for some reason, it can be disabled.

I have also clarified in the documentation that, if you want to capture standard output and error from jobs, `workDir` will need to be on a shared file system. Things should still work if `workDir` is not a shared filesystem, e.g. `/tmp`, but what will probably happen is that standard output and error will be written onto a local (to the node) file system, and then lost once the job ends. (That's what was happening with my Slurm testing mentioned above.) So the requirement of a shared filesystem is only if you actually want to capture the standard output and error from jobs; if you don't care about it, `workDir` can still be on a local filesystem.

adamnovak:
This will fix #2665.

PR #2664, opened 2019-05-07
Move wallTime to a label so we can use exact timestamps.

### LINKED ISSUE #2526: Timestamps are not set in the Mesos dashboard
Timestamps are not set in the Mesos dashboard (localhost:5050) when trying the CWL runner on AWS. It shows "49 years ago" (aka epoch) for everything.

I think it's caused by [this commit](https://github.com/DataBiosphere/toil/commit/8efb7cabcc04b3b052cc1d8a1319ca19d260d723) as the build from the previous commit (just after the 3.18.0 release) works fine, but i haven't figured out the root cause yet.

@adamnovak FYI, in case you have any ideas.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-51)
┆Issue Number: TOIL-51

arostamianfar:
The `timestamp` field is used in the Mesos dashboard to display the status of tasks. However, in commit https://github.com/DataBiosphere/toil/commit/19fd1cc2a2cdf9b67f90608aa2297940db1dde2d , `wallTime` (i.e. task duration) was moved from the `data` field to the `timestamp` field. This results in Mesos showing tasks that started 49 years ago in the dashboard as the "timestamp" now becomes the task duration (note: using this new field was later fixed in https://github.com/DataBiosphere/toil/commit/2b6c60247b378f180e7166a3b17fe5427e28c3b4 to make things work properly, but the dashboard is still incorrect).

I think we should ensure the timestamp field is untouched and move our `wallTime` value to some other field. Pymesos sets the timestamp by default [here](https://github.com/douban/pymesos/blob/c1c95a1d9ec241f2e76be5d61d955e0fb4158d37/pymesos/executor.py#L234), so there is no need to set it ourselves.

I chose to use [`labels`](https://github.com/apache/mesos/blob/5ca56257e9377c36c595cc83ce2eeb98a166945e/include/mesos/mesos.proto#L2683) instead of the `data` as I found that it's a bit more descriptive and easier to work with than bytes, but please let me know if you like to use the `data` field instead.

Resolves #2526


arostamianfar:
@DailyDreaming @adamnovak thanks for the review! It looks like python3 tests are failing due to an unrelated reason. Do you suggest merging this or wait until python3 tests pass again?

adamnovak:
It looks like the Python 3 tests failed because some Ubuntu packages disappeared while we were trying to install them. I've kicked the test off again, so hopefully it will not have that problem this time.

arostamianfar:
@adamnovak thanks so much! both tests passed :)

PR #2671, opened 2019-05-14
Terminate the leader first in a running cluster.

### LINKED ISSUE #2670: Race condition when destroying a running cluster
When destroying a running cluster with worker nodes, all instances (including the leader) are terminated at the same time. However, the processes in the leader may still be running while the workers are being terminated, so the leader (itself in the process of being terminated) may spin up more worker instances to reach the desired cluster size again.

This results in orphan workers being left after the cluster is destroyed[*].

I think the simplest solution is to terminate the leader first and then any workers that may be leftover.

[*] Actually, the destroy cluster fails as the security group is still associated with orphan worker nodes.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-358)
┆Issue Number: TOIL-358

arostamianfar:
Making sure that the leader is terminated first results in a guarantee that no other worker has started in the meantime (from when the destroyCluster call is made until when the instances are actually terminated). This is only an issue in a cluster where a workflow is actively running.

Resolves #2670

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -94,6 +94,10 @@ def awsRetry(f):
     return wrapper
 
 
+class InvalidClusterStateException(Exception):
+    pass
+
+
 class AWSProvisioner(AbstractProvisioner):
     """
     Implements an AWS provisioner using the boto libraries.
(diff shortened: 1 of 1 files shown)


adamnovak:
Some of the Gitlab tests are failing here because the Docker intermediate images for apt-get updates are being sourced from the cache (i.e. the update is not actually running), while the apt-get upgrade or install commands that run later are not. So apt is working with outdated package listings and can't find the packages.

If the tests run on the autoscaling runner they work, so I'm running them again.

PR #2675, opened 2019-05-16
Make sure we always have current package lists (fixes #2674)

### LINKED ISSUE #2674: Docker build allows apt-get updates to come from cache and apt-get upgrades/installs to run based on them
It's possible that the `apt-get update` commands in our Dockerfile don't actually run before the apt-get upgrade commands they are supposed to do work for, because the image they produce might be put in the Docker cache and then reused.

We should always apt-get update on the same line as any other apt operation.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-361)
┆Issue Number: TOIL-361

adamnovak:
This makes sure apt package lists can't get cached in the docker cache separately from the command to install packages using them. It also makes sure that they don't go into the stack of layers used to build the final image.

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -60,9 +60,7 @@ motd = ''.join(l + '\\n\\\n' for l in motd.splitlines())
 print(heredoc('''
     FROM ubuntu:16.04
     
-    RUN apt-get -y update && apt-get -y upgrade
-
-    RUN apt-get -y update --fix-missing && apt-get -y install apt-transport-https ca-certificates software-properties-common
+    RUN apt-get -y update --fix-missing && apt-get -y upgrade && apt-get -y install apt-transport-https ca-certificates software-properties-common && apt-get clean && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


adamnovak:
I don't think `--fix-missing` does anything on `apt-get update`.

Also, in the last test run, the CWL conformance tests timed out on Python 2. Maybe we need to raise the timeout?

DailyDreaming:
@adamnovak While I agree your fix is probably more appropriate method, in a purely empirical test, `--fix-missing` does something on my system because it passes with it and fails without it.

I agree, we should probably raise our timeout then.  Those tests take a ghastly amount of time but it's easy to increase the timeout.

DailyDreaming:
@adamnovak This PR works for me without `--fix-missing`.  However with the change I made earlier, it needed `--fix-missing` in order to work.  I understand why you thought it didn't matter now.

DailyDreaming:
@adamnovak A rebase from master should fix the tests.

PR #2677, opened 2019-05-16
Subprocess dockercall bytestring compatibility for py2/3.

DailyDreaming:
Ugh, only tested in py2 so missed this.

### DIFF
--- a/src/toil/lib/docker.py
+++ b/src/toil/lib/docker.py
@@ -12,6 +12,7 @@ import base64
 import requests
 import logging
 import os
+import sys
 from docker.errors import create_api_error_from_http_exception
 from docker.errors import ContainerError
 from docker.errors import ImageNotFound
@@ -229,7 +230,8 @@ def apiDockerCall(job,
                                         user=user,
                                         environment=environment,
                                         **kwargs)
-
+            if sys.version_info >= (3, 0):
+                out = out.decode('utf-8')
             return out
         else:
             if (stdout or stderr) and log_config is None:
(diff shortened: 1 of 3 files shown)


src/toil/lib/docker.py
```diff
@@ -229,7 +230,8 @@ def toil_job(job):
                                         user=user,
                                         environment=environment,
                                         **kwargs)
-
+            if sys.version_info >= (3, 0):
+                out = out.decode('utf-8')
```
adamnovak: I don't think it's a good idea to do this here. What if someone is capturing binary data on stdout that is not valid UTF-8?

DailyDreaming: Python 2 works differently so that the output works with "normal" strings which is why the tests failed.  You're right though, we should probably model the behavior off of the interior functions and change the tests instead.  Will change accordingly.

DailyDreaming: Fixed.

DailyDreaming:
@adamnovak Guessing you are gone for the day and since this fixes a failing test on master I'm going to merge.  Let me know if you see anything egregious and want modifications.

adamnovak:
@DailyDreaming After merging, the Python 3 test run failed 7 tests, running on the shell runner.

One of them looked like the shell runner is just running low on free disk, so I've paused it and I'm re-running the tests. If it does better we can just fully move to the docker runners from now on, which I think will be more consistent (since the machines get torn down and rebuilt constantly).

If master passes all the Python tests except the CWL conformance test due to timeout, then we can have @jeffrey856 rebase #2678 onto master.

adamnovak:
The test job to watch is https://ucsc-ci.com/databiosphere/toil/-/jobs/10117

DailyDreaming:
@adamnovak Cool, looks like it succeeded!  (though with 2308 warnings XD)  I'll let Jeffrey know he should rebase.

PR #2684, opened 2019-05-21
Update to ubuntu18.04

jeffrey856:
Updated to 18.04 

Added :
 RUN add-apt-repository ppa:xapienz/curl34
 RUN apt-get update

To be able to avoid conflicts between libcurl4 and libcurl3.


adamnovak:
OK, this turns out not to work:

```
Step 9/30 : RUN apt-get -y install libffi-dev python3.6 python3.6-dev python-dev python-pip python3-pip libssl-dev wget curl openssh-server mesos nodejs rsync screen && apt-get clean && rm -rf /var/lib/apt/lists/*
 ---> Running in 7e58cc3539ad
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mesos : Depends: libcurl3
E: Unable to correct problems, you have held broken packages.
```

@jeffrey856 I think you need to re-install libcurl4 first for this to work. I unpacked the libcurl4 from the PPA and it does properly announce itself as providing libcurl3, but your system thinks it already has libcurl4 installed and apt isn't clever enough to work out that it needs to update (or maybe even downgrade) libcurl4 to the PPA version in order to provide libcurl3, especially when libcurl3 is also its own real package.

DailyDreaming:
@adamnovak @jeffrey856 I can verify that it seems to work with `libcurl4-openssl-dev` included.

Running the batch system tests, one fails however, as running:

`(venv) quokka@qcore:~/toil$ make test tests=src/toil/test/batchSystems/batchSystemTest.py`

gives me this output:

[output.txt](https://github.com/DataBiosphere/toil/files/3213579/output.txt)

adamnovak:
I haven't tested it separately from the CI.

PR #2685, opened 2019-05-22
Fix security group removal when destroying a cluster.

### LINKED ISSUE #2670: Race condition when destroying a running cluster

arostamianfar:
Fixes a bug that was introduced in PR #2671 where the security group was not deleted if only the leader was running in the cluster.

Tested:
* Only the leader running
* Only workers running
* Both leader and workers running
* No instances running (this is a no-op).

Resolves #2670

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -214,8 +214,10 @@ class AWSProvisioner(AbstractProvisioner):
 
         # We should terminate the leader first in case a workflow is still running in the cluster.
         # The leader may create more instances while we're terminating the workers.
+        vpcId = None
         try:
             leader = self.getLeader(returnRawInstance=True)
+            vpcId = leader.vpc_id
             logger.info('Terminating the leader first ...')
             self._terminateInstances(instances=[leader])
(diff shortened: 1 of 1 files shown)


arostamianfar:
@DailyDreaming thanks for the quick review! But I just noticed that the same is true for `_deleteIAMProfiles` 😢 (it wasn't caught in my testing as we use custom IAM roles internally).

I'm trying to see if there is a more unified way of fixing this rather than special-casing IAM profiles as well. It may be better to revert my previous PR and submit another one that's bug free.

PR #2686, opened 2019-05-22
Revert "Terminate the leader first in a running cluster."

arostamianfar:
Reverts DataBiosphere/toil#2671

The original PR introduced a bug about cleaning up security groups and IAM profiles if the leader is the only instance running (most common case). Reverting this now as I like to take some time to figure out the best way to address this and hope to avoid special-casing things in two places.

Issue #2670

### DIFF
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -94,10 +94,6 @@ def awsRetry(f):
     return wrapper
 
 
-class InvalidClusterStateException(Exception):
-    pass
-
-
 class AWSProvisioner(AbstractProvisioner):
     """
     Implements an AWS provisioner using the boto libraries.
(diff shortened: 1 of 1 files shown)


adamnovak:
One of the Gitlab tests failed because its connection dropped when installing dependencies from Pip. I have restarted it.

DailyDreaming:
@arostamianfar Merging this then as it fixes a bug.  Thanks for catching this.

arostamianfar:
@adamnovak thanks for restarting the test! Is there a way that I could do it myself without pushing a new change? I don't see that option on gitlab.
@DailyDreaming I'm still looking into good ways of addressing this in a clean way. It's a bit tricky as the leader must be terminated first and *only then* we should list the remaining instances (otherwise, the leader may have created more instances). Perhaps special-casing based on whether a single instance is in the cluster (regardless of whether it's the leader) may be a cleaner way. Lemme know if you have better ideas!

adamnovak:
I think we have to add you as a user to our Gitlab. I believe we can
do that; you have to log into the Gitlab instance with your Github
account, whereupon you will be automatically banned. Then we un-ban
you.

PR #2688, opened 2019-05-23
Rearrange Gitlab Testing.

DailyDreaming:
Separating tests out into  sensible groups.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,7 @@ after_script:
 
 
 stages:
+  - main_tests
   - test
 
 #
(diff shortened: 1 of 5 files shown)


DailyDreaming:
@adamnovak Is `src/toil/test/spark/__init__.py` functional?  It doesn't mesh exactly with this PR but it didn't seem to be used anywhere so I removed it.

They're mostly separated by directory... but I'd also like to roughly sort by runtime.  I'll think about better arranging them.  Some things can be commented out (like Azure testing).  Other things like the jenkins stuff can probably be removed.

adamnovak:
> @adamnovak Is `src/toil/test/spark/__init__.py` functional? It doesn't mesh exactly with this PR but it didn't seem to be used anywhere so I removed it.

It makes the directory a module, right? Without it, toil.test.spark.whatever can't be referred to from Python.

PR #2690, opened 2019-05-23
Clean up temporary directories cwltoil uses to hold output streams

### LINKED ISSUE #2601: Temporary CWL workdir files are not cleaned up
The CWL runner creates temporary dirs in `workdir` to use as temporary output dirs ([code](https://github.com/DataBiosphere/toil/blob/6a18b8a6a4ad885729ef9fa53542e0a6fd4b6c0a/src/toil/cwl/cwltoil.py#L586)). These are added to `openTempDirs` variable and are supposed to be cleaned up after the workflow is done through the [`cleanTempDirs`](https://github.com/DataBiosphere/toil/blob/6a18b8a6a4ad885729ef9fa53542e0a6fd4b6c0a/src/toil/cwl/cwltoil.py#L1031) function. However, this function is only added as a [followOn to the first CWLJob](https://github.com/DataBiosphere/toil/blob/6a18b8a6a4ad885729ef9fa53542e0a6fd4b6c0a/src/toil/cwl/cwltoil.py#L1273) and not for the nested workflows [*]. This results in intermediate files being left on the workers, which can take up a lot of space.

[*] It doesn't even seem to work for the first one as [`wf1 is CWLJob`](https://github.com/DataBiosphere/toil/blob/6a18b8a6a4ad885729ef9fa53542e0a6fd4b6c0a/src/toil/cwl/cwltoil.py#L1272) is always false. Should it be `isinstance(wf1, CWLJob)` instead?

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-35)
┆Issue Number: TOIL-35

adamnovak:
This should fix #2601 

cwltoil has a whole system for creating job-specific temporary directories
under its work directory. It should instead probably be using Toil's built-in
concept of a work directory, and not maintaining its own. After #2587 merges I
think they will always be under the same path, so then we can rip out the
second system and see if anything breaks.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -561,6 +561,8 @@ class CWLJob(Job):
         self.runtime_context = runtime_context
         self.step_inputs = step_inputs or self.cwltool.tool["inputs"]
         self.workdir = runtime_context.workdir
+        # This tracks temporary directories that we create while running this job.
+        # They should be cleaned up by us when we finish.
         self.openTempDirs = []
 
     def run(self, file_store):
@@ -582,14 +584,20 @@ class CWLJob(Job):
             self.runtime_context.tmpdir or file_store.getLocalTempDir())
         outdir = os.path.join(file_store.getLocalTempDir(), "out")
         os.mkdir(outdir)
+        # Create per-job temporary output directories under the CWL workdir,
+        # which we also are using as our job store.
+        #
+        # TODO: Why don't we just use Toil's file_store.getLocalTempDir() under
+        # Toil's --workDir? Why do we have two notions of workdir?
         top_tmp_outdir = self.workdir or os.environ["TMPDIR"]
         tmp_outdir_prefix = os.path.join(
             _makeNestedTempDir(top=top_tmp_outdir, seed=outdir, levels=2),
             "out_tmpdir")
-        self.openTempDirs.append(top_tmp_outdir)
+        self.openTempDirs.append(tmp_outdir_prefix)
 
         index = {}
         existing = {}
+        # Prepare the run instructions for cwltool
         runtime_context = self.runtime_context.copy()
         runtime_context.basedir = os.getcwd()
         runtime_context.outdir = outdir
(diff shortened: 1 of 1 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -582,14 +584,20 @@ def run(self, file_store):
...
+        # which we also are using as our job store.
+        #
+        # TODO: Why don't we just use Toil's file_store.getLocalTempDir() under
+        # Toil's --workDir? Why do we have two notions of workdir?
         top_tmp_outdir = self.workdir or os.environ["TMPDIR"]
         tmp_outdir_prefix = os.path.join(
             _makeNestedTempDir(top=top_tmp_outdir, seed=outdir, levels=2),
             "out_tmpdir")
-        self.openTempDirs.append(top_tmp_outdir)
+        self.openTempDirs.append(tmp_outdir_prefix)
```
arostamianfar: @adamnovak  FYI, this may not actually work as the temporary directory is something like `...out_tmpdir<tmpname>` (e.g. `/var/lib/toil/tmpPbBjlT/3/a/out_tmpdiraitngE/`).
Ideally, we'd want to remove the top-level `out_tmpdiraitngE` directory, but I'm not sure how we can get that as the suffix is generated by CWL. We also don't want to delete the upstream directory as it can be used by other tasks on the same worker.
My workaround so far has been to use `cwl_executor.output_dir` variable and remove those. It works, but it still leaves the temporary directory paths on the workers (I think that's fine though as they don't actually take up any space). See https://github.com/DataBiosphere/toil/commit/e149d827e9b07febf64d3bb9a2242165b25863fb for reference.

adamnovak: The tmp_outdir_prefix is job-specific, right? And it is no longer needed at the end of the job. I can just add another level of subdirectory and delete the parent of whatever we pass along to CWL.

arostamianfar: @adamnovak  Your idea works too! However, `_makeNestedTempDir` doesn't actually use `mkdtemp`, so collisions are possible and likely as it's just based on the first two characters of a hash. Could you please add another layer that uses `mkdtemp` to ensure folders are indeed unique?

adamnovak:
#2587 is in now, and I am taking a different approach here. There's no real reason I could see for `_makeNestedTempDir` to exist at all; all it is used for is these output temp directories, and they are limited in lifetime to the job that creates them. So I am just putting the output temp directories under a Toil local temp directory, which is also limited to the job's lifetime, but which is managed by Toil itself. That lest me delete a bunch of code here and should hopefully make things harder to mess up.

mr-c:
Thanks!

PR #2698, opened 2019-05-27
Updated azure docs

jeffrey856:
Currently updating Azure documentations. (in progress)


adamnovak:
@jeffrey856 I'm worried that this isn't the right approach.

I don't think we want to de-document the Azure-based functionality that we do have entirely. I think what we want to do is to document that, while Azure ought to work, we can't promise our users that it actually does work, that it will stay working, or that the development team will devote any time to fixing it when it breaks.

We should maybe say that it is "unsupported" or "unmaintained". So removing it from the list of "supported" cloud providers like you did makes sense, but removing all the documentation for what we have and how it is supposed to work is probably not what we want, especially if we aren't simultaneously removing the applicable code.

@DailyDreaming, what do you think?

diekhans:
We have several system interfaces that barely work. The LSF Toil
module got one group temporally banned from the EBI cluster.

Rather than removing, I think it would be better to clearly
documented the limited support and actively ask for maintainers
in a positive way.

adamnovak:
Yeah, it would definitely be good to solicit maintainers for these subsystems.

docs/running/cloud/cloud.rst
```diff
@@ -24,7 +24,7 @@ Toil can launch and manage a cluster of virtual machines to run using the *provi
 distributed over several nodes. The provisioner also has the ability to automatically scale up or down the size of
 the cluster to handle dynamic changes in computational demand (autoscaling).
 
-Currently we have working provisioners with AWS and GCE with Azure coming soon.
+Currently we have working provisioners with AWS and GCE. Azure provisioner works but has limited support and is unmaintained.
```
DailyDreaming: Probably can't say it's `working`.

Something more like:
`The azure provisioner has limited support and is unmaintained.  We've left our (previously working) Azure support in place, but cannot guarantee it will work in the future as we no longer have the resources to continue testing Azure.`

docs/running/introduction.rst
```diff
@@ -71,7 +70,7 @@ The :ref:`clusterRef` are command line tools used to provision nodes in your des
 They allows you to launch nodes, ssh to the leader, and rsync files back and forth.
 
 For detailed instructions for using the provisioner see :ref:`runningAWS` or :ref:`runningGCE`. An
-Azure provisioner is in the works and coming soon. For more details see the `Azure provisioner github ticket`_.
+Azure provisioner is working but has limited support. For more details see the `Azure provisioner github ticket`_.
```
DailyDreaming: I would just say limited support and leave out the `working` part since we can't really know without CI/CD in place for it.

Issue #2700, opened 2019-05-30
Minimum permissions required by AWS role


tthyer:
I am using the new option to specify the AWS role when launching a cluster (thanks to @arostamianfar). I have a policy that's working, but would like to tighten it down further if possible. I also suggest that the set of required permissions when specifying a pre-created role would be a useful addition to the documentation. Has anyone already explored what the minimum set is? Below is what I have so far for the policy. I will almost certainly have to tighten this down further by specifying specific resources rather than `"Resource": "*"`, but this set is at least working.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "ec2:CancelSpotInstanceRequests",
                "ec2:CreateSecurityGroup",
                "ec2:CreateTags",
                "ec2:DeleteSecurityGroup",
                "ec2:DescribeAvailabilityZones",
                "ec2:DescribeImages",
                "ec2:DescribeInstances",
                "ec2:DescribeInstanceStatus",
                "ec2:DescribeKeyPairs",
                "ec2:DescribeSecurityGroups",
                "ec2:DescribeSpotInstanceRequests",
                "ec2:DescribeSpotPriceHistory",
                "ec2:DescribeVolumes",
                "ec2:RequestSpotInstances",
                "ec2:RunInstances",
                "ec2:StartInstances",
                "ec2:StopInstances",
                "ec2:TerminateInstances",
                "iam:PassRole"
                "s3:AbortMultipartUpload",
                "s3:CreateBucket",
                "s3:DeleteBucket",
                "s3:DeleteObject",
                "s3:DeleteObjectVersion",
                "s3:GetBucketLocation",
                "s3:GetBucketVersioning",
                "s3:GetObject",
                "s3:GetObjectVersion",
                "s3:GetObjectVersionTagging",
                "s3:HeadBucket",
                "s3:ListAllMyBuckets",
                "s3:ListBucket",
                "s3:ListBucketMultipartUploads",
                "s3:ListBucketVersions",
                "s3:ListMultipartUploadParts",
                "s3:PutBucketTagging",
                "s3:PutBucketVersioning",
                "s3:PutObject",
                "s3:PutObjectTagging",
                "s3:PutObjectVersionTagging",
                "s3:ReplicateTags",
                "sdb:*",
            ],
            "Resource": "*"
        }
    ]
}
```

arostamianfar:
[...]

PR #2703, opened 2019-05-31
Add documentation about --awsEc2ProfileArn.

arostamianfar:
Add documentation about the new `--awsEc2ProfileArn` flag. I added an appendix section so that we can include all customized permissions when @tthyer finalizes the list in #2700.


adamnovak:
Oh, that's what #2700 is all about: finding the minimum list. I guess we can just leave the text this way and update the policy when we are sure of it.

PR #2708, opened 2019-06-07
Break up the fileStore file into multiple files and factor out deferred functions

adamnovak:
I might want to move them from `toil.fileStore` to `toil.fileStores` eventually. But I want this broken up before I really tinker with the caching.

Now we have one file for the interface, and one for each implementation, and they're no longer one extremely long file.


src/toil/fileStore/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+
+__all__ = ['fileStore', 'nonCachingFileStore', 'cachingFileStore']
+from toil.fileStore.fileStore import *
```
DailyDreaming: I'm a bit leery of using `*`, especially since importing all of these names into one file seems to be inviting collision.

Would you be amenable to something like:
```
import toil.fileStore.fileStore as fileStore
import toil.fileStore.nonCachingFileStore as nonCachingFileStore
import toil.fileStore.cachingFileStore as cachingFileStore
```

adamnovak:
I think it would end up being something like "from
toil.fileStore.fileStore import FileID, FileStore" and so on for all
the types that other parts of the codebase actually need. Or, we could
go through and rewrite all the users to import things from their new
real paths.

DailyDreaming:
I think the latter would make more sense.

adamnovak:
OK, I've also pulled out the deferred function system into its own class, so it isn't part of the file store anymore. This removes one moving part from the file store for the actual caching refactor.

adamnovak:
I still haven't rewritten everything else to use the new import paths yet.

adamnovak:
@mr-c Since @DailyDreaming is out at a conference, would you have time to review this PR?

mr-c:
Hello @adamnovak I don't have any experience with this part of the codebase

PR #2716, opened 2019-06-16
Delete Redundant Tests  

jeffrey856:
Looking for and deleting potential redundant tests.


src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1458,27 +1456,6 @@ def testInvalidJobStoreName(self):
...
-@slow
-class EncryptedAWSJobStoreTest(AWSJobStoreTest, AbstractEncryptedJobStoreTest.Test):
-    pass
-
-
-@needs_azure
-@needs_encryption
-@slow
-class EncryptedAzureJobStoreTest(AzureJobStoreTest, AbstractEncryptedJobStoreTest.Test):
-    pass
```
adamnovak: Are you deleting these just because the class body is empty? I think the EncryptedAzureJobStoreTest and the EncryptedAWSJobStoreTest classes really do create actual tests that can be run, because they inherit other classes and compose them together. Basically each of these says "run this set of tests using this job store implementation".

I suspect that we want to keep these (or at least the AWS one) to be part of the integration tests that @DailyDreaming is getting set up again.

src/toil/test/mesos/mesosTest.py
```diff
@@ -1,70 +0,0 @@
-# Copyright (C) 2015-2016 Regents of the University of California
```
adamnovak: @jeffrey856 Why did you decide to delete this test? What makes it redundant? Is Mesos sufficiently tested by some other tests?

jeffrey856: I assumed the tests from the batchstytemTest.py under test would run a similar kind of job since it does run setup and creates a cluster, teardown, and a test within MesosBatchSystemTest() and or MesosBatchSystemJobTest() classes.

src/toil/test/jobStores/jobStoreTest.py
jeffrey856: That's my bad. I didn't realize it was running the AbstractEncryptedJobStoreTest. I will revert this change.

src/toil/test/mesos/mesosTest.py
adamnovak: OK, that makes sense.

PR #2720, opened 2019-06-23
Updated dockerfile

jeffrey856:
updated docker file due to gitter issue.

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -70,10 +70,11 @@ print(heredoc('''
         && apt-key adv --keyserver keyserver.ubuntu.com --recv 68576280
 
     RUN add-apt-repository -y ppa:jonathonf/python-3.6
-    
-    RUN add-apt-repository -y ppa:xapienz/curl34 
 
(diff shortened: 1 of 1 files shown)


jeffrey856:
@adamnovak Shouldn't the star (*)regex be enough? Since I looked over the previous versions, it seems like the creator uses the same naming scheme but with different numbers at the end after ubuntu. I.e. **7.58.0-2ubuntu**3.6ppa1 vs **7.59.0-2ubuntu**3.5ppa1.

adamnovak:
> @adamnovak Shouldn't the star (*)regex be enough? Since I looked over the previous versions, it seems like the creator uses the same naming scheme but with different numbers at the end after ubuntu. I.e. **7.58.0-2ubuntu**3.6ppa1 vs **7.58.0-2ubuntu**3.5ppa1.

It [does look like you can use wildcards in those version strings](https://askubuntu.com/a/603966/170317), so I guess you could do `*ppa*` as the version specifier.

But see my other comment; I don't think this will really solve @ionox0's problem.

Also, we really should make an issue on Github to track the problem; just having a PR like this is against our own contributing rules, and saying it fixes an unelaborated and un-hyperlinked "gitter issue" makes it very difficult to review, because it's not clear what the problem it is trying to fix actually is.

Issue #2724, opened 2019-06-27
Decode lshosts output in lsf.py


adamnovak:
@drkennetz suggests that we add some decoding logic to `lsf.py`, because apparently it doesn't work in Pythion 3.

```
        for line in p.stdout:
                split_items = line.strip().split()
#logic          items = [item.decode() for item in split_items if isinstance(item, bytes)]
                for value in items:
                    if isinstance(value, bytes):
                        value = value.decode()
                if len(items) < num_columns:
                        RuntimeError("lshosts output has a varying number of "
                                     "columns")
#logic      if items[cpu_index] != '-' and int(items[cpu_index]) > int(maxCPU):
                        maxCPU = items[cpu_index]
                if (items[mem_index] != '-' and
                        MemoryString(items[mem_index]) > maxMEM):
                    maxMEM = MemoryString(items[mem_index])
```

See [this comment](https://github.com/DataBiosphere/toil/issues/2523#issuecomment-506395221).

As suggested, we can also explore if there's some way to wrap or set a default encoding on all our text streams with one bit of code.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-388)
┆Issue Number: TOIL-388

drkennetz:
I am going to continue to look into what is going on here. I am tracking the issues as I work through them. I want to make sure this is reproducible for different versions of python from a pip install. I am seeing some other downstream errors as well like:
```
Traceback (most recent call last):
  File "/hpcf/apps/python/install/3.5.2/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/users/dkennetz/toil_test/lib/python3.5/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 235, in run
    activity = self.checkOnJobs()
  File "/users/dkennetz/toil_test/lib/python3.5/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 210, in checkOnJobs
    status = with_retries(self.getJobExitCode, batchJobID)
  File "/users/dkennetz/toil_test/lib/python3.5/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 42, in with_retries
    return operation(*args, **kwargs)
  File "/users/dkennetz/toil_test/lib/python3.5/site-packages/toil/batchSystems/lsf.py", line 98, in getJobExitCode
    if "Done successfully" in line:
[...]

PR #2726, opened 2019-06-27
made changes to lsf.py to make it compatible on lsf9 and lsf10 with p…

drkennetz:
…ython3x


drkennetz:
@drjrm3
@mr-c

src/toil/batchSystems/lsf.py
```diff
@@ -117,13 +118,12 @@ def getJobExitCode(self, lsfJobID):
...
                     return 1
-                elif line.find("Started on ") > -1:
+                elif line.find("Started on ") > -1 or line.find(": Started") > -1:
                     started = 1
-
-            if started == 1:
-                logger.debug("bjobs detected job started but not completed: "
-                             "{}".format(job))
-                return None
+                    if started == 1:
```
mr-c: Why the indent here?

src/toil/batchSystems/lsf.py
```diff
@@ -167,18 +169,18 @@ def prepareBsub(cpu, mem, jobID):
...
                 bsubMem = []
             bsubCpu = [] if cpu is None else ['-n', str(math.ceil(cpu))]
-            bsubline = ["bsub", "-cwd", ".", "-J", "toil_job_{}".format(jobID)]
+            bsubline = ["bsub", "-cwd", ".",
+                        "-J", "toil_job_{}".format(jobID),
+                        "-o", "toil_job_{}.out".format(jobID),
+                        "-e", "toil_job_{}.err".format(jobID)]
             bsubline.extend(bsubMem)
             bsubline.extend(bsubCpu)
-            stdoutfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_output')
```
mr-c: Why are these function calls dropped?

drkennetz: Because the stdout and stderr were added to the bsub command line call above

src/toil/batchSystems/lsf.py
drkennetz: Because this should only be called if started was found in the line

drjrm3: I agree. It looks to me like this indent should not be here. Right now it renders the `if` evaluation useless because you set `started = 1` and then evaluate the state of `started`. I think this is the only place that `started` is used so perhaps could remove that variable assignment / check completely and jump straight to the `logger` + `return None` at this point.

src/toil/batchSystems/lsf.py
drjrm3: I prefer to keep the `formatStdOutErrPath` call. It will add `-e` and `-o` flags like this does but it uses the abstract method which would be consistent with other schedulers. Unless there is a reason this method doesn't work correctly for LSF I'm in favor of keeping it as it was.

drkennetz: I can add it back, I just didn't understand why I needed it if I had the -o and -e included in the line above. Should I remove it from the line:

``` 
          bsubline = ["bsub", "-cwd", ".", 
                        "-J", "toil_job_{}".format(jobID), 
                        "-o", "toil_job_{}.out".format(jobID), 
                        "-e", "toil_job_{}.err".format(jobID)]
```
I guess I am not understanding what the `self.boss.formatStdOutErrPath` is doing

src/toil/batchSystems/lsf.py
drkennetz: The way it was before just didn't make a lot of sense to me, so instead of removing it I put it where it made sense. The only place `if started == 1:` should be used is inside the `elif line.find("Started on ") > -1:` so instead of just removing it I indented it. I can remove or keep.

src/toil/batchSystems/lsf.py
mr-c: @drkennetz It names the output files in a consistent manner across batch schedulers and puts them in the work directory instead of the current working directory as your code does

https://github.com/DataBiosphere/toil/blob/ff31650c0eb0b5438b89002f789ce46b85cccc6a/src/toil/batchSystems/abstractBatchSystem.py#L304

https://github.com/DataBiosphere/toil/blob/ff31650c0eb0b5438b89002f789ce46b85cccc6a/src/toil/common.py#L1009:9

src/toil/batchSystems/lsf.py
mr-c: The outer loop runs across all entries in the output log. It has several exits (`return 0`, `return None`, etc). Only if "Started on" was seen AND none of the other conditions are met do we want to issue the error message.

So please undo the indents to this section :-)

src/toil/batchSystems/lsf.py
drkennetz: my code does indeed. I'll add the two recommendations above back into the code, run them again and see if outputs are affected. If not I'll commit changes back.

Thanks for your help! @drjrm3 @mr-c

drkennetz: Sorry for not documenting this as I was going through the first time. This is why I changed the code here:

```
            stdoutfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_output')
            stdoutfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_output')
            bsubline.extend(['-o', stdoutfile, '-e', stderrfile])
```
`self` is not defined:

```
  File "/hpcf/apps/python/install/3.5.2/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/users/dkennetz/mytoil/lib/python3.5/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 234, in run
    activity |= self.createJobs(newJob)
  File "/users/dkennetz/mytoil/lib/python3.5/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 136, in createJobs
    subLine = self.prepareSubmission(cpu, memory, jobID, command)
  File "/users/dkennetz/mytoil/lib/python3.5/site-packages/toil/batchSystems/lsf.py", line 71, in prepareSubmission
    return self.prepareBsub(cpu, memory, jobID) + [command]
  File "/users/dkennetz/mytoil/lib/python3.5/site-packages/toil/batchSystems/lsf.py", line 182, in prepareBsub
    stdoutfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_output')
NameError: name 'self' is not defined
```

drkennetz: I have fixed the indent on the other comment, but after changing the `bsubline` to my fix, I get a successful workflow.

src/toil/batchSystems/lsf.py
drjrm3: Thanks for clarifying @mr-c. I was confusing this with `status <RUN>` and thought that this line should be mutually exclusive from all others. Speaking of which - in LSF 10 `Started on` will not be found, but rather `Started 1 Task(s) on ...` so we may need to change this line still.

src/toil/batchSystems/lsf.py
```diff
@@ -93,21 +93,22 @@ def getJobExitCode(self, lsfJobID):
                          "{}".format(job))
             process = subprocess.Popen(args, stdout=subprocess.PIPE,
                                        stderr=subprocess.STDOUT)
-            started = 0
-            for line in process.stdout:
-                if "Done successfully" in line:
+            output = process.stdout.read().decode('ascii').replace("\n                     ", "")
+            process_output = output.split('\n')
+            for line in process_output:
```
drjrm3: Should we be more explicit with the statuses (`<DONE>`, `<PEND>`, `<EXIT>`)? What happens in the current code if someone submits a job with name `DONE` to queue `PEND` and names their job `EXIT` while it is still running? An edge case, I know, but is there anything wrong with putting a check for `Status <DONE>` instead?

src/toil/batchSystems/lsf.py
drkennetz: @drjrm3 
```
                elif line.find("Started on ") > -1 or line.find(": Started") > -1:
                    started = 1
```
This should work right?

src/toil/batchSystems/lsf.py
drkennetz: I am going to submit these to our normal queue as well and get the status updates for all cases. I think we could make them a bit more specific. I like looking for `"Status <DONE>"` and I have tested it with success. I am adding `"Status "` to the beginning of the other searches as well as it appears in all of them.

src/toil/batchSystems/lsf.py
drjrm3: @mr-c - now I'm confused again. Shouldn't we only enter this if the job is in `RUN` state? And isn't it fully expected that this function will be invoked at times when the job is in `RUN` state? In this case, shouldn't we modify this to be in line with the other states to check on all mutually exclusive, exhaustive conditions? I'm almost leaning towards checking the job status initially with `line.split('Status <')[1].split('>')[0]` and then looping over the mutually exclusive cases (including `RUN`, `ZOMBI`, `UNKWN`, etc.). What benefits does it provide to look over all other cases and then mark it as running at the end?

src/toil/batchSystems/lsf.py
drjrm3: @drkennetz I think `self` is not defined because this is defined as a `@staticmethod`. I'm not sure what benefit the static nature is providing but I think we will need to remove that if we want to use `self.boss.formatStdOutErrPath`. @mr-c - thoughts?

drkennetz: Yeah this makes sense. If I remove `@staticmethod` and add `self` into the function arguments I get success.

drkennetz: The log files look like:
```
[dkennetz@hpc02  toil-f654dd2a-ae66-492f-a246-3a15c1dee365-cc963444-5205-4a81-8502-3445f286bed5]$ ls
toil_job_1_batch_lsf_81672902_std_error.log   toil_job_2_batch_lsf_81672903_std_error.log   toil_job_3_batch_lsf_81672904_std_error.log
toil_job_1_batch_lsf_81672902_std_output.log  toil_job_2_batch_lsf_81672903_std_output.log  toil_job_3_batch_lsf
```
and persist as long as whatever I have the `--cleanWorkDir` flag set to.

drkennetz:
@adamnovak When you get a chance could you take a look. This is to incorporate python3.x compatibility for lsf9/10 integration.

src/toil/batchSystems/lsf.py
```diff
@@ -148,8 +151,7 @@ def getJobExitCode(self, lsfJobID):
         """
         Implementation-specific helper methods
         """
-        @staticmethod
```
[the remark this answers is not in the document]
drkennetz: Yeah, later in lines 181 and 182:
```
            stdoutfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_output')
            stderrfile = self.boss.formatStdOutErrPath(jobID, 'lsf', '%J', 'std_error')
```
is called for naming. the code errors out if `@staticmethod` is not removed and self isn't added as an argument to the function. Alternatively I thought about hard-coding stdout and stderr to line 178 but @mr-c and @drjrm3 recommended that this method is more compliant with multiple batch systems.

src/toil/batchSystems/lsf.py
```diff
@@ -93,21 +93,23 @@ def getJobExitCode(self, lsfJobID):
                          "{}".format(job))
             process = subprocess.Popen(args, stdout=subprocess.PIPE,
                                        stderr=subprocess.STDOUT)
+            output = process.stdout.read().decode('ascii').replace("\n                     ", "")
```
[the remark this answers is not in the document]
drkennetz: I am comfortable with changing encoding to utf-8, @drjrm3 what do you think? Any reason to stick with ascii?

src/toil/batchSystems/lsf.py
```diff
@@ -213,15 +215,16 @@ def obtainSystemConstants(cls):
...
-                if len(items) < num_columns:
-                        RuntimeError("lshosts output has a varying number of "
-                                     "columns")
-                if items[cpu_index] != '-' and items[cpu_index] > maxCPU:
-                        maxCPU = items[cpu_index]
-                if (items[mem_index] != '-' and
-                        MemoryString(items[mem_index]) > maxMEM):
-                    maxMEM = MemoryString(items[mem_index])
+            split_items = line.strip().split()
+            items = [item.decode('ascii') for item in split_items if isinstance(item, bytes)]
```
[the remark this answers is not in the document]
drkennetz: I am comfortable with changing encoding to utf-8, @drjrm3 what do you think? Any reason to stick with ascii?

src/toil/batchSystems/lsf.py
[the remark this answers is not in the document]
drkennetz: The line format is because of the actual output of `bjobs -l` formatting. It looks like:
```
..., Started on <node123>, 1 Task Started on node123, Status <RUN>, Memory...,
                                           (other reported metrics), (other reported metrics).....,
                                           (continue the report here), (continue the report here),
                                           Task <task-name> Done successfully..."
```
@drjrm3 wrote a pipeline in house that uses something similar for job monitoring and saw some issues with the number of spaces in line continuations. Perhaps he can comment on this a bit more?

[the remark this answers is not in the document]
drjrm3: `utf-8` seems fine.

For the bunch of spaces + newline, it allows us to unfold each entry in `bjobs -l`. As it is without that transformation, we may not be able to detect things like `<DONE>` since it could be folded. By unwrapping that and removing the newline + large indent, it allows us to be sure we can capture the reported states to make decisions.

drkennetz:
@adamnovak pushed changes for character encodings.

src/toil/batchSystems/lsf.py
```diff
@@ -167,7 +169,7 @@ def prepareBsub(cpu, mem, jobID):
                     mem = old_div(float(mem), 1024**3)
                     mem_resource = parse_memory_resource(mem)
                     mem_limit = parse_memory_limit(mem)
-                bsubMem = ['-R', 'select[type==X86_64 && mem > {m}] '
+                bsubMem = ['-R',
```
mr-c: Is `select[mem > ` not compatible with lsf10?

drkennetz: Not compatible with all architectures. Would break if machine type isn’t x86_64. I don’t think there is a reason to hard-code this as it will be picked up as a system default, for example our default is select[type==rhel7_enterprise] but will adopt the hard-coded value when it is in the code. Mem is whatever is specified by default mem (—defaultMem) on the command line, or whatever is set by the system default (ours is 2.5G) for a job with no -R specified.

drkennetz: https://github.com/DataBiosphere/toil/issues/2725

drkennetz: This created a conflict which I just resolved based on your fix in master. This is going through travis again.

adamnovak:
This PR will fix #2724

drkennetz:
@adamnovak any status update on this?

adamnovak:
OK, I've imported this in for full testing. When that passes we can go ahead and merge it.

mr-c:
Thank you @drkennetz !

drkennetz:
Thanks for your help all! @mr-c @adamnovak @drjrm3

PR #2728, opened 2019-07-02
Prevent Toil from zipping entire dirs

jeffrey856:
Added a check to not allow user to execute toil in a dir such as /tmp or /var.

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -287,8 +287,8 @@ class DirectoryResource(Resource):
         else:
             # This is a simple user script (with possibly a few helper files)
             rootDir = path
-        blacklistDir = ['/tmp', '/var', '/etc', '/bin', '/sbin', '/home', '/dev', '/sys', '/usr', '/run']
-        if path not in blacklistDir:
+        skipdirList = ['/tmp', '/var', '/etc', '/bin', '/sbin', '/home', '/dev', '/sys', '/usr', '/run']
+        if path not in skipdirList:
             with ZipFile(file=bytesIO, mode='w') as zipFile:
                 for dirName, _, fileList in os.walk(path):
                     for fileName in fileList:
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -287,11 +287,21 @@ class DirectoryResource(Resource):
         else:
             # This is a simple user script (with possibly a few helper files)
             rootDir = path
-        with ZipFile(file=bytesIO, mode='w') as zipFile:
-            for dirName, _, fileList in os.walk(path):
-                for fileName in fileList:
-                    fullPath = os.path.join(dirName, fileName)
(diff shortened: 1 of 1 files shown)


src/toil/resource.py
```diff
@@ -287,11 +287,21 @@ def _load(cls, path):
         else:
             # This is a simple user script (with possibly a few helper files)
             rootDir = path
-        with ZipFile(file=bytesIO, mode='w') as zipFile:
-            for dirName, _, fileList in os.walk(path):
-                for fileName in fileList:
-                    fullPath = os.path.join(dirName, fileName)
-                    zipFile.write(fullPath, os.path.relpath(fullPath, rootDir))
+        blacklistDir = ['/tmp', '/var', '/etc', '/bin', '/sbin', '/home', '/dev', '/sys', '/usr', '/run']
```
mr-c: let's call this the skipdirList, that'd be clearer

jeffrey856: got it! updated it in new commit

src/toil/resource.py
```diff
@@ -287,11 +287,21 @@ def _load(cls, path):
...
+        if path not in skipdirList:
+            with ZipFile(file=bytesIO, mode='w') as zipFile:
+                for dirName, _, fileList in os.walk(path):
+                    for fileName in fileList:
+                        try:
+                            fullPath = os.path.join(dirName, fileName)
+                            zipFile.write(fullPath, os.path.relpath(fullPath, rootDir))
+                        except IOError:
+                            log.critical('Cannot access and read the file at address: %s' % fullPath)
+                            sys.exit()
```
adamnovak: This should be `sys.exit(1)`. As is, we will exit with success.

src/toil/resource.py
```diff
@@ -287,11 +287,21 @@ def _load(cls, path):
...
+        else:
+            log.critical("Couldn't package the directory at %s for hot deployment. Would recommend to create a \
+                subdirectory (ie %s/MYDIR_HERE/)" % (path, path))
+            sys.exit()
```
adamnovak: This should also be `sys.exit(1)`.

src/toil/resource.py
```diff
@@ -287,11 +287,21 @@ def _load(cls, path):
...
+        skipdirList = ['/tmp', '/var', '/etc', '/bin', '/sbin', '/home', '/dev', '/sys', '/usr', '/run']
+        if path not in skipdirList:
+            with ZipFile(file=bytesIO, mode='w') as zipFile:
+                for dirName, _, fileList in os.walk(path):
+                    for fileName in fileList:
+                        try:
+                            fullPath = os.path.join(dirName, fileName)
+                            zipFile.write(fullPath, os.path.relpath(fullPath, rootDir))
+                        except IOError:
+                            log.critical('Cannot access and read the file at address: %s' % fullPath)
```
adamnovak: Say `path` instead of `address` here.

Issue #2648, opened 2019-04-24
Jobs "ended successfully" when they fail


adamnovak:
We report jobs in the log as having "ended successfully" even when the jobs have actually failed. We should not lie to our users; they complain to me that they get confused. We should change that message to "has stopped running" or something.

```
2018-07-03 12:46:55,043 - toil.leader - INFO - Job ended successfully: ...
2018-07-03 12:46:55,050 - toil.leader - WARNING - The job seems to have left a log file, indicating failure: ...
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-3)
┆Issue Number: TOIL-3

jeffrey856:
Made a PR on #2729

adamnovak:
We should probably leave this open until #2729 actually is merged.

If you write "Closes #XXXX" in your PR description, the issue will automatically be closed when the PR is merged.

DailyDreaming:
#2729 was merged.  Closing this.

PR #2729, opened 2019-07-02
changed log output to avoid confusion 

jeffrey856:
Changed potential confusing message

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -491,7 +491,7 @@ class Leader(object):
             if result == 0:
                 cur_logger = (logger.debug if str(updatedJob.jobName).startswith(CWL_INTERNAL_JOBS)
                               else logger.info)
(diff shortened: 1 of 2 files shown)


adamnovak:
I don't think this is what I meant in #2648. I'm after these log messages:

```
2018-07-03 12:46:55,043 - toil.leader - INFO - Job ended successfully: 'file:///<base-path>/touch.cwl' touch y/T/jobPTqCjV
...
2018-07-03 12:46:55,050 - toil.leader - WARNING - The job seems to have left a log file, indicating failure: 'file:///<base-path>/touch.cwl' touch y/T/jobPTqCjV
```

See for example the log in [this comment](https://github.com/DataBiosphere/toil/issues/2274#issue-337978962).

So I guess we're looking for "completed successfully" and not "finished successfully". I will update the issue.

PR #2731, opened 2019-07-08
Check for Partial Download

jeffrey856:
Added few checks to avoid partial download.

### DIFF
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -1161,7 +1161,7 @@ class AWSJobStore(AbstractJobStore):
                     elif info.version:
                         headers = info._s3EncryptionHeaders()
                         key = info.outer.filesBucket.get_key(bytes(info.fileID), validate=False)
-                        with open(writable, 'wb') as fileObject
+                        with open(writable, 'wb') as fileObject:
                             handler = boto.s3.resumable_download_handler.ResumableDownloadHandler(num_retries=10)
                             handler.get_file(key, fileObject, headers, version_id=info.version)
                         if handler.content_length != self.fileID.size:
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -1148,11 +1148,7 @@ class AWSJobStore(AbstractJobStore):
                                                      version_id=self.version,
                                                      headers=headers)
                         if self.fileID.size != key.content_length:
-                            boto.s3.resumable_download_handler.ResumableDownloadHandler.get_file(key)
-                    if self.fileID.size == key.content_length:
-                        break
-                    else:
-                        continue
+                            continue
             else:
                 assert False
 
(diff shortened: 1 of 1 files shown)


src/toil/jobStores/aws/jobStore.py
```diff
@@ -1161,11 +1161,11 @@ def writeTo(self, writable):
                     elif info.version:
                         headers = info._s3EncryptionHeaders()
                         key = info.outer.filesBucket.get_key(bytes(info.fileID), validate=False)
-                        for attempt in retry_s3():
-                            with attempt:
-                                key.get_contents_to_file(writable,
-                                                         headers=headers,
-                                                         version_id=info.version)
+                        with open(writable, 'wb') as fileObject:
```
adamnovak: I don't think this is going to work. Writable is already a file object; you probably can't `open()` it again.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1142,11 +1142,11 @@ def download(self, localFilePath):
...
                 key = self.outer.filesBucket.get_key(bytes(self.fileID), validate=False)
-                for attempt in retry_s3():
-                    with attempt:
-                        key.get_contents_to_filename(localFilePath,
-                                                     version_id=self.version,
-                                                     headers=headers)
+                with open(localFilePath, 'wb') as fileObject:
+                    handler = boto.s3.resumable_download_handler.ResumableDownloadHandler(num_retries=10)
+                    handler.get_file(key, fileObject, headers, version_id=info.version)
+                if handler.content_length != self.fileID.size:
```
adamnovak: I don't see a `content_length` property on the ResumableDownloadHandler documented at https://www.pydoc.io/pypi/boto-2.42.0/autoapi/s3/resumable_download_handler/index.html#s3.resumable_download_handler.ResumableDownloadHandler or implemented at https://chromium.googlesource.com/external/boto/+/77d18c583b43265962e628d8597713841e7dfd62/boto/s3/resumable_download_handler.py so I don't think this check is going to work. Since the resumable download handler handles all the resuming, I think it will just raise its own error if the whole file can't be downloaded, so you might not need this check.

arostamianfar:
@jeffrey856 FYI, this fails due to `AttributeError: 'module' object has no attribute 'resumable_download_handler'` ...and when I add `import boto.s3.resumable_download_handler` , it still fails with:
```
   File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
     self.run()
   File "/usr/local/lib/python2.7/dist-packages/toil/lib/threading.py", line 73, in run
     self.tryRun( )
   File "/usr/local/lib/python2.7/dist-packages/toil/lib/threading.py", line 79, in tryRun
     super( ExceptionalThread, self ).run( )
   File "/usr/lib/python2.7/threading.py", line 754, in run
     self.__target(*self.__args, **self.__kwargs)
   File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/utils.py", line 203, in _writer
     self.writeTo(writable)
   File "/usr/local/lib/python2.7/dist-packages/toil/jobStores/aws/jobStore.py", line 1164, in writeTo
     handler.get_file(key, writable, headers, version_id=info.version)
   File "/usr/local/lib/python2.7/dist-packages/boto/s3/resumable_download_handler.py", line 276, in get_file
     had_file_bytes_before_attempt = get_cur_file_size(fp)
   File "/usr/local/lib/python2.7/dist-packages/boto/s3/resumable_download_handler.py", line 80, in get_cur_file_size
     cur_pos = fp.tell()
 IOError: [Errno 29] Illegal seek
```

Tested on a simple CWL pipeline.

adamnovak:
Uh-oh. @DailyDreaming did we not manage to test this ourselves with the new integration tests?

@jeffrey856 We should probably roll this back until we can actually verify that it works. We might need to wrap the user-submitted output stream in something that counts bytes and provides `tell()` when the user stream doesn't have it, because it looks like the downloader just uses `tell()` on its output destination instead of tracking the bytes downloaded itself.

PR #2737, opened 2019-07-12
Aws python3

TMiguelT:
I alternately used `str()` instead of `bytes()` where the API seemed to want a string type, and `some_string.encode()` when the API wanted a binary type.

This seems to fix the main issues on Python 3, but I'm currently hitting another error that I need to verify (https://github.com/boto/botocore/issues/1780)

Then I need to write a test and check this is all Python 2 compatible as well (assuming we still care about Python 2).

I'm making this a draft PR in case I can't get around to finishing it off, at least it's mostly done here.

### DIFF
--- a/src/toil/batchSystems/mesos/batchSystem.py
+++ b/src/toil/batchSystems/mesos/batchSystem.py
@@ -660,7 +660,7 @@ class MesosBatchSystem(BatchSystemLocalSupport,
         """
         
         # Take it out of base 64 encoding from Protobuf
-        message = decode_data(message)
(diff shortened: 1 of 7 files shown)


DailyDreaming:
@TMiguelT Thanks for your work on this, it served as the basis for the beginning work!  Toil should be fully compatible with python 3.6 now (and is dropping 2.7 support), so I think this PR can be closed.  If you have any concerns or comments, please let us know!

TMiguelT:
Great, thanks!

PR #2746, opened 2019-07-22
Revert "Check for Partial Download"

jeffrey856:
Reverts DataBiosphere/toil#2731


jeffrey856:
@adamnovak I have access to toil-dev aws.

PR #2749, opened 2019-07-25
WIP Run aws jobstore tests.

DailyDreaming:


Makefile
```diff
@@ -158,18 +154,14 @@ test_offline: check_venv check_build_reqs
 
 # The auto-deployment test needs the docker appliance
 test: check_venv check_build_reqs docker
-	TOIL_APPLIANCE_SELF=$(docker_registry)/$(docker_base_name):$(docker_tag) \
-	TRAVIS=true \
```
adamnovak: If you drop TRAVIS=true, then running make test locally skips a bunch of the tests that we dropped from Gitlab by annotating them as Travis-only.

We probably need to change the travis annotation to something like not-Gitlab, so it is clear that we still want them to run when running tests manually.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -765,8 +745,7 @@ def testImportHttpFile(self):
                     httpThread.join()
             finally:
                 http.server_close()
-       
-        @travis_test
```
adamnovak: How come you don't want these to be Travis tests anymore?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -801,7 +779,6 @@ def testFileDeletion(self):
                     # NB: the fooStream() methods return context managers
                     self.assertRaises(NoSuchFileException, self.jobstore_initialized.readFileStream(fileID).__enter__)
 
-        @slow
```
adamnovak: Is this no longer slow?

Makefile
DailyDreaming: I agree with that.  I was trying to think of a better way of doing it, but like your suggestion.  You're right, I didn't expect a review yet.  I'll add the WIP.

src/toil/test/jobStores/jobStoreTest.py
DailyDreaming: It's still slow.  Just cruft since I didn't expect a review yet.  I'll add the WIP.

src/toil/test/jobStores/jobStoreTest.py
DailyDreaming: These tests are inherited by the AWS Jobstore Class which means that some of the AWS jobstore tests will only run on travis, but travis doesn't have `aws` so they're skipped there too.

DailyDreaming:
@adamnovak Sorry, not sure why I flagged the review last week.  I was a bit overzealous.  Still running into issues currently.

adamnovak:
@DailyDreaming Is this PR still needed after #2769?

DailyDreaming:
@adamnovak Nope.

PR #2766, opened 2019-08-16
Py2 jobstore tests.

DailyDreaming:
Implementing this because we really need these before https://github.com/DataBiosphere/toil/pull/2765 can be properly evaluated.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,10 +14,10 @@ after_script:
 stages:
   - main_tests
   - test
+  - integration
 
-#
+
 # Python2.7
-#
 py2_batch_systems:
   stage: test
   script:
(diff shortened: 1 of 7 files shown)


DailyDreaming:
@adamnovak Python 2 jobstore tests are passing.  I had to do a `git push --force` to squash since I accidentally included some binaries so it's deviated from gitlab (so tests won't run).  If you remember how to bring them back in sync, I'd appreciate it!  Otherwise I'll try and solve it sometime between now and Monday afternoon.

There seems to be an option to automatically overwrite diverged branches when mirroring, but I haven't found it.  https://docs.gitlab.com/ee/workflow/repository_mirroring.html#overwrite-diverged-branches-starter  I suspect it might involve re-mirroring and I don't want to chance messing up the webhooks right now, but may later this week.

DailyDreaming:
Closing this in favor of: https://github.com/DataBiosphere/toil/pull/2769

adamnovak:
I just ran into this too with a branch I force pushed. The workaround
is to go to Repository ->< Branches in Gitlab, delete the diverged
branch, and have Gitlab re-check the source repo. The right solution
is to delete and recreate the mirroring as you suggest.

PR #2781, opened 2019-09-04
Add --writeLogsFromAllJobs option.

### LINKED ISSUE #2779: Add an option to log output from all jobs without --logDebug
It's often useful to see the logs from all jobs for debugging. Currently, the only way to turn this on is through `--logDebug`, which enables all kinds of other debug logging and is too noisy.

We can add a new option, `--writeLogsFromAllJobs`, to enable logging from all jobs without enabling debug mode.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-413)
┆Issue Number: TOIL-413

arostamianfar:
Useful option for debugging without enabling `--logDebug`. This is especially useful when trying to get raw logs from several thousand jobs (`--logDebug` is too verbose and slow).

Fixes #2779

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -123,6 +123,7 @@ class Config(object):
         self.maxLogFileSize = 64000
         self.writeLogs = None
         self.writeLogsGzip = None
+        self.writeLogsFromAllJobs = False
         self.sseKey = None
         self.cseKey = None
         self.servicePollingInterval = 60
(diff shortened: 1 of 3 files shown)


arostamianfar:
@DailyDreaming thanks! On our branch, I also added a `failed_` prefix to filenames of failed jobs, so it's easier to distinguish failed logs. Is that something you'd be interested in? I can send a PR for it (also open to other ideas).

DailyDreaming:
@DailyDreaming I think that sounds like a good convention, so yes, I'd welcome a PR adding that change.  Any small quality of life improvements in the code I'm probably going to support.  Thanks again.

Issue #2717, opened 2019-06-20
Kubernetes Batch System


cricketsloan:
Since our lab is probably going to deploy a Kubernetes cluster, and since we have access to the Pacific Research Platform which uses Kubernetes, we ought to have Kubernetes support in Toil.

The easy part is probably the batch system. There's another harder question of providing Toil's Docker functions in an environment where the Docker daemon is not allowed to start and all Docker access must be mediated by Kubernetes, but that's probably a separate issue to work on after this one.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-385)
┆Issue Number: TOIL-385

adamnovak:
@jeffrey856 Here's an example script I put together to talk to Kubernetes, using the credentials and namespace in your `~/.kube/config`.

It might work out of the box with Minikube; for the PRP/Optiputer system I had to create a service account and set up my Kubernetes to use it:

```
kubectl apply -f - <<EOF
apiVersion: v1
kind: ServiceAccount
metadata:
  name: adamnovak-service
EOF

kubectl apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: adamnovak-service-secret
  annotations:
    kubernetes.io/service-account.name: adamnovak-service
type: kubernetes.io/service-account-token
EOF

kubectl apply -f - <<EOF
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: ucsc-cgl
  name: job-runner
rules:
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get", "list", "watch"]
- apiGroups: ["batch"]
  resources: ["jobs"]
  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
EOF

kubectl create rolebinding adamnovak-service-user \
  --role=job-runner \
  --serviceaccount=ucsc-cgl:adamnovak-service \
  --namespace=ucsc-cgl

kubectl describe secrets/adamnovak-service-secret
```

And then I put the key in my `~/.kube/config`:

```
apiVersion: v1
clusters:
- cluster:
[...]

PR #2786, opened 2019-09-05
Kubernetes BatchSystem

jeffrey856:
issues #2717


adamnovak:
OK, this *almost* works.

I had to fix some unrelated issues. For example, I fixed some Python 3 encoding problems in our AWS code, and I enabled caching of MFA logins in boto3 and not just in our boto2-to-3 adapter code. I also added a tiny example workflow from the documentation in `examples/`, because I'm always digging around for one when testing stuff like this.

I had to add a couple hacks, which I guess need to become environment variables:

1. I changed `resolveEntryPoint` in `__init__.py` to always return the bare command, and not a full path, even if the leader is in a virtualenv. With Kubernetes workers, it seems like we will expect people to run the leader locally in a virtualenv, while the workers run on the appliance, so there's no sense looking for the user's virtualenv in the appliance. We probably want to let the old behavior (useful for HPC clusters) be available behind an environment variable.

2. I added code when submitting Kubernetes jobs to mount an AWS access secret, and hardcoded in the name of the one we're using for our GI Kubernetes cluster. We need to make the secret optional (for people using Kubernetes with IAM roles to get at the AWSJobStore, or for people using other cloud job stores), and we need to make the name of the secret to mount selectable by environment variable.

In addition to that, we have some other work to do:

1. We have to grant SimpleDB access for the Kubernetes nodes in the GI cluster. I've e-mailed Erich to do that.

2. We have to set up the Toil appliance to run Python 3, or dual stack, and not Python 2 only. I had to bump down the pickle protocol I am using to talk to the workers because our Docker build process installs Toil only for Python 2.

3. We ought to write a job store that doesn't need SimpleDB.

Here's the command I have been using to test this:

```
TOIL_APPLIANCE_SELF=quay.io/adamnovak/toil:3.21.0a1-1979688a9e3c72efae6a64482a3fde3e30abffd6 python examples/hello.py --logInfo --realTimeLogging --batchSystem kubernetes aws:us-west-2:adamnovak-toiltest || toil clean aws:us-west-2:adamnovak-toiltest
```

Right now this gets as far as shipping the job off to Kubernetes and running it, but then the job can't talk to the job store (because of a lack of SimpleDB authorization), so it fails. The leader successfully detects the failure, dumps the pod log, retries a bit, and eventually gives up and fails the workflow, cleaning up all the failed jobs.

While we wait for Erich to set up SimpleDB access, maybe @jeffrey856 can add in the environment variable hooks I mentioned?

adamnovak:
I've added a commit that fixes the weird errors PRs are getting with the mocks for the cluster scaler tests.

This branch now works with toil-vg set up to use Singularity, run on Kubernetes, and it installs Singularity 3 from source in the Toil appliances.

It still needs some extra logic around Kubernetes job completion/deletion. Either multiple threads can be looking for updated jobs simultaneously, or Kubernetes doesn't really let us block until a job is deleted, so we're getting the same finished job picked up multiple times. We may need a running set of handled completed job IDs so we can only complete jobs once. Alternately, we could refactor to a watch/event-based way of handling completed jobs, but I'd like to merge something that works to master before optimizing in that way.

adamnovak:
The other extension to this is a way to deliver Singularity support via Toil, backporting some of the toil-vg Singularity/Docker switching logic into Toil itself, so workflows don't have to juggle Singularity manually. This would probably let us support CWL Docker requirements on SIngularity, on Kubernetes.

mr-c:
> This would probably let us support CWL Docker requirements on SIngularity, on Kubernetes.


As a reminder, `toil-cwl-runner` already supports using the Singularity runtime to execute Docker format software containers via the `--singularity` option.

Happy to move our docker/Singularity/udocker code to a shared library and merge it with others, yep!

adamnovak:
@mr-c Can toil-cwl-runner's Singularity code handle running inside of an unprivileged container? To get `toil-vg` to work in that context, I had to make sure to always use the `-u` option, to tell Singularity to do all its work with user namespaces, and to use sandbox directories instead of .sif files, because user namespace mode doesn't work with sif files.

Even though you have root in the container, there are things you aren't allowed to do by the kernel, related to container setup, due to being in a container. I think this is so you can't just de-containerize yourself. But Singularity will try to do them by default and give me permission errors.

mr-c:
@adamnovak Not as it is today, but it could easily be taught to do so: https://github.com/common-workflow-language/cwltool/blob/master/cwltool/singularity.py

adamnovak:
@mr-c You mean by porting logic from that file into it? It looks like the linked part of cwltool has autodetection of user namespace support, which is probably something we will need.

It also appears to do its own pulling and caching, which is something I had to implement too.

Does toil-cwl-runner get at Docker through Toil's Docker API? Because I feel like we would want to build all the Singularity management into a similar Toil Singularity API, rather than putting it in at the CWL level only.

mr-c:
@adamnovak Yes, I'm proposing to pull out our Singularity/Docker logic into a separate Python library to share with Toil and whoever.

> Does toil-cwl-runner get at Docker through Toil's Docker API? 

Nope, it uses the `cwltool` code at https://github.com/common-workflow-language/cwltool/blob/master/cwltool/docker.py

adamnovak:
OK, I think this is ready to merge as at least a first cut of Kubernetes support, if it passes tests.

It still needs some of its own tests, and some larger workflows run through it, but I want to get it in master so it's easy for other groups in our lab to test out.

@DailyDreaming Would you have a chance to review this any time soon?

adamnovak:
@mr-c I suppose a separate library makes sense; then the same code could be used in Toil/Toil workflows and in cwlrunner itself.

Do you have a timeline proposal for that project?

examples/hello.py
```diff
@@ -0,0 +1,19 @@
+from toil.common import Toil
```
DailyDreaming: It looks like this file is already at: https://github.com/DataBiosphere/toil/blob/d23f7ec46d2006c136a2a5b4e57eadfb44a606b7/src/toil/test/docs/scripts/tutorial_arguments.py

Can we pull from there?  I'd rather not have two of the same script.

src/toil/batchSystems/kubernetes.py
```diff
@@ -0,0 +1,668 @@
...
+        # The active context gets us our namespace.
+        contexts, active_context = kubernetes.config.list_kube_config_contexts()
+        if not contexts:
+            raise RuntimeError("No Kubernetes contexts available in ~/.kube/config or $KUBECONFIG")
+            
+        # Identify the namespace to work in
+        self.namespace = active_context.get('context', {}).get('namespace', 'default')
+
+        # Create a prefix for jobs, starting with our username
+        self.jobPrefix = '{}-toil-{}-'.format(getpass.getuser(), uuid.uuid4())
```
DailyDreaming: Nit: Might want to limit the length of this?

src/toil/batchSystems/kubernetes.py
```diff
@@ -0,0 +1,668 @@
...
+        """
+
+        return int(jobObject.metadata.name[len(self.jobPrefix):])
+    
+
+    def _genPods(self):
+        for j in _ourJobObjects:
+            yield self._getPodForJob(j)
+
+    def getUpdatedBatchJob(self, maxWait):
```
DailyDreaming: Nit: It might be worth breaking this up.

examples/hello.py
adamnovak: Hm, I didn't know that was there. I can make this a symlink probably.

src/toil/batchSystems/kubernetes.py
adamnovak: I want to leave it as the full UUID, I think. [Kubernetes says we can have 253 characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) and a UUID is pretty short compared to that, and guaranteed to be unique. If we limit it to something more like a Git commit hash, we start getting non-astronomically-improbably collision probabilities (although you'd still have to run a billion Toil workflows at once on the same cluster).

We could cut the user name to some non-excessively-long substring. The other problem is keeping non-ASCII username characters (supported on some not-quite-POSIX systems) out of Kubernetes. I think I might make another PR for this.

src/toil/batchSystems/kubernetes.py
adamnovak: I think we can address this when we hopefully swap over to a watch-based implementation.

adamnovak:
I'm going to merge this and make some followup PRs to address some of the nits. I also want to talk to @DailyDreaming, our Integration Testing Expert, about the best way to set up testing.

PR #2807, opened 2019-10-02
Allow the use of user-provided provisioner (resolves #2806).

### LINKED ISSUE #2806: Allow user-provided provisioner
Provisioner are limited and hard coded to 'aws' 'gce' and 'azure'.
We want to use our own implementation of AbstractProvisioner tu work on OVH infracstructure for instance.

saimeCS:


DailyDreaming:
I agree with the spirit of this but I'm wondering if adding the provisioner you want to use directly wouldn't be easier?  That way unittests could help ensure it's maintained and other people might get the benefit of it.

adamnovak:
So the idea here is that if you pass the name of a class in an installed module to use as a provisioner, it will use that?

I think that makes some sense, but if we're going to add a plugin system here I think we need a bit more documentation than just what's in the option explanation. There ought to be a new documentation page that explains the pluggable provisioner system, describes what the requirements for a provisioner are (i.e. it has to extend AbstractProvisioner, it has to call the base class constructor with super, it has to live as a top-level class in the module it is in, etc.). There should also be a minimal working example provisioner plugin (I suppose with a setup.py and directory structure to install it as the right module name?) that people can use as a base.

I'm also not sure about the "Invalid provisioner" message; since it's going to show up when the requested provisioner string is a perfectly fine specification but the provisioner it mentions is not installed/missing a dependency/has a typo in its source code, it probably should be more like "Unable to load provisioner".

We might want to think about making the batch systems (and job stores?) pluggable like this as well, in which case we'd want one unified way to load these plugins by name. On the other hand, those need to be accessed from the worker as well, which means we'd have the problem of hot-deploying the plugins that would be responsible for telling us how to hot-deploy, whereas the provisioner operates only on the leader.

adamnovak:
I'm going to close this out, since we never actually did the work to pull it in.

I think the project plan right now is to lean more on clusters that know how to scale themselves (via e.g. the Kubernetes-integrated autoscaler) and to try and remove Toil's integrated autoscaler, and eventually (if we can find a replacement) get Toil itself out of the business of building and tearing down clusters. So while we could do #2806 it might be better to have someone who would want to write a Toil provisioner instead write a Kubernetes cluster autoscaler backend, and then run Toil on that. OVH, who was the original target for this work, [seems to support autoscaling Kubernetes clusters](https://help.ovhcloud.com/csm/en-public-cloud-kubernetes-configure-cluster-autoscaler?id=kb_article_view&sysparm_article=KB0049656).

The problem there is getting a Toil job store that works well on Kubernetes outside AWS; we need #3569 to be able to use *just* an S3 API for storage, and we might need some extra logic in the KubernetesBatchSystem to handle auth that doesn't look exactly like the auth on real AWS.

PR #2809, opened 2019-10-05
Pass cmd args to toil-wdl-runner, resolves #2721

azzaea:

PR #2824, opened 2019-10-28
Pass wdl args

adamnovak:
This is a new PR with the commits from #2809, in order to see if I can get the Travis PR tests to not complain that they can't get the merged version of the PR from Github.

I think the problem might be that #2823 was opened and closed, and it confused Travis as to what PR the commits belong to.

### DIFF
--- a/src/toil/wdl/toilwdl.py
+++ b/src/toil/wdl/toilwdl.py
@@ -68,7 +68,7 @@ def main():
     parser.add_argument('secondary_file', help='A secondary data file (json).')
     parser.add_argument("--jobStore", type=str, required=False, default=None)
     parser.add_argument('-o',
-                        '--outdir',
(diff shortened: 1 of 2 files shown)


adamnovak:
@azzaea This still doesn't appear to work. I think https://github.com/DataBiosphere/toil/pull/2809#issuecomment-548007092 is still the problem.

@DailyDreaming, did you ever figure out what we need to rework in order to keep the jobstore generation?

azzaea:
Thank you @adamnovak.
You are right, I got the same failing tests too, tried to get them to work, but eventually had to let it be.

However, I'm still very much interested in this

adamnovak:
I'm going to close this and #2809 because we now run WDL with an interpreter instead of the old code-generation method, and the interpreter handles command line arguments properly.

Issue #2827, opened 2019-10-29
toil launch-cluster --provisioner gce fails with "The referenced image resource cannot be found"


dleehr:
With a pip-installed toil 3.2.0, attempting to launch a Google Cloud cluster fails due to referencing a CoreOS boot image (coreos-stable-1576-4-0-v20171206) that is no longer available:

```
$ toil --version
3.20.0

$ toil launch-cluster gce-demo --provisioner gce --leaderNodeType n1-standard-1 --keyPairName dcl9 --zone us-central1-b
INFO:toil:Using default docker registry of quay.io/ucsc_cgl as TOIL_DOCKER_REGISTRY is not set.
INFO:toil:Using default docker name of toil as TOIL_DOCKER_NAME is not set.
INFO:toil:Overriding docker appliance of quay.io/ucsc_cgl/toil:3.20.0-cf34ca3416697f2abc816b2538f20ee29ba16932 with quay.io/ucsc_cgl/toil:latest from TOIL_APPLIANCE_SELF.
INFO:toil:Using default user-defined custom docker init command of  as TOIL_CUSTOM_DOCKER_INIT_COMMAND is not set.
INFO:toil:Using default docker registry of quay.io/ucsc_cgl as TOIL_DOCKER_REGISTRY is not set.
INFO:toil:Using default docker name of toil as TOIL_DOCKER_NAME is not set.
INFO:toil:Overriding docker appliance of quay.io/ucsc_cgl/toil:3.20.0-cf34ca3416697f2abc816b2538f20ee29ba16932 with quay.io/ucsc_cgl/toil:latest from TOIL_APPLIANCE_SELF.
Traceback (most recent call last):
  File "/Users/dcl9/Code/toil/env/bin/toil", line 8, in <module>
    sys.exit(main())
  File "/Users/dcl9/Code/toil/env/lib/python2.7/site-packages/toil/utils/toilMain.py", line 35, in main
    module.main()
  File "/Users/dcl9/Code/toil/env/lib/python2.7/site-packages/toil/utils/toilLaunchCluster.py", line 162, in main
    awsEc2ProfileArn=config.awsEc2ProfileArn)
  File "/Users/dcl9/Code/toil/env/lib/python2.7/site-packages/toil/provisioners/gceProvisioner.py", line 157, in launchCluster
    ex_preemptible=False)
[...]

PR #2829, opened 2019-10-30
Specify GCE coreos image by family name instead of specific version

dleehr:
When launching a cluster in Google Cloud, uses the family name `coreos-stable` rather than a specific image release `coreos-stable-1576-4-0-v20171206`, which is not currently available


dleehr:
Fixes #2827

PR #2838, opened 2019-11-01
Write tests for Kubernetes

### LINKED ISSUE #2837: Kubernetes batch system is missing CI tests
Someone needs to write CI tests for the Kubernetes batch system.

adamnovak:
This is the first part of #2837.

I still need the code to setup CI to run the tests, but these tests should test the Kubernetes batch system.

I had to tweak the batch system tests slightly to avoid assuming the batch system runs commands on the machine the tests are running on, and to separate different kinds of AWS access.


adamnovak:
@DailyDreaming This has a couple bugfixes for the actual Kubernetes batch system to make it pass the tests. Most importantly it can now return more than one job as being currently running. I want to get it in before the next release.

PR #2841, opened 2019-11-07
Configure Kubernetes for testing

### LINKED ISSUE #2837: Kubernetes batch system is missing CI tests

adamnovak:
This should fully fix #2837 by enabling the Kubernetes batch system CI tests.

It also moves most of the batch system tests off of Travis, since Travis was always skipping them due to not having e.g. mesos installed. Gitlab at least might run them.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -167,7 +167,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
             # We just need the namespace string
             if config_source == 'in_cluster':
                 # Our namespace comes from a particular file.
-                with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", 'r') as fh
+                with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", 'r') as fh:
                     return fh.read().strip()
             else:
                 # Find all contexts and the active context.
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -434,7 +434,6 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
                                 return result
                             else:
                                 continue
-
         else:
             # Try polling instead
             while result is None and (datetime.datetime.now() - entry).total_seconds() < maxWait:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,9 +22,15 @@ py2_batch_systems:
   stage: test
   script:
     - pwd
-    - virtualenv -p python2.7 venv && . venv/bin/activate && make prepare && make develop extras=[all] && pip install htcondor
-    - python -m pytest src/toil/test/batchSystems/batchSystemTest.py
-    - python -m pytest src/toil/test/mesos/MesosDataStructuresTest.py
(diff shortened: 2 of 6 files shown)


src/toil/batchSystems/kubernetes.py
```diff
@@ -154,10 +161,12 @@ def issueBatchJob(self, jobNode):
...
             # https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1Job.md
 
-            # Make a definition for the container's resource requirements
+            # Make a definition for the container's resource requirements.
+            # Don't let people request too-small amounts of memory or disk.
+            # Kubernetes needs some lower limit to run the pod at all without OOMing.
             requirements_dict = {'cpu': jobNode.cores,
-                                 'memory': jobNode.memory,
-                                 'ephemeral-storage': jobNode.disk}
+                                 'memory': max(jobNode.memory, 1024 * 1024 * 512),
```
DailyDreaming: This should probably log a warning to the user if they've specified resources too low for our minimum, otherwise I could see this as as undocumented and confusing reconfiguration of resources.

adamnovak: Do we want the user to raise their Toil job's resource requirement just because Kubernetes can't start up a Toil pod with so little memory? Should the memory requirement for a job cover just the job, or the job and also the Toil overhead necessary to run it?

I was getting OOMs in the tests that asked for 100 MB of memory, so I upped it right to 512. But I don't think the test code in the jobs was itself using anywhere near 100 MB of memory, and the tests run fine wherever else they run (which actually might not be many places, now that I am looking at the batch system tests).

DailyDreaming: @adamnovak I am thinking more of someone who is possibly attempting to optimize by setting the runtime requirements lower and seeing it magically/silently run jobs with an undocumented default.

I think it's sensible, I'm just wondering what the best way to present this to the user is.  A warning just being logged would be pretty immediately obvious, but maybe this could be a note in the documentation?  Do we need documentation on running on kubernetes?

adamnovak: We probably do need some Kubernetes documentation. It's not necessarily immediately obvious that you want to use the AWS job store, for example (or whatever cloud/network jobs tore is appropriate for where your nodes are), or that we have special AWS secret glue.

adamnovak:
@DailyDreaming Can you give this another look? I changed it around a bit.

adamnovak:
This is now also affected by #2873.

DailyDreaming:
@adamnovak I suspect this is due to the addition of `awscli==1.16.278`.  It's needed to fetch secrets but I hadn't tested it on the normal unit tests.

adamnovak:
#2873 was fixed, so I've updated this to current master for re-testing.

arostamianfar:
@adamnovak could you change `export TOIL_DOCKER_REGISTRY` and `TOIL_DOCKER_NAME` to use `?=` instead of `=` like it was before?
The reason is that we're overriding the registry using an env variable and running `make`, so we like to keep the override from an earlier export (outside of make).

adamnovak:
@arostamianfar Sorry. I just opened #2908 to put that back the way it was. Most of the build changes in here I stole from one of Lon's unfinished PRs and I didn't really work out if they were good or not.

PR #2843, opened 2019-11-08
Drop the long-deprecated cwltoil

mr-c:
Same entry point is provided by `toil-cwl-runner`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -169,7 +169,6 @@ def runSetup():
             'console_scripts': [
                 'toil = toil.utils.toilMain:main',
                 '_toil_worker = toil.worker:main',
-                'cwltoil = toil.cwl.cwltoil:main [cwl]',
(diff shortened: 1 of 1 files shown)


DailyDreaming:
@mr-c We'll bump `toil` to 4.0 once we move to only python3.6+ around New Year's.  I'd like to defer breaking changes until then.  Otherwise I agree it's better to just have `toil-cwl-runner`!

mr-c:
@DailyDreaming Sure, that is fine by me

mr-c:
@DailyDreaming can I merge now?

DailyDreaming:
@mr-c There's been a slight hold-up as we transition cactus, which relies on toil on our end, to python3 as well.  I'm told 2-3 weeks?  Will post here when we're ready to release 4.0.0 and include this, I'm eager to remove it too!

PR #2844, opened 2019-11-08
CWL: handle metadata + merge nested

### LINKED ISSUE #2846: CWL: mergenested broken
As reported by @kaushik-work  (thanks!)

```
class: Workflow
cwlVersion: v1.0
inputs:
  in1: string
  in2: string

steps:

  step1:
    in:
      in1: [in1, in2]
    run: list-in.cwl
    out: [out1]

outputs:
  out1: 
    type: string[]
    outputSource:
        - step1/out1
        - in1

requirements: 
  MultipleInputFeatureRequirement: {}
```

```
class: CommandLineTool
cwlVersion: v1.0
inputs:
  in1: string[]

baseCommand: [echo]

outputs:
  out1:
    type: string
    outputBinding:
      outputEval: a_$(inputs.in1)_b
```

Works with plain `cwltool` but not `toil-cwl-runner`

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-447)
┆Issue Number: TOIL-447

mr-c:
Fixes #2846

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -112,7 +112,13 @@ class MergeInputsNested(MergeInputs):
     based on the merge_nested behavior (as described in the CWL spec).
     """
     def resolve(self):
-        return [v[1][v[0]] for v in self.sources]
+        result = []
+        for v in self.sources:
+            if isinstance(v[1], tuple):
+                result.append(v[1][0][v[0]])
+            else:
+                result.append(v[1][v[0]])
+        return result
 
 
 class MergeInputsFlattened(MergeInputs):


kaushik-work:
Request that the fixes be done in Conditionals branch. #2845 Happy to help put the `metadata` (provenance) propagation back. Do request a better propagation mechanism than passing back tuples which (as we have seen) can pollute and corrupt the rest of the code. Thanks!

mr-c:
@kaushik-work What is your proposed other mechanism for the metadata? We only get one return value. We could use a dictionary instead, I guess. Might make the breakage more explicit and the code cleaner.

kaushik-work:
@mr-c I'm thinking of two things. (BTW, should be renamed to provenance rather than `metadata`. `metadata` is used somewhere else too, and it was not clear to me if these two were the same thing.)

(A) Yes, a dictionary would be fine, looking like:

```
{
   "value": jobobj,
   "provenance": provenance
}
```

The important thing is to always pass this dictionary back so that the **data type is consistent** - part of the earlier problem was that the returned value could sometimes be a tuple, sometimes not and so on.

Then wherever this is used, deconstruct the dictionary right at the start of the function and reconstruct it at the end.

(B) Another mechanism, less prone to massive refactoring might be to have each Process object carry a `self.provenance` object. This object is filled out as things progress and parent steps can retrieve this member variable when the step is done.

I feel this is more explicit and less prone to interfering with existing code.

Thanks!

mr-c:
@kaushik-work 
> BTW, should be renamed to `provenance` rather than `metadata`

we named it `metadata` so that it would be a catch-all for any others that needed passing around.

> `metadata` is used somewhere else too, and it was not clear to me if these two were the same thing

The only other reference to `metadata` I can find in `cwltoil.py` is `cwltool.loading_context.metadata`..

How about `toil_metadata` to keep it differentiated from `cwltool`'s `metadata`?

> data type is consistent

Agreed. We implemented the tuple by introducing that and fixing things as they broke with the eventual goal of removing the `isinstance(x, tuple)` calls once everything was normalized. Guess we didn't finish the job!

> Another mechanism, less prone to massive refactoring might be to have each Process object carry a `self.provenance` object.

We looked into that; wasn't clear that would survive the remote execution

mr-c:
regarding the naming, `metadata` vs `provenance` vs `toil_metadata` doesn't matter if we switch to a dictionary, as we can keep `value` for the main result and elevate the other elements of the `metadata` dictionary to top level entries.

kaushik-work:
@mr-c I'll give a shot using a new `ResultWithMetadata` class that subclasses dict and adds a `self.provenance` attribute. This way we won't have to refactor the working code. I'll work on that in a bit.

kaushik-work:
@mr-c how does the `metadata` variable get saved? It disappears after this line: https://github.com/DataBiosphere/toil/blob/f162afe7d4340e9a59f4be8590e62f1230f75411/src/toil/cwl/cwltoil.py#L1365 Thanks!

mr-c:
@kaushik-work we haven't implemented that yet

PR #2848, opened 2019-11-08
added kubernetes watch

jeffrey856:
Added Kubernetes watch


src/toil/batchSystems/kubernetes.py
```diff
@@ -496,47 +492,126 @@ def _getUpdatedBatchJobImmediately(self):
...
+                    waitingInfo = getattr(pod.status.container_statuses[0].state, 'waiting')
+                    if waitingInfo is not None and waitingInfo.reason == 'ImagePullBackOff':
+                        # Assume it will never finish, even if the registry comes back or whatever.
+                        # We can get into this state when we send in a non-existent image.
+                        # See https://github.com/kubernetes/kubernetes/issues/58384
+                        jobObject = j
+                        chosenFor = 'stuck'
+                        logger.warning('Failing stuck job; did you try to run a non-existent Docker image?'
+                                       ' Check TOIL_APPLIANCE_SELF.')
+                        break
```
adamnovak: You've copy-pasted and duplicated about half of the method at a deeper indentation level, before you get to setting up the watch. See https://github.com/DataBiosphere/toil/pull/2848/files#diff-183d11f3af830162fadf362f8c641fdfL499-R537

I don't think you want to do that. This probably resulted form a conflict, since I think I de-indented most of the method when I broke out `_getUpdatedBatchJobImmediately()`.

src/toil/batchSystems/kubernetes.py
```diff
@@ -496,47 +492,126 @@ def _getUpdatedBatchJobImmediately(self):
...
+           
+            w = kubernetes.watch.Watch()    
+            if jobObject is None:
```
adamnovak: Really you need to change the new `getUpdatedBatchJob`; merging has dropped your changes in `_getUpdatedBatchJobImmediately` instead.

See: https://github.com/DataBiosphere/toil/pull/2848/files#diff-183d11f3af830162fadf362f8c641fdfR391-R412

You probably just want to cut-and-paste your new version of this code up there.

src/toil/batchSystems/kubernetes.py
```diff
@@ -496,47 +492,126 @@ def _getUpdatedBatchJobImmediately(self):
...
+                # Nothing is ready yet.
+
+                if self.enableWatching:
+                    for j in self._ourJobObjects():
+                        print(j.spec.template.metadata.labels[u'job-name'], type(j.spec.template.metadata.labels[u'job-name']))
```
adamnovak: We don't want to merge this in with prints in it. They should become `logger.debug` calls probably.

src/toil/batchSystems/kubernetes.py
```diff
@@ -496,47 +492,126 @@ def _getUpdatedBatchJobImmediately(self):
...
+                                    return None
+                                elif job.status.phase == 'Succeeded':
+                                    logger.info("SUCCEEDED NAME: %s EXIT CODE: %s TOTAL TIME: %s" % (job.metadata.name,\
+                                        job.status.container_statuses[0].state.terminated.exit_code, \
+                                        (job.status.container_statuses[0].state.terminated.finished_at - \
+                                        job.status.container_statuses[0].state.terminated.started_at).total_seconds()))
+                                    return (job.metadata.name,\
+                                        job.status.container_statuses[0].state.terminated.exit_code, \
+                                        (job.status.container_statuses[0].state.terminated.finished_at - \
+                                                job.status.container_statuses[0].state.terminated.started_at).total_seconds())
```
adamnovak: Here you're returning the job's name as the first item, where you need to be returning the job's ID number (an int).

```
jobID = int(job.metadata.name[len(self.jobPrefix):])
```

You also have this big long tuple of things that reach deep into `job` repeated twice. You probably want to condense this down:

```
jobID = int(job.metadata.name[len(self.jobPrefix):])
terminated = job.status.container_statuses[0].state.terminated
runtime = (termianted.finished_at - terminated.started_at).total_seconds()
result = (jobID, terminated.exit_code, runtime)
```

And then use `result` instead of the giant expression in both places.

You also need to make sure to [delete the job that has finished, and wait for it to no longer exist](https://github.com/DataBiosphere/toil/pull/2848/files#diff-183d11f3af830162fadf362f8c641fdfR631-R650), before returning. Otherwise, the next time someone asks for an updated job, that job we already reported (when we saw it finish) will still be there in Kubernetes in the finished state, so we'll grab it and report it as having finished again, which we aren't supposed to do.

src/toil/batchSystems/kubernetes.py
```diff
@@ -496,47 +492,126 @@ def _getUpdatedBatchJobImmediately(self):
...
+                        print(j.spec.template.metadata.labels[u'job-name'], type(j.spec.template.metadata.labels[u'job-name']))
+                        for event in w.stream(self.coreApi.list_namespaced_pod, self.namespace, timeout_seconds=maxWait):
+                            job = event['object']
+                            if job.metadata.name.startswith(self.jobPrefix):
+                                logger.info("Event: %s %s %s" % (event['type'],event['object'].kind, event['object'].metadata.name))
+                                if job.status.phase == 'Failure':
+                                    logger.info("FAILED")
+                                    logger.warning(job.status.container_status[0].state.reason,
+                                        job.status.container_statuses[0].state.terminated.exit_code)
+                                    return None
```
adamnovak: The method is supposed to tell the caller about failed jobs; they do in fact count as "updated". Here if the job fails you're just returning None, which isn't right.

jeffrey856:
It ran perfectly fine on my end. I guess when I tried to resolve the merge conflict I rushed it and made a big mistake and created a larger problem... Sorry that was bad.

src/toil/batchSystems/kubernetes.py
```diff
@@ -395,25 +391,31 @@ def getUpdatedBatchJob(self, maxWait):
...
-                else:
-                    continue
-            return None
+            w = kubernetes.watch.Watch()    
+
+            if self.enableWatching:
+                for j in self._ourJobObjects():
+                    logger.debug(j.spec.template.metadata.labels[u'job-name'], type(j.spec.template.metadata.labels[u'job-name']))
+                    for event in w.stream(self.coreApi.list_namespaced_pod, self.namespace, timeout_seconds=maxWait):
+                        job = event['object']
```
adamnovak: You should call this `pod` instead of `job` since it holds a pod.

src/toil/batchSystems/kubernetes.py
```diff
@@ -395,25 +391,31 @@ def getUpdatedBatchJob(self, maxWait):
...
+                                    job.status.container_statuses[0].state.terminated.exit_code)
+                                jobID = int(job.metadata.owner_references[0].name[len(self.jobPrefix):])
+                                terminated = job.status.container_statuses[0].state.terminated
+                                runtime = (terminated.finished_at - terminated.started_at).total_seconds()
+                                result = (jobID, terminated.exit_code, runtime)
+                                # Deleting not Working 404 client.rest.ApiException
+                                self.batchApi.delete_namespaced_job(job.metadata.owner_references[0].name,
+                                                                    self.namespace,
+                                                                    propagation_policy='Foreground')
+
```
adamnovak: After you call `delete_namespaced_job` you need to call `_waitForJobDeath` to wait around for it to go away. Otherwise the job you just asked to delete can be sent back to you when you list jobs later, since it isn't deleted yet.

adamnovak:
@jeffrey856 Since the Kubernetes stuff isn't under proper CI yet, can you run the Kubernetes batch system tests against the cluster to make sure your fixes pass them? Once you make sure you are logged into Quay (`docker login quay.io`), and that you are in a virtualenv, it works something like:

```
git checkout 91250d3cb373c4b7669766acbca87bfb27e00be4
TOIL_DOCKER_REGISTRY=quay.io/<YOUR_USERNAME> make docker
TOIL_DOCKER_REGISTRY=quay.io/<YOUR_USERNAME> make push_docker

TOIL_AWS_SECRET_NAME=shared-s3-credentials TOIL_APPLIANCE_SELF=quay.io/<YOUR_USERNAME>/toil:3.21.0a1-91250d3cb373c4b7669766acbca87bfb27e00be4 TRAVIS=true python -m pytest -vv -r s --timeout=530 src/toil/test/batchSystems/batchSystemTest.py::KubernetesBatchSystemTest
```

adamnovak:
@jeffrey856 It looks like this is failing because it doesn't have a commit that's in master that fixed a test failure that got into master. You should merge master into your branch.

PR #2851, opened 2019-11-12
Try service account configuration for Kubernetes

### LINKED ISSUE #2850: Kubernetes batch system can't pick up config from a pod's service account
The config loading method we're using to connect to Kubernetes, as described in the module's README, only works when your credentials come from a config file. When they come from a pod's service account, `kubectl` finds them automatically, but the `kubernetes` module needs you to manually call a differnet utility method (`config.load_incluster_config()`) instead.

See:

* Example: https://github.com/kubernetes-client/python/blob/master/examples/in_cluster_config.py

* Implementation/utility method docs: https://github.com/kubernetes-client/python-base/blob/4c1ab55553fa7d0479bd5b270c2b303cb056dd4b/config/incluster_config.py

* Kubernetes issue: https://github.com/kubernetes-client/python/issues/1005

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-449)
┆Issue Number: TOIL-449

adamnovak:
`kubectl` is clever enough to pick up its config and credentials from an enclosing pod, but the `kubernetes` Python module can't do it by itself.

This PR adds some code to try and grab the Kubernetes configuration form the currently running pod if it can't be obtained from a config file.

This fixes #2850.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -73,19 +73,32 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
         logging.getLogger('requests_oauthlib').setLevel(logging.ERROR)
 
         try:
-            # Load ~/.kube/config
+            # Load ~/.kube/config or KUBECONFIG
             kubernetes.config.load_kube_config()
-        except TypeError:
-            raise RuntimeError('Could not load Kubernetes configuration. Does ~/.kube/config or $KUBECONFIG exist?')
-
-        # Find all contexts and the active context.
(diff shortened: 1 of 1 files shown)


adamnovak:
I'm also dropping some more fixes in here for cases where Kubernetes sends us pods that don't have any container statuses.

adamnovak:
OK, it looks like the cases where we get pods with no container status are:

1. We're polling jobs and the job isn't really running yet

2. The pod was stopped (or is stopping?) due to exceeding a (always disk?) limit, and we see it as having failed

PR #2857, opened 2019-11-14
[WIP] record the maximum memory used by LSF jobs

mr-c:
- [ ] merge this `bhist` call with the `bjobs` calls below?
- [ ] parse the "human friendly" memory value into an int/long?
- [ ] refactor to return the result code & the memory used together?

### DIFF
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -98,12 +98,15 @@ class LSFBatchSystem(AbstractGridEngineBatchSystem):
                 meminfo = re.search(memregex, bhist)
                 command = re.search(r".*Command <(\S*)>.*\n", bhist)
                 if meminfo:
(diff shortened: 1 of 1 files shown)


mr-c:
@drkennetz would you have any interest in taking over this PR?

drkennetz:
@mr-c yeah I'll take it over! should be easy enough. Max mem can be determined from bjobs -l which I think is already called somewhere. Is there any value in actually returning the Max Mem used, rather than just logging it? I bring that up because this would have to be reflected in leader.py and job.py, which I think would result in some ugly and perhaps unnecessary isintance checks until this change was synced across all schedulers.

mr-c:
@drkennetz you can log it for now; some day we will want to return it for provenance recording ; thanks!

drkennetz:
@mr-c I've started work on this yesterday. I'll probably open a new PR when ready.

PR #2868, opened 2019-11-18
Raise job limits to be more than their requests

adamnovak:
This fixes #2866 and also includes some quieting of the logging from event watching, which was swamping the actual output from my jobs.

PR #2869, opened 2019-11-18
Constantly reload Kubernetes config

### LINKED ISSUE #2867: Kubernetes credentials can only be refreshed once per Kubernetes API object, and can eventually expire, leading to Unauthorized errors from the Kubernetes API
See https://github.com/kubernetes-client/python/issues/741

The Kubernetes module establishes some credentials at startup but never refreshes them during the life of the program.

When the credentials expire, (and we start getting `401 Unauthorized`-s), we need to call `load_kube_config()` again (if that's how we got our credentials) and replace all the API clients.

Alternately, we could re-load the config and re-make the client at every use of Kubernetes, or just reload the config "sufficiently often" and remake the clients on every use, or something.

@glennhickey is seeing problems due to this (I think) on our Kubernetes cluster after an hour.

A workaround is running with service account credentials, which don't expire.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-460)
┆Issue Number: TOIL-460

adamnovak:
This (written on top of #2868) will fix #2867 by reloading the Kubernetes config and making a new API clients every 5 minutes, as suggested in https://github.com/kubernetes-client/python/issues/741#issuecomment-518052241.

We'll still get in trouble if we have a single watch or loop over pods that runs for more than the credential expiration time, but that seems unlikely to happen and would be its own slowness problem anyway.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -71,34 +71,14 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
         # reveal on CI.
         logging.getLogger('kubernetes').setLevel(logging.ERROR)
         logging.getLogger('requests_oauthlib').setLevel(logging.ERROR)
-
-        try:
-            # Load ~/.kube/config or KUBECONFIG
-            kubernetes.config.load_kube_config()
-            
-            # We loaded it; we need to figure out our namespace the config-file way
-            
-            # Find all contexts and the active context.
-            # The active context gets us our namespace.
-            contexts, activeContext = kubernetes.config.list_kube_config_contexts()
-            if not contexts:
-                raise RuntimeError("No Kubernetes contexts available in ~/.kube/config or $KUBECONFIG")
-                
-            # Identify the namespace to work in
-            self.namespace = activeContext.get('context', {}).get('namespace', 'default')
-            
-        except TypeError:
-            # Didn't work. Try pod-based credentials in case we are in a pod.
-            try:
-                kubernetes.config.load_incluster_config()
-                
-                # We got pod-based credentials. Our namespace comes from a particular file.
-                self.namespace = open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", 'r').read().strip()
-                
-            except kubernetes.config.ConfigException:
-                raise RuntimeError('Could not load Kubernetes configuration from ~/.kube/config, $KUBECONFIG, or current pod.')
-
         
+        # This will hold the last time our Kubernetes credentials were refreshed
+        self.credential_time = None
+        # And this will hold our cache of API objects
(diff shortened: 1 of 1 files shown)


adamnovak:
@DailyDreaming I got what I think is an unrelated test failure from the integration tests:

```
_ EncryptedAWSJobStoreTest.test_testImportSharedFile__otherCls_awsjobstoretest _
  File "/builds/databiosphere/toil/src/toil/test/jobStores/jobStoreTest.py", line 1061, in setUp
    super(AbstractEncryptedJobStoreTest.Test, self).setUp()
  File "/builds/databiosphere/toil/src/toil/test/jobStores/jobStoreTest.py", line 124, in setUp
    self.jobstore_initialized.initialize(self.config)
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 148, in initialize
    super(AWSJobStore, self).initialize(config)
  File "/builds/databiosphere/toil/src/toil/jobStores/abstractJobStore.py", line 146, in initialize
    self.writeConfig()
  File "/builds/databiosphere/toil/src/toil/jobStores/abstractJobStore.py", line 154, in writeConfig
    pickle.dump(self.__config, fileHandle, pickle.HIGHEST_PROTOCOL)
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 552, in writeSharedFileStream
    yield writable
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 1122, in uploadStream
    assert bool(self.version) == (self.content is None)
AssertionError
```

I'm going to rerun the test, but we also might have a bug.

src/toil/batchSystems/kubernetes.py
```diff
@@ -125,11 +105,79 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
...
+        self._apis['core'] = kubernetes.client.CoreV1Api()
+        
+        # And save the time
+        self.credential_time = now
+        
+        if kind == 'namespace':
+            # We just need the namespace string
+            if config_source == 'in_cluster':
+                # Our namespace comes from a particular file.
+                return open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", 'r').read().strip()
```
DailyDreaming: `.close()`

adamnovak: The file should close when it goes out of scope (at least on CPython), but I've put in a `with` here.

DailyDreaming: Thanks!

PR #2871, opened 2019-11-21
Fix BotoCredentialAdapter

jeffrey856:
Attempts to fix Cactus that is trying to pickle an open Boto connection


src/toil/__init__.py
```diff
@@ -456,214 +452,223 @@ def str_to_datetime(s):
...
+                        log.debug('Wrote credentials to %s. Renaming to %s.', tmp_path, path)
+                    os.rename(tmp_path, path)
+                    return
+                finally:
+                    if fd is not None:
+                        os.close(fd)
+                        
+
+# Now we have defined the adapter class. Patch the Boto module so it replaces the default Provider when Boto makes Providers.
+provider.Provider = BotoCredentialAdapter
```
adamnovak: This line should stay inside the monkey patching function.

src/toil/__init__.py
```diff
@@ -456,214 +452,223 @@ def str_to_datetime(s):
...
-    class BotoCredentialAdapter(provider.Provider):
-        """
-        Adapter to allow Boto 2 to use AWS credentials obtained via Boto 3's
-        credential finding logic. This allows for automatic role assumption
-        respecting the Boto 3 config files, even when parts of the app still use
-        Boto 2.
+
+from boto import provider
+
+class BotoCredentialAdapter():
```
adamnovak: This still needs to inherit from the provider class.

src/toil/__init__.py
```diff
@@ -456,214 +452,223 @@ def str_to_datetime(s):
...
         return datetime.strptime(s, datetime_format)
 
-    class BotoCredentialAdapter(provider.Provider):
-        """
-        Adapter to allow Boto 2 to use AWS credentials obtained via Boto 3's
-        credential finding logic. This allows for automatic role assumption
-        respecting the Boto 3 config files, even when parts of the app still use
-        Boto 2.
+
+from boto import provider
```
adamnovak: You can't import from boto at module scope if boto isn't installed. You need something like:

```
try:
    from boto import provider
    base_class = provider.Provider
except ImportError:
    base_class = object

class BotoCredentialAdapter(base_class):
```

src/toil/__init__.py
```diff
@@ -456,215 +453,229 @@ def str_to_datetime(s):
...
+    credential finding logic. This allows for automatic role assumption
+    respecting the Boto 3 config files, even when parts of the app still use
+    Boto 2.
+    
+    This class also handles cacheing credentials in multi-process environments
+    to avoid loads of processes swamping the EC2 metadata service.
+    """
+    
+    from botocore.session import Session
+    from botocore.credentials import create_credential_resolver, RefreshableCredentials, JSONFileCache
```
adamnovak: You might want to move these imports into the `try` above. I don't think they'll work here.

jeffrey856:
new clean PR at #2904

Issue #2873, opened 2019-11-21
One of our integration tests has started failing on unrelated PRs


adamnovak:
The Python 2 integration tests are failing with this:

```
_____ AWSJobStoreTest.test_testImportSharedFile__otherCls_filejobstoretest _____
Traceback (most recent call last):
  File "/builds/databiosphere/toil/src/toil/test/jobStores/jobStoreTest.py", line 124, in setUp
    self.jobstore_initialized.initialize(self.config)
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 148, in initialize
    super(AWSJobStore, self).initialize(config)
  File "/builds/databiosphere/toil/src/toil/jobStores/abstractJobStore.py", line 146, in initialize
    self.writeConfig()
  File "/builds/databiosphere/toil/src/toil/jobStores/abstractJobStore.py", line 154, in writeConfig
    pickle.dump(self.__config, fileHandle, pickle.HIGHEST_PROTOCOL)
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 552, in writeSharedFileStream
    yield writable
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/builds/databiosphere/toil/src/toil/jobStores/aws/jobStore.py", line 1122, in uploadStream
    assert bool(self.version) == (self.content is None)
AssertionError
```

This affects #2869 and #2872. I think there is something wrong with the code as it exists in master; AWS may have changed behavior.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-462)
┆Issue Number: TOIL-462

PR #2875, opened 2019-11-22
Set an older awscli version to test Lon't theory that it is breaking our boto

adamnovak:


adamnovak:
I'm going to merge this without review to make master stop failing its tests.

adamnovak:
This happened to pass its own tests, but did not actually fix #2873 and the master build resulting from the merge failed. Our problem is intermittent and apparently not caused by a botocore change after November 1st 2019.

PR #2877, opened 2019-12-05
Change Kubernetes job sizing logic

adamnovak:
This will fix #2876 by not floor-ing the CPU requirement.

It now uses the same requirements and limits, to make our admins happy, and also now adds half a gig of memory and storage for Toil overhead (like the hot-deployed user scripts, or the logs), instead of using that as a minimum job size, which didn't quite make sense.

PR #2880, opened 2019-12-06
Up the disk to the default as well as the memory

### LINKED ISSUE #2879: Have a way to increase the disk requirement of failed jobs when retrying them
In #2876 @glennhickey complained that, with Kubernetes now enforcing disk limits, you can get in trouble if your workflow has a job with a too-low disk limit, because there's no way to increase it once the workflow has started.

There should be a way to recover a workflow that has a job that can't complete with the amount of disk it thinks it needs, instead of starting over completely.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-465)
┆Issue Number: TOIL-465

adamnovak:
This should fix #2879; you should be able to `--restart` with a higher `--defaultDisk` (and a nonzero number of retries) and at least eventually get your job to run through, after one more failure at the old disk limit.

What we might really want is a separation between the default job size and the max job size we will increase to, and some kind of doubling logic that is ideally aware of what limit a job hit when it failed. But this should at least allow recovering a stuck workflow.

### DIFF
--- a/src/toil/jobGraph.py
+++ b/src/toil/jobGraph.py
@@ -117,8 +117,13 @@ class JobGraph(JobNode):
         # batch system)
         if self.memory < config.defaultMemory:
             self._memory = config.defaultMemory
-            logger.warn("We have increased the default memory of the failed job %s to %s bytes",
+            logger.warn("We have increased the memory of the failed job %s to the default of %s bytes",
                         self, self.memory)
+            
+        if self.disk < config.defaultDisk:
+            self._disk = config.defaultDisk
+            logger.warn("We have increased the disk of the failed job %s to the default of %s bytes",
+                        self, self.disk)
 
     def restartCheckpoint(self, jobStore):
         """Restart a checkpoint after the total failure of jobs in its subtree.


adamnovak:
CI says:

```
_____________________ JobServiceTest.testServiceRecursive ______________________
  File "/builds/databiosphere/toil/src/toil/test/src/jobServiceTest.py", line 133, in testServiceRecursive
    self.runToil(t)
  File "/builds/databiosphere/toil/src/toil/test/src/jobServiceTest.py", line 185, in runToil
    self.fail() #Exceeded a reasonable number of restarts
  File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
AssertionError: None
```

This is the same problem as #2877; definitely a bug, either in the tests, our infrastructure, or something we did in master that managed not to break master.

adamnovak:
I merged #2883 into this PR in hopes that it will make the tests pass. Though really I think we have an intermittent problem in master...

adamnovak:
I got this from the Python 3 jobstore tests:

```
_ EncryptedAWSJobStoreTest.testSDBDomainsDeletedOnFailedJobstoreBucketCreation _
  File "/builds/databiosphere/toil/src/toil/test/jobStores/jobStoreTest.py", line 1258, in testSDBDomainsDeletedOnFailedJobstoreBucketCreation
    s3.delete_bucket(bucket=bucket)
  File "/builds/databiosphere/toil/venv/lib/python3.6/site-packages/boto/s3/connection.py", line 648, in delete_bucket
    response.status, response.reason, body)
boto.exception.S3ResponseError: S3ResponseError: 404 Not Found
<Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><BucketName>domain-test-3a585564-91d7-496a-96ba-cc913cef84d6--files</BucketName><RequestId>80B61A7C59F3A543</RequestId><HostId>GxTgOE/cFzGeJekWxLlDBsTWzzXgLqIR1h4BOoGQ5ejpyr4FvWjznfgCuJD4w36zsue/dDlg8i8=</HostId></Error>
```

I don't think that has anything to do with this PR.

PR #2889, opened 2019-12-18
Add checksum verification for AWS job store.

### LINKED ISSUE #2870: Add checksum verification to AWS S3 jobstore
Large files downloaded from S3 can get corrupted for a number of reasons (e.g. due to network or disk issues) and we've seen this happen very rarely when running large scale pipelines. However, when it does happen, it can lead to unexpected failures, or worse, incorrect results [*].

S3 does not have a good checksum value for large files that get uploaded via multipart uploads (see: https://teppen.io/2018/06/23/aws_s3_etags/ ), so the easiest way is to add the `sha1` checksum as an additional value to the `SimpleDB` data and use that when downloading/uploading files to/from S3. I propose using `sha1` as it's also used in CWL.

Ideally other jobstores should also have this, but I've limited the scope to S3 for this issue.

[*] I had added a checksum check on our internal branch to see whether it would ever fire, and just saw an instance of it happening in a production Toil pipeline! So, decided to prioritize adding it to master as well.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-461)
┆Issue Number: TOIL-461

arostamianfar:
This ensures that files are not corrupted during transfer. Since this is a very rare, but important event, I figured it's simpler to just fail and let the pipeline retry the task rather than trying to re-download the file. Also, since this is a safety check, I think it should be enabled by default and only disabled if explicitly specified.

I used `sha1` to match the checksums from CWL.

Tested on various pipelines for a couple of months on our branch. So far, we've seen at least one instance of this happening in a production run when handling large files.
 
Fixes #2870

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -120,6 +120,7 @@ class Config(object):
         # Misc
         self.disableCaching = True
         self.disableChaining = False
+        self.disableJobStoreChecksumVerification = False
         self.maxLogFileSize = 64000
         self.writeLogs = None
         self.writeLogsGzip = None
@@ -268,6 +269,7 @@ class Config(object):
         setOption("maxLocalJobs", int)
         setOption("disableCaching")
         setOption("disableChaining")
+        setOption("disableJobStoreChecksumVerification")
         setOption("maxLogFileSize", h2b, iC(1))
         setOption("writeLogs")
         setOption("writeLogsGzip")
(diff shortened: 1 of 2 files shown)


diekhans:
I would agree with enabled by default.  Much better to be safe and the user explicitly disable at their own risk.

src/toil/common.py
```diff
@@ -539,6 +541,11 @@ def _addOptions(addGroupFn, config):
     addOptionFn('--disableChaining', dest='disableChaining', action='store_true', default=False,
                 help="Disables chaining of jobs (chaining uses one job's resource allocation "
                 "for its successor job if possible).")
+    addOptionFn("--disableJobStoreChecksumVerification", dest="disableJobStoreChecksumVerification",
+                default=False, action="store_true",
+                help=("Disables checksum verification for files transferred to/from the job store. "
+                      "Checksum verification is a safety check to ensure the data is not corrupted "
+                      "during transfer. Currently only supported for non-streaming AWS files."))
```
adamnovak: This is kind of ambiguous between the verification or the disabling as the thing being only supported for non-streaming AWS files.

arostamianfar: ah, good point! Will watch out for this in the future!

PR #2896, opened 2019-12-19
Add --enableUnlimitedPreemptableRetries option.

### LINKED ISSUE #2890: Add an option to not count preemptible failures towards task failures
Currently preemptible/spot failures also count towards task failures. As a result, `retryCount`  needs to be set to a reasonably high value when using spot/preemptible instances, which can result in many unnecessary retries when tasks are actually failing due to other reasons (this can become a problem if there are a large number of similar tasks that fail for the same reason).

Ideally, spot failures should be separated out from other types of failures. Some other workflow engines have a "max preemptible try" feature that allocates tasks on regular instances after X retries, but implementing that on Toil is more involved.

In the short term, I propose adding a flag (e.g. `--enableUnlimitedPreemptibleRetries`) that does not count preemptible failures (i.e. anything with `TASK_LOST` status) towards failures. This is a relatively simple change and allows setting `retryCount` to a low value.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-471)
┆Issue Number: TOIL-471

arostamianfar:
This change adds the option to set `--retryCount` to a low value since spot/premptable terminations no longer count as failures.

I decided to use `2550` as the exit code for lost jobs, but open to other suggestions as well! Rationale:
* Using anything but an int can result in unexpected bugs, since there could be assumptions in the code about the nature of the arg (now or in the future).
* Using anything in 0-255 can be problematic as it can conflict with other values.
* Using `-1` is also a good option, but I figured it's a 'common' special value and someone else could use it for something else accidentally.

Tested with and without `--enableUnlimitedPreemptableRetries` on a pipeline and forced terminations.

Fixes #2890

### DIFF
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -37,6 +37,10 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+# Special (fake) exit code for jobs that are lost due to preemption.
+# We use a code that is outside of the normal 0-255 range on purpose here to ensure it
+# doesn't conflict with other exit codes.
+LOST_JOB_EXIT_STATUS = 2550
 
 # A class containing the information required for worker cleanup on shutdown of the batch system.
 WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -40,7 +40,9 @@ logger = logging.getLogger(__name__)
 
 UpdatedBatchJobInfo = namedtuple('UpdatedBatchJobInfo', (
     'jobID',
-    'exitStatus',  # The exit status (integer value) of the job. 0 implies successful.
+    # The exit status (integer value) of the job. 0 implies successful.
+    # EXIT_STATUS_UNAVAILABLE_VALUE is used when the exit status is not available (e.g. job is lost).
+    'exitStatus',
     'exitReason',  # The exit reason, if available. One of BatchJobExitReason enum.
     'wallTime'))
 
@@ -52,6 +54,9 @@ class BatchJobExitReason(enum.Enum):
     KILLED = 4  # Job killed before finishing.
     ERROR = 5  # Internal error.
 
+
+# Value to use as exitStatus in UpdatedBatchJobInfo.exitStatus when status is not available.
+EXIT_STATUS_UNAVAILABLE_VALUE = 255
 
 # A class containing the information required for worker cleanup on shutdown of the batch system.
 WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
--- a/src/toil/batchSystems/abstractBatchSystem.py
+++ b/src/toil/batchSystems/abstractBatchSystem.py
@@ -16,6 +16,7 @@ from future import standard_library
 standard_library.install_aliases()
 from future.utils import with_metaclass
 from builtins import object
+import enum
 import os
 import shutil
 import logging
@@ -37,10 +38,20 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
-# Special (fake) exit code for jobs that are lost due to preemption.
-# We use a code that is outside of the normal 0-255 range on purpose here to ensure it
-# doesn't conflict with other exit codes.
-LOST_JOB_EXIT_STATUS = 2550
+UpdatedBatchJobInfo = namedtuple('UpdatedBatchJobInfo', (
+    'jobID',
+    'exitStatus',  # The exit status (integer value) of the job. 0 implies successful.
+    'exitReason',  # The exit reason, if available. One of BatchJobExitReason enum.
+    'wallTime'))
+
+
+class BatchJobExitReason(enum.Enum):
+    FINISHED = 1  # Successfully finished.
+    FAILED = 2  # Job finished, but failed.
+    LOST = 3  # Preemptable failure.
+    KILLED = 4  # Job killed before finishing.
+    ERROR = 5  # Internal error.
+
 
 # A class containing the information required for worker cleanup on shutdown of the batch system.
 WorkerCleanupInfo = namedtuple('WorkerCleanupInfo', (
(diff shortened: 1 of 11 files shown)


src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -37,6 +37,10 @@
 
 logger = logging.getLogger(__name__)
 
+# Special (fake) exit code for jobs that are lost due to preemption.
+# We use a code that is outside of the normal 0-255 range on purpose here to ensure it
+# doesn't conflict with other exit codes.
+LOST_JOB_EXIT_STATUS = 2550
```
adamnovak: I still think this should be a non-integer, and we should go through and change things to understand that the return value isn't always an exit code. Right now the Kubernetes batch system spits out -1 in cases where it couldn't get an exit code; we could also change that to something more descriptive.

I think to make it work all we would really have to change is https://github.com/DataBiosphere/toil/blob/8e1d8c0f45f6210bbd0900380b62c5a1c1a6cdc7/src/toil/leader.py#L498

Even if we leave it like this, we shouldn't call this an EXIT_STATUS; it should be something else (stop reason?) and the getUpdatedBatchJob docs should be changed accordingly.

src/toil/common.py
```diff
@@ -520,6 +522,11 @@ def _addOptions(addGroupFn, config):
     addOptionFn("--retryCount", dest="retryCount", default=None,
                 help=("Number of times to retry a failing job before giving up and "
                       "labeling job failed. default=%s" % config.retryCount))
+    addOptionFn("--enableUnlimitedPreemptableRetries", dest="enableUnlimitedPreemptableRetries",
```
adamnovak: Maybe this should be `--preemptibleRetryCount` and accept a value of `inf` instead?

src/toil/leader.py
```diff
@@ -827,7 +828,9 @@ def processFinishedJob(self, batchSystemID, resultStatus, wallTime=None):
                             else:
                                 logger.warn('The batch system left an empty file %s' % batchSystemFile)
 
-                jobGraph.setupJobAfterFailure(self.config)
+                # Do not decrement retry count for lost jobs (i.e. due to preemptable failures).
+                decrementRetryCount = not self.config.enableUnlimitedPreemptableRetries or resultStatus != LOST_JOB_EXIT_STATUS
```
adamnovak: I think this check should live in `setupJobAfterFailure`. It makes more sense to pass it the failure reason for the job than a single-purpose flag, I think, and it already has access to the config.

src/toil/batchSystems/abstractBatchSystem.py
arostamianfar: I still feel uneasy about changing the type to be `stri|int`: while we can ensure the code in the current state can handle that, any future implementation requires thinking about both cases separately and may introduce bugs.

What do you think about adding another parameter to the tuple, `exitReason: str|None`, that's independent of `exitStatus`? The new field could be none for majority of cases, but we could have special ones for cases that we care about.

Doing this properly would require refactoring all use cases to (ideally) use a class/namedtuple [*], so that the code is more readable, but I think it's worth the effort and we could add additional fields easier in the future.

[*] Using namedtuple would be ideal, but defaults are not easily supported before python3.7. We could require all setters to set all fields, but it would make it more difficult if we wanted to add new fields later.

adamnovak:
Maybe the Right Way to do this is a better structured job result tuple, with both exit code (or None) and exit reason (or None), with the latter coming from a restricted set of machine readable reason strings (WORKER_HOST_GONE, etc).

NamedTuple wouldn't be too bad; we only manufacture them inside batch systems, of which there are only a few.

Then we can hand the whole tuple to setupJobAfterFailure if we have it, as a second argument with default value None, and it can take appropriate action based on the kind of failure, if known.

How urgently do you want this merged in? We could go with fake exit codes for now, because the new system might take a while to write.

arostamianfar:
SGTM! Not urgent. I've already merged the fake exit code version on our batch that we use. This is just to sync it with master and enable others to use the feature too. I can work on it after the holidays!

src/toil/common.py
arostamianfar: This is the only one I haven't changed: mainly b/c it's a lot of work to actually implement proper counting and the specs for having such a value aren't clear (do we start using normal instances after that count or just fail?), and using a float for this felt awkward.
I figured when/if we have the proper implementation for counting failures, we could make it so that setting this bool sets the retry count to a large value and/or deprecate it.
Anyhow lemme know if you feel strongly about it and I can change it :)

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -37,6 +38,20 @@
 
 logger = logging.getLogger(__name__)
 
+UpdatedBatchJobInfo = namedtuple('UpdatedBatchJobInfo', (
+    'jobID',
+    'exitStatus',  # The exit status (integer value) of the job. 0 implies successful.
```
adamnovak: The Mesos batch system puts a 255 here if the exit status is unavailable, whereas it looks like the Kubernetes batch system and single machine batch system use -1. We should probably settle on one (or say that either is allowed) and document that as part of the spec for all the batch systems.

src/toil/common.py
adamnovak: I think this is fine, if we can't actually count it. We just have to be careful about jobs that manage to take down their machines; we'll retry a `shutdown -h now` forever.

adamnovak:
We don't actually have testing set up for the grid engine based batch systems, since we don't have a grid engine cluster laying around. It's a bit of a to do item to get one set up, because a lot of users care about them.

We do have CI on the Kubernetes batch system set up, so that should be tested by Gitlab.

src/toil/batchSystems/mesos/batchSystem.py
```diff
@@ -647,12 +648,15 @@ def jobEnded(_exitStatus, wallTime=None):
                             update.message, update.reason,
                             update.executor_id, update.agent_id)
             
-            jobEnded(exitStatus)
-        elif update.state in ('TASK_LOST', 'TASK_KILLED', 'TASK_ERROR'):
+            jobEnded(exitStatus, exitReason=BatchJobExitReason.FAILED)
+        elif update.state == 'TASK_LOST':
+            log.warning("Job %i is lost. Retrying.", jobID)
```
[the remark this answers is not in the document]
arostamianfar: Good point! Done.

src/toil/common.py
arostamianfar: hmm...interesting, since we run everything in docker on Mesos, it presumably shouldn't allow that level of access to the instance (although, not sure what happens if a task manages to crash our docker; would it get the LOST status or ERROR?). Anyhow, I think that risk should be pretty low at least on AWS/Mesos side.

src/toil/batchSystems/abstractBatchSystem.py
arostamianfar: Good point! Created a constant (decided to use 255) and used it everywhere.

PR #2903, opened 2020-01-03
Set up Gitlab testing for Kubernetes runners

### LINKED ISSUE #2902: Gitlab runners should be moved to Kubernetes and our .gitlab-ci.yml should support that
I've already got a runner set up on our Kubernetes cluster, but I need to work around https://gitlab.com/gitlab-org/gitlab-runner/issues/4125 which prevents the Docker daemon from automatically starting up.

I think I am going to tie the .gitlab-ci.yml specifically to the image we are using, and let it just start and stop Docker itself in the pre/post build steps.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-477)
┆Issue Number: TOIL-477

adamnovak:
This will fix #2902.

When nobody else is using CI, I am going to want to manually pause the non-Kubernetes runner, unpause the Kubernetes runner, and verify that these changes are indeed sufficient to make jobs run through on the Kubernetes runner. I'm not sure if I want to do that before or after we merge this; there should be no trouble with these changes on the existing non-Kubernetes runner.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -34,40 +34,40 @@ py2_batch_systems:
     - cp "$GITLAB_SECRET_FILE_KUBE_CONFIG" ~/.kube/config
     - mkdir -p ~/.aws
     - cp "$GITLAB_SECRET_FILE_AWS_CREDENTIALS" ~/.aws/credentials
-    - python -m pytest -r s src/toil/test/batchSystems/batchSystemTest.py
-    - python -m pytest -r s src/toil/test/mesos/MesosDataStructuresTest.py
+    - python -m pytest -v -r s src/toil/test/batchSystems/batchSystemTest.py
+    - python -m pytest -v -r s src/toil/test/mesos/MesosDataStructuresTest.py
 
 py2_cwl:
   stage: test
   script:
     - pwd
     - virtualenv -p python2.7 venv && . venv/bin/activate && make prepare && make develop extras=[all] && pip install htcondor
-    - python -m pytest src/toil/test/cwl/cwlTest.py
+    - python -m pytest -v src/toil/test/cwl/cwlTest.py
 
 py2_wdl:
   stage: test
   script:
     - pwd
     - virtualenv -p python2.7 venv && . venv/bin/activate && make prepare && make develop extras=[all] && pip install htcondor
-    - python -m pytest src/toil/test/wdl/toilwdlTest.py
+    - python -m pytest -v src/toil/test/wdl/toilwdlTest.py
 
 py2_jobstore_and_provisioning:
   stage: test
   script:
     - pwd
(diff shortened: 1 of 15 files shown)


adamnovak:
OK, the CI for this is running on Kubernetes and appears to be at least vaguely working. Other CI runs might not work until this is merged in or we swap the runners back.

adamnovak:
OK, the CI doesn't run through on Kubernetes because Gitlab can't set a disk space limit (https://gitlab.com/gitlab-org/gitlab-runner/issues/6184) and it seems like our cluster is assigning a default of 500 Mi, which is too small. I'm talking to the admins to get it raised to something that will let the jobs run, while we wait for Gitlab to get support to actually specify it.

adamnovak:
OK, here's a real failure on the Kubernetes CI that doesn't happen on the VM-based CI, in `py3_main`:

```
_________ SingleMachinePromisedRequirementsTest.testConcurrencyDynamic _________
  File "/builds/databiosphere/toil/src/toil/test/src/promisedRequirementTest.py", line 58, in testConcurrencyDynamic
    self.assertEqual(maxValue, old_div(self.cpuCount, coresPerJob))
  File "/usr/lib/python3.6/unittest/case.py", line 829, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.6/unittest/case.py", line 822, in _baseAssertEqual
AssertionError: 12 != 96
```

We might be having trouble because Kubernetes is limiting our CPU usage (to 4 cores, I think), but the underlying machine has 96 cores. Also, I don't think Kubernetes' 4000 milli-CPU limit *actually* says anything about how many of our threads can be executing at any given time; I think it just limits our overall CPU usage to 400% over some timeframe. So that test might need to be adjusted to pick up the core count of the container instead of the host, and also maybe to tolerate more parallelism than expected given the core count.

adamnovak:
OK I have a fix for the CPU count thing, but I also need to re-0work the way we send secrets to CI, since the Kubernetes Gitlab runners don't have an AMI role to use to download secrets from AWS's secrets manager.

I've loaded all the secrets into Gitlab as protected variables instead. I'm going to let the pipeline run through on the old runners to make sure that that works.

The other problem I've run into is that our Kubernetes cluster is full, and like pretty much everything else, Gitlab doesn't handle that well. It always picks up jobs thinking it can run them, and times out in a minute or two if it can't make a pod, rather than letting the Gitlab jobs sit as waiting for a runner.

I'll have to talk to Erich about whether/how we can ensure that Gitlab doesn't fail our CI jobs when our Kubernetes cluster is busy.

adamnovak:
OK, I think I have to not make the Gitlab variables "protected", because then they are only visible to protected branches, not to random issue branches.

Hopefully having the secrets live in files, and Gitlab's natural reluctance to show you variable-substituted shell commands, will keep them out of our logs.

PR #2911, opened 2020-01-09
Issues/2897 atomic jobstore

### LINKED ISSUE #2897: clarify whether JobStore.updateFile is atomic or not
Code in Cactus currently assumes that JobStore.updateFile is atomic, but this appears not to always be the case in practice, since the fileJobStore implementation just uses Python's `shutil.copyfile` (with no further interlocks).

It would be useful to clarify the intended behavior.

Here's the related Cactus code: https://github.com/ComparativeGenomicsToolkit/cactus/blob/master/src/cactus/pipeline/cactus_workflow.py#L301-L309

See also https://github.com/ComparativeGenomicsToolkit/cactus/issues/114

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-474)
┆Issue Number: TOIL-474

diekhans:
Use atomic file copy when writing to posix file system.
Fixes #2897

### DIFF
--- a/contrib/admin/test-pr
+++ b/contrib/admin/test-pr
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# test-pr: submit a PR from an external contributor for Gitlab testing
+
+set -e
+
+# Check arguments
+if [[ "${#}" -ne 3 ]] ; then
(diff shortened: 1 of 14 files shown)


diekhans:
closing this in favor of a new branch because of the merge getting confused

Issue #2910, opened 2020-01-08
"database is locked" error when using the new SQLite caching db


arostamianfar:
The new SQLite-based caching keeps failing when there are several tasks running on the same worker (stack trace below). Looks like SQLite can't handle many concurrent calls (see https://stackoverflow.com/questions/26862809/operational-error-database-is-locked/26864360 ).

@adamnovak not sure if there is an easy way to limit concurrency as workers are on different processes. Perhaps the easiest thing could be to retry with a random exponential backoff and also increase the timeout as suggested in the thread? 

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/cwltool/executors.py", line 168, in run_jobs
    for job in jobiter:
  File "/usr/local/lib/python2.7/dist-packages/cwltool/command_line_tool.py", line 455, in job
    reffiles, builder.stagedir, runtimeContext, True)
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 274, in make_path_mapper
    runtimeContext.toil_get_file)
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 228, in __init__
    referenced_files, basedir, stagedir, separateDirs=separateDirs)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/pathmapper.py", line 262, in __init__
    self.setup(dedup(referenced_files), basedir)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/pathmapper.py", line 319, in setup
    self.visit(fob, stagedir, basedir, copy=fob.get("writable"), staged=True)
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 262, in visit
    stagedir, basedir, copy=copy, staged=staged)
  File "/usr/local/lib/python2.7/dist-packages/cwltool/pathmapper.py", line 267, in visitlisting
    self.visit(ld, stagedir, basedir, copy=ld.get("writable", copy), staged=staged)
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 255, in visit
    resolved = self.get_file(loc) if self.get_file else loc
  File "/usr/local/lib/python2.7/dist-packages/toil/cwl/cwltoil.py", line 309, in toil_get_file
    src_path = file_store.readGlobalFile(FileID.unpack(file_store_id[7:]))
[...]

PR #2914, opened 2020-01-09
Make sure to commit when giving away files

adamnovak:
This might help with #2910.


adamnovak:
Our Docker builds seem unable to reliably pull one particular base image today. I have code in #2903 that should fix that problem by pre-pulling with retries.

adamnovak:
Looks like the Kubernetes batch system tests can sometimes fail because they kill a job and it comes back with a 0 wall clock time:

```
____________________ KubernetesBatchSystemTest.testRunJobs _____________________
  File "/builds/databiosphere/toil/src/toil/test/batchSystems/batchSystemTest.py", line 188, in testRunJobs
    self.assertTrue(wallTime > 0)
  File "/usr/lib/python2.7/unittest/case.py", line 422, in assertTrue
AssertionError: False is not true
```

PR #2916, opened 2020-01-10
Actually call the implemented size polling method

### LINKED ISSUE #2915: We call jobStore.getGlobalFileSize when we mean jobStore.getFileSize
https://github.com/DataBiosphere/toil/blob/7471248edc8569e37ec3d4c6161c65e9330b54e9/src/toil/fileStores/abstractFileStore.py#L306

The actual implemented method is `getFileSize`:

https://github.com/DataBiosphere/toil/blob/7471248edc8569e37ec3d4c6161c65e9330b54e9/src/toil/jobStores/abstractJobStore.py#L912

I don't think Cactus will work with the caching file store the way it is right now.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-482)
┆Issue Number: TOIL-482

adamnovak:
This will fix #2915.

### DIFF
--- a/src/toil/fileStores/abstractFileStore.py
+++ b/src/toil/fileStores/abstractFileStore.py
@@ -303,7 +303,7 @@ class AbstractFileStore(with_metaclass(ABCMeta, object)):
         if size is None:
             # It fell off
             # Someone is mixing FileStore and JobStore file APIs, or serializing FileIDs as strings.
-            size = self.jobStore.getGlobalFileSize(fileStoreID)
+            size = self.jobStore.getFileSize(fileStoreID)
 
         return size
 


mr-c:
```
=================================== FAILURES ===================================
____________________ KubernetesBatchSystemTest.testRunJobs _____________________
  File "/builds/databiosphere/toil/src/toil/test/batchSystems/batchSystemTest.py", line 188, in testRunJobs
    self.assertTrue(wallTime > 0)
  File "/usr/lib/python2.7/unittest/case.py", line 422, in assertTrue
AssertionError: False is not true
```

adamnovak:
This is the same problem as https://github.com/DataBiosphere/toil/pull/2914#issuecomment-573237243 where the Kubernetes batch system tests intermittently fail. I've reported this as https://github.com/DataBiosphere/toil/issues/2919 and I don't think it is a problem with this PR. I will mash rerun.

PR #2931, opened 2020-01-22
Use an http-parser for pymesos that builds on 3.7+

### LINKED ISSUE #2924: Toil fails on python 3.7 due to dependency
Toil has an indirect dependency on  http_parser package which is no longer maintained on pypi (last update 2013).  This package does not compile with python 3.7.  It is required by pymesos

However, it does not appear that pymesos has released a version that corrects this.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-489)
┆Issue Number: TOIL-489

adamnovak:
This should fix #2924 (as long as your pip is new enough to respect [PEP 508](https://www.python.org/dev/peps/pep-0508/)).
If someone with permissions forked http-parser into databiosphere, or if https://github.com/benoitc/http-parser/pull/88 were to be merged, we could use a better place than my fork of `http-parser` as the source for this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,11 @@ def runSetup():
     futures = 'futures==3.1.1'
     pycryptodome = 'pycryptodome==3.5.1'
     pymesos = 'pymesos==0.3.7'
+    # We need this specific http-parser that still claims to be version 0.8.3
+    # but which builds on Python 3.7+, to satisfy pymesos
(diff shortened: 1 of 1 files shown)


diekhans:
I think we should leave this open, as this workaround is less than ideal.  I am waiting to see if anyone in pymesos cares about the ticket that was filed.

PR #2933, opened 2020-01-22
Fix cpu_count() to actually read quotas

adamnovak:
This will fix #2932 by making sure to actually read the quota files when trying to get CPU counts.

That should fix both the total cores it is expecting, and the number of jobs it actually attempts to run in parallel.

PR #2936, opened 2020-01-22
Fix Intermittent Failures

jeffrey856:
Increased numb of retries and the intervals between fails.


adamnovak:
Sorry @jeffrey856, one of your test jobs got picked up by the Kubernetes runner that I turned on to test #2933. I told it to rerun on the old, definitely-working runner, so hopefully it will pass now.

Issue #2742, opened 2019-07-18
Upgrade Toil appliance to Python 3


adamnovak:
We want the Toil appliance Docker image to be on Python 3 *before* Python 2 goes out of support.

This involves modifying its build process, and also fixing bugs in Toil's AWS support on Python 3, which most users haven't encountered before because nobody uses the AWS job store and batch system and provisioner without the (Python 2) appliance images.

This will probably involve starting from https://github.com/DataBiosphere/toil/pull/2737 and will fix https://github.com/DataBiosphere/toil/issues/2736.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-399)
┆Issue Number: TOIL-399

arostamianfar:
Curious if there are any updates/ETA on this?

DailyDreaming:
@arostamianfar Not yet.  Some on setting up the testing infrastructure, but this should be next on the priority queue after that.

adamnovak:
I'm a little concerned that we might have to keep the Python 2 version around at least until we properly end Python 2 support, because we need to support Cactus/Kubernetes, and AFAIK Cactus is still all Python 2 with no immediate plan in place to port it. We want at least one Toil release that Cactus can run on on Kubernetes before January.

PR #2937, opened 2020-01-23
Make an appliance for Python 3

adamnovak:
This is a start on #2742. @DailyDreaming I know it is assigned to you, but I was waiting around for tests to run and I figured I would see if it is actually easy.

### DIFF
--- a/version_template.py
+++ b/version_template.py
@@ -22,9 +22,12 @@ import the expand_ function and invoke it directly with either no or exactly one
 
 # Note to maintainers:
 #
-#  - don't import at module level unless you want the imported value to be included in the output
-#  - only import from the Python standard run-time library (you can't have any dependencies)
-
+#  - don't import at module level unless you want the imported value to be
+#    included in the output
+#  - only import from the Python standard run-time library (you can't have any
+#    dependencies)
+#  - don't import even standard modules at global scope without renaming them
+#    to have leading/trailing underscores
 
 baseVersion = '3.22.0a1'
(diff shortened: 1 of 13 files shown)


adamnovak:
OK, looks like it is not easy. But maybe this is a starting point? It builds a Python 3 appliance when I `make docker`.

adamnovak:
OK, I think this is only part of #2937, because we still need to actually test the appliance, but it looks like this successfully builds the appliance for Python 3 while calling pip3 to install things, and pushes it to the normal Docker tag, while the Python 2 appliance gets "-py2" appended.

I'm not so sure I like the way I've stuffed things like "what Python are we" into the version file, and I suspect that with multiple minor versions of Python 3 in the world we might want one appliance Docker for each of them, so that might be what review of this should think about.

version_template.py
```diff
@@ -61,23 +64,67 @@ def distVersion():
...
+    workflows to work with a variety of leader Python versions.
+    """
+
+    import sys
+    
+    if sys.version_info[0] == 3:
+        # Ignore minor version
+        return 'python3'
+    else:
+        return exactPython()
```
DailyDreaming: Shouldn't we drop py2.7?

adamnovak:
I want to make sure Python 3 is working for everything in master
before we rip out anything we need for Python 2.7 to also work.
Including having manually run a workflow, probably on a Mesos
autoscaling cluster, using the appliance.

PR #2945, opened 2020-01-30
Include jobNode.jobName in submitted slurm job name

dleehr:
Updates SlurmBatchSystem to include a more descriptive slurm job name, particularly handy for monitoring workflows on slurm clusters

Before: `toil_job_4`
After: `toil_job_4_tools/picard/picard-BedToIntervalList.cwl`


mr-c:
@dleehr This is great, thanks! Can you extend this to be for all the `abstractGridEngineBatchSystem`s and not just slurm? That would probably make the code cleaner and more useful.

adamnovak:
Can we guarantee that these names with slashes in them are allowed for all the `abstractGridEngineBatchSystem`s?

dleehr:
Thanks for the feedback. I agree that AbstractGridEngineBatchSystem is the place to add this data, even if the other batch systems don't consume it yet.

adamnovak:
I merged #2946; this code is now in master.

PR #2946, opened 2020-01-30
Slurm Job Names

adamnovak:
This is #2945 with master merged in so it will actually pass CI tests.

### DIFF
--- a/src/toil/batchSystems/abstractGridEngineBatchSystem.py
+++ b/src/toil/batchSystems/abstractGridEngineBatchSystem.py
@@ -130,10 +130,10 @@ class AbstractGridEngineBatchSystem(BatchSystemLocalSupport):
             while len(self.waitingJobs) > 0 and \
                     len(self.runningJobs) < int(self.boss.config.maxLocalJobs):
                 activity = True
-                jobID, cpu, memory, command = self.waitingJobs.pop(0)
+                jobID, cpu, memory, command, jobName = self.waitingJobs.pop(0)
 
                 # prepare job submission command
-                subLine = self.prepareSubmission(cpu, memory, jobID, command)
+                subLine = self.prepareSubmission(cpu, memory, jobID, command, jobName)
                 logger.debug("Running %r", subLine)
                 batchJobID = with_retries(self.submitJob, subLine)
                 logger.debug("Submitted job %s", str(batchJobID))
@@ -237,7 +237,7 @@ class AbstractGridEngineBatchSystem(BatchSystemLocalSupport):
                     logger.debug('No activity, sleeping for %is', self.boss.sleepSeconds())
 
         @abstractmethod
-        def prepareSubmission(self, cpu, memory, jobID, command):
+        def prepareSubmission(self, cpu, memory, jobID, command, jobName):
             """
             Preparation in putting together a command-line string
             for submitting to batch system (via submitJob().)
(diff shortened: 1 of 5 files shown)


adamnovak:
Here's an error I haven't seen before:

```
___________________________ SortTest.testFileSingle ____________________________
  File "/builds/databiosphere/toil/src/toil/test/sort/sortTest.py", line 211, in testFileSingle
    self._toilSort(jobStoreLocator=self._getTestJobStorePath(), batchSystem='singleMachine')
  File "/builds/databiosphere/toil/src/toil/test/sort/sortTest.py", line 161, in _toilSort
    with runMain(options):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/builds/databiosphere/toil/src/toil/test/sort/sortTest.py", line 56, in runMain
    main(options)
  File "/builds/databiosphere/toil/src/toil/test/sort/sort.py", line 254, in main
    sortedFileID = workflow.restart()
  File "/builds/databiosphere/toil/src/toil/common.py", line 834, in restart
    self._shutdownBatchSystem()
  File "/builds/databiosphere/toil/src/toil/common.py", line 1080, in _shutdownBatchSystem
    self._batchSystem.shutdown()
  File "/builds/databiosphere/toil/src/toil/batchSystems/singleMachine.py", line 289, in shutdown
    BatchSystemSupport.workerCleanup(self.workerCleanupInfo)
  File "/builds/databiosphere/toil/src/toil/batchSystems/abstractBatchSystem.py", line 320, in workerCleanup
    AbstractFileStore.shutdownFileStore(workflowDir, info.workflowID)
  File "/builds/databiosphere/toil/src/toil/fileStores/abstractFileStore.py", line 146, in shutdownFileStore
    CachingFileStore.shutdown(cacheDir)
  File "/builds/databiosphere/toil/src/toil/fileStores/cachingFileStore.py", line 1705, in shutdown
    cls._ensureTables(con)
  File "/builds/databiosphere/toil/src/toil/fileStores/cachingFileStore.py", line 281, in _ensureTables
    """)
DatabaseError: database disk image is malformed
```

I think this problem is caused by this code, where during batch system shutdown we open the database via a hard link that is not the same filename as the one that we use when actually using it for caching, because that name is supposed to depend on the workflow attempt number:

https://github.com/DataBiosphere/toil/blob/422a73cc32fcbcde880eab6c41a97550ac473119/src/toil/fileStores/cachingFileStore.py#L1687-L1691

According to [the docs](https://sqlite.org/tempfiles.html) and also [StackOverflow](https://stackoverflow.com/q/12906867):

> A rollback journal is a temporary file used to implement atomic commit and rollback capabilities in SQLite. (For a detailed discussion of how this works, see the separate document titled Atomic Commit In SQLite.) The rollback journal is always located in the same directory as the database file and has the same name as the database file except with the 8 characters "-journal" appended.

This failure is happening during the bad worker test, where our workers signal themselves and die randomly, to suss out problems like this. I think a journal file is getting left behind by a writer, and then not being replayed by the reader during shutdown, because it is accessing the database through a different filename. So it sees the database as corrupted and dies.

I think we need to rip out the symlink hack and find a better way to locate the most recent database when shutting down.

I'm going to rerun the offending test because I don't think it has anything to do with this PR.

adamnovak:
OK, this passed, and I have what should be a fix for the intermittent failure in another PR. I will merge this, as it has already been reviewed in #2945

PR #2948, opened 2020-01-31
Scan for the latest caching database instead of using a hard link

### LINKED ISSUE #2947: We can lose SQLite journals and see a corrupted database when shutting down the caching file store, because we access the database through a different hard link
SQLite cares about filenames, and it uses them to connect databases to the journal files that allow it to ensure atomicity for writes.

When we access the caching database through a hard link, we don't necessarily see the journal file, and we are able to see partially-completed writes (which look like database corruption, or worse). See https://github.com/DataBiosphere/toil/pull/2946#issuecomment-580530181

We need to change how shutdown works so it doesn't rely on a hard link to find the database.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-497)
┆Issue Number: TOIL-497

adamnovak:
This should fix #2947, assuming that workflow attempt numbers are increasing integers, and nobody is trying to race us to clean up the workflow when there are old databases from previous attempts, or dropping databases from new attempts in while we are cleaning up.

### DIFF
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1676,7 +1676,7 @@ class CachingFileStore(AbstractFileStore):
                in the database.
         """
 
-        if os.path.exists(dir_) and os.path.isdir(dir_):
+        if os.path.isdir(dir_):
             # There is a caching directory to clean up
        
        
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -32,6 +32,7 @@ import errno
 import hashlib
 import logging
 import os
+import re
 import shutil
 import sqlite3
 import stat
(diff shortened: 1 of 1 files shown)


src/toil/fileStores/cachingFileStore.py
```diff
@@ -1684,33 +1676,65 @@ def shutdown(cls, dir_):
...
-            except:
-                # Probably someone deleted it.
-                pass
-            else:
-                # We got a database connection
-
-                # Create the tables if they don't exist so deletion of dead
-                # jobs won't fail.
-                cls._ensureTables(con)
+        if os.path.exists(dir_) and os.path.isdir(dir_):
```
DailyDreaming: `os.path.isdir(dir_)` alone should be sufficient, as it will return `False` if it doesn't exist.

PR #2955, opened 2020-02-05
Fix database being locked when trying to write in the CachingFileStore

### LINKED ISSUE #2910: "database is locked" error when using the new SQLite caching db

adamnovak:
This should fix #2910, but harder.

I've wrapped all the writes to the database to go through a single function, which uses Toil's retry system to retry forever, with increasing backoff, if it tries to write and the database is locked. It also takes care of committing.

I've also inserted some sleeps into the loops that poll the database with conditional writes. This should reduce the portion of the time that the database is locked, overall.

I've also added a test that was able to reproduce the database is locked error with the sqlite timeout set to 5 seconds. I'm bumping it back up to 60, so I'm not sure how good the test will be anymore at checking the logic that deals with timeouts from sqlite, but the higher timeout should provide better protection from timeouts on the read side (which we haven't encountered yet).

If something really does hold a database lock forever, I would expect this new code to deadlock, or to spit out a timeout on a read operation. But it should now be impossible for a timeout on a write operation to fail a job, because of the indefinite retry. Plus, I have checked for database-modifying statements that don't go through the new system, and there are none, so it should be impossible for us to forget to commit now.

### DIFF
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -221,7 +221,7 @@ class CachingFileStore(AbstractFileStore):
         # Connect to the cache database in there, or create it if not present
         self.dbPath = os.path.join(self.localCacheDir, 'cache-{}.db'.format(self.workflowAttemptNumber))
         # We need to hold onto both a connection (to commit) and a cursor (to actually use the database)
-        self.con = sqlite3.connect(self.dbPath, timeout=60)
+        self.con = sqlite3.connect(self.dbPath, timeout=5.0)
         self.cur = self.con.cursor()
 
         # Note that sqlite3 automatically starts a transaction when we go to
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -308,7 +308,7 @@ class CachingFileStore(AbstractFileStore):
                             args = item[1]
                         # Do it
                         cur.execute(command, args)
-                except e:
+                except Exception as e:
                     logging.error('Error talking to caching database: %s', str(e))
 
                     # Try to make sure we don't somehow leave anything part-done if a
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1028,10 +1028,7 @@ class CachingFileStore(AbstractFileStore):
         return FileID.forPath(fileID, absLocalFileName)
 
     def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=False, symlink=False):
-        if not isinstance(fileStoreID, FileID):
-            # Don't let the user forge File IDs.
-            raise TypeError('Received file ID not of type FileID: {}'.format(fileStoreID))
-
+        
         if str(fileStoreID) in self.filesToDelete:
             # File has already been deleted
             raise FileNotFoundError('Attempted to read deleted file: {}'.format(fileStoreID))
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -185,11 +185,15 @@ class CachingFileStore(AbstractFileStore):
 
     """
 
-    def __init__(self, jobStore, jobGraph, localTempDir, waitForPreviousCommit, forceNonFreeCaching=False):
+    def __init__(self, jobStore, jobGraph, localTempDir, waitForPreviousCommit):
         super(CachingFileStore, self).__init__(jobStore, jobGraph, localTempDir, waitForPreviousCommit)
 
         # For testing, we have the ability to force caching to be non-free, by never linking from the file store
-        self.forceNonFreeCaching = forceNonFreeCaching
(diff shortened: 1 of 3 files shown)


adamnovak:
@arostamianfar Could you maybe try testing this out, if it passes CI? This shouldn't be able to produce the errors you have shown me, although if it then starts producing timeouts on the read side, that would need to be addressed.

src/toil/fileStores/cachingFileStore.py
```diff
@@ -259,6 +270,84 @@ def __init__(self, jobStore, jobGraph, localTempDir, waitForPreviousCommit, forc
...
+                            item = (item,)
+                        # Parse out the command and the variables to substitute
+                        command = item[0]
+                        if len(item) < 2:
+                            args = ()
+                        else:
+                            args = item[1]
+                        # Do it
+                        cur.execute(command, args)
+                except e:
```
arostamianfar: Looks like this needs a small fix:
```
  File "/usr/local/lib/python2.7/dist-packages/toil/fileStores/cachingFileStore.py", line 341, in _write
    return self._staticWrite(self.con, self.cur, operations)
  File "/usr/local/lib/python2.7/dist-packages/toil/fileStores/cachingFileStore.py", line 303, in _staticWrite
    except e:
NameError: global name 'e' is not defined
```

arostamianfar:
Thanks, @adamnovak ! I patched this PR together with the small fix on the exception, and it worked!

Issue #2855, opened 2019-11-13
toil-cwl-runner on slurm fails to run jobs on multiple nodes


johnbradley:
I have been trying to run a CWL workflow with toil and Slurm.
I do so by running an sbatch script containing `toil-cwl-runner`.
When the job runs entirely on one Slurm node it succeeds.
If slurm chooses to run `_toil_worker` on a different node than the node running `toil.leader` the `_toil_worker` job fails and the following is included in `toil-cwl-runner` output:
```
MainThread WARNING toil.leader: No log file is present, despite job failing...
```
If you look at the failed job with `sacct` the `State` is `FAILED` and `ExitCode` is `1:0`.
I have not been able to find any output logs on the nodes attempting to run `_toil_worker`.
I have been using toil from the master branch.

NOTE: If you specify `--workDir` with a path shared across all nodes this problem does not occur.

---

I added some logging and found that the `sbatch` command issued by `toil.leader` is attempting to write stderr and stdout to files within a directory in /tmp that does not exist on the node trying to run `_toil_worker`:

```
sbatch 
...
-o /tmp/toil-882d58aa-9bc5-4faa-abf5-4906fc65e8af-8c697123-fa43-4d9b-9948-405a15a8957d/toil_job_0_batch_slurm_%j_std_output.log
 -e /tmp/toil-882d58aa-9bc5-4faa-abf5-4906fc65e8af-8c697123-fa43-4d9b-9948-405a15a8957d/toil_job_0_batch_slurm_%j_std_error.log
--wrap=_toil_worker...
```
This `/tmp/toil-<uuid>/` directory does get created on the `toil.leader` node.

I think the following code specifies the file paths for the `-o` and `-e` arguments:
https://github.com/DataBiosphere/toil/blob/81a543d0e1a8e6c299f22bb8e862d34097e1f0bc/src/toil/batchSystems/slurm.py#L190-L191

[sbatch documentation](https://slurm.schedmd.com/sbatch.html)

## Reproducing
This requires a slurm cluster and toil installed from the master branch (to get another bug fix).
Identify two nodes you wish run `toil.leader` and `_toil_worker` on which I will refer to as `c1-leader` and `c1-worker` from now on.

Create a sbatch script `toil.sbatch` with the following contents:
```
#!/usr/bin/env bash

source env/bin/activate
export TOIL_SLURM_ARGS="-w c1-worker"

time toil-cwl-runner \
  --jobStore ${HOME}/jobstore \
[...]

PR #2959, opened 2020-02-07
Attempt to make batch system more robust and debuggable

diekhans:
**WARNING: this has not been tests I have no access to test batch systems.  People who use them will have to test them before merging into master.**

Log results from HPC batch system commands before parsing.

Check status of calls to batch system commands.

Add popen_communicate that does both subprocess Popen and communicate and
checks for errors.  Also includes stderr in exception that is raised.

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -123,7 +123,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                 logger.debug("sacct exit code is %s, returning status %d", exitcode, status)
                 return (state, status)
             logger.debug("Did not find exit code for job in sacct output")
-            return None
+            return (None, None)
 
         def _getJobDetailsFromScontrol(self, slurmJobID):
             args = ['scontrol',
@@ -136,7 +136,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
 
             if rc != 0:
                 # no accounting system or some other error
-                return (None, -999)
+                return (None, None)
 
             job = dict()
             for line in process.stdout:
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -19,9 +19,9 @@ from past.utils import old_div
 import logging
 import os
 from pipes import quote
-import subprocess
 import time
 import math
+from toil.lib.misc import call_command, CalledProcessErrorStderr
 
 # Python 3 compatibility imports
 from six.moves.queue import Empty, Queue
@@ -46,9 +46,8 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
             # -h for no header
             # --format to get jobid i, state %t and time days-hours:minutes:seconds
 
-            lines = subprocess.check_output(['squeue', '-h', '--format', '%i %t %M']).decode('utf-8').split('\n')
+            lines = call_command(['squeue', '-h', '--format', '%i %t %M']).split('\n')
             for line in lines:
-                logger.debug("squeue output %s", line)
                 values = line.split()
                 if len(values) < 3:
                     continue
@@ -60,15 +59,14 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
             return times
 
         def killJob(self, jobID):
-            subprocess.check_call(['scancel', self.getBatchSystemID(jobID)])
+            call_command(['scancel', self.getBatchSystemID(jobID)])
 
         def prepareSubmission(self, cpu, memory, jobID, command, jobName):
             return self.prepareSbatch(cpu, memory, jobID, jobName) + ['--wrap={}'.format(command)]
 
         def submitJob(self, subLine):
             try:
-                output = subprocess.check_output(subLine, stderr=subprocess.STDOUT).decode('utf-8')
-                logger.debug("%s output %s", subLine[0], output)
+                output = call_command(subLine)
                 # sbatch prints a line like 'Submitted batch job 2954103'
                 result = int(output.strip().split()[-1])
                 logger.debug("sbatch submitted job %d", result)
@@ -80,9 +78,10 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
         def getJobExitCode(self, slurmJobID):
             logger.debug("Getting exit code for slurm job %d", int(slurmJobID))
 
-            state, rc = self._getJobDetailsFromSacct(slurmJobID)
-
-            if rc == -999:
+            try:
+                state, rc = self._getJobDetailsFromSacct(slurmJobID)
+            except CalledProcessErrorStderr:
+                # no accounting system or some other error
                 state, rc = self._getJobDetailsFromScontrol(slurmJobID)
 
             logger.debug("s job state is %s", state)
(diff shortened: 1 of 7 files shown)


johnbradley:
Testing a simple CWL workflow using slurm toil never finished. This job is running a simple CWL tar workflow created by adding the `--noStdOutErr` flag to the with instructions here https://github.com/DataBiosphere/toil/issues/2855#issue-522401082.
In slurm (sacct) I can see that the workflow command (tar) had finished successfully but the logs kept print out a lines like the following again and again:
```
... 2020-02-10 09:29:20,787 MainThread DEBUG toil.batchSystems.slurm: squeue output 10245772_[0%20] PD 0:00
```
I think toil kept trying (and failing) to determine that the job was completed.

src/toil/batchSystems/slurm.py
```diff
@@ -127,26 +130,32 @@ def _getJobDetailsFromScontrol(self, slurmJobID):
...
             for line in process.stdout:
+                logger.debug("%s output %s", args[0], line)
                 values = line.decode('utf-8').strip().split()
-    
+
                 # If job information is not available an error is issued:
                 # slurm_load_jobs error: Invalid job id specified
                 # There is no job information, so exit.
                 if len(values)>0 and values[0] == 'slurm_load_jobs':
                     return (None, None)
```
johnbradley: I think the early exit `return (None, -999)` might be related to [an issue where the toil never finishes](https://github.com/DataBiosphere/toil/pull/2959#issuecomment-584207037) on this branch.
The comment at the bottom seems to say that it should `return (None, None)`.

For completed jobs slurm responds with the following for scontrol:
```
$ scontrol show job 15671080
slurm_load_jobs error: Invalid job id specified
```

diekhans: thanks, yes only _getJobDetailsFromSacct() should return -999 magic number..

diekhans:
I fixed the incorrect return and added logging for the commands
being executed.  Perhaps give it another try..

johnbradley:
I retested and all the scenarios I tried functioned as expected.
Scenarios tested with `--logDebug`:
- toil-cwl-runner on Slurm using a shared workdir - Succeeded
- toil-cwl-runner on Slurm using default workdir (/tmp on the worker node) and specifying `--noStdOutErr` flag  - Succeeded
- toil-cwl-runner on Slurm using node local storeage and without specifying `--noStdOutErr` flag  - Failed as expected but with better logging `sacct output b'FAILED|1:0\n'`

I used the same CWL workflow as before detailed here: https://github.com/DataBiosphere/toil/issues/2855#issue-522401082

src/toil/lib/misc.py
```diff
@@ -243,3 +254,53 @@ def atomic_copyobj(src_fh, dest_path, length=16384):
...
+            err = self.stderr if isinstance(self.stderr, str) else self.stderr.decode("ascii", errors="replace")
+            return "Command '%s' exit status %d: %s" % (self.cmd, self.returncode, err)
+
+
+def popen_communicate(cmd, input=None, timeout=None, **kwargs):
+    """Start a process with subprocess.Popen, run the process and return
+    (stdout, stderr). Raise an exception if the process exits non-zero. Kwargs
+    are as passed to subprocess.Popen, with input and timeout being passed to
+    communicate.  The PY3.6 Popen encode and error arguments are allowed and
+    implemented.  Generally, Toil should used encoding='latin1'.
```
adamnovak: Shouldn't we be using `utf8` throughout instead of `latin1`? As soon as someone runs Toil against a grid engine scheduler that titles its columns in Chinese or something, we're going to get in trouble with `latin1`.

diekhans: I am not convinced latin1 is right either ;-)

The problem, which I should have documented, is that we don't want to compound the
error reporting problem by having garbage on stderr cause a decoding exception with an invalid UTF8 character.  latin1 can handle all bytes without error.

Probably better would be to use utf-8 and set error to ignore or replace.

I have kind of decide that this is change is both a bit more code than it needs to be and doesn't go far enough.  I am thinking of making a wrapper around the new subprocess.run function that all of toil will use to call process. This way we  would get logging even of all programs executed plus provide the recommended encoding/error as defaults.

subprocess is somewhat limited in that you really want stderr to be strings even if stdin/stdout is binary ..

sigh, why can't ASCII be the universal language

diekhans:
thanks @johnbradley, 

this is interesting:
> * toil-cwl-runner on Slurm using node local storeage and without specifying `--noStdOutErr` flag  - Failed as expected but with better logging `sacct output b'FAILED|1:0\n'`

not clear why this is happening, as the exec of sacct should have overridden the redirecting of stdout/error that happens  when running the command.

src/toil/lib/misc.py
adamnovak: If we want to handle all possible bytes, it would make sense to be passing around bytes objects; we can repr() them for logging. But yeah, ignoring errors in the decoding would solve the same problem too. I just want to avoid a situation where we have to re-encode the library function's return value and decode it as utf-8 because some batch system prints out happy emoji for completed jobs or something.

adamnovak:
> not clear why this is happening, as the exec of sacct should have overridden the redirecting of stdout/error that happens when running the command.

I don't think that is true; the standard output/error redirection for the job itself is accomplished by telling Slurm to redirect the streams when it actually gets around to running the job. I think that in this case, doing that still makes the job fail, and then the `sacct` output describing the failed job isn't parsed by Toil, but it is logged.

diekhans:
Ah, maybe I misinterpreted @johnbradley description as indicating -noStdOutErr was required to trigger the failure, but maybe that isn't the only variable.

johnbradley:
> Ah, maybe I misinterpreted @johnbradley description as indicating -noStdOutErr was required to trigger the failure, but maybe that isn't the only variable.

@diekhans sorry for the confusion. When running toil with the slurm backend and using local storage for workdir the toil worker will likely fail to start. I created issue #2855 for this problem. `sbatch` is the slurm command used in toil to run the toil worker. The directory that should contain the stdout and stderr logs for `sbatch` likely will not exist yet on whatever node the worker is run on. `sbatch` quits immediately if it can't write to it's stdout and stderr logs.

If a user specifies the `--noStdOutErr` flag this failure doesn't happen. For a user to figure that out with the current logging wasn't easy. With the new logging changes you have made here it is easer to see where the failure is happening. The `sacct` command displays the accounting/historical database for slurm jobs. The `sacct` command is working fine and reporting that `sbatch` has previously failed with the slurm exit code of `1:0`.

mr-c:
I can test on LSF. @johnbradley what is your testing regime?

src/toil/lib/misc.py
```diff
@@ -10,11 +13,19 @@
 import socket
 from contextlib import contextmanager
 
+# can't do from 'toil import subprocess' as __init__.py import this module!
+if os.name == 'posix' and six.PY2:
```
mr-c: Aren't we py3 only now?

johnbradley:
> I can test on LSF. @johnbradley what is your testing regime?

@mr-c See my notes above https://github.com/DataBiosphere/toil/pull/2959#issuecomment-584639194

mr-c:
I'm running this branch with the CWL v1.0 conformance tests on LSF, it looks fine to me

adamnovak:
I merged master into this branch; that should fix the missing-pytz problem.

src/toil/lib/misc.py
diekhans: this was done before py3 only; I am going to update and simplify

diekhans:
also fixes #3027

src/toil/lib/misc.py
diekhans: it now reads as utf-8,  however it also sets locale to C, due to Perl locale issues on one system.

diekhans:
I have made several changes to this PR to simplify it and improve logging  GridEngine (UGE) seems to be working well on a system I happen to have access to.

If others could try out that current version in other environments, that would be most helpful.

diekhans:
fixes  #3027

PR #2960, opened 2020-02-07
Source toil environment as a configuration file.

DailyDreaming:
So... the intention is to better centralize setting the environment variables in toil.  This will hopefully make things clearer when first attempting to run toil.  `Make` commands now check that this has been sourced before allowing a user to proceed.

I've made 2 files: one for test-specific environment variables (`environment-dev.sh`), and one for configuring normal toil run environment variables (`environment.sh`).  Running `source environment-dev.sh` for testing will also `source environment.sh`.

I've also cleaned up the `Makefile` a bit so that it has to get its environment variables from the environment files and cleaned up some commands that either no longer exist or no longer make sense.

I also got rid of BUILD_NUMBER and the short and minimal docker tags, which serve no purpose I can ascertain.

I also deleted `jenkins.sh` and `run_tests.py` (which was made to produce xml results for jenkins when running tests).

This also pulls in changes from https://github.com/DataBiosphere/toil/pull/2958 .


DailyDreaming:
@diekhans It surprised me too, but on my system it actually does not set the env vars to empty strings as you'd expect and they remain unset.  I'm not sure why this is, but I can update it so that they are commented out, and perhaps lay it out a bit better.

diekhans:
save below this as barney.sh and do
   bash barney.sh
and you can see that the empty string is set in the environment.


`set -a
TRY_SET="why me"
TRY_EMPTY=

env | fgrep TRY
`

DailyDreaming:
@diekhans I was using an old file that had `set a-` instead of `set -a`.  >.<  I've made the changes if I could get a re-review.

DailyDreaming:
@diekhans Thanks!  Merging away...

DailyDreaming:
@adamnovak This should be ready for a re-review.

I've made changes so that this is just a `common.mk` file after our talk and does not try to be a central source for config, only filling things in when running `make`.  I think that this could/should still be changed to be something we agree on and more in line with how I think it should be managed, but I'll have to think about it some more after your comments and Mark's.

PR #2962, opened 2020-02-07
[easy] Update EC2 instance types.

DailyDreaming:
Just some housekeeping.  I've updated the EC2 instance types that are available and also added a short bit to parse new disk syntax.


DailyDreaming:
I noticed that too, definitely some weirdness on AWS's part.  Some stuff like the `hs1.8xlarge` going from 17 to 16 cores looks like they may have been correcting a typo?  The cores are mainly what's changed, and there's no parsing there, just `i["vcpu"]` from their sheet so I think we'll just have to remember that we're trusting their json.

Will merge once tests pass!

PR #2966, opened 2020-02-10
Specify py3.6 or py3.x instead of just py3.

### LINKED ISSUE #2964: Python 3.6 doesn't work well in a Toil Appliance
When I use `TOIL_APPLIANCE_SELF=quay.io/ucsc_cgl/toil:3.24.1a1-3475cb8e60da9e064296dd96b6460562aebb0356 toil launch-cluster...`

I get something with python3 defaulting to `3.5.2`.  It is also where Toil seems to get installed.  Python3 support in Cactus requires Python 3.6.  This means I can' use (non singleMachine) batch systems with Cactus under Python3

@adamnovak mentions 

> Hm, maybe Toil is installing in 3.5 instead of 3.6. We should change https://github.com/DataBiosphere/toil/blob/3475cb8e60da9e064296dd96b6460562aebb0356/docker/Dockerfile.py#L25 to use something like https://github.com/DataBiosphere/toil/blob/3475cb8e60da9e064296dd96b6460562aebb0356/version_template.py#L67-L75 but to select the major and minor version of pip to use.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-507)
┆Issue Number: TOIL-507

DailyDreaming:
Should fix #2964, but needs to be tested.  The image I pushed is:
```quay.io/ucsc_cgl/toil:3.24.1a1-efeaa75465e9c6dd3df8e4535a810a47b17f0561-dirty-py3.6```

### DIFF
--- a/docker/Dockerfile.py
+++ b/docker/Dockerfile.py
@@ -22,7 +22,7 @@ sdistName = os.environ['_TOIL_SDIST_NAME']
 
 # Make sure to install packages into the pip for the version of Python we are
 # building for.
-pip = 'pip{}'.format(sys.version_info[0])
+pip = 'python{}.{} -m pip'.format(sys.version_info[0], sys.version_info[1])
 
 
 dependencies = ' '.join(['libffi-dev',  # For client side encryption for extras with PyNACL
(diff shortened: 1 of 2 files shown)


DailyDreaming:
@adamnovak I'll create an issue and should be able to add that later today.  Should be a quick fix, but will need to test it.

PR #2975, opened 2020-02-14
Actually fail on rev and not du in CWL restart test.

DailyDreaming:
A minor bug now that I understand it, however this should amend the test to actually work in the way it intended.  Issue is here for more info: https://github.com/DataBiosphere/toil/issues/2974


diekhans:
actually, I suspect pytest is capturing enough info that extra logging is unnecessary.  If you have a specific example where pytest isn't generating enough info, lets look.

DailyDreaming:
@diekhans It's mostly that I see the error and it isn't immediately obvious to me which test the error came from since it prints the test class and not the function it came from.

I've added a `setup_method(self, method)` which seems to be what we want and should be inherited by all toil tests.  Pytest itself doesn't seem to have a good CLI option that I can find for doing this.

It seems that we have similar code to print this info in `setUp()` but it doesn't seem to do what I want.

Issue #2928, opened 2020-01-20
toil leader thread model uses an insane number of threads


diekhans:
Toil's use of threads causes fork failures on very beefy machines.

This needs to scale better :-(

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-493)
┆Issue Number: TOIL-493

diekhans:
I justed DoS-ed myself on hgwdev with toil creating 20K threads

adamnovak:
20k threads is a lot of threads. We are supposed to be making ~10 per
core for the weirdly thread-intensive single machine batch system, and
hgwdev doesn't have 2k cores.

Did you try and run several workflows at once in single machine mode
without limiting the max cores of each? Or is there some source of
extremely numerous threads other than the single machine batch
system's weird way of allocating cores to jobs that is tripping you
up?

diekhans:
I am trying to paralyzed the cactus tests, so I was potentially
running 10 test batches at once.  So that is like wanting to
allocate 6400 threads, which is awful but not 20K horrible.

PR #2979, opened 2020-02-19
Raise default min CPU size of single machine jobs

adamnovak:
This should stop us spawning quite so many threads on single machine, and help with #2928. The cost is that jobs that ask for less than a core will get an entire core to themselves. A real solution without all these placeholder threads would address that, but this change is easy and should be a good stopgap.

### DIFF
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -55,13 +55,14 @@ class SingleMachineBatchSystem(BatchSystemSupport):
 
     numCores = cpu_count()
 
-    minCores = 0.1
     """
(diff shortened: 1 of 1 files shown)


DailyDreaming:
So long as tests pass, this seems alright to try out.

adamnovak:
Nope. It looks like the [testSimultaneousReadsUncachedStream test](https://github.com/DataBiosphere/toil/blob/1da749a756d124f1171d825d26ca9365e7e056cb/src/toil/test/src/fileStoreTest.py#L1093) and also the service tests that use ToyService depend on being able to run multiple jobs per core simultaneously to run in a reasonable length of time. We won't be able to adopt this simple fix after all; we need a real reimplementation of how the single machine batch system thinks about managing cores.

PR #2982, opened 2020-02-20
Use file locks instead of PID polling to see if other processes sharing the cache are alive

### LINKED ISSUE #2926: Caching's PID polling system has never been correct and will not work between Kubernetes pods
As an alternative to #2923, we can get improved efficiency by letting Kubernetes schedule our pods, but keeping Toil's file cache in a place where multiple pods on the same node can access it.

However, our system of polling for PIDs to see if the job that was downloading a file has died cannot possibly work in this case, because Kubernetes containers all have their own PID namespaces, and can't see each other's processes. Moreover, it has never been correct as implemented anyway, because of the possibility of PID re-use by other processes.

What I think we should do is port the whole system over to file locks, which at least have a chance of working between containers (although this needs to be tested). Each process that wants to talk to the cache assigns itself a unique identifire (based on a temp file name, or a UUID or something). That ID is used to identify the process in the caching database, and the process keeps a file by that name open, with an exclusive lock on it. To poll to see if the process is alive, other processes try to take a nonblocking shared lock on its identifying file. If successful, the process is known to be dead (and the identifying file can be cleaned up). If taking the shared lock would block, the process is still alive.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-491)
┆Issue Number: TOIL-491

adamnovak:
This should fix #2926, by making checks in the fileStore that used to use polling of other processes by PID instead poll based on ability to lock files. This at least stands a chance of working across container boundaries.


adamnovak:
@arostamianfar You have big workflows that heavily exercise the caching corner cases; any chance you could run one of them with this before we merge it?

arostamianfar:
Will do by EOD and update here! Thanks for letting me know about this change!

diekhans:
It would be good to include an loud notice in the doc to say the
Toil only supports POSIX file systems with file locking.  Some
people don't setup lockd for NFS.

diekhans:
never mind on the lockd warning, this uses fcntl , which is only local to a given system and doesn't work across hosts.

diekhans:
This locking is only on between processes on a single host, correct?
Have you RTFM-ed that fcntl locks work between containers

I have no concerns about file lock and lots of jobs.

DailyDreaming:
@diekhans Toil wouldn't work if locks didn't work between containers.  Have you RTFM-ed that: `On at least some systems, LOCK_EX can only be used if the file descriptor refers to a file opened for writing.` and that **ulimit** sets a limit on open file descriptors?

I believe on my computer this is:

```
quokka@qcore:~$ cat /proc/sys/fs/file-max
6558120
```

It seems like a difficult limit to reach, but I'm unsure if:

A. cactus would ever lock this many files at once
B. we're operating cactus on a system where this might open file descriptors to acquire locks and potentially hit this limit

diekhans:
I had not been able to find clear doc that states if fcntl locks work
on imported volumes, just kind of rambling discussions like:

  https://stackoverflow.com/questions/55361769/docker-and-file-locking
  https://forums.docker.com/t/docker-volume-file-permission-and-locking/24335/3


The default limit on centos is "unlimited" ..

Cactus requires a reasonable high-end machines.  It is reason to
require the limit to be increased if it is too low.  Not worried here.

adamnovak:
These locks aren't per-job but per-worker-process, and any given worker should only need to have two at a time (if it is only running a single Toil workflow): one for the worker's own name to signal it is alive, and one for polling other workers).

I believe that I have the file descriptor read/write modes set up correctly (i.e. I believe that `mkstemp` opens for writing) so that `LOCK_EX` will work where it is used.

I also haven't been able to find anything about whether file locking works across container boundaries, but from what I know of the implementation I think that it should. It's the same kernel making the same file at the same inode visible to two or more processes; those processes just happen to be confined by cgroups in terms of what they can do and see. At any rate, it can't be worse than the current implementation, which definitely doesn't work across containers, and which isn't strictly correct even without containers due to the possibility of PID re-use.

It does require that fcntl locking work within the current machine on the mount point used for the Toil workdir. If there's some way to set up NFS so you just can't lock at all, even locally, that would be a problem if you then put the workdir on NFS.

DailyDreaming:
@diekhans Those are "file lock" limits.  It is also unlimited on my machine:

```
quokka@qcore:~$ ulimit -n
1024
quokka@qcore:~$ ulimit -x
unlimited
```

Please search for what an "open file descriptor" is and how it differs from a file lock.

DailyDreaming:
@adamnovak That sounds reasonable.  I've already approved.  Merge at your leisure.

Issue #2917, opened 2020-01-11
on restart, avoid long sleeps while checking job store


michaelkarlcoleman:
Not sure whether this is a cactus or toil issue, but on a --restart, it appears that the jobStore is being checked.  For each missing job file, a 30s sleep loop occurs.  The point seems to be to wait for a job file that might appear, but on a full restart, this seems pointless.  For large jobs, running through these can take hours.  Perhaps this could be avoided for the startup case?

(Also, that '#' character seems spurious.)

WARNING:toil.jobStores.fileJobStore:Job file `/gpfs/projects/hpcrcf/mkc/cactus-20191116_84way/jobStore/tmp/n/4/jobb23bLC/job` for job `n/4/jobb23bLC` does not exist (yet). We will try #29 more times with 1s intervals.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-483)
┆Issue Number: TOIL-483

adamnovak:
This should be fixed by #2990 which I just merged.

PR #2990, opened 2020-02-26
Eliminate sleep time on --restart

jeffrey856:
This should pass the stage of needing to check for jobstore on `--restart`.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -309,9 +309,13 @@ class FileJobStore(AbstractJobStore):
         :param str url: A path as a string of the file to be read from.
         :param object writable: An open file object to write to.
         """
+        
         # we use a ~10Mb buffer to improve speed
         with open(cls._extractPathFromUrl(url), 'rb') as readable:
             shutil.copyfileobj(readable, writable, length=cls.BUFFER_SIZE)
+            # Return the number of bytes we read when we reached EOF.
+            return readable.tell()
+        
 
     @classmethod
     def _writeToUrl(cls, readable, url):
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -643,7 +643,6 @@ class FileJobStore(AbstractJobStore):
         if not self.restart:
             if not self.waitForExists(jobStoreID,30):
                 raise NoSuchJobException(jobStoreID)
-        raise NoSuchJobException(jobStoreID)
 
     def _getFilePathFromId(self, jobStoreFileID):
         """
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -248,7 +248,6 @@ class FileJobStore(AbstractJobStore):
                     try:
                         if self.exists(jobId):
                             yield self.load(jobId)
-                        raise NoSuchJobException(jobId)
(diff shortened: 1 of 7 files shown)


adamnovak:
Wait actually I'm not quite sure thing is the right approach now. We don't want to turn off the sleep check for the entire duration of Toil runs started with --restart. We just want to do it for the initial part where we are going through all the jobs on the leader, as described in #2917. So the flag should be something like `fastExistenceCheck` and not `restart`, and we need some way to clear it out eventually.

src/toil/jobStores/fileJobStore.py
```diff
@@ -95,6 +95,8 @@ def __init__(self, path, fanOut=1000):
 
         self.linkImports = None
 
+        self.restart = False
+
```
adamnovak: This isn't used anymore and should be removed.

src/toil/jobStores/fileJobStore.py
```diff
@@ -172,6 +174,7 @@ def waitForExists(self, jobStoreID, maxTries=35, sleepTime=1):
         over the course of large workflows with a jobStore on a busy NFS."""
         for iTry in range(1,maxTries+1):
             jobFile = self._getJobFileName(jobStoreID)
+            logging.debug("being called waitforexists for {}".format(jobStoreID))
```
adamnovak: Are you sure you want to leave this line in?

src/toil/jobStores/fileJobStore.py
```diff
@@ -244,8 +247,12 @@ def jobs(self):
                 logger.warning('Job Dir: %s' % i)
                 if i.startswith(self.JOB_DIR_PREFIX):
                     # This is a job instance directory
+                    jobId = self._getJobIdFromDir(os.path.join(tempDir, i))
                     try:
-                        yield self.load(self._getJobIdFromDir(os.path.join(tempDir, i)))
+                        if self.exists(jobId):
+                            yield self.load(jobId)
+                        else:
+                            raise NoSuchFileException(jobId)
```
adamnovak: This would be `NoSuchJobException` if you want to hit the `except` later. You could also just delete the whole `else` block.

As is it looks like when the job doesn't exist, you'll abort the whole workflow with a `NoSuchFileException`.

src/toil/jobStores/fileJobStore.py
```diff
@@ -244,8 +244,11 @@ def jobs(self):
                 logger.warning('Job Dir: %s' % i)
                 if i.startswith(self.JOB_DIR_PREFIX):
                     # This is a job instance directory
+                    jobId = self._getJobIdFromDir(os.path.join(tempDir, i))
                     try:
-                        yield self.load(self._getJobIdFromDir(os.path.join(tempDir, i)))
+                        if self.exists(jobId):
+                            yield self.load(jobId)
+                        raise NoSuchJobException(jobId)
```
adamnovak: Now it's going to raise and immediately catch this exception for every job, whether it exists and got yielded or not. It shouldn't do that.

adamnovak:
@jeffrey856 This is failing some Python 2 tests that shouldn't really be there anymore. I think you have to rebase onto master to make the tests pass.

adamnovak:
OK, IDK why Github shows the diff so weird, but it looks like this PR is really the right set of changes, and it passes tests. I'm going to merge it.

This should fix #2990.

Issue #2991, opened 2020-02-27
Toil process does not terminate when mesos batch system can't create new thread. 


glmxndr:
When launching a lot of parallel processes, if the subprocess ulimit is reached, the toil process does not stop properly.

Sample of stack:

```
Traceback (most recent call last):
  File "/home/gandrieu/.local/bin/cwltoil", line 11, in <module>
    sys.exit(main())
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/cwl/cwltoil.py", line 1508, in main
    outobj = toil.start(wf1)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/common.py", line 783, in start
    self._batchSystem = self.createBatchSystem(self.config)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/common.py", line 928, in createBatchSystem
    return batchSystemClass(**kwargs)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/mesos/batchSystem.py", line 87, in __init__
    super(MesosBatchSystem, self).__init__(config, maxCores, maxMemory, maxDisk)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/abstractBatchSystem.py", line 339, in __init__
    config, config.maxLocalJobs, maxMemory, maxDisk)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/singleMachine.py", line 144, in __init__
    worker.join()
  File "/usr/lib/python2.7/threading.py", line 929, in join
    raise RuntimeError("cannot join thread before it is started")
RuntimeError: cannot join thread before it is started
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-519)
┆Issue Number: TOIL-519

adamnovak:
I think we might fix #1902 soon, which would solve the underlying problem (the fact that each `cwltoil` wants to run 10+ threads per core) that is causing the ulimit to be exhausted in the first place. But we should also handle hitting the thread ulimit better.

PR #2992, opened 2020-02-27
Prevent zombie process when mesos batch system can't create new thread

glmxndr:


mr-c:
Is this really specific to mesos?

glmxndr:
Can't say, but that's the only batch system I'm able to test against. I'm not sure which others call the singleMachine code.

adamnovak:
I don't think this is Mesos-specific; the problem should affect all uses of the single machine batch system (both as the main batch system and for leader-only jobs in all the other batch systems that use it).

That being said, I don't really understand what the problem is (a deadlock between **threads** in the main Toil **process**? A dead **thread** in the main toil **process** causing the **process** to continue on with a bunch of non-daemon **threads** not doing anything? Or an unreaped ("zombie") child **process** that gets forked off somewhere and never `wait()`ed?) or how this PR is meant to fix it.

@glmxndr It looks like your stack trace in #2991 is *from* this PR (it is complaining about a join on line 144 of singleMachine.py, which isn't there until this patch is applied).

Is this supposed to result in all the previously-started workers being stopped as soon as a worker thread fails to start?

glmxndr:
I'm not sure either as I don't know exactly what the threading model in Python is, and only know parts of toil's code. What I know is that we use toil like this: a Java service is notified of a job to run, launches a Process calling a bash script itself launching cwltoil. When launching a lot of processes in parallel, many Process instances never terminate. 

So I just launched a bash script going like this:
```
for iter in {0..100}
do
  cwltoil ... &
done
``` 

Some of them worked correctly. For some of though, I had in their logs the stacktrace I gave in the #2991 ticket. Doing a `ps ux | grep cwl` lists them as still up. "Zombie process" is probably not the correct term, but they are hanging when they should have been terminated.

You're absolutely right that the stacktrace I gave in the #2991 ticket is the wrong one, sorry about that, I'll fix this tomorrow.

The fix I propose here is probably not the right thing to do, but it prevented the cwltoil process to hang. I'll try to make a cleaner proposal tomorrow.

PR #2998, opened 2020-03-02
Set worker threads as dameon to prevent hanging process on error

### LINKED ISSUE #2996: Launching many cwltoil processes in parallel leads to hanging processes
I have a bash script launching many cwltoil instances in parallel, like so:

```bash
for iter in {0..100}
do
  cwltoil ... &
done
```

When doing so, most of the processes are left hanging, with this stacktrace at the end of the logs.

```
  File "/home/gandrieu/.local/bin/cwltoil", line 11, in <module>
    result = main()
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/cwl/cwltoil.py", line 1508, in main
    outobj = toil.start(wf1)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/common.py", line 782, in start
    self._batchSystem = self.createBatchSystem(self.config)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/common.py", line 927, in createBatchSystem
    return batchSystemClass(**kwargs)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/mesos/batchSystem.py", line 87, in __init__
    super(MesosBatchSystem, self).__init__(config, maxCores, maxMemory, maxDisk)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/abstractBatchSystem.py", line 339, in __init__
    config, config.maxLocalJobs, maxMemory, maxDisk)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/batchSystems/singleMachine.py", line 140, in __init__
    worker.start()
  File "/usr/lib/python2.7/threading.py", line 736, in start
    _start_new_thread(self.__bootstrap, ())
thread.error: can't start new thread
```

I noticed that if I happen to configure the line with an inexisting input file, the behaviour is the same (many hanging processes) with this stacktrace.

```
  File "/home/gandrieu/.local/bin/cwltoil", line 11, in <module>
    result = main()
  File "/home/gandrieu/.local/lib/python2.7/site-packages/toil/cwl/cwltoil.py", line 1426, in main
    loading_context.overrides_list, tool_file_uri)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/cwltool/main.py", line 258, in load_job_order
    job_order_object, _ = loader.resolve_ref(job_order_file, checklinks=False)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 747, in resolve_ref
    doc = self.fetch(doc_url, inject_ids=(not mixin))
  File "/home/gandrieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 1212, in fetch
    text = self.fetch_text(url)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/schema_salad/ref_resolver.py", line 213, in fetch_text
    raise_from(ValidationException(Text(err)), err)
  File "/home/gandrieu/.local/lib/python2.7/site-packages/future/utils/__init__.py", line 483, in raise_from
    raise e
schema_salad.exceptions.ValidationException: [Errno 2] No such file or directory: '/home/gandrieu/Dev/inputs/args.json'
```

The hanging processes are still using around 90 megabytes of RAM each, no CPU activity whatsoever, and a `ps huH p $PID` lists 90 threads for each. I am able to kill these processes with a `killall cwltoil`.

Using pycharm's "Attach to process" functionality (allowing debugging of a running process) on one of these hanging processes reaches a connection timeout.

I could verify by modifying the cwltoil script that the main thread is not blocked, and that the `system.exit()` call is reached.

It looks like one or several threads are launched and are not joined properly.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-522)
┆Issue Number: TOIL-522

glmxndr:
fix #2996

### DIFF
--- a/src/toil/batchSystems/mesos/executor.py
+++ b/src/toil/batchSystems/mesos/executor.py
@@ -79,7 +79,6 @@ class MesosExecutor(Executor):
         log.debug("Registered executor %s with framework", self.id)
         self.address = socket.gethostbyname(agentInfo.hostname)
         nodeInfoThread = threading.Thread(target=self._sendFrameworkMessage, args=[driver], daemon=True)
-        nodeInfoThread.daemon = True
         nodeInfoThread.start()
 
     def reregistered(self, driver, agentInfo):
--- a/src/toil/batchSystems/mesos/executor.py
+++ b/src/toil/batchSystems/mesos/executor.py
@@ -78,7 +78,7 @@ class MesosExecutor(Executor):
         
         log.debug("Registered executor %s with framework", self.id)
         self.address = socket.gethostbyname(agentInfo.hostname)
-        nodeInfoThread = threading.Thread(target=self._sendFrameworkMessage, args=[driver])
+        nodeInfoThread = threading.Thread(target=self._sendFrameworkMessage, args=[driver], daemon=True)
         nodeInfoThread.daemon = True
         nodeInfoThread.start()
 
@@ -230,7 +230,7 @@ class MesosExecutor(Executor):
 
             driver.sendStatusUpdate(update)
 
-        thread = threading.Thread(target=runTask)
+        thread = threading.Thread(target=runTask, daemon=True)
         thread.start()
 
     def frameworkMessage(self, driver, message):
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1702,7 +1702,7 @@ class CachingFileStore(AbstractFileStore):
         self.waitForCommit()
 
         # Start the commit thread
-        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,), daemon=True)
+        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,))
         self.commitThread.start()
 
     def startCommitThread(self, jobState):
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1702,7 +1702,7 @@ class CachingFileStore(AbstractFileStore):
         self.waitForCommit()
 
         # Start the commit thread
-        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,))
+        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,), daemon=True)
         self.commitThread.start()
 
     def startCommitThread(self, jobState):
--- a/src/toil/provisioners/gceProvisioner.py
+++ b/src/toil/provisioners/gceProvisioner.py
@@ -368,7 +368,7 @@ class GCEProvisioner(AbstractProvisioner):
 
         threads = []
         for instance in instances:
-            t = threading.Thread(target=worker, args=(self._gceDriver,instance), daemon=True)
+            t = threading.Thread(target=worker, args=(self._gceDriver,instance))
             threads.append(t)
             t.start()
 
--- a/src/toil/provisioners/gceProvisioner.py
+++ b/src/toil/provisioners/gceProvisioner.py
@@ -368,7 +368,7 @@ class GCEProvisioner(AbstractProvisioner):
 
         threads = []
         for instance in instances:
-            t = threading.Thread(target=worker, args=(self._gceDriver,instance))
+            t = threading.Thread(target=worker, args=(self._gceDriver,instance), daemon=True)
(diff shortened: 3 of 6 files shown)


mr-c:
From https://docs.python.org/3.6/library/threading.html#thread-objects

> Note
>
> Daemon threads are abruptly stopped at shutdown. Their resources (such as open files, database transactions, etc.) may not be released properly. If you want your threads to stop gracefully, make them non-daemonic and use a suitable signalling mechanism such as an [`Event`](https://docs.python.org/3.6/library/threading.html#threading.Event). 

So probably better to use `Event` so that they can shut down gracefully.

adamnovak:
@mr-c I'm not sure a system to catch all exceptions in the leader thread and signal the worker threads to shut down is really going to pull its weight.

For one thing, it would take a neat trick to catch exceptions happening outside the batch system and ship them into any SingleMachineBatchSystem that happens to exist so it can notify its threads.

For another, we already have to be able to take a `kill -9` to the Toil leader without corrupting the job store or anything; you never know when your leader will suffer a critical existence failure or a power outage or drop off the network that connects it to the job store.

Jumping through a lot of hoops to avoid worker threads just vanishing when a Python exception is thrown won't actually free us from having to deal with the threads just vanishing in other cases.

That note in the docs is just saying that you can't rely on Python's close-and-flush behavior or any sort of context manager here, and that anything that isn't actually protected by a transaction mechanism could misbehave. The OS will take care of preventing us from doing something actually dangerous like using up all a user's open file descriptors with dead Toil processes.

Since I think we're going to pull out all the worker threads soon anyway, I don't think engineering a nicer system to notify them of failures and shut them down cleanly will be worth it.

mr-c:
Okay, I'm convinced!

adamnovak:
I've pulled this in for testing.

When reviewing, we should check if the commit thread in the worker really wants to be a daemon, or if it guaranteed to finish eventually.

glmxndr:
I admit I have not discriminated much... :) I wasn't sure exactly which were the ones misbehaving so I went a bit berserk.

src/toil/batchSystems/mesos/executor.py
```diff
@@ -78,7 +78,7 @@ def registered(self, driver, executorInfo, frameworkInfo, agentInfo):
         
         log.debug("Registered executor %s with framework", self.id)
         self.address = socket.gethostbyname(agentInfo.hostname)
-        nodeInfoThread = threading.Thread(target=self._sendFrameworkMessage, args=[driver])
+        nodeInfoThread = threading.Thread(target=self._sendFrameworkMessage, args=[driver], daemon=True)
         nodeInfoThread.daemon = True
```
adamnovak: We're making this a daemon twice now.

src/toil/fileStores/cachingFileStore.py
```diff
@@ -1702,7 +1702,7 @@ def startCommit(self, jobState=False):
         self.waitForCommit()
 
         # Start the commit thread
-        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,))
+        self.commitThread = threading.Thread(target=self.startCommitThread, args=(jobState,), daemon=True)
```
adamnovak: The commit thread runs `startCommitThread()` which just runs through and doesn't loop waiting to talk to other threads. So I think this shouldn't actually be a daemon.

src/toil/provisioners/gceProvisioner.py
```diff
@@ -368,7 +368,7 @@ def worker(driver, instance):
 
         threads = []
         for instance in instances:
-            t = threading.Thread(target=worker, args=(self._gceDriver,instance))
+            t = threading.Thread(target=worker, args=(self._gceDriver,instance), daemon=True)
```
adamnovak: These threads are also one-shot threads, and it's somewhat important for them to execute as much as they can, because they delete cloud instances that cost money. I don't think they should be daemons.

Really this should be a thread pool; who knows when we'll have more instances to shut down than we have thread ulimit available.

adamnovak: I went to turn this into a thread pool, but we don't actually have this under CI or any Google money, so I'd rather not refactor it without being able to test it.

glmxndr:
Thanks for the cleanup!

PR #2999, opened 2020-03-05
Cut the Threads

### LINKED ISSUE #1902: singleMachine batchSystem may be really inefficient
Currently, the singleMachine batch system launches a huge number of threads: 10 times the maxCores number. All these threads contend on a single queue waiting for jobs to appear. When they do get a job, they hold onto it and try to contend until a "resource pool" has enough resources to continue, and they can get an exclusive lock on the pool so they can decrement it.

This seems pretty inefficient. We could get the same behavior by just having one thread spin waiting for resources and monitoring the Popen objects. It's not totally clear how much speed we'd get from doing it with fewer threads, but it definitely shouldn't hurt. Plus, this will get rid of the uber-annoying "Recieved termination signal" tens or hundreds of times when exiting a single-machine Toil workflow.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-208)
┆Issue Number: TOIL-208

adamnovak:
This should fix #1902 and a swathe of related issues, by refactoring the SingleMachineBatchSystem to just use one "daddy" thread to mind all its child processes.

It basically keeps all the jobs in a delay line of a queue until they have the resources they need, which isn't efficient, but I doubt it's worse then 640 contending threads, and it won't stop scheduling small jobs if they queue up behind a lot of jobs that can't run yet.

It passes some local tests but I want to see what the real tests think of it.

### DIFF
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -184,8 +184,6 @@ class SingleMachineBatchSystem(BatchSystemSupport):
             while not self.shuttingDown.is_set():
                 # Main loop
 
-                log.debug('Daddy looking for new jobs.')
-                
                 while not self.shuttingDown.is_set():
                     # Try to start as many jobs as we can try to start
                     try:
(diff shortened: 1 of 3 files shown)


adamnovak:
I haven't actually run local tests on Linux yet, only Mac, and there's a different code path for polling the processes on Linux that I thought would be more efficient than actually polling each PID. But there's some risk of child terminations getting lost if someone else in the Toil process wants to use the `waitid` system call, so we should decide whether we want to just `popen.poll()` on all the children in a loop until one is done, or implement @diekhans's file descriptor based approach to child polling.

PR #3000, opened 2020-03-05
[Part 2] CWL v1.1 Support

DailyDreaming:
This adds support for the EnvVarRequirement portion of the following CWL1.1 tests:
 - [x] Test [227/253] Test requirements in input document via EnvVarRequirement
 - [x] Test [228/253] Test conflicting requirements in input document via EnvVarRequirement and expression
 - [x] Test [229/253] Test conflicting requirements in input document via EnvVarRequirement

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -37,7 +37,8 @@ from typing import (Text,
                     Any,
                     Iterator,
                     TextIO,
-                    Set)
+                    Set,
+                    Tuple)
 
 # Python 3 compatibility imports
 from six import iteritems, string_types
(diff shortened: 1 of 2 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -757,15 +797,58 @@ def __init__(self, tool, cwljob, runtime_context, step_inputs=None, conditional=
...
+            required_env_vars[k] = v  # will tell cwltool which env vars to take from the environment
+            os.environ[k] = v  # needs to actually be populated in the environment as well or they're not used
+
+        # EnvVarRequirement env vars take priority over those specified with "requirements"
+        # so cwltool.requirements need to be overwritten if an env var with the same name is found
+        for req in self.cwltool.requirements:
+            for env_def in req.get('envDef', {}):
+                env_name = env_def.get('envName', '')
+                if env_name in required_env_vars:
+                    env_def['envValue'] = required_env_vars[env_name]
```
adamnovak: You're sure that modifying the "requirements" env vars won't affect any other jobs? Jobs have both "requirements" and "EnvVarRequirements" just as part of the job, but if they conflict (because the author weirdly tried to use them both), "EnvVarRequirements" win?

DailyDreaming: Test 228 is the one where the author weirdly tried to use both: `Test conflicting requirements in input document via EnvVarRequirement and expression`

So I'm overriding based on that test.  I'm relying a bit on the conformance tests to help me understand what needs to work in what way, but this seems to be how cwltool overrides.  Maybe @mr-c can comment?

adamnovak:
I'm going to go ahead and merge this; it seems like it is working.

PR #3001, opened 2020-03-05
[easy] Raise and warn the user if using "cwltoil".

DailyDreaming:
Using `cwltoil` now can result in a confusing error (see: https://github.com/DataBiosphere/toil/issues/2987).

```
      File "/home/ubuntu/toil/src/toil/worker.py", line 366, in workerScript
          job._runner(jobGraph=jobGraph, jobStore=jobStore, fileStore=fileStore, defer=defer)
      File "/home/ubuntu/toil/src/toil/job.py", line 1392, in _runner
          returnValues = self._run(jobGraph, fileStore)
      File "/home/ubuntu/toil/src/toil/job.py", line 1329, in _run
          return self.run(fileStore)
      File "/home/ubuntu/toil/src/toil/cwl/cwltoil.py", line 809, in run
          if self.conditional.is_false(cwljob):
      AttributeError: 'CWLJob' object has no attribute 'conditional'
```

This adds a function to be called when `cwltoil` is run to raise and warn the user.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -90,8 +90,8 @@ CWL_INTERNAL_JOBS = ("CWLJobWrapper", "CWLWorkflow", "CWLScatter", "CWLGather",
 # output object to the correct key of the input object.
 
 
-def cwltoil_is_deprecated():
-    raise RuntimeError('Please run with "toil-cwl-runner" instead of "cwltoil" (deprecated).')
+def cwltoil_was_removed():
+    raise RuntimeError('Please run with "toil-cwl-runner" instead of "cwltoil" (which has been removed).')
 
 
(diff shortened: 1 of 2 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -90,6 +90,10 @@
 # output object to the correct key of the input object.
 
 
+def cwltoil_is_deprecated():
+    raise RuntimeError('Please run with "toil-cwl-runner" instead of "cwltoil" (deprecated).')
```
adamnovak: It's not deprecated; it's gone and doesn't work anymore. We should say it has been removed.

DailyDreaming: @adamnovak Updated!

Issue #2884, opened 2019-12-09
Big kubernetes job ends cryptically


glennhickey:
After swamping the cluster which a bunch of small jobs which seem to run fine, my cactus run aborts with the following connection error.
```
Traceback (most recent call last):
  File "/venv2/bin/cactus", line 8, in <module>
    sys.exit(main())
  File "/venv2/local/lib/python2.7/site-packages/cactus/progressive/cactus_progressive.py", line 496, in main
    runCactusProgressive(options)
  File "/venv2/local/lib/python2.7/site-packages/cactus/progressive/cactus_progressive.py", line 546, in runCactusProgressive
    halID = toil.start(RunCactusPreprocessorThenProgressiveDown(options, project, memory=configWrapper.getDefaultMemory()))
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 795, in start
    self._shutdownBatchSystem()
  File "/usr/local/lib/python2.7/dist-packages/toil/common.py", line 1073, in _shutdownBatchSystem
    self._batchSystem.shutdown()
  File "/usr/local/lib/python2.7/dist-packages/toil/batchSystems/kubernetes.py", line 666, in shutdown
    propagation_policy='Background')
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/apis/batch_v1_api.py", line 306, in delete_namespaced_job
    (data) = self.delete_namespaced_job_with_http_info(name, namespace, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/apis/batch_v1_api.py", line 406, in delete_namespaced_job_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/api_client.py", line 334, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/api_client.py", line 168, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/api_client.py", line 400, in request
    body=body)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/rest.py", line 256, in DELETE
    body=body)
  File "/usr/local/lib/python2.7/dist-packages/kubernetes/client/rest.py", line 166, in request
    headers=headers)
[...]

PR #3002, opened 2020-03-07
Kubernetes connection with Leader

jeffrey856:
Fix connection between Kubernetes and Leader.  #2884

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -793,13 +793,13 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
                 # Synthesize an exit code
                 exitCode = EXIT_STATUS_UNAVAILABLE_VALUE
                 # Say it ran from when the job was submitted to when the pod got stuck
-                runtime = slow_down((utc_now() - jobSubmitTime).totalSeconds())
+                runtime = slow_down((utc_now() - jobSubmitTime).total_seconds())
         else:
             # The pod went away from under the job.
             logging.warning('Exit code and runtime unavailable; pod vanished')
             exitCode = EXIT_STATUS_UNAVAILABLE_VALUE
             # Say it ran from when the job was submitted to when the pod vanished
-            runtime = slow_down((utc_now() - jobSubmitTime).totalSeconds())
+            runtime = slow_down((utc_now() - jobSubmitTime).total_seconds())
         
         
         try:
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -41,7 +41,6 @@ import sys
 import uuid
 import time
 import urllib3
-import http
 
 from kubernetes.client.rest import ApiException
 from six.moves.queue import Empty, Queue
@@ -63,7 +62,7 @@ def retryable_kubernetes_errors(e):
     A function that determins whether or not Toil should retry or stop given 
     exceptions thrown by Kubernetes. 
     """
-    if isinstance(e, urllib3.exceptions.MaxRetryError) or isinstance(e, http.client.GATEWAY_TIMEOUT) or isinstance(e, ApiException):
+    if isinstance(e, urllib3.exceptions.MaxRetryError) or isinstance(e, urllib3.exceptions.HTTPError(504)) or isinstance(e, ApiException):
         return False
     return True
 
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -454,7 +454,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
             kwargs = {'label_selector': query}
             if token is not None:
                 kwargs['_continue'] = token
-            results = self.try_kubernetes(self._api('core').list_namespaced_pod, self.namespace)
+            results = self.try_kubernetes(self._api('core').list_namespaced_pod, self.namespace, **kwargs)
             
             for pod in results.items:
                 # Return the first pod we find
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -63,10 +63,9 @@ def retryable_kubernetes_errors(e):
     exceptions thrown by Kubernetes. 
     """
     if isinstance(e, urllib3.exceptions.MaxRetryError) or \
-        getattr(getattr(e, 'response', None), 'status_code', None) == 504 or \
         isinstance(e, ApiException):
-        return False
-    return True
+        return True
+    return False
 
 def retry_kubernetes(retry_while=retryable_kubernetes_errors):
     """
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -455,7 +455,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
             if token is not None:
                 kwargs['_continue'] = token
             results = self.try_kubernetes(self._api('core').list_namespaced_pod, self.namespace, **kwargs)
-            
+        
             for pod in results.items:
                 # Return the first pod we find
                 return pod
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -41,6 +41,7 @@ import sys
 import uuid
 import time
 import urllib3
+import http
 
 from kubernetes.client.rest import ApiException
 from six.moves.queue import Empty, Queue
@@ -58,7 +59,7 @@ from toil.lib.retry import retry
 logger = logging.getLogger(__name__)
      
 def exception_found(e):
-    if e is urllib3.exceptions.MaxRetryError or e.code == 504 or e is kubernetes.client.rest.ApiException:
+    if e is urllib3.exceptions.MaxRetryError or e is http.client.GATEWAY_TIMEOUT or e is kubernetes.client.rest.ApiException:
         return False
     return True
 
@@ -755,7 +756,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
                     exitCode = EXIT_STATUS_UNAVAILABLE_VALUE
                     # Say it stopped now and started when it was scheduled/submitted.
                     # We still need a strictly positive runtime.
-                    runtime = slow_down((utc_now() - startTime).totalSeconds())
+                    runtime = slow_down((utc_now() - startTime).total_seconds())
                 else:
                     # Get the termination info from the pod's main (only) container
                     terminatedInfo = getattr(getattr(containerStatuses[0], 'state', None), 'terminated', None)
@@ -765,7 +766,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
                         exitCode = EXIT_STATUS_UNAVAILABLE_VALUE
                         # Say it stopped now and started when it was scheduled/submitted.
                         # We still need a strictly positive runtime.
-                        runtime = slow_down((utc_now() - startTime).totalSeconds())
+                        runtime = slow_down((utc_now() - startTime).total_seconds())
                     else:
                         # Extract the exit code
                         exitCode = terminatedInfo.exit_code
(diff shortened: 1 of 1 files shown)


src/toil/batchSystems/kubernetes.py
```diff
@@ -205,11 +218,15 @@ def _api(self, kind, max_age_seconds = 5 * 60):
                 
         else:
             # We need an API object
+            # WIP -------------------------------------------------------------------- 
+            for attempt in retry_http(predicate=handle_retriable_exception):
+                with attempt:
+                    api_get = self._apis[kind]
```
adamnovak: This is wrapping the dict access in the retry, not any actual use of the API object.

But we could turn `_api()` into a context manager, and make everything that uses a Kubernetes connection be inside a `with self._api('whatever'):`, and let `_api()` handle all the retries maybe?

Or you could apply this `with attempt in retry_http...` to all the places where we call any methods on the PAI objects we get from `_api()`.

src/toil/batchSystems/kubernetes.py
```diff
@@ -49,13 +49,21 @@
...
+    return e.error_code.endswith('.ApiException') or e.error_code.endswith('.MaxRetryError')
+
+def retry_kubernetes(t=5, retry_for=10 * 5, retry_while=exception_found):
+    return retry(delays=(t, t, t * 2, t * 4),
+                 timeout=retry_for,
+                 predicate=retry_while)
 
 def slow_down(seconds):
-    """
+    """f
```
adamnovak: Should this f be here?

src/toil/batchSystems/kubernetes.py
```diff
@@ -49,13 +49,21 @@
...
 from toil.lib.humanize import human2bytes
 from toil.resource import Resource
-from toil.lib.retry import retry_http
+
+from toil.lib.retry import retry
 
 logger = logging.getLogger(__name__)
 
+def exception_found(e):
+    return e.error_code.endswith('.ApiException') or e.error_code.endswith('.MaxRetryError')
```
adamnovak: You might still want to enforce that only some API exceptions (GATEWAY_TIMEOUT, MAX_RETRY_ERROR) are recognized here. We don't want to retry when our code makes a genuine mistake using the API. If we try to make a job in a nonexistent namespace, for example, we should just fail right away, I think, instead of retrying a bunch.

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +54,19 @@
 from toil.lib.humanize import human2bytes
 from toil.resource import Resource
 
+from toil.lib.retry import retry
+
 logger = logging.getLogger(__name__)
+     
+def exception_found(e):
+    if e is urllib3.exceptions.MaxRetryError or e is http.client.GATEWAY_TIMEOUT or e is kubernetes.client.rest.ApiException:
```
adamnovak: Shouldn't these be `isinstance` checks like [the S3 code has](https://github.com/DataBiosphere/toil/blob/e6dea9602ca05dcf28b190be97673fd57f69022f/src/toil/jobStores/aws/utils.py#L381-L393)? I think `e` is the exception object and not the class of the exception.

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +54,19 @@
 from toil.lib.humanize import human2bytes
 from toil.resource import Resource
 
+from toil.lib.retry import retry
+
 logger = logging.getLogger(__name__)
+     
+def exception_found(e):
```
adamnovak: I would call this something like `retryable_kubernetes_errors`. There's always definitely an exception.

Also, this function needs a docstring!

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +54,19 @@
...
+    if e is urllib3.exceptions.MaxRetryError or e is http.client.GATEWAY_TIMEOUT or e is kubernetes.client.rest.ApiException:
+        return False
+    return True
 
+def retry_kubernetes(t=5, retry_for=300, retry_while=exception_found):
+    return retry(delays=(t, 1*t, 1*t, 4*t, 16*t, 64*t),
```
adamnovak: We shouldn't hardcode all these numbers.

Since `retry_kubernetes` isn't called with non-default timeouts, we could just drop all the timeout numbers and parameters and call `retry` with only `predicate=` passed.

Or, we could factor out the defaults from [here](https://github.com/DataBiosphere/toil/blob/e6dea9602ca05dcf28b190be97673fd57f69022f/src/toil/jobStores/aws/utils.py#L344-L345) into `toil.lib.retry` and pull them in here. There's no reason I can see for Kubernetes retries to have different timing characterisitcs than other retries Toil does.

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +54,19 @@
...
+from toil.lib.retry import retry
+
 logger = logging.getLogger(__name__)
+     
+def exception_found(e):
+    if e is urllib3.exceptions.MaxRetryError or e is http.client.GATEWAY_TIMEOUT or e is kubernetes.client.rest.ApiException:
+        return False
+    return True
 
+def retry_kubernetes(t=5, retry_for=300, retry_while=exception_found):
```
adamnovak: This function could also use a docstring.

src/toil/batchSystems/kubernetes.py
```diff
@@ -88,6 +101,12 @@ def supportsAutoDeployment(cls):
     @classmethod
     def supportsWorkerCleanup(cls):
         return False
+
+    def try_kubernetes(self, method, *args, **kwargs):
```
adamnovak: This function needs a docstring that explains why we have it and admonishes everyone to use it instead of making calls directly. It could also have an example of how to use it in conjunction with `self._api`.

It also might be worth renaming it to `_try_kubernetes`, since it shouldn't really be called outside the class. Or, since this is the Kubernetes batch system anyway, we could call it `_retry` or `_retry_call` or something.

And I think it should move down. It shouldn't come before `__init__`. I would put it right after `_api`.

src/toil/batchSystems/kubernetes.py
```diff
@@ -492,7 +512,7 @@ def _isPodStuckOOM(self, podObject, minFreeBytes=1024 * 1024 * 2):
 
         # Look for it
         # TODO: When the Kubernetes Python API actually wraps the metrics API, switch to that
-        response = self._api('customObjects').list_namespaced_custom_object('metrics.k8s.io', 'v1beta1',
+        response = self.try_kubernetes(self._api('customObjects').list_namespaced_custom_object, 'metrics.k8s.io', 'v1beta1',
```
adamnovak: This line is a bit long and has incurred the horizontal-scrolling ire of Github's PR display. Some of the args on the first line should move down to their own line(s).

src/toil/batchSystems/kubernetes.py
```diff
@@ -583,9 +603,9 @@ def getUpdatedBatchJob(self, maxWait):
                                 terminated = pod.status.container_statuses[0].state.terminated
                                 runtime = slow_down((terminated.finished_at - terminated.started_at).total_seconds())
                                 result = UpdatedBatchJobInfo(jobID=jobID, exitStatus=terminated.exit_code, wallTime=runtime, exitReason=None)
-                                self._api('batch').delete_namespaced_job(pod.metadata.owner_references[0].name,
-                                                                         self.namespace,
-                                                                         propagation_policy='Foreground')
+                                self.try_kubernetes(self._api('batch').delete_namespaced_job, pod.metadata.owner_references[0].name,
```
adamnovak: This line could also wrap.

src/toil/batchSystems/kubernetes.py
```diff
@@ -799,7 +819,7 @@ def _getUpdatedBatchJobImmediately(self):
             self._waitForJobDeath(jobObject.metadata.name)
                     
         except kubernetes.client.rest.ApiException:
-            # TODO: check to see if this is a 404 on the thing we tried to delete
+            # TODO: check to see if this is a 404 on the thing wgit add e tried to delete
```
adamnovak: This "git add" has gotten lost and does not want to be here.

src/toil/batchSystems/kubernetes.py
```diff
@@ -211,7 +232,21 @@ def _api(self, kind, max_age_seconds = 5 * 60):
...
+    
+    def _try_kubernetes(self, method, *args, **kwargs):
+        """
+        Kubernetes API can end abruptly when it could wait and retry until successful or terminate immediatley if it encounters 
+        an unretryable error such as gateway timeout or max number of retries reached. 
+        For example, in a traditional case where we call self._api('batch').create_namespaced_job(self.namespace, job)
+        without this wrapper, Kubernetes can behave irregularly and inconsistently given a large job.
         
+        This new function, given a large volume of incoming jobs, we can give Kubernetes more time in between tries 
+        to check if the API is executable or not by calling self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, **kwargs).
```
adamnovak: This docstring sort of communicates the point of the function, but these are not really good sentences.

It sounds like a gateway timeout is unretryable, for example, but this function does indeed retry when there is a gateway timeout.

Also, the last sentence has two subjects ("This new function" and "we").

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +53,25 @@
...
 logger = logging.getLogger(__name__)
+     
+def retryable_kubernetes_errors(e):
+    """
+    A function that determins whether or not Toil should retry or stop given 
+    exceptions thrown by Kubernetes. 
+    """
+    if isinstance(e, urllib3.exceptions.MaxRetryError) or isinstance(e, urllib3.exceptions.HTTPError(504)) or isinstance(e, ApiException):
+        return False
+    return True
```
adamnovak: OK, now I think there are 2 new problems.

1. The sense is backward here. the function is supposed to return True if the error observed is likely to be temporary, and False if it's a permanent error that can't be fixed by retrying. As written, you're retrying for all errors *except* those on the list, so e.g. `NameError` (caused by using an undefined variable) will result in a retry, while `urllib3.exceptions.MaxRetryError` will not result in a retry. I think you need to swap False and True.

2. `urllib3.exceptions.HTTPError(504)` is an instance of `HTTPError` with `504` as its first argument. It is not a class that other exceptions are instances of. You can match all `HTTPError` instances with `isinstance(e, urllib3.exceptions.HTTPError)`. There's no documentation I can find about `urllib3` putting HTTP status codes into `HTTPError` anyway, so I'm not sure it makes any sense to try to catch only the `HTTPError`s that correspond to 504s coming from the server.

How did you decide to catch HTTPError here anyway? It didn't come up in https://github.com/DataBiosphere/toil/issues/2884

src/toil/batchSystems/kubernetes.py
```diff
@@ -52,8 +53,25 @@
...
 logger = logging.getLogger(__name__)
+     
+def retryable_kubernetes_errors(e):
+    """
+    A function that determins whether or not Toil should retry or stop given 
+    exceptions thrown by Kubernetes. 
+    """
+    if isinstance(e, urllib3.exceptions.MaxRetryError) or e.response.status_code == 504 or isinstance(e, ApiException):
+        return False
+    return True
```
adamnovak: The sense is still backward here, unfortunately.

This version of the check is better, but what if `e` doesn't have a `response`, or its `response` is some type that doesn't have a `status_code`? You should do something like `getattr(getattr(e, 'response', None), 'status_code', None) == 504` so it's true if the value is present and also equal to 504.

jeffrey856: Right thank you ....

adamnovak:
The exception checks look sensible now. If it turns out we're catching too many/few things, we'll fix it later. Maybe when we have a way to stress a Kubernetes cluster on purpose.

diekhans:
Also, isinstance can take a tuple of to check.  

>>> isinstance(1, (str, int))
True
>>> isinstance('foo', (str, int))
True

Issue #3004, opened 2020-03-09
Need a good way to read/find failed job logs


johnbradley:
When a CWL workflow fails it can be difficult to find the error messages generated by the failing command. Is there any guidance on how this data could be pulled out from larger logs?

As a simple example here is a simple CWL tool that has an invalid argument:
```
#!/usr/bin/env cwl-runner

cwlVersion: v1.0
class: CommandLineTool
baseCommand: wc
arguments: ["-c", "--fake-arg"]
inputs:
  input_file:
    type: File
    inputBinding:
      position: 1
outputs: []
```

You can run it with toil like so:  `toil-cwl-runner wc-tool.cwl --input_file wc-tool.cwl`
The following output is printed:
```
GCB-118-B.local 2020-03-09 12:16:11,597 MainThread INFO cwltool: Resolved 'wc-tool.cwl' to 'file:///Users/john/Documents/work/toil_log_fun/wc-tool.cwl'
GCB-118-B.local 2020-03-09 12:16:12,413 MainThread WARNING toil.batchSystems.singleMachine: Limiting maxCores to CPU count of system (8).
GCB-118-B.local 2020-03-09 12:16:12,413 MainThread WARNING toil.batchSystems.singleMachine: Limiting maxMemory to physically available memory (17179869184).
GCB-118-B.local 2020-03-09 12:16:12,413 MainThread WARNING toil.batchSystems.singleMachine: Limiting maxDisk to physically available disk (122682159104).
GCB-118-B.local 2020-03-09 12:16:12,428 MainThread INFO toil: Running Toil version 3.24.0-de586251cb579bcb80eef435825cb3cedc202f52.
GCB-118-B.local 2020-03-09 12:16:12,431 MainThread INFO toil.leader: Issued job 'file:///Users/john/Documents/work/toil_log_fun/wc-tool.cwl' wc kind-file_Users_john_Documents_work_toil_log_fun_wc-tool.cwl/instancepb96z3fs with job batch system ID: 0 and cores: 1, disk: 3.0 G, and memory: 2.0 G
INFO:toil.worker:Redirecting logging to /var/folders/fs/vv08ppmd1mj7g35hyq5m_7nr0000gp/T/toil-9d39cb7f-94e2-4093-8fa2-18e0f3ed86f9-268678283177431/tmp1xq4ln5c/worker_log.txt
[...]

PR #3005, opened 2020-03-10
fix python3 string/bytes error when using --printLogs

johnbradley:
Adds `decode('utf-8')` method call to fix the following error:
```
  .../toil/utils/toilStatus.py", line 83, in printJobLog
    msg += fH.read()
TypeError: can only concatenate str (not "bytes") to str
```
This fix is already being used in leader.py FailedJobsException:
https://github.com/DataBiosphere/toil/blob/993be0c3d95c83ca969d783e708c48d918407b16/src/toil/leader.py#L79-L80
Let me know if you would prefer a new function to abstract out this duplication. 

For testing the bug I added an example CWL workflow that always fails.

Fixes problem raised in #3004

### DIFF
--- a/src/toil/test/cwl/alwaysfails.cwl
+++ b/src/toil/test/cwl/alwaysfails.cwl
@@ -1,0 +1,11 @@
+#!/usr/bin/env cwl-runner
+# Command that will always fail
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: invalidcommand
+inputs:
+  message:
+    type: string
+    inputBinding:
+      position: 1
+outputs: []
--- a/src/toil/test/utils/utilsTest.py
+++ b/src/toil/test/utils/utilsTest.py
@@ -40,6 +40,7 @@ from toil.utils.toilStats import getStats, processData
 from toil.common import Toil, Config
 from toil.provisioners import clusterFactory
 from toil.version import python
+from mock import patch, mock_open
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


adamnovak:
The Gitlab tests are all failing due to the absence of the `pytz` module. It isn't installed because installing our dependencies bails out, because `http-parser` can't install, becuase it goes looking for `setuptools.Feature` and can't find it.

```
Collecting http-parser
  Downloading http-parser-0.8.3.tar.gz (83 kB)
    ERROR: Command errored out with exit status 1:
     command: /builds/databiosphere/toil/venv/bin/python3.6 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-1bte35b_/http-parser/setup.py'"'"'; __file__='"'"'/tmp/pip-install-1bte35b_/http-parser/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-1bte35b_/http-parser/pip-egg-info
         cwd: /tmp/pip-install-1bte35b_/http-parser/
    Complete output (5 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-1bte35b_/http-parser/setup.py", line 19, in <module>
        from setuptools import setup, find_packages, Extension, Feature
    ImportError: cannot import name 'Feature'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

Between the [last successful job](https://ucsc-ci.com/databiosphere/toil/-/jobs/44524) and [this one](https://ucsc-ci.com/databiosphere/toil/-/jobs/44938) we changed from setuptools 45 to 56. In 46 they removed `Feature` (see https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v4600 and  https://github.com/pypa/setuptools/issues/65); now http-parser is completely broken and not installable.

We either need to make Toil require setuptools <46, or fix http-parser and pymesos.

adamnovak:
OK I stacked this on top of current master which passed CI; I think this will pass too.

PR #3012, opened 2020-03-18
Kubernetes Shared Caching

### LINKED ISSUE #2988: Implement and test inter-pod caching on Kubernetes
We have a plan to make Toil work efficiently on Kubernetes without having to constantly re-download things.

Now that #2926 has been fixed, we *ought* to be able to share caches between Kubernetes pods on the same host, by setting the Toil `workDir` to be on a HostPath volume that will be shared between all the pods.

Someone needs to try actually doing this. It probably will require some patching the Kubernetes batch system to actually create the HostPath volume, and we need to see if the Kubernetes containers on the host all think they are the same machine or not, and thus whether they deign to actually share the cache.

We should also produce a test for this. It might involve hacking the Kubernetes batch system to give pods affinity for each other, and checking to make sure they actually cache things.

If we can actually make this work we may not need to do #2923.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-518)
┆Issue Number: TOIL-518

adamnovak:
This will eventually fix #2988.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -47,6 +47,7 @@ from six.moves.queue import Empty, Queue
 from toil import applianceSelf, customDockerInitCmd
 from toil.batchSystems.abstractBatchSystem import (AbstractBatchSystem,
                                                    BatchSystemLocalSupport)
+from toil.common import Toil
 from toil.lib.humanize import human2bytes
 from toil.lib.threading import LastProcessStandingArena
 from toil.resource import Resource
@@ -985,9 +986,12 @@ def executor():
     # that knows where the work dir is.
     cleanupInfo = job['workerCleanupInfo']
     
+    # We need to get the real workDir, not just the override from cleanupInfo.
+    workDir = Toil.getToilWorkDir(cleanupInfo.workDir)
+    
     # Join a Last Process Standing arena, so we know which process should be
     # responsible for cleanup.
-    arena = LastProcessStandingArena(cleanupInfo.workDir, 
+    arena = LastProcessStandingArena(workDir, 
         cleanupInfo.workflowID + '-kube-executor')
     arena.enter()
     
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -312,7 +312,8 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
             if self.host_path is not None:
                 # Provision Toil WorkDir from a HostPath volume, to share with other pods
                 host_path_volume_name = 'workdir'
-                host_path_volume_source = kubernetes.client.V1HostPathVolumeSource(path=self.host_path)
+                # Use type='Directory' to fail if the host directory doesn't exist already.
+                host_path_volume_source = kubernetes.client.V1HostPathVolumeSource(path=self.host_path, type='Directory')
                 host_path_volume = kubernetes.client.V1Volume(name=host_path_volume_name,
                                                              host_path=host_path_volume_source)
                 volumes.append(host_path_volume)
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -48,6 +48,7 @@ from toil import applianceSelf, customDockerInitCmd
 from toil.batchSystems.abstractBatchSystem import (AbstractBatchSystem,
                                                    BatchSystemLocalSupport)
 from toil.lib.humanize import human2bytes
+from toil.lib.threading import LastProcessStandingArena
 from toil.resource import Resource
 
 logger = logging.getLogger(__name__)
@@ -85,7 +86,7 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
 
     @classmethod
     def supportsWorkerCleanup(cls):
-        return False
+        return True
    
     def __init__(self, config, maxCores, maxMemory, maxDisk):
         super(KubernetesBatchSystem, self).__init__(config, maxCores, maxMemory, maxDisk)
@@ -249,6 +250,10 @@ class KubernetesBatchSystem(BatchSystemLocalSupport):
             job = {'command': jobNode.command,
                    'environment': self.environment.copy()}
             # TODO: query customDockerInitCmd to respect TOIL_CUSTOM_DOCKER_INIT_COMMAND
+            
+            # Send the worker cleanup info so that the last worker on a
+            # host to shut down can clean up if warranted.
+            job['workerCleanupInfo'] = self.workerCleanupInfo
 
             if self.userScript is not None:
                 # If there's a user script resource be sure to send it along
@@ -958,10 +963,13 @@ def executor():
 
     logging.basicConfig(level=logging.DEBUG)
     logger.debug("Starting executor")
+    
+    # If we don't manage to run the child, what should our exit code be?
+    exit_code = 1
 
     if len(sys.argv) != 2:
         logger.error('Executor requires exactly one base64-encoded argument')
-        sys.exit(1)
+        sys.exit(exit_code)
 
     # Take in a base64-encoded pickled dict as our first argument and decode it
     try:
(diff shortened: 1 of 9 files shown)


adamnovak:
OK, I can use this, along with `TOIL_KUBERNETES_HOST_PATH=/data/scratch`, to get caching working between pods. It does add some overhead though on my test caching-benchmark workflow:

```
- With caching and sharing:
    real	5m16.515s
    user	0m31.565s
    sys	0m2.832s
- With no caching and no sharing:
    real	5m24.036s
    user	0m31.381s
    sys	0m2.754s
- Only idna fix (baseline):
    real	5m1.922s
    user	0m30.138s
    sys	0m2.872s
```

Hopefully this can be made up for by the benefits of caching.

Right now the `Resource` system doesn't use a shared directory across pods. It could be adapted to do it, though, if it didn't insist on a fresh temp dir every time it started up but instead knew how to connect to one.

adamnovak:
I don't think the new shared directory caching is under CI right now. I still have to figure out how to hook that up.

src/toil/lib/threading.py
```diff
@@ -335,9 +333,222 @@ def process_name_exists(workDir, name):
...
+        os.unlink(lock_filename)
+        fcntl.lockf(fd, fcntl.LOCK_UN)
+        # Note that we are unlinking it and then unlocking it; a lot of people
+        # might have opened it before we unlinked it and will wake up when they
+        # get the worthless lock on the now-unlinked file. We have to do some
+        # stat gymnastics above to work around this.
+        os.close(fd)
+        
 
+class LastProcessStandingArena:
```
DailyDreaming: This is excellent.

src/toil/lib/threading.py
```diff
@@ -335,9 +333,222 @@ def process_name_exists(workDir, name):
...
+        
+        log.debug('Leaving arena %s', self.lockfileDir)
+        
+        with global_mutex(self.workDir, self.mutex):
+            # Now nobody else should also be trying to join or leave.
+            
+            # Take ourselves out.
+            try:
+                os.unlink(self.lockfileName)
+            except:
```
DailyDreaming: What happens if this has an error and cannot delete the lock file?

Is this worth posting a debug log message when this occurs?

src/toil/lib/threading.py
```diff
@@ -335,9 +333,222 @@ def process_name_exists(workDir, name):
...
+            else:
+                # Nothing alive was found. Nobody will come in while we hold
+                # the global mutex, so we are the Last Process Standing.
+                log.debug('We are the Last Process Standing in arena %s', self.lockfileDir)
+                yield True
+                
+                try:
+                    # Delete the arena directory so as to leave nothing behind.
+                    os.rmdir(self.lockfileDir)
+                except:
```
DailyDreaming: I suppose I'll ask a similar question here, what happens if this has an error and cannot delete the lock dir?

Is this worth posting a debug log message when this occurs?

src/toil/test/docs/scripts/example_cachingbenchmark.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def root(job, options):
+    # Make a file
+    with job.fileStore.writeGlobalFileStream() as (stream, file_id):
+        stream.write(('This is a test of the Toil file caching system. ' +
+            'Had this been an actual file, its contents would have been more interesting.').encode('utf-8'))
+    
+    child_rvs = []
+    for i in range(100):
```
DailyDreaming: How long does this test take to run?

src/toil/test/src/threadingTest.py
```diff
@@ -0,0 +1,145 @@
...
+                # Make sure all workers say they succeeded
+                self.assertEqual(item, True)
+                
+    @travis_test
+    def testLastProcessStanding(self):
+        for it in range(10):
+            log.info('Iteration %d', it)
+        
+            scope = self._createTempDir()
+            arena_name = 'thunderdome'
```
DailyDreaming: Heh, +1.

src/toil/test/src/threadingTest.py
```diff
@@ -0,0 +1,145 @@
...
+            out_stream.write(str(number))
+            out_stream.close()
+            
+            # Wait
+            time.sleep(random.random() * 0.01)
+            
+            # Make sure our name is still there
+            in_stream = open(potato, 'r')
+            seen = in_stream.read().rstrip() 
+            assert seen == str(number), "We are {} but {} stole our potato!".format(number, seen)
```
DailyDreaming: I'm not sure how python handles open file handles upon an exception exit these days, but could I suggest either moving the `in_stream.close()` above the assert statement, or using the context manager `with open(file, 'r') as f:` (possibly for the code several lines above it as well where it will mostly just save a line of code).

src/toil/test/src/threadingTest.py
```diff
@@ -0,0 +1,145 @@
...
+            # Put our name there
+            out_stream = open(my_precious, 'w')
+            out_stream.write(str(number))
+            out_stream.close()
+            
+            # Wait
+            time.sleep(random.random() * 0.01)
+            
+            # Make sure our file is still there unmodified
+            assert os.path.exists(my_precious), "Precious file {} has been stolen!".format(my_precious)
```
DailyDreaming: Mostly a nit, and somewhat DRY, but this code seems to be repeating the potato experience above.  Would it be useful to be centralized in a single, say precious potato, function they will both be forced to share?

src/toil/batchSystems/kubernetes.py
```diff
@@ -943,10 +981,13 @@ def executor():
 
     logging.basicConfig(level=logging.DEBUG)
     logger.debug("Starting executor")
+    
+    # If we don't manage to run the child, what should our exit code be?
```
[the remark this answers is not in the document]
adamnovak: I picked this because it's already (now) specified as what get_updated_batch_job ought to return if it doesn't have an exit code for the job. I figured the case of being able to start the pod but not whatever the user asked for ought to be similar to other weird conditions like the pod vanishing.

src/toil/batchSystems/kubernetes.py
```diff
@@ -955,27 +996,60 @@ def executor():
...
+                # Do batch system cleanup.
+                logger.debug('Cleaning up worker')
+                BatchSystemSupport.workerCleanup(cleanupInfo)
+    finally:
+        logger.debug('Cleaning up resources')
+        # TODO: Change resource system to use a shared resource directory for everyone.
+        # Then move this into the last-process-standing cleanup
+        Resource.cleanSystem()
+        logger.debug('Shutting down')
+        sys.exit(exit_code)
```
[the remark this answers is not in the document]
adamnovak: Nope. We define it up near the top of the executor.

src/toil/lib/threading.py
```diff
@@ -335,9 +333,222 @@ def process_name_exists(workDir, name):
...
+            path_stats = None
+        
+        if path_stats is None or fd_stats.st_dev != path_stats.st_dev or fd_stats.st_ino != path_stats.st_ino:
+            # The file we have a lock on is not the file linked to the name (if
+            # any). This usually happens, because before someone releases a
+            # lock, they delete the file. Go back and contend again. TODO: This
+            # allows a lot of queue jumping on our mutex.
+            fcntl.lockf(fd, fcntl.LOCK_UN)
+            os.close(fd)
+            continue
```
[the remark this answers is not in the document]
adamnovak: We could say something like "the default timeout for waiting for a mutex is X minutes" or something. But I figured that the probability of someone actually managing to starve a process for X minutes and coming back to complain in a bug report was pretty high, especially if we ever re-use this for something else. Plus, generally you expect a mutex to block until it successfully locks; that's its job.

There is some probability that a worker manages to hang in some deferred function cleanup forever and blocks all the other workers. But no particular timeout is going to be "correct" either.

src/toil/lib/threading.py
adamnovak: I figured unlink would only error out if the file was already gone. But it is possible the delete will somehow not be allowed.

In that case I think the file just gets left behind; the impact is that everyone else who leaves will see that the file is there, see that it isn't locked, try to delete it, fail to delete it, and continue on. There will be a lot of wasted time attempting deletes, and all the files will be left behind when the workflow finishes, but I don't think anything will actually *break*.

We could print a warning, but it would be a warning along the lines of "the filesystem doesn't work". I wouldn't add a debug message here unless we actually go to debug a problem where the workdir isn't properly being cleaned.

src/toil/lib/threading.py
adamnovak: Here it might be sensible to have a warning, because we don't expect anything to still be there, and if something made it in, then somebody didn't respect the locks.

src/toil/test/docs/scripts/example_cachingbenchmark.py
adamnovak: I timed it at about 5 minutes on our cluster.

src/toil/test/src/threadingTest.py
adamnovak: That makes sense.

src/toil/test/src/threadingTest.py
adamnovak: Maybe if we use it again. I kept it separate because even though the contents logic is the same, when the files are created and destroyed is different, and a write-wait-read-back function seemed a bit of a weird unit.

src/toil/lib/threading.py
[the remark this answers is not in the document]
DailyDreaming: Hmmm... alright, I'll defer to you on this one since it is trying to be OS-like and therefore could reasonably be expected to block infinitely.  In general I like to have an escape hatch, but I can't really think of a good default timeout either.  Something to keep in mind though if we ever see a cached kube job somehow frozen on one worker.

Issue #3020, opened 2020-03-31
Update conda recipe to support python3


ifiddes:
The current conda recipe for toil is version 0.24.1, but requires python be 2.*.

https://bioconda.github.io/recipes/toil/README.html

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-533)
┆Issue Number: TOIL-533

adamnovak:
It looks like @matthdsm and @chapmanb might be the Bioconda recipe maintainers.

It also looks like Bioconda does some magic reading from PyPI or something to track our releases. Maybe it will be able to pick up that we [ask for Py3.6 via classifier](https://github.com/DataBiosphere/toil/blob/eda5d4cc05518b761c6a4e0c1905286ffd139552/setup.py#L127) when we do the 4.0 release?

adamnovak:
I'm trying explaining to setuptools that we want Python 3.6. Maybe bioconda can figure that out?

matthdsm:
The latest version of toil in bioconda (3.24.0) is supposed to be noarch, which means it's compatible with py2 and py3.
If there's a hard requirement for py2, then it's derived from one of the listed dependencies or pinned versions of those dependencies

You can check the listed dependencies [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/toil/meta.yaml). If there's anything that doesn't need to be pinned or any dependencies that are deprecated, please remove them from the yaml file. It'll help the build process and improve compatibility.

Cheers
M

matthdsm:
actually, since v4.0 drop py2 support, it might be an ideal time to clean up the dependencies! Huraah for progess! 🥇 

M

PR #3021, opened 2020-03-31
Make dependency on Python3.6+ machine-readable

adamnovak:
I don't want to pin to 3.6 exactly because we can work fine on newer versions without the Mesos extra.

This might make #3020 be magically solved by a Conda bot.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -134,6 +134,7 @@ def runSetup():
           'Topic :: System :: Distributed Computing',
           'Topic :: Utilities'],
         license="Apache License v2.0",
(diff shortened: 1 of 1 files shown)


adamnovak:
I think the Kubernetes batch system test failed here because our Kubernetes cluster was too busy for its jobs to run in time to satisfy the test. I'm going to rerun it. We should think about whether/how we could reserve capacity for these tests, and if we wanted to how we would pay for it. I suspect running another Kubernetes inside the first Kubernetes would be a Bad Idea, but...

PR #3025, opened 2020-04-02
Do More Hashing

### LINKED ISSUE #2375: Toil AWS JobStore does not protect workflow from truncated downloads
I just ran a workflow that failed when one of my jobs encountered a partially-downloaded file.

I imported this from S3:

```
2018-09-03 13:59:35 9786900533 platinum-WG38_NA12878_haplo_thread_1.xg
```

And I got this in the job store bucket:

```
2018-09-20 17:21:21 9786900533 f506b22a-c4b8-4376-835d-85e822b8e7e3
```

But after downloading it with `job.fileStore.readGlobalFile` the file was truncated. This caused the `vg` command my job was running on the file to crash, and when my code dumped out the file it as it saw it, I got this:

```
2018-09-20 19:14:44 9786896570 index.xg
```

This file is 3963 bytes shorter than, and a prefix of, the file I tried to download from the file store:

```
cmp index.xg platinum-WG38_NA12878_haplo_thread_1.xg
cmp: EOF on index.xg
```

I believe this is related to [this sort of behavior from boto](https://forums.aws.amazon.com/thread.jspa?messageID=701669); downloads from S3 can appear to succeed, with a 200 HTTP code, but with only part of the file contents.

The Toil code to download the file just runs Boto until it gets no error:
https://github.com/DataBiosphere/toil/blob/5c7b39501338fc0d01a600846d04f6a98b259cce/src/toil/jobStores/aws/jobStore.py#L1148-L1152

It needs to check Boto's work to verify that the entire contents of the file were obtained, checking the bytes downloaded against the expected file length. If the file data was truncated, Toil needs to transparently resume the download where Boto left off, to hide this partial download behavior from application code.

Toil *especially* needs to make sure that partially-downloaded files don't make it into the cache.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-279)
┆Issue Number: TOIL-279

adamnovak:
This should fix #2375, at least enough to satisfy me. We now compute checksums even for streamed uploads and downloads (as long as the downloads read through to EOF). We also retry non-streaming downloads if the checksum fails.

This doesn't quite implement the resume-from-last-obtained-byte functionality that #2375 wants,. If we want to do that, I think we need to track expected sizes in the job store, in addition to checksums. But even without the resume feature, this should protect user code from ever seeing a silent partial download.

If we run into a file that's so enormous that AWS keeps failing during the download *and* retrying the download from the beginning stops us from actually making progress, then we can implement byte range resumes of partial downloads.

### DIFF
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -1113,12 +1113,16 @@ class AWSJobStore(AbstractJobStore):
                 expected = parts[1]
             
             hasher = getattr(hashlib, algorithm)()
+            
+            log.debug('Starting %s checksum to match %s', algorithm, expected)
+            
             return (algorithm, hasher, expected)
             
         def _update_checksum(self, checksum_in_progress, data):
             """
             Update a checksum in progress from _start_checksum with new data.
             """
+            log.debug('Updating checksum with %d bytes', len(data))
             checksum_in_progress[1].update(data)
         
         def _finish_checksum(self, checksum_in_progress):
@@ -1128,6 +1132,8 @@ class AWSJobStore(AbstractJobStore):
             """
             
             result_hash = checksum_in_progress[1].hexdigest()
+            
+            log.debug('Completed checksum with hash %s vs. expected %s', result_hash, checksum_in_progress[2])
             
             if checksum_in_progress[2] is not None:
                 # We expected a particular hash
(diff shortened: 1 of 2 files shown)


adamnovak:
I think running the CI tests filled up our AWS account with abandoned job stores. I think something is crashing and leaving the tests unable to do their cleanup. I should test this manually against AWS, one job store at a time, until it works properly.

adamnovak:
This isn't exploding and throwing out a hundred abandoned buckets when I run it manually, now. Hopefully this will pass.

PR #3036, opened 2020-04-13
Use the original pymesos.

DailyDreaming:
This gets rid of the toil-pymesos package stuff we originally created as a work around.


DailyDreaming:
@diekhans Kind of agree.  I'll make another PR after this to pare things down.

PR #3041, opened 2020-04-15
always honor TOIL_GRIDENGINE_PE and never assume a site-dependent default

### LINKED ISSUE #3040: SGE batch system doesn't always honor TOIL_GRIDENGINE_PE
the TOIL_GRIDENGINE_PE environment parameter is only used for when number of job CPUs is great than one.  Some configurations require -pe  even with one CPU.  Also, if multiple CPUs are requested and TOIL_GRIDENGINE_PE is not set, it defaults to 'shm'.  There is no requirement to have such a PE.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-541)
┆Issue Number: TOIL-541

diekhans:
fixes #3040

### DIFF
--- a/src/toil/batchSystems/gridengine.py
+++ b/src/toil/batchSystems/gridengine.py
@@ -134,9 +134,9 @@ class GridEngineBatchSystem(AbstractGridEngineBatchSystem):
                     if arg.startswith(("vf=", "hvmem=", "-pe")):
                         raise ValueError("Unexpected CPU, memory or pe specifications in TOIL_GRIDGENGINE_ARGs: %s" % arg)
                 qsubline.extend(sgeArgs)
-            if cpu is not None and math.ceil(cpu) > 1:
-                peConfig = os.getenv('TOIL_GRIDENGINE_PE') or 'shm'
-                qsubline.extend(['-pe', peConfig, str(int(math.ceil(cpu)))])
+            if os.getenv('TOIL_GRIDENGINE_PE') is not None:
+                peCpu = int(math.ceil(cpu)) if cpu is not None else 1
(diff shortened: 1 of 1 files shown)


diekhans:
Doh!  Good eye.  Yea, PE is the only way in for the number of CPUs.

I changed it to require PE if cpu is > 1.

adamnovak:
So how come there's no way to pick a sensible default? Is there a way to list the PEs available on a cluster? When a user encounters a `must specify PE in TOIL_GRIDENGINE_PE` error, what procedure exactly are they supposed to follow to get a working value? Open a support ticket with their site's admins?

It looks like [maybe](https://docs.oracle.com/cd/E19957-01/820-0698/6ncdvjcmd/index.html) we want to tell the user to run `qconf -spl` to list parallel environments available. In which case we might as well run it ourselves and either default to the only one or print a list of options for the user.

diekhans:
The PEs are dependent on how the cluster is configured, they are just names that map to configurations. and the project and queue selected.  I think we would be in danger of making it worse by making suggestions.  The best is to give an error to get them to consult their local
documentation or admins.

adamnovak:
I think we should at least suggest the `qconf -spl` command. It's not going to be a big problem if it somehow doesn't work, and it might save them digging through a bunch of documentation or having to open a support ticket (or just giving up on Toil).

diekhans:
I suggested qconf and reading the local documentation.  If they
only have one PE, then qconf is good; otherwise they need to
read the doc to figure out the right one to use.  The system I
am on has five.

adamnovak:
Sounds good!

PR #3043, opened 2020-04-22
Detect insufficient resources and deadlocks more usefully

### LINKED ISSUE #3016: Tell user when job hangs due to lack of resources
When Toil can't schedule a job due to lack of resources, it just hangs indefinitely.  If the user is not running `--logDebug`, then they never get told why their job has stopped.  

This comes up frequently in Cactus, ex:
https://github.com/ComparativeGenomicsToolkit/cactus/issues/150
https://github.com/ComparativeGenomicsToolkit/cactus/issues/182

Can Toil please make a clear warning for these situations at a higher log level?  

In general it's hard to detect a deadlock, but I think on `singleMachine` if it's in a state where no jobs are running, and no jobs can be run, then exiting with a clear error would be much better than the current behaviour.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-530)
┆Issue Number: TOIL-530

adamnovak:
This fixes #3016. 
That issue is really caused by Cactus setting the deadlock checking time too high, so it takes us an hour to even see the oops-all-services deadlock, and another hour to do anything about it.

This PR changes Toil to separate the time we wait to terminate the workflow from the time we wait to see if there's a deadlock. It adds some warning-level log messages when the deadlock checker *starts* to think there's a deadlock, and it is able to pull a hint from the batch system (currently only implemented for `SingleMachineBatchSystem`) that can suggest to the user why the deadlock might be happening.

With Cactus installed, I did this to produce a service deadlock:

```
mkdir -p ~/mnt/tmpfs
sudo mount -t tmpfs -o size=2200000000 tmpfs ~/mnt/tmpfs
TMPDIR=$HOME/mnt/tmpfs cactus --workDir=$HOME/mnt/tmpfs ./tree examples/evolverMammals.txt examples/evolverMammals.hal --root mr
```

And now I get these messages about it:

```
Issued job 'CactusSetupPhase' kind-CactusSetupPhase/instance-juylrkhr with job batch system ID: 2 and cores: 1, disk: 2.0 G, and memory: 3.3 G
Potential deadlock detected! All 1 running jobs are service jobs, with no normal jobs to use them! The batch system reports: Not enough disk to run job 2
Potentially deadlocked for 30 seconds. Waiting at most 3570 more seconds for any of 1 issued non-service jobs to schedule and start. The batch system reports: Not enough disk to run job 2
Potentially deadlocked for 60 seconds. Waiting at most 3540 more seconds for any of 1 issued non-service jobs to schedule and start. The batch system reports: Not enough disk to run job 2
...
```

It might be noisy in big cluster runs, but you can (manually) up `--deadlockCheckInterval` if you know you're on a batch system that takes a while to schedule jobs and don't want to hear about it. It might make sense to up it (and the deadlock wait too) when running with autoscaling, automatically, if we can work out how.

### DIFF
--- a/src/toil/batchSystems/__init__.py
+++ b/src/toil/batchSystems/__init__.py
@@ -12,19 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import absolute_import
-from past.builtins import cmp
-from builtins import str
-from builtins import object
 import sys
 
-if sys.version_info >= (3, 0):
+from functools import total_ordering
 
-    # https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons
-    def cmp(a, b):
-        return (a > b) - (a < b)
-
-class MemoryString(object):
+@total_ordering
+class MemoryString:
+    """
+    Represents an amount of bytes, as a string, using suffixes for the unit.
+    
+    Comparable based on the actual number of bytes instead of string value.
+    """
     def __init__(self, string):
         if string[-1] == 'K' or string[-1] == 'M' or string[-1] == 'G' or string[-1] == 'T': #10K
             self.unit = string[-1]
@@ -55,8 +53,8 @@ class MemoryString(object):
         elif self.unit == 'T':
             return self.val * 1099511627776
 
-    def __cmp__(self, other):
-        return cmp(self.bytes, other.bytes)
+    def __eq__(self, other):
+        return self.bytes == other.bytes
 
-    def __gt__(self, other):
-        return self.bytes > other.bytes
+    def __lt__(self, other):
+        return self.bytes < other.bytes
--- a/src/toil/batchSystems/__init__.py
+++ b/src/toil/batchSystems/__init__.py
@@ -16,6 +16,21 @@ import sys
 
 from functools import total_ordering
 
+class DeadlockException(Exception):
+    """
+    Exception thrown by the Leader or BatchSystem when a deadlock is encountered due to insufficient
+    resources to run the workflow
+    """
+    def __init__(self, msg):
+        self.msg = "Deadlock encountered: " + msg
+        super().__init__()
+
+    def __str__(self):
+        """
+        Stringify the exception, including the message.
+        """
+        return self.msg
+
 @total_ordering
 class MemoryString:
     """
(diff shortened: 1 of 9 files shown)


adamnovak:
@diekhans Does this inspire you to bliss?

diekhans:
It would inspire me to wait and see what happens :-)

src/toil/batchSystems/__init__.py
```diff
@@ -12,19 +12,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import absolute_import
-from past.builtins import cmp
-from builtins import str
-from builtins import object
 import sys
 
-if sys.version_info >= (3, 0):
```
DailyDreaming, line 21: It always feels good seeing these removed.  Thanks!

src/toil/batchSystems/__init__.py
```diff
@@ -12,19 +12,32 @@
...
 
-class MemoryString(object):
+    def __str__(self):
+        """
+        Stringify the exception, including the message.
+        """
+        return self.msg
+
+@total_ordering
+class MemoryString:
```
DailyDreaming, line 35: Created an issue based on this since I think some of this is duplicated in a few places: https://github.com/DataBiosphere/toil/issues/3046

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -180,6 +180,26 @@ def getUpdatedBatchJob(self, maxWait):
                  that were killed.
         """
         raise NotImplementedError()
+        
+    def getSchedulingHint(self):
+        """
+        Get a hint for the user about anything that might be going wrong in the
```
DailyDreaming, line 186: Hints in CWL are resource requirements, so to avoid confusion, could this explicitly state somewhere in the first sentence that this is intended to be a log message?  Something like:

s/`Get a hint for the user`/`Get a log message string serving as a hint for the user`/g ?

adamnovak:
OK, I've revised this in light of your suggestions @DailyDreaming. I've changed the documentation around for the deadlock checking time parameters, so it should be clearer what they are actually there for, and I've replaced the "hints" with "details" and "status messages".

PR #3044, opened 2020-04-23
Replace reissued jobs message with better progress indicators

adamnovak:
This removes `Reissued any over long jobs`, everyone's favorite message to incorrectly blame their problems on.

When the reissue machinery actually reissues things, that should still be reported. But when it does nothing we get no message.

To take over its role as "the message you see when nothing happens", I added a periodic message shoeing the number of jobs running and the number of jobs issued but not running. This should help debug secret deadlocks, or jobs that run forever.

I've also added a progress bar:
image

This will tell you if your job completion rate tanks, and also tracks total failed jobs (although it doesn't know if they're completely failed vs. retried). It just compares ever-finished and ever-issued job counts; it doesn't walk the job store to find jobs that can't be issued yet due to dependencies. Is that too misleading? I can replace it with a simpler counter that just shows jobs finished/issued/failed and maybe jobs per second, without implying we have any idea when the workflow will be done.

I *wanted* to show the number of running vs. issued jobs as a status line, so you can see right away if your jobs aren't scheduling/when they schedule, but out batch systems don't expose that information in a good way. We can poll all the running jobs constantly (and possibly bowl over the backing scheduler), but we can't get events when jobs go from issued to running in order to update our progress bar/status line in a timely fashion. We should think about maybe making `getUpdatedBatchJob` tell you when a job goes from issued to running, or else some way of caching the running job info or indicating which batch systems won't fall over if you query it every second.


mr-c:
@adamnovak This looks great, thanks! What would appear in a log of stdout/err if `--disableProgress` is specified?

adamnovak:
@mr-c Right now it doesn't print any particular log messages in lieu of the progress/status bar, if you turn it off. You'd just have the hourly report that replaces the old reissued message; it looks something like "5 jobs are running, 2 jobs are issued and waiting to run".

PR #3048, opened 2020-04-30
Support for new resource syntax in PBSPro

Phhere:
PBSPro has deprecated the support for nodes=..:ppn=... since some versions and you should use the new syntax for it.


Phhere:
I have found some more issues with PBSPro and will fix them

adamnovak:
@Phhere Thanks for working on this. Is it still WIP, or is it ready for me to pull into the Toil repo for integration testing?

Phhere:
I am currently testing it togeather with cactus. when the pipeline is done, it can be merged i think

mr-c:
Any update @phhere? Thanks again for your contribution!

adamnovak:
I've set this up for testing. If it passes and you think it works @mr-c, I suppose we can just pull it in.

src/toil/batchSystems/torque.py
```diff
@@ -163,12 +163,21 @@ def prepareQsub(self, cpu, mem, jobID):
...
-                memStr = str(old_div(mem, 1024)) + 'K'
-                reqline.append('mem=' + memStr)
+            if self._version == "pro":
+                request = 'select=1'
+                if mem is not None:
+                    memStr = str(old_div(mem, 1024)) + 'K'
+                    request +=':mem=' + memStr
+                if cpu is not None and math.ceil(cpu) > 1:
+                    request +=':ncpus=' + str(int(math.ceil(cpu)))
+                 reqline.append(request)
```
mr-c, line 173: error: unindent does not match any outer indentation level

mr-c:
I pushed this branch to our repo as `issues/3863-newer-pbspro` so that GitLabCI would run

PR #3053, opened 2020-05-06
Stop CWL from trying to copy downloaded directories from somewhere local

### LINKED ISSUE #2801: Toil tries to move directories by copy instead of through the FileStore in CWL conformance tests
As discovered in #2770, this CWL conformance test will pass with job chaining enabled on Toil master:

```
wget https://github.com/common-workflow-language/common-workflow-language/archive/a062055fddcc7d7d9dbc53d28288e3ccb9a800d8.zip
unzip a062055fddcc7d7d9dbc53d28288e3ccb9a800d8.zip
mv common-workflow-language-a062055fddcc7d7d9dbc53d28288e3ccb9a800d8/ test/
toil-cwl-runner --disableCaching=True --jobStore ./chain --logDebug --realTimeLogging --outdir=/tmp/tmpb5MyuT2 test/v1.0/v1.0/iwdr_with_nested_dirs.cwl test/v1.0/v1.0/empty.json
```

But it will fail when chaining is disabled:

```
toil-cwl-runner --disableCaching=True --disableChaining --jobStore ./nochain --logDebug --realTimeLogging --outdir=/tmp/tmpb5MyuT2 test/v1.0/v1.0/iwdr_with_nested_dirs.cwl test/v1.0/v1.0/empty.json
```

It is failing because, when setting up the second job cwltool ends up trying to recursively copy a directory form the first job's local temp directory. With chaining on this directory happens to still exist, but with chaining off it will have already been cleaned up. In general, two jobs might run on completely different machines, so cwltool needs to be shipping the directory tree around via Toil's FileStore.

The exceptions look like this:

```
  File "/Users/anovak/Documents/Workspace/toil/venv35/lib/python3.5/site-packages/cwltool/executors.py", line 182, in run_jobs
    job.run(runtime_context)
  File "/Users/anovak/Documents/Workspace/toil/venv35/lib/python3.5/site-packages/cwltool/job.py", line 445, in run
    symlink=True, secret_store=runtimeContext.secret_store)
  File "/Users/anovak/Documents/Workspace/toil/venv35/lib/python3.5/site-packages/cwltool/process.py", line 253, in stage_files
    shutil.copytree(entry.resolved, entry.target)
  File "/Users/anovak/Documents/Workspace/toil/venv35/lib/python3.5/shutil.py", line 309, in copytree
    names = os.listdir(src)
FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/ws/t7b9wm2j2b7fy4_xyq0n5vh00000gn/T/toil-dfba1124-76bd-45d9-ada7-12e2e8d2020f-106618767760070/tmp5tgq0qpe/3f3d9b8e-b1cd-4ff7-ac5a-2dd0a8c05afc/tcc4hil93/tmp-outnngwbct7/deeply'
```

@mr-c Is there something you can do about this in cwltool? Or is there something wrong in cwltoil that is causing cwltool to try and do these local copies when it needs to be shipping around tarballs or rebuilding the directory structure instead?

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-424)
┆Issue Number: TOIL-424

adamnovak:
[`cwltool` seems to treat `_:` as a special protocol for directories that need to be created rather than resolved to a filesystem path that gets copied.](https://github.com/common-workflow-language/cwltool/blob/9f3b9e7b74d5a904b12674dfd1300b56a48c3d33/cwltool/process.py#L311-L313) Since the Toil file store only stores files, and CWL ships around the whole metadata tree of files and directories needed to fulfill an InitialWorkDir input, we can just tell CWL that all the directories in the metadata come from the `_:` protocol.

For some reason, we can't just set all the directories' locations to just `_:`. They need to be unique for the `PathMapper` to be able to keep them straight. So I'm just prefixing `_:` onto the source `file:` URL that the directory had on the originating machine. This seems to be enough for CWL to both know that the directories are themselves and know that they don't need to be copied from anywhere.

I haven't tested this with any CWL workflows that actually drag along files in their directories (CI tests that, right?), but it works with the empty directory test from #2801. This should fix #2801.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -674,9 +674,9 @@ def prepareDirectoryForUpload(directory_metadata: dict,
                 
     # The metadata for a directory is all we need to keep around for it. It
     # doesn't have a real location. But each directory needs a unique location
-    # or CWL won't ship the metadata along. CWL takes "_:" as a signal to make
-    # directories instead of copying from somewhere. So we give every directory
-    # a unique _: location and CWL's machinery Just Works.
+    # or cwltool won't ship the metadata along. cwltool takes "_:" as a signal
+    # to make directories instead of copying from somewhere. So we give every
+    # directory a unique _: location and cwltool's machinery Just Works.
     directory_metadata["location"] = "_:" + directory_metadata["location"]
     
     logger.debug("Sending directory: %s", directory_metadata)
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -928,6 +928,14 @@ class CWLJob(Job):
         return required_env_vars
 
     def run(self, file_store: AbstractFileStore) -> Any:
+        
+        # Adjust cwltool's logging to conform to Toil's settings.
+        # We need to make sure this happens in every worker process before we
+        # do CWL things.
+        cwllogger.removeHandler(defaultStreamHandler)
+        cwllogger.setLevel(logger.getEffectiveLevel())
+        
+        
         cwljob = resolve_dict_w_promises(self.cwljob)
         fill_in_defaults(
             self.step_inputs,
@@ -1411,6 +1419,7 @@ def remove_unprocessed_secondary_files(unfiltered_secondary_files: dict) -> list
 
 def main(args: Union[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
     """Main method for toil-cwl-runner."""
+    # Remove cwltool logger's stream handler so it uses Toil's
     cwllogger.removeHandler(defaultStreamHandler)
     config = Config()
     config.cwl = True
@@ -1580,6 +1589,8 @@ def main(args: Union[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
     use_container = not options.no_container
 
     if options.logLevel:
+        # Make sure cwltool uses Toil's log level.
+        # Applies only on the leader.
         cwllogger.setLevel(options.logLevel)
 
     outdir = os.path.abspath(options.outdir)
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -133,7 +133,7 @@ def filter_skip_null(name: str, value: Any) -> Any:
     err_flag = [False]
     value = _filter_skip_null(value, err_flag)
     if err_flag[0]:
-        cwllogger.warning(
+        logger.warning(
             "In %s, SkipNull result found and cast to None. \n"
             "You had a conditional step that did not run, "
             "but you did not use pickValue to handle the skipped input." % name)
@@ -312,7 +312,7 @@ class ResolveSource:
             return values
 
         if not isinstance(values, list):
-            cwllogger.warning(
+            logger.warning(
                 "pickValue used but input %s is not a list." % self.name)
             return values
 
@@ -474,7 +474,6 @@ class ToilPathMapper(PathMapper):
             
     def mapper(self, src: str) -> MapperEnt:
         result = super(ToilPathMapper, self).mapper(src)
-        logger.debug("ToilPathMapper: Mapped %s to %s", src, result)
         return result
 
     def visit(self,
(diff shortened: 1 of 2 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -631,6 +646,40 @@ def write_file(writeFunc: Any, index: dict, existing: dict, file_uri: str) -> st
...
+            return
+        else:
+            raise cwltool.errors.WorkflowException(
+                "Directory is missing: %s" % directory_metadata["location"])
+                
+    # The metadata for a directory is all we need to keep around for it. It
+    # doesn't have a real location. But each directory needs a unique location
+    # or CWL won't ship the metadata along. CWL takes "_:" as a signal to make
+    # directories instead of copying from somewhere. So we give every directory
+    # a unique _: location and CWL's machinery Just Works.
```
mr-c, line 679: s/CWL/cwltool/g

src/toil/cwl/cwltoil.py
```diff
@@ -82,6 +82,8 @@
 from toil.common import Config, Toil, addOptions
 from toil.version import baseVersion
 
+logger = logging.getLogger(__name__)
```
mr-c, line 85: Why not use `_logger` from https://github.com/DataBiosphere/toil/blob/dd7322725bac8fb65aacd70c3f3514982a2c8b2d/src/toil/cwl/cwltoil.py#L68 ?

adamnovak: I set this up the way other Toil components do logging. I've never done logging the cwltool way before.

Presumably we should use only one of these in this module.

I actually would favor the Toil style as the right one to use here. The [cwltool logger setup](https://github.com/common-workflow-language/cwltool/blob/master/cwltool/loghandler.py) is pretty simple; it hardcodes INFO as the log level, logs everything as coming from "`cwltool`", and sends it to a default stream handler. The Toil style is to have loggers that report what file is doing the logging, and it obeys some options (`--logDebug`, `--logInfo`, etc.) to let the user control the log level.

Since Toil's `cwltoil.py` isn't actually `cwltool`, I think it makes sense for it to not log via `cwltool`'s internal logger. Moreover, since I'm adding some debug-level messages, I need a logger that I can set to debug level from the command line to actually see them, and the Toil logger setup has that, while I don't think cwltool's does.

Should Toil maybe be doing something to cwltool's `_logger` to adopt it and get its log messages to go through the same stream handler as Toil's?

adamnovak: Yeah, if I send my debug-level logging to `cwllogger`, I don't see it even when running with `--logDebug`.

adamnovak:
OK @mr-c, I think I've addressed your comments.

I changed some references to "CWL" in general to now be about cwltool specifically.

I've also settled on pointing all the logging that `cwltoil.py` does itself to a Toil-style logger initialized with `__name__`, instead of creating log messages that claim to come from the `cwltool` module. But I've also fixed up the logging reconfiguration for the `cwltool` module so it happens in the worker processes as well as in the leader process. This makes Toil's `--logDebug` suddenly start showing the debug-level logging messages from `cwltool` in the individual job logs, like it should have been doing this whole time. It turns out `cwltool` logs a bunch of info about Files and Directories at DEBUG level already, so I cut back on some of my own added logging in favor of that.

Now it looks something like this:

```
DEBUG:cwltool:[job 3a931e67-06b1-4290-9e4d-4b6d9b4d585b] outputs {
    "ya": {
        "location": "file:///tmp/node-a9f628d3-fd18-44ac-aea0-8169c3407e32-11c2d78d2bcc48f28d74bd1f7dc0a45f/tmp6j7b_4tk/50a1e33b-5ccb-4bc2-bb1e-7f018716ad41/t2kvh9tuy/tmp-outuvoqo1t7/deeply/nested/dir/structure/ya",
        "basename": "ya",
        "nameroot": "ya",
        "nameext": "",
        "class": "File",
        "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709",
        "size": 0,
        "http://commonwl.org/cwltool#generation": 0
    }
}
DEBUG:cwltool:[job 3a931e67-06b1-4290-9e4d-4b6d9b4d585b] Removing input staging directory /tmp/tmpvb4k026e
DEBUG:toil.cwl.cwltoil:Sending file at: toilfs:0:files/for-job/kind-file_z_home_anovak_workspace_toil_test_v1.0_v1.0_iwdr_with_nested_dirs.cwl_second_3a931e67-06b1-4290-9e4d-4b6d9b4d585b/instance-395aiuba/file-nbpf8szh/ya
DEBUG:toil.fileStores.abstractFileStore:LOG-TO-MASTER: Job files/for-job/kind-CWLWorkflow/instance-2ssncyrt/cleanup/file-epir9qk1/stream used 0.00% (52.0 KB [53248B] used, 3.0 GB [3221225472B] requested) at the end of its run.
```

PR #3065, opened 2020-05-20
Don't log in the inner scheduling loop

### LINKED ISSUE #3064: Debug logging with single machine is now extremely spammy because it logs on every scheduling loop.
Trying to run Cactus with --logDebug on single machine gives me:

```
Tried to start job 41 and got: None
Not enough cores to run job 38
Tried to start job 38 and got: None
Not enough cores to run job 41
Tried to start job 41 and got: None
Not enough cores to run job 38
Tried to start job 38 and got: None
Not enough cores to run job 41
Tried to start job 41 and got: None
...
```

I get many messages like this per second.

This is my own dang fault for spinning so fast/logging so much in the inner loop of the single machine scheduler. I think we should just remove these particular debug log lines as we aren't currently trying to debug the single machine scheduler itself.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-555)
┆Issue Number: TOIL-555

adamnovak:
Logging about trying and scheduling to start jobs, at 0.01 second intervals, makes `--logDebug` useless for debugging anything except the scheduler.

Fixes #3064.

### DIFF
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -207,8 +207,6 @@ class SingleMachineBatchSystem(BatchSystemSupport):
                         result = self._startChild(jobCommand, jobID,
                             coreFractions, jobMemory, jobDisk, environment)
 
-                        log.debug('Tried to start job %s and got: %s', jobID, str(result))
-                        
                         if result is None:
                             # We did not get the resources to run this job.
                             # Requeue last, so we can look at the next job.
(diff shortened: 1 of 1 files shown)


adamnovak:
@DailyDreaming Do you have any idea what went wrong with this test?

https://ucsc-ci.com/databiosphere/toil/-/jobs/52814

adamnovak:
I'm going to rerun the test and see if it's an intermittent problem. But we don't want those, so we still need to figure out what actually went wrong.

PR #3067, opened 2020-05-21
Handle anticipated 404 errors without retrying

### LINKED ISSUE #3066: Kubernetes batch system retries when it is expecting exceptions
When a Kubernetes job finishes, we delete it:

https://github.com/DataBiosphere/toil/blob/ba1f150dbb2f28081227e9f0d52748928d0be76c/src/toil/batchSystems/kubernetes.py#L862-L865

And then we wait for it to be gone, which is accomplished by polling for it and waiting for the polling to fail:

https://github.com/DataBiosphere/toil/blob/ba1f150dbb2f28081227e9f0d52748928d0be76c/src/toil/batchSystems/kubernetes.py#L885-L904

When we added Kubernetes retry logic, we also added retry logic to these cases, where we want the job to be gone and we are at some point hoping to get back a 404 from the Kubernetes API. Right now we retry the 404-ing requests until timeout, and then, when `retry` gives up, we see the exception we were expecting and continue. This makes the Kubernetes batch system so slow that it is kind of useless, but not so slow as to break our tests that only run a couple jobs.

We should only retry non-404 exceptions in these instances (`kubernetes.client.rest.ApiException` with a [`status`](https://github.com/kubernetes-client/python/blob/7763e37b7d8ca6e966716ad82d54f82f3e341cd4/kubernetes/client/rest.py#L305) of `404`)

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-556)
┆Issue Number: TOIL-556

adamnovak:
This should fix #3066.

When the Docker builds I still need to actually try it on Kubernetes to make sure it really does pass along the errors promptly instead of retrying them.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -70,14 +70,18 @@ def retryable_kubernetes_errors(e):
         isinstance(e, ApiException):
         return True
     return False
-
-def retry_kubernetes(retry_while=retryable_kubernetes_errors):
+    
+def retryable_kubernetes_errors_expecting_gone(e):
     """
-    A wrapper that sends retryable Kubernetes predicates into a context-manager which will allow 
-    Kubernetes to keep retrying until a False or an executable method is seen.  
+    A function that determins whether or not Toil should retry or stop given 
+    exceptions thrown by Kubernetes, when Toil is expecting a 404 indicating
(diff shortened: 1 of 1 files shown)


adamnovak:
The Python 3.8 appliance build failed because it couldn't fetch a key from a keyserver. I've kicked it.

PR #3073, opened 2020-05-27
Unify logging format

### LINKED ISSUE #3072: Worker log has no timestamps
I'm working on figuring out where the time is going in Cactus on Toil Kubernetes, and I want to see when the worker logs things, but it does its own plain basicConfig and doesn't include timestamps. It should include timestamps with its log messages, and use the same format as the rest of Toil.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-560)
┆Issue Number: TOIL-560

adamnovak:
This fixes #3072 by making sure that messages logged in the worker have timestamps on them.

I'm also trying to follow the recommendations at https://reflectoring.io/logging-format/ for "Human-Readable Logging Formats" by reformatting the logging messages. I've made them look like this on both the leader and the worker:

```
[2020-05-27 14:43:04,578] [MainThread] [W] [toil.resource] Can't globalize module ModuleDescriptor(dirPath='/z/home/anovak/workspace/cactus/venv/lib/python3.6/site-packages', name='cactus.progressive.cactus_progressive', fromVirtualEnv=True).
```

The thread name is clipped at/padded to 10 characters, and the level is reduced to a single character. All the fixed-length fields come first so they are lined up. All the metadata is in `[]` to distinguish it visually from the messages. The hostname is removed since I don't think it bought us anything; it doesn't look like it reflects the source hostname for real-time logging messages. I've made the workers log their hostnames when they log the Toil version, in case we want to see what host a weird worker's log came from.

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -18,6 +18,7 @@ import errno
 import logging
 import os
 import requests
+import socket
 import sys
 import time
 from datetime import datetime
@@ -415,7 +416,7 @@ def logProcessContext(config):
     # toil.version.version (string) cannot be imported at top level because it conflicts with
     # toil.version (module) and Sphinx doesn't like that.
     from toil.version import version
-    log.info("Running Toil version %s.", version)
+    log.info("Running Toil version %s on host %s.", version, socket.gethostname())
     log.debug("Configuration: %s", config.__dict__)
 
 
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -53,6 +53,8 @@ from toil.batchSystems.abstractBatchSystem import (AbstractBatchSystem,
                                                    EXIT_STATUS_UNAVAILABLE_VALUE,
                                                    UpdatedBatchJobInfo)
 from toil.common import Toil
+from toil.lib.bioio import configureRootLogger
+from toil.lib.bioio import setLogLevel
 from toil.lib.humanize import human2bytes
 from toil.lib.threading import LastProcessStandingArena
 from toil.resource import Resource
@@ -1039,7 +1041,8 @@ def executor():
 
     """
 
-    logging.basicConfig(level=logging.DEBUG)
+    configureRootLogger()
+    setLogLevel("DEBUG")
     logger.debug("Starting executor")
     
     # If we don't manage to run the child, what should our exit code be?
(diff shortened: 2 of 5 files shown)


diekhans:
I am of the opinion that ISO time format should always be used,
it is easy to parse, fairly readable and widely used standard.
Doing all logging in GMT good as you remove all need to deal
with time zones.

While human readability is nice, I think that machine
readability is more important, simple because you can build
different tools on top of it, including ones to make it human
readable.  For instance there are several emacs packages
for field log files and I believe there are vim ones as well.

So while abbreviating the log level is fine, as it can be
unambiguously parsed, truncating fields is losing information.

adamnovak:
I'm not sure how Toil determines the time format now. I don't think I'm changing it here, but we could adjust it.

I find that I have to deal with more time zones when logging is in GMT than when it is in my machine's time zone, whatever it is. If I see a message that says `11:00`, and it's now `11:30` according to my desktop clock, how old is the message? If it's in GMT regardless of my system tome zone, it depends on where I'm sitting, and I have to do math. If it uses the system time zone, it's either in GMT (for most servers) or local time (for people's local machines), so at least sometimes it's an easily human-readable time. But we should definitely include the time zone in case anyone does want to parse the logs, and so people can tell the difference.

I guess I can let the thread names get longer and just bump the other fields over if they're too long.

adamnovak:
OK, how about this:

```
[2020-05-27T16:07:21-0700] [MainThread] [D] [botocore.credentials] Looking for credentials via: env
```

As far as I can tell that's an ISO-compliant way to write the time, and (although you can't see it) I'm no longer truncatint the thread names, just padding them.

diekhans:
ISO local time is in the form: 2020-05-27T19:03:57-0400 Time
zone abbreviations are evil, since they are not unique and then
you get into the DST mess.

I use these bash aliases (and equivalent emacs commands) frequently
  alias idate='date +%Y-%m-%d'
  alias itime='date +%Y-%m-%dT%T%z'
  alias igmt='date -u +%Y-%m-%dT%T%z'

I just type igmt if I am looking at GMT, since I can't do simple
math.

The thing I don't understand about the python logging convention
is how one is suppose to turn the logger(__name__) into useful
filtering.  Not that I have a better idea.  All I see is
recommendations on to use it, not how to actually make it useful.

diekhans:
> [2020-05-27T16:07:21-0700] [MainThread] [D] [botocore.credentials] Looking for credentials via: env

this looks both readable and parsable

adamnovak:
Hey @diekhans, do you want to formally review this PR on Github so I can merge it?

PR #3076, opened 2020-06-03
get wait duration is far too short of a rest for lsf.

drkennetz:
We use LSF on prem and the 8 second wait duration has caused our HPC admins some headaches because bjobs is being polled to frequently which can cause lsf daemons to be occupied and not available to submit jobs. There is no reason to poll every 8 seconds, so I have changed it to 60 on prem. It has not had any negative consequences in terms of toil performance and it has since made our HPC admins happier.

### DIFF
--- a/src/toil/batchSystems/lsf.py
+++ b/src/toil/batchSystems/lsf.py
@@ -181,7 +181,7 @@ class LSFBatchSystem(AbstractGridEngineBatchSystem):
 
     def getWaitDuration(self):
         """We give LSF a second to catch its breath (in seconds)"""
-        return 8
+        return 60
 
     @classmethod
     def obtainSystemConstants(cls):


diekhans:
the SLUM batch system has a way to determine the wait period by querying the configuration.  Would it be possible to do this for LSF?

There is also a --statePollingWait option, but it is not clear to me how this relates to the wait period in the batch system,  Maybe these should be unified?

drkennetz:
@diekhans I'm not entirely sure what we would query the configuration for. I believe SLURM uses `sacct` right? The LSF mbatch daemon has configurations of which it checks the state of jobs and a separate parameter to accept new jobs and it does it fairly often (every 5-10 seconds). When you query bjobs, you are separately telling the mbatchd to check to see the status of these jobs. If you query it too frequently you are getting in the way of the scheduler and interrupting its functionality by forcing checks every 8 seconds. This isn't a problem when it is just 1 job polling, but it becomes a problem at scale. 

The --statePollingWait changes the default. So for LSF the default is 8. I am arguing that you should set the default higher to be safe and allow users to lower it at their own risk.

mr-c:
Thank you @drkennetz !

drkennetz:
@mr-c happy to help where I can!

Issue #3082, opened 2020-06-09
Toil does user work in local jobs in CWL workflows. Is chaining to blame?


mr-c:
Running Toil version 4.2.0a1-3f4e790c1fa32a2c2257bd5cd203a3b51d5d5661
Batch scheduler: LSF

> Toil is thinking that there are running jobs, but we don’t have them in bqueue

looks like
```
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
6 jobs are running, 1 jobs are issued and waiting to run
```
https://github.com/EBI-Metagenomics/pipeline-v5/blob/master/workflows/subworkflows/raw_reads/functional_annotation_raw.cwl

https://github.com/EBI-Metagenomics/pipeline-v5/blob/master/workflows/conditionals/raw-reads/raw-reads-2.cwl

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-566)
┆Issue Number: TOIL-566

KeteSakharova:
Hi, 

Maybe this also could help:
I killed main job and restarted again

```bash
Got exit code 1 (indicating failure) from job _toil_worker CWLWorkflow file:/nfs/public/release/metagenomics_scratch/pipeline-5/job-store-ERR1697183 kind-CWLWorkflow/instance-2kiod88f.
Got exit code 1 (indicating failure) from job _toil_worker CWLWorkflow file:/nfs/public/release/metagenomics_scratch/pipeline-5/job-store-ERR1697183 kind-CWLWorkflow/instance-3na_vy1m.
[...]

PR #3085, opened 2020-06-11
Log local jobs more comprehensively and make extra sure not to lose them

adamnovak:
This might help with #3082.


adamnovak:
I think @mr-c fixed the underlying issue for #3082, which was CWL local jobs chaining to non-local jobs. I'm pretty sure that the jobs I was concerned about were still running and not actually lost.

The extra checks I added to make sure the daddy thread is running cause some tests to fail, because they don't actually start the daddy thread before interrogating the batch system, I think.

I think I will abandon these changes.

PR #3088, opened 2020-06-18
Document Kubernetes

### LINKED ISSUE #3068: Kubernetes documentation.
Our documentation is a bit light, for reference: https://toil.readthedocs.io/en/latest/search.html?q=kubernetes&check_keywords=yes&area=default

@adamnovak We should probably add this next sprint if possible I think.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-557)
┆Issue Number: TOIL-557

adamnovak:
I'm writing some long-form Toil-on-Kubernetes documentation.

This will fix #3068.


adamnovak:
These docs are now fully written. They should still be run through on `minikube` to make sure they actually work.

adamnovak:
I had a weird test failure that appeared to result from multiple job updates being in flight at the same time. It may have something to do with the new nonzero worker exit statuses. I've adjusted the leader to try and deduplicate new job states, but later ones might now be getting lost; I think we need to redo the whole system to not represent an update to a jobgraph as a mutated jobgraph.

adamnovak:
OK, I've pulled out my patch for the intermittent test failure I was investigating into #3108. Hopefully this PR will still happen to pass without it, and we can merge that fix separately.

adamnovak:
@DailyDreaming This should be ready for review when you get a chance. I still want to manually test it on Minikube but I'm not sure I'll be able to get to that before our sprint is over, and I'm not sure we should wait.

PR #3096, opened 2020-06-26
Add --nodeStorageOverrides option.

### LINKED ISSUE #3084: Add a `--nodeStorageOverride` option to customize disk per node type
Currently, all nodes have a uniform disk size specified via the `--nodeStorage` flag. However, some jobs can be heterogeneous requiring vastly different disk requirements. For example, we have a pipeline where some tasks require 2TB+ of disk per task, but the rest require much smaller (e.g. 100GB) amount of disk per task, so setting 2TB of disk for all workers is wasteful especially when there are hundreds of such workers running.

Since the auto scaler is disk-aware, then adding the override to the `nodeShape` definition will automatically let the autoscaler choose the node type that has the override for the particular task (typically, these tasks also require large number of CPUs/RAM, so in practice, they turn out to be "extra large nodes in all aspects").

Example use case: `--nodeTypes c5.9xlarge:1.53,m4.10xlarge:2.0 --nodeStorage 200 --nodeStorageOverride m4.10xlarge:2000`

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-568)
┆Issue Number: TOIL-568

arostamianfar:
Adds an option to override `nodeStorage` for specific `nodeTypes`. Useful for heterogeneous jobs where some tasks need much more disk than others.
Example use case:
`--nodeTypes c5.2xlarge,c5.9xlarge,m4.10xlarge --nodeStorage 100 --nodeStorageOverrides c5.9xlarge:2000`

Fixes #3084

### DIFF
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -189,6 +189,11 @@ the logging module:
                         when they are launched in gigabytes. You may want to
                         set this if your jobs require a lot of disk space. The
                         default value is 50.
+  --nodeStorageOverrides NODESTORAGEOVERRIDES
+                        (optional) Comma-separated list of nodeType:nodeStorage that
+                        are used to override the default value from --nodeStorage for
+                        the specified nodeType(s). This is useful for heterogeneous jobs
+                        where some tasks require much more disk than others.
   --metrics             Enable the prometheus/grafana dashboard for monitoring
                         CPU/RAM usage, queue size, and issued jobs.
   --defaultMemory INT   The default amount of memory to request for a job.
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -190,9 +190,9 @@ the logging module:
                         set this if your jobs require a lot of disk space. The
                         default value is 50.
   --nodeStorageOverrides NODESTORAGEOVERRIDES
-                        (optional) Comma-separated list of nodeType:nodeStorage that
-                        are used to override the default value from --nodeStorage for
-                        the specified nodeType(s). This is useful for heterogeneous jobs
+                        Comma-separated list of nodeType:nodeStorage that are used
+                        to override the default value from --nodeStorage for the
+                        specified nodeType(s). This is useful for heterogeneous jobs
                         where some tasks require much more disk than others.
   --metrics             Enable the prometheus/grafana dashboard for monitoring
                         CPU/RAM usage, queue size, and issued jobs.
--- a/docs/running/cloud/clusterUtils.rst
+++ b/docs/running/cloud/clusterUtils.rst
@@ -222,9 +222,9 @@ exist yet, Toil will create it for you.
                         any worker instances created when using the -w flag.
                         This is an EBS volume.
   --nodeStorageOverrides NODESTORAGEOVERRIDES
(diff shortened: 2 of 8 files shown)


docs/running/cliOptions.rst
```diff
@@ -189,6 +189,11 @@ the logging module:
                         when they are launched in gigabytes. You may want to
                         set this if your jobs require a lot of disk space. The
                         default value is 50.
+  --nodeStorageOverrides NODESTORAGEOVERRIDES
+                        (optional) Comma-separated list of nodeType:nodeStorage that
```
adamnovak, line 193: We shouldn't say `(optional)` here; it has override in the name and none of the other optional options say `(optional)`.

docs/running/cloud/clusterUtils.rst
```diff
@@ -221,6 +221,11 @@ exist yet, Toil will create it for you.
                         Specify the size (in gigabytes) of the root volume for
                         any worker instances created when using the -w flag.
                         This is an EBS volume.
+  --nodeStorageOverrides NODESTORAGEOVERRIDES
+                        (optional) Comma-separated list of nodeType:nodeStorage that
```
adamnovak, line 225: We should also drop `(optional)` here.

src/toil/common.py
```diff
@@ -458,6 +460,11 @@ def _addOptions(addGroupFn, config):
                 help=("Specify the size of the root volume of worker nodes when they are launched "
                       "in gigabytes. You may want to set this if your jobs require a lot of disk "
                       "space. The default value is 50."))
+    addOptionFn('--nodeStorageOverrides', default=None,
+                help="(optional) Comma-separated list of nodeType:nodeStorage that are used to "
```
adamnovak, line 464: And here.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -118,6 +118,10 @@ def __init__(self, clusterName=None, zone=None, nodeStorage=50):
         self.clusterName = clusterName
         self._zone = zone
         self._nodeStorage = nodeStorage
+        self._nodeStorageOverrides = {}
+        for override in nodeStorageOverrides or []:
```
adamnovak, line 122: If the default was `[]` instead of `None` we wouldn't need this `or`.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -118,6 +118,10 @@ def __init__(self, clusterName=None, zone=None, nodeStorage=50):
...
+            nodeShape, storageOverride = override.split(':')
+            self._nodeStorageOverrides[nodeShape] = int(storageOverride)
```
adamnovak, line 124: Since the override string is coming straight from the user, we might want to have some error handling here so we report a badly written entry instead of just crashing with a stacktrace.

docs/running/cliOptions.rst
arostamianfar: Good point! Removed from here and elsewhere.

src/toil/provisioners/abstractProvisioner.py
arostamianfar: That's a good point! I decided to add it as a `checkFn` in common.py as it seems like most arg error checking is done there anyway.

Issue #3106, opened 2020-07-08
Leader passes around updates as mutated JobGraphs it tries to ensure single ownership of


adamnovak:
The leader handles updates by mutating a JobGraph and putting it in one of several places, and hoping that it only ever actually exists in one place at a time, so that multiple mutations in flight don't conflict.

We should rearchitect this; it causes test failures like this one: https://ucsc-ci.com/databiosphere/toil/-/jobs/54698

PR #3108, opened 2020-07-09
Deduplicate updated jobs

adamnovak:
As is, multiple JobGraphs for the same job ID can be updated in the same tick.
I don't think multiple job graphs for the same job ID are supposed to exist at
all.

I'm now deduplicating everything by job ID. The first update with the first
status will win.

Really we should change the whole system to pass around just the IDs of updated
things, and store a single source of truth for the job graph's state, so it is
impossible to have conflicting updates in flight.

This is a workaround for some test failures caused by #3106 in #3088.

### DIFF
--- a/src/toil/jobGraph.py
+++ b/src/toil/jobGraph.py
@@ -103,6 +103,34 @@ class JobGraph(JobNode):
 
     def __hash__(self):
         return hash(self.jobStoreID)
+        
+    def __eq__(self, other):
+        return (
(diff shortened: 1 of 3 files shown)


mr-c:
Any update on this? I think getting this reviewed and merged will help us merge the CWL v1.2.0-dev4 PR and others

adamnovak:
@mr-c Do you want to review this? Then we could merge it.

PR #3105, opened 2020-07-07
upgrade cwltool to cwl v1.2.0-dev4

mr-c:

PR #3114, opened 2020-07-17
More robust CWL filepaths and secondary files.

DailyDreaming:
Addresses #3105 

Toil filepaths were missing the import of secondary files (and thus the mapping).


mr-c:
Huzzah! Does this cause additional tests to pass?

DailyDreaming:
@mr-c Ha, maybe.  I guess we'll see.

DailyDreaming:
Specifically I've only tested #3105 and can verify it works for that workflow.  I'll test the two failing 1.1+ conformance tests with this fix.

DailyDreaming:
@mr-c Unfortunately, 213 and 236 are still failing.

PR #3123, opened 2020-07-29
[easy] Replace custom mkdir_p with the built in.

DailyDreaming:
"exist_ok" was added in Python 3.2, so I think we're okay.


DailyDreaming:
https://github.com/ComparativeGenomicsToolkit/cactus/pull/283

and

https://github.com/vgteam/toil-vg/pull/820

were merged and tests passed so I'm merging this now.

adamnovak:
@DailyDreaming I don't think tests actually pass. Don't we have to actually point at the new Cactus version?

PR #3126, opened 2020-07-29
Wrap Kubernetes watches to fix #3125

### LINKED ISSUE #3125: Dropped Kubernetes connection during a watch can crash the KubernetesBatchSystem
We have retries in place in other parts of the Kubernetes batch system, but as shown in https://github.com/DataBiosphere/toil/pull/3124#issuecomment-665954608 (which is about https://ucsc-ci.com/databiosphere/toil/-/jobs/56257), if our connection to Kubernetes goes wonky while we're watching for events, the whole batch system can crash.

We should probably just treat it the same as a watch timing out without reporting anything.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-591)
┆Issue Number: TOIL-591

adamnovak:
This should fix #3125 by wrapping calls to the Kubernetes watch API with something that will suppress network-y errors that happen inside the `stream()` generator itself, so that a disconnect during a stream operation doesn't crash the whole workflow. Instead, it will just end the watch early (rather than starting a new watch for any remaining timeout, which is more work).

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -69,6 +69,7 @@ def retryable_kubernetes_errors(e):
     exceptions thrown by Kubernetes. 
     """
     if isinstance(e, urllib3.exceptions.MaxRetryError) or \
+       isinstance(e, urllib3.exceptions.ProtocolError) or \
         isinstance(e, ApiException):
         return True
     return False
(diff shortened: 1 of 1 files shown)


src/toil/batchSystems/kubernetes.py
```diff
@@ -286,6 +287,53 @@ def _try_kubernetes_expecting_gone(self, method, *args, **kwargs):
...
+                # For everything the watch stream gives us
+                try:
+                    # Show the item to user code
+                    yield item
+                except Exception as e:
+                    # If we get an error from user code, skip our catch around
+                    # the Kubernetes generator.
+                    userError = True
+                    raise
+        except Exception as e:
```
DailyDreaming, line 320: Hmmm... this seems to be equivalent to:

```
        except Exception as e:
            if retryable_kubernetes_errors(e):
                    # This is just cloud weather.
                    # TODO: We will also get an APIError if we just can't code good against Kubernetes. So make sure to warn.
                    logger.warning("Received error from Kubernetes watch stream: %s", e)
                    # Just end the watch.
                    return
            # Something actually weird is happening.
            raise
```

Unless `if UserError` had some special log message here?  I'm not actually completely sure what `userError` is doing here, unless I was overzealous and reviewed this PR too early.

Issue #3129, opened 2020-07-31
Log file already exists


ionox0:
We are experiencing an issue where a chained CWL job is trying to create a duplicate link to one of the log files. 

The offending portion of the CWL workflow has these two steps:
```
  find_covered_intervals:
    run: ../../cwl_tools/gatk/FindCoveredIntervals.cwl
    in:
      bams: bams
      reference_sequence: reference_fasta
      ignore_misencoded_base_qualities: fci__basq_fix
      out:
        valueFrom: ${return inputs.patient_id + '.fci.list'}
    out: [fci_list]

  list2bed:
    run: ../../cwl_tools/python/list2bed.cwl
    in:
      input_file: find_covered_intervals/fci_list
      output_filename:
        valueFrom: ${return inputs.input_file.basename.replace('.list', '.bed.srt')}
    out: [output_file]
```

The result is the following error:
```
Traceback (most recent call last):
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/site-packages/toil/statsAndLogging.py", line 153, in statsAndLoggingAggregator
    break
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/site-packages/toil/jobStores/fileJobStore.py", line 549, in readStatsAndLogging
    callback(fH)
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/site-packages/toil/statsAndLogging.py", line 146, in callback
    message='Received Toil worker log. Disable debug level logging to hide this output')
  File "miniconda3/envs/ACCESS_1.3.40/lib/python2.7/site-packages/toil/statsAndLogging.py", line 109, in writeLogFiles
    os.symlink(os.path.relpath(fullName, path), name)
OSError: [Errno 17] File exists
```

When I put some debugging in the code to see the file name it appears as so:
`file:--workflows-access_workflows-v1-pipeline_1.3.39-ACCESS--Pipeline-cwl_tools-python-list2bed.cwl_list2bed_kind--file_work_000.log`

And I can see that indeed that file is already present in the `--logDir`

[...]

PR #3130, opened 2020-07-31
potential fix for log file being linked more than once

ionox0:
Potential fix for https://github.com/DataBiosphere/toil/issues/3129

### DIFF
--- a/src/toil/statsAndLogging.py
+++ b/src/toil/statsAndLogging.py
@@ -142,7 +142,8 @@ class StatsAndLogging( object ):
             # There are chained jobs in this output - indicate this with a symlink
             # of the job's name to this file
             name = createName(path, alternateName, extension, failed)
-            os.symlink(os.path.relpath(fullName, path), name)
+            if not os.path.exists(name):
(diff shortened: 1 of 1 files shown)


adamnovak:
The tests for this hit https://github.com/DataBiosphere/toil/issues/3127 again.

adamnovak:
@ionox0 Did this actually fix #3129?

ionox0:
For our workflow it completes successfully after this change, however we also normally use `--disableChaining` to avoid this issue

Can you help us to understand the purpose of using symlinks for log files so we can look into this?

DailyDreaming:
@ionox0 Adam may have more context, but I'm not sure symlinking is necessary for log files.  Looking at the original PR I don't believe they explained it: https://github.com/DataBiosphere/toil/pull/1304/files

I'm guessing, but it may simply be a reference when searching a directory for that worker's log file.  So when multiple workers share a log file, a bunch of symlinks are created, one for each worker's name, as a way to find that worker's log file?  I'm guessing that's how it reaps the correct log file at the end when it sees that a worker has failed.  If that's the case, I would support reworking how we log.

PR #3145, opened 2020-08-20
Use boto3 to create our leader node.

DailyDreaming:


adamnovak:
@DailyDreaming Is this ready to review or should it be a draft?

adamnovak (approved):
I'm not really a boto3 expert, but this looks like it plausibly works.

Issue #2707, opened 2019-06-06
Thread-safe file import on S3


glennhickey:
I often find that importing large files at the beginning of my workflow to be a large time bottleneck.  For instance, if I want to map a pair of fastq files from the EBI FTP site, it can take several hours import each one using `toil.importFile()`.  Same issue for, say, the phased chromosome 1000 Genomes VCFs. 

This time could be cut in half if I could import the two fastq files at once.  This actually works fine on a local job store, but does not work on a S3 job store.  If I recall, it's because the boto session object is not thread safe.  But I don't think it's a major change to use multiple sessions (or maybe there's a thread-safe API somewhere?).  It would be a big speedup for importing large files. 

Coupling this with some kind of asynchronous / batch import interface would be even better.

adamnovak:
Maybe we can solve this, #2704, and #2706 all at the same time by moving all our S3 uploads/downloads to boto3: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#client

It has its own logic for multithreaded uploads and downloads, and for uploads and downloads to and from streams.

Though to really solve this issue, we probably have to make importFile asynchronous. There's no way to make S3 itself download from a URL, and I don't think there's an easy way to do byte range reads from a remote URL to support a boto3-managed multipart upload.

glennhickey:
+1 for the boto3 and for asynchronous importFile!

I was testing this out with a quick (and bad) asynchronous wrapper for
importFile

https://github.com/vgteam/toil-vg/blob/master/src/toil_vg/vg_common.py#L956-L1027

but it never worked because the current implementation of importFile is not
thread-safe.

DailyDreaming:
@adamnovak Reference is here: https://boto3.amazonaws.com/v1/documentation/api/1.14.31/guide/session.html#multithreading-or-multiprocessing-with-sessions

DailyDreaming:
@w-gao @adamnovak Reference is here: https://boto3.amazonaws.com/v1/documentation/api/1.14.31/guide/session.html#multithreading-or-multiprocessing-with-sessions

Issue #3140, opened 2020-08-18
TypeError during sock.sendall("CONNECT %s HTTP/1.0\r\n" % host)


fbemm:
I am trying to launch a cluster with:

`toil launch-cluster --zone eu-central-1c --keyPairName aws --leaderNodeType t2.micro cactus --logDebug`

But keep getting the following error:

```
Traceback (most recent call last):
  File "/cactus/env/bin/toil", line 11, in <module>
    sys.exit(main())
  File "/cactus/env/lib/python3.7/site-packages/toil/utils/toilMain.py", line 25, in main
    module.main()
  File "/cactus/env/lib/python3.7/site-packages/toil/utils/toilLaunchCluster.py", line 154, in main
    awsEc2ExtraSecurityGroupIds=config.awsEc2ExtraSecurityGroupIds)
  File "/cactus/env/lib/python3.7/site-packages/toil/provisioners/aws/awsProvisioner.py", line 153, in launchCluster
    profileArn = kwargs.get('awsEc2ProfileArn') or self._getProfileArn()
  File "/cactus/env/lib/python3.7/site-packages/toil/provisioners/aws/awsProvisioner.py", line 95, in wrapper
    return f(*args, **kwargs)
  File "/cactus/env/lib/python3.7/site-packages/toil/provisioners/aws/awsProvisioner.py", line 652, in _getProfileArn
    iamRoleName = self._ctx.setup_iam_ec2_role(role_name=_INSTANCE_PROFILE_ROLE_NAME, policies=policy)
  File "/cactus/env/lib/python3.7/site-packages/toil/lib/context.py", line 535, in setup_iam_ec2_role
    "Action": ["sts:AssumeRole"]}
  File "/cactus/env/lib/python3.7/site-packages/boto/iam/connection.py", line 1155, in create_role
    return self.get_response('CreateRole', params)
  File "/cactus/env/lib/python3.7/site-packages/boto/iam/connection.py", line 85, in get_response
    response = self.make_request(action, params, path, verb)
  File "/cactus/env/lib/python3.7/site-packages/boto/connection.py", line 1117, in make_request
    return self._mexe(http_request)
  File "/cactus/env/lib/python3.7/site-packages/boto/connection.py", line 914, in _mexe
    self.is_secure)
  File "/cactus/env/lib/python3.7/site-packages/boto/connection.py", line 705, in get_http_connection
    return self.new_http_connection(host, port, is_secure)
[...]

PR #3153, opened 2020-08-21
Move AWS job store uploads/downloads to boto3

### LINKED ISSUE #2704: Use boto3 API for s3 uploads/downloads
Currently, uploading/downloading large files to/from the S3 jobstore is slow as it's single threaded. The new `boto3` API actually has a [nice interface](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/s3/transfer.html) with multithreaded uploads/downloads that can dramatically speed up the process especially for large files. It also simplifies the logic as we no longer need to add custom logic to specify the chunkSize ourselves.

Related tutorial: https://medium.com/@niyazi_erd/aws-s3-multipart-upload-with-python-and-boto3-9d2a0ef9b085

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-378)
┆Issue Number: TOIL-378

w-gao:
This PR moves uploads and downloads for the AWS job store to boto3. ~Note that a new job store (aws3) is introduced to prevent workflows from breaking if they depend on the "private" methods in the original AWS job store.~

This can possibly close #2704 and `importFile()` _should_ be thread-safe with the current implementation according to the [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-and-multiprocessing). Although, `importFile` has not been made asynchronous as suggested in #2707 since that would make a major update to the `abstractJobStore` API. 

~See [here](https://github.com/DataBiosphere/toil/compare/699cb4b8b27d6f588fae455a47c3d710efa12b53...issues/2707-move-to-boto3) for a more informative review of the changes to the AWS job store.~ I will be adding unit tests, though it would be helpful if I could get some feedback first. 

Related tickets: #2704, #2707


adamnovak:
One thing that came up at stand-up today is #3140. If we can rip out all of Boto and move everything to Boto3, including the provisioner, we can work properly behind a proxy. But the stuff we've been leaving on Boto (the provisioner and SimpleDB) is also the stuff we're planning to kill/completely redo anyway, so it's not clear we want to spend the time porting what's there to Boto3.

src/toil/jobStores/aws3/jobStore.py
```diff
@@ -0,0 +1,1626 @@
...
+        url = self.s3_client.generate_presigned_url('get_object',
+                                                    Params={'Bucket': self.filesBucket.name,
+                                                            'Key': compat_bytes(jobStoreFileID),
+                                                            'VersionId': info.version},
+                                                    ExpiresIn=self.publicUrlExpiration.total_seconds())
+
+        # boto doesn't properly remove the x-amz-security-token parameter when
+        # query_auth is False when using an IAM role (see issue #2043). Including the
+        # x-amz-security-token parameter without the access key results in a 403,
+        # even if the resource is public, so we need to remove it.
```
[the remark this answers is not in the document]
w-gao: It seems like boto3 includes the `x-amz-security-token` _with_ the access key, but the link still results in a 403. Without the parameter works fine.

w-gao:
Marked as Draft because I want to perform more tests on encryption. 

A few notes:
- Only S3 related operations have been updated to use boto3. Like Adam said, boto2 is still used in the provisioner and SimpleDB. 
- Most of the retries for s3 have been removed since boto3 has its own [retry mechanism](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html) for client errors

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1276,19 +1277,7 @@ def testSDBDomainsDeletedOnFailedJobstoreBucketCreation(self):
...
-                except S3ResponseError as e:
-                    # The actual HTTP code of the error is in status.
-                    # See https://github.com/boto/boto/blob/91ba037e54ef521c379263b0ac769c66182527d7/boto/exception.py#L77-L80
-                    # See also: https://github.com/boto/boto/blob/91ba037e54ef521c379263b0ac769c66182527d7/boto/exception.py#L154-L156
-                    if e.status == 404:
-                        # The bucket doesn't exist; maybe a failed delete actually succeeded.
-                        pass
-                    else:
-                        raise
+                client.delete_bucket(Bucket=bucketName)
```
adamnovak, line 1280: This is now safe when the bucket is gone already?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -63,30 +65,37 @@
...
+        'assume-role').cache = botocore.credentials.JSONFileCache()
+    return boto3.Session(botocore_session=botocore_session)
+
+
 # Make sure to use credential caching when talking to Amazon via boto3
 # See https://github.com/boto/botocore/pull/1338/
-botocore_session = botocore.session.get_session()
-botocore_session.get_component('credential_provider').get_provider('assume-role').cache = botocore.credentials.JSONFileCache()
-boto3_session = boto3.Session(botocore_session=botocore_session)
+boto3_session = get_boto3_session()
```
adamnovak, line 86: Why do we call `get_boto3_session()` again in some places, but use this module-level session in others?

src/toil/test/jobStores/jobStoreTest.py
w-gao: Good catch. It seems like botocore does throw an error if the bucket doesn't exist, which wasn't obvious from the documentation. I shall create an update along with a fix for the `JobFileStoreTest.testJobFileStore` travis test failure.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1212,34 +1218,35 @@ def readFrom(self, readable):
...
 
-                            for attempt in retry_s3():
-                                with attempt:
-                                    log.debug('Attempting to complete upload...')
-                                    completed = upload.complete_upload()
-                                    log.debug('Completed upload object of type %s: %s', str(type(completed)), repr(completed))
-                                    info.version = completed.version_id
-                                    log.debug('Completed upload with version %s', str(info.version))
+                            log.debug('Attempting to complete upload...')
+                            completed = client.complete_multipart_upload(
```
DailyDreaming, line 1234: The docs recommend some error handling here: https://ibm.github.io/ibm-cos-sdk-python/reference/services/s3.html#S3.Client.complete_multipart_upload

Could we add some retries for 404, 500, 502, 503, and 504?

src/toil/jobStores/aws/jobStore.py
w-gao: `get_boto3_session()` is called again in places that may need to be on its own thread (uploads and downloads). Since, according to the [boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-and-multiprocessing),

> It is recommended to create a resource instance for each thread / process in a multithreaded or multiprocess application rather than sharing a single instance among the threads / processes.

and

> ... each thread would have its own Boto3 session and its own instance of the S3 resource.

The module-level session is only used by the module-level client, `s3_boto3_client`, which is only used in the static method `__getBucketRegion()` and when destroying a job store. This should be fine since low-level clients are thread safe.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1260,32 +1267,34 @@ def readFrom(self, readable):
...
 
-                        for attempt in retry_s3():
-                            with attempt:
-                                log.debug('Uploading single part of %d bytes', dataLength)
-                                assert dataLength == key.set_contents_from_file(fp=buf,
-                                                                                headers=headers)
-                                log.debug('Upload received version %s', str(key.version_id))
-                        info.version = key.version_id
+                        log.debug('Uploading single part of %d bytes', dataLength)
+                        client.upload_fileobj(Bucket=bucket_name, Key=compat_bytes(info.fileID),
```
DailyDreaming, line 1281: I would also keep the retries here for 404, 500, 502, 503, 504.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1186,24 +1185,31 @@ def readFrom(self, readable):
...
-                                for attempt in retry_s3():
-                                    with attempt:
-                                        log.debug('Uploading part %d of %d bytes', part_num + 1, len(buf))
-                                        upload.upload_part_from_file(fp=BytesIO(buf),
-                                                                     # part numbers are 1-based
-                                                                     part_num=part_num + 1,
-                                                                     headers=headers)
-                                
+                                log.debug('Uploading part %d of %d bytes', part_num + 1, len(buf))
+                                part = client.upload_part(Bucket=bucket_name, Key=compat_bytes(info.fileID),
```
DailyDreaming, line 1207: I would keep the retries here for 404, 500, 502, 503, 504.  I would also a comment `TODO` to include the `Content-MD5` header in the future: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.complete_multipart_upload

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1394,25 +1394,21 @@ def writeTo(self, writable):
...
-                        headers = info._s3EncryptionHeaders()
-                        key = info.outer.filesBucket.get_key(compat_bytes(info.fileID), validate=False)
-                        for attempt in retry_s3():
-                            with attempt:
-                                key.get_contents_to_file(writable,
-                                                         headers=headers,
-                                                         version_id=info.version)
+                        headerArgs = info._s3EncryptionArgs()
+                        obj = info.outer.filesBucket.Object(compat_bytes(info.fileID))
+                        obj.download_fileobj(writable, ExtraArgs={'VersionId': info.version, **headerArgs})
```
DailyDreaming, line 1399: I would keep 404, 500, 502, 503, 504 retries for this.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1448,10 +1444,9 @@ def delete(self):
                             compat_bytes(self.fileID),
                             expected_values=['version', self.previousVersion])
                 if self.previousVersion:
-                    for attempt in retry_s3():
-                        with attempt:
-                            store.filesBucket.delete_key(key_name=compat_bytes(self.fileID),
-                                                         version_id=self.previousVersion)
+                    store.s3_client.delete_object(Bucket=store.filesBucket.name,
```
DailyDreaming, line 1447: I would keep 404, 500, 502, 503, 504 retries for this.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1333,39 +1322,31 @@ def _prepareTestFile(self, bucket, size=None):
...
 
     def _cleanUpExternalStore(self, bucket):
         try:
-            for key in bucket.list():
-                key.delete()
+            bucket.objects.all().delete()
             bucket.delete()
-        finally:
-            bucket.connection.close()
+        except:
```
DailyDreaming, line 1348: You no longer need the `try` here and if bucket deletion fails, it should error.

Formerly this was a `try` `finally`, which only gaurantees that the `finally` clause runs in the event of an error, but stills errors.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1333,39 +1322,31 @@ def _prepareTestFile(self, bucket, size=None):
...
-        try:
-            return s3.create_bucket(bucket_name='import-export-test-%s' % uuid.uuid4(),
-                                    location=region_to_bucket_location(self.awsRegion()))
-        except:
-            with panic(log=logger):
-                s3.close()
+        from toil.jobStores.aws.jobStore import get_boto3_session
+        resource = get_boto3_session().resource('s3', region_name=self.awsRegion())
+        bucket = resource.Bucket('import-export-test-%s' % uuid.uuid4())
+        bucket.create(CreateBucketConfiguration={'LocationConstraint': region_to_bucket_location(self.awsRegion())})
```
DailyDreaming, line 1340: Probably needs a 500, 502, 503, 504 retry here.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1333,39 +1322,31 @@ def _prepareTestFile(self, bucket, size=None):
             return url
         read_type = 'r' if USING_PYTHON2 else 'rb'
         with open('/dev/urandom', read_type) as readable:
-            if USING_PYTHON2:
-                bucket.new_key(fileName).set_contents_from_string(readable.read(size))
-            else:
-                bucket.new_key(fileName).set_contents_from_string(str(readable.read(size)))
-        return url, hashlib.md5(bucket.get_key(fileName).get_contents_as_string()).hexdigest()
+            bucket.put_object(Key=fileName, Body=str(readable.read(size)))
```
DailyDreaming, line 1325: Probably needs a 500, 502, 503, 504 retry here.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1333,39 +1322,31 @@ def _prepareTestFile(self, bucket, size=None):
             return url
         read_type = 'r' if USING_PYTHON2 else 'rb'
```
DailyDreaming, line 1323: This can be removed and amended below.

src/toil/test/jobStores/jobStoreTest.py
DailyDreaming, line 1280: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1237,23 +1240,21 @@ def testSDBDomainsDeletedOnFailedJobstoreBucketCreation(self):
...
             testJobStoreUUID = str(uuid.uuid4())
+            bucketName = 'domain-test-' + testJobStoreUUID + '--files'
             # Create the bucket at the external region
-            s3 = S3Connection()
-            for attempt in retry_s3(delays=(2, 5, 10, 30, 60), timeout=600):
-                with attempt:
-                    bucket = s3.create_bucket('domain-test-' + testJobStoreUUID + '--files',
-                                              location=externalAWSLocation)
+            client = get_boto3_session().client('s3', region_name=externalAWSLocation)
+            client.create_bucket(Bucket=bucketName,
```
DailyDreaming, line 1255: Probably needs a 500, 502, 503, 504 retry here.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1237,23 +1240,21 @@ def testSDBDomainsDeletedOnFailedJobstoreBucketCreation(self):
         different region with the same name.
         """
         from boto.sdb import connect_to_region
-        from boto.s3.connection import Location, S3Connection
-        from boto.exception import S3ResponseError
         from toil.jobStores.aws.jobStore import BucketLocationConflictException
-        from toil.jobStores.aws.utils import retry_s3
-        externalAWSLocation = Location.USWest
+        from toil.jobStores.aws.jobStore import get_boto3_session
+        externalAWSLocation = 'us-west-1'
```
DailyDreaming, line 1245: Nit: This could probably be set in the main toil test config and imported to ensure all of our tests are in the same region.

src/toil/jobStores/aws/utils.py
```diff
@@ -302,18 +299,18 @@ def copyKeyMultipart(srcBucketName, srcKeyName, srcKeyVersion, dstBucketName, ds
...
     object_summary.wait_until_exists(**destEncryptionArgs)
 
     # Unfortunately, boto3's managed copy doesn't return the version
     # that it actually copied to. So we have to check immediately
     # after, leaving open the possibility that it may have been
     # modified again in the few seconds since the copy finished. There
     # isn't much we can do about it.
-    info = boto3.client('s3').head_object(Bucket=dstObject.bucket_name, Key=dstObject.key,
-                                          **destEncryptionArgs)
+    info = resource.meta.client.head_object(Bucket=dstObject.bucket_name, Key=dstObject.key, **destEncryptionArgs)
```
DailyDreaming, line 310: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
...
+def uploadFile(readable, resource, bucketName, fileID, headerArgs=None, partSize=50 << 20):
+    """
+    Upload a readable object to s3, using multipart uploading if applicable.
+
+    :param readable: a readable stream or a file path to upload to s3
+    :param S3.Resource resource: boto3 resource
+    :param str bucketName: name of the bucket to upload to
+    :param str fileID: the name of the file to upload to
+    :param dict headerArgs: http headers to use when uploading - generally used for encryption purposes
+    :param int partSize: max size of each part in the multipart upload, in bytes
```
DailyDreaming, line 231: Including this typing in the args would be awesome.  For example:

```
from typing import Optional  # place at top of file

uploadFile(readable, resource, bucketName: str, fileID: str, headerArgs: Optional[dict] = None, partSize: int = 50 << 20):
```

The objects are a little trickier and would have to be hunted down.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
...
-                version = upload.complete_upload().version_id
-    return version
+        client.upload_fileobj(Fileobj=readable, Bucket=bucketName, Key=compat_bytes(fileID),
+                              ExtraArgs=headerArgs, Config=config)
+
+    # Wait until the object exists before calling head_object
+    object_summary = resource.ObjectSummary(bucketName, compat_bytes(fileID))
+    object_summary.wait_until_exists(**headerArgs)
+
+    info = client.head_object(Bucket=bucketName, Key=compat_bytes(fileID), **headerArgs)
```
DailyDreaming, line 255: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
...
+    )
+    if isinstance(readable, str):
+        client.upload_file(Filename=readable, Bucket=bucketName, Key=compat_bytes(fileID),
+                           ExtraArgs=headerArgs, Config=config)
     else:
-        for attempt in retry_s3():
-            with attempt:
-                version = upload.complete_upload().version_id
-    return version
+        client.upload_fileobj(Fileobj=readable, Bucket=bucketName, Key=compat_bytes(fileID),
```
DailyDreaming, line 248: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
...
-            version = chunkedFileUpload(f, bucket, fileID, file_size, headers, partSize)
-    for attempt in retry_s3():
-        with attempt:
-            key = bucket.get_key(compat_bytes(fileID),
-                                 headers=headers,
-                                 version_id=version)
-    assert key.size == file_size
+
+    version = uploadFile(localFilePath, resource, bucketName, fileID, headerArgs, partSize)
+    info = client.head_object(Bucket=bucketName, Key=compat_bytes(fileID), VersionId=version, **headerArgs)
```
DailyDreaming, line 213: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
...
-        with open(localFilePath, 'rb') as f:
-            version = chunkedFileUpload(f, bucket, fileID, file_size, headers, partSize)
-    for attempt in retry_s3():
-        with attempt:
-            key = bucket.get_key(compat_bytes(fileID),
-                                 headers=headers,
-                                 version_id=version)
-    assert key.size == file_size
+
+    version = uploadFile(localFilePath, resource, bucketName, fileID, headerArgs, partSize)
```
DailyDreaming, line 212: Probably needs a 404, 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/utils.py
```diff
@@ -193,78 +190,80 @@ def fileSizeAndTime(localFilePath):
     return file_stat.st_size, file_stat.st_mtime
 
 
-def uploadFromPath(localFilePath, partSize, bucket, fileID, headers):
+def uploadFromPath(localFilePath, resource, bucketName, fileID, headerArgs=None, partSize=50 << 20):
```
DailyDreaming, line 193: Including this typing in the args would be awesome.  For example:

```
from typing import Optional  # place at top of file

uploadFile(localFilePath: str, resource, bucketName: str, fileID: str, headerArgs: Optional[dict] = None, partSize: int = 50 << 20):
```

The objects are a little trickier and would have to be hunted down.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1559,21 +1547,24 @@ def _delete_domain(self, domain):
...
+                for u in uploads:
+                    s3_boto3_client.abort_multipart_upload(Bucket=bucket.name, Key=u["Key"], UploadId=u["UploadId"])
+
+            bucket.objects.all().delete()
+            bucket.object_versions.delete()
+            bucket.delete()
+        except s3_boto3_client.exceptions.NoSuchBucket:
+            pass
+        except ClientError as e:
+            if e.response.get('Error', {}).get('Code') != 'NoSuchBucket':
```
DailyDreaming, line 1566: Would this be a '404'?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1559,21 +1547,24 @@ def _delete_domain(self, domain):
...
-                    bucket = s3_boto3_resource.Bucket(compat_bytes(b.name))
-                    bucket.objects.all().delete()
-                    bucket.object_versions.delete()
-                    bucket.delete()
-                except s3_boto3_resource.meta.client.exceptions.NoSuchBucket:
-                    pass
-                except S3ResponseError as e:
-                    if e.error_code != 'NoSuchBucket':
-                        raise
+    def _delete_bucket(self, bucket):
```
DailyDreaming, line 1550: Probably needs a 500, 502, 503, 504 retry here.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1504,27 +1496,23 @@ def __repr__(self):
...
-                return self.versionings[status['Versioning']] if status else False
+        status = self.s3_resource.BucketVersioning(bucket_name).status
+        return self.versionings[status] if status else False
 
-    def __getBucketRegion(self, bucket):
-        for attempt in retry_s3():
-            with attempt:
-                return bucket_location_to_region(bucket.get_location())
+    @staticmethod
+    def __getBucketRegion(bucket_name):
```
DailyDreaming, line 1513: Potentially rename this to just `def getBucketRegion(bucket_name)` without the `__` as it's not appropriate here imo.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -410,126 +421,90 @@ def getEmptyFileStoreID(self, jobStoreID=None, cleanup=False, basename=None):
...
-                                       (keyName, bucketName))
-            elif existing is False:
-                if key is not None:
-                    raise RuntimeError("Key '%s' exists in bucket '%s'." %
-                                       (keyName, bucketName))
-            elif existing is None:
-                pass
+            obj.load()
+        except ClientError as e:
+            if e.response.get('Error', {}).get('Code') == '404':
```
DailyDreaming, line 491: Are you sure that this shouldn't be:

```
if e.response.get('ResponseMetadata', {}).get('HTTPStatusCode') == '404':
```

?

I would expect a human readable error from `e.response.get('Error', {}).get('Code')`, like `BucketNotFound`.

See: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html

DailyDreaming (changes requested):
Alright, I did a first pass.  I believe we still need retries on portions of the code that were removed.  I'm not confident in boto3 retrying and believe that these still need to be caught.

Overall, it looks like a great first refactor though.  Thanks for taking this one on!

w-gao:
@DailyDreaming Thanks for the review! Will address the review comments in a new PR.

adamnovak:
That makes sense. Maybe there should be a comment in the source
explaining why we need to make fresh ones when we might be running at
the same time as other code. This could get simplified away by
somebody who doesn't see a good reason for it to be the way it is.

src/toil/test/jobStores/jobStoreTest.py
w-gao: I'm not entirely sure what this is referring to.

PR #3171, opened 2020-09-01
Add WDL built-in tests 'ceil' and 'floor'.

DailyDreaming:
Eventually I'd like to use something like: https://github.com/DailyDreaming/wdl-conformance-tests

And make it general and robust... but in the meantime, this PR is probably how we should add wdl/json file toil tests in the meantime.

In general, "File" outputs are probably how we should test WDL files for the WDL built-ins until we implement support for cromwell's `-m` option.


src/toil/test/wdl/standard_library/ceil.wdl
```diff
@@ -0,0 +1,16 @@
+workflow ceilWorkflow {
+  Float num
+  call get_ceil { input: num=ceil(num) }
```
w-gao, line 3: I wonder if the `ceil` call should be inside `get_ceil`

w-gao:
Thanks for this! Just to clarify, we would have wdl function tests in `toilwdlTest.py` and the actual workflow tests in `builtinTest.py`?

src/toil/test/wdl/standard_library/ceil.wdl
DailyDreaming: Testing it inside of the call might even be a different test and we could include both.  I think I'll add it as a separate test this time, but as you add in the future, really only one should be necessary since we want to test the function and not necessarily permutations of the syntax.

Cromwell returns a 12.0 (float) in this case and toil returns a 12 (int), which means that they're not really obeying their spec it seems: https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#int-floorfloat-int-ceilfloat-and-int-roundfloat and I imagine we'll find other cases where this occurs as well.  We'll also see the holes in toil as well going through this and how large those holes are.

DailyDreaming:
@w-gao If you want to add a new Test Class inside of `builtinTest.py` called something like `WdlStandardLibraryFunctionsTest` and move the appropriate wdl function unit tests from `toilwdlTest.py`, I think that that would make a lot of sense.  Then the first file could be renamed to `integrationTest.py` or `workflowsTest.py` or something to better separate things out.

src/toil/test/wdl/builtinTest.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+import subprocess
+import shutil
+import uuid
+
+from toil.version import exactPython
+from toil.test import ToilTest
+
+
+class WdlStandardLibraryWorkflowsTest(ToilTest):
```
adamnovak, line 11: How big is the WDL standard library? Are we going to end up with 50 test functions in this class? Should we maybe (later?) switch to 50 test files instead? Or find a way to condense those 50 functions down as much as possible, so it's more like 50 asserts?

src/toil/test/wdl/builtinTest.py
```diff
@@ -0,0 +1,51 @@
...
+            shutil.rmtree(self.output_dir)
+        unittest.TestCase.tearDown(self)
+
+    def test_ceil_as_input(self):
+        wdl_file = os.path.abspath('src/toil/test/wdl/standard_library/ceil_as_input.wdl')
+        json_file = os.path.abspath('src/toil/test/wdl/standard_library/ceil.json')
+        subprocess.check_call([exactPython, self.program, wdl_file, json_file, '-o', self.output_dir])
+        output = os.path.join(self.output_dir, 'the_ceiling.txt')
+        with open(output, 'r') as f:
+            assert float(f.read()) == 12
```
adamnovak, line 39: When we get more functions to test, is there a way to turn this into one call that's like `test_wdl_function('ceil', 12)`? Otherwise we'll be copy-pasting this many times.

src/toil/test/wdl/standard_library/ceil_as_command.wdl
```diff
@@ -0,0 +1,16 @@
...
+
+task get_ceil {
+  Float num
+
+  command {
+    echo ${ceil(num)} > the_ceiling.txt
+  }
+
+ output {
+    File the_ceiling = 'the_ceiling.txt'
```
adamnovak, line 14: This wants to be a standardized/generic `output.txt` or something so it can be the same file to read the result from for all these WDL tests.

src/toil/test/wdl/builtinTest.py
adamnovak: Maybe:

```
with get_wdl_test_output('ceil') as f:
    assert float(f.read()) == 12
```

DailyDreaming:
@adamnovak Those are all great suggestions.  Will implement.  Thanks for the review!

DailyDreaming:
Only https://ucsc-ci.com/databiosphere/toil/-/jobs/59283 left which should be unaffected by this PR so I'm merging this.  Might be hanging as it usually takes ~85 minutes and it's more than twice that now.

src/toil/test/wdl/standard_library/ceil.wdl
w-gao: I've been testing with Cromwell and it seems like Cromwell's `ceil` does return an Int. In this case, I think Cromwell's behavior (returning 12.0) is actually correct- we're expecting a Float `num` output. I don't know if it's worth it to investigate how Toil does type casting and make it behave exactly like Cromwell. It doesn't matter here, but I suppose there are instances where typing is important.

Issue #3174, opened 2020-09-01
Kubernetes getUpdatedBatchJob watches each job in sequence and loops over the timeout, potentially exceeding maxWait


adamnovak:
I think this might be the underlying cause of  #3104

https://github.com/DataBiosphere/toil/blob/ab42ab9e7ee078c3d27395e2a8b0041e4d7b8c72/src/toil/batchSystems/kubernetes.py#L688-L693

What we *want* to do is watch once for any pod in any of our jobs completing, with one timeout. But what we really do is, for each job, watch its pod for the duration of the timeout. If nothing is happening with most jobs, we'll get stuck here watching every job we have have for the timeout's amount of time, until we hit one that happens to, I think, not only be done, but change state while we're watching it.

We also check enableWatching twice for no apparent reason.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-624)
┆Issue Number: TOIL-624

adamnovak:
We still want to fix the watch code to not have this problem, so we can turn it on again.

Issue #3104, opened 2020-07-06
Kubernetes batch system takes a prohibitively long time to collect completed jobs


adamnovak:
My workflow has been sitting for hours with a Kubernetes namespace full of ~1200 completed jobs. It seems to only be able to process one completed job every couple of minutes, which is going to make my workflow take an unacceptably long time to run.

We need to change how the Kubernetes batch system collects completed jobs so that it is faster.

Here's the log: [kube-log.txt](https://github.com/DataBiosphere/toil/files/4880668/kube-log.txt)

Here's an exerpt showing the problem:

```
[2020-07-06T17:09:01+0000] [MainThread] [I] [toil.leader] 0 jobs are running, 1364 jobs are issued and waiting to run
[2020-07-06T17:09:03+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 0e8ad5fc-85fe-409f-83cd-bbfeebebd1de
[2020-07-06T17:11:04+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 7d257e52-c5b4-49f4-93ad-389c3dad10db
[2020-07-06T17:13:06+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 8acafd49-4fed-4f06-9cae-11b512e0ad36
[2020-07-06T17:15:05+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' e99416a0-fc51-4945-94ac-af20a43aa2c6
[2020-07-06T17:17:06+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 325b4cf3-1811-4f96-ab37-fcb56bfaead8
[2020-07-06T17:19:09+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' e7c465e2-82a7-4324-a6e0-eb3aa0aa42d7
[2020-07-06T17:21:09+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 63e86b64-4589-499d-b13b-e13ea5872740
[2020-07-06T17:23:08+0000] [MainThread] [I] [toil.leader] Job ended: 'run_construct_region_graph' 2d2bb6ba-f366-49ed-938a-3a8d47232fe7
[...]

PR #3175, opened 2020-09-01
Turn off watch code in Kubernetes

adamnovak:
It may be dramatically slowing job collection. This will work around #3174 and might help with #3104.

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -179,7 +179,8 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         self.awsSecretName = os.environ.get("TOIL_AWS_SECRET_NAME", None)
 
         # Set this to True to enable the experimental wait-for-job-update code
-        self.enableWatching = True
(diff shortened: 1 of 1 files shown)


adamnovak:
Turning this off does nothing for #3104 that I can tell.

src/toil/batchSystems/kubernetes.py
```diff
@@ -179,7 +179,8 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
         self.awsSecretName = os.environ.get("TOIL_AWS_SECRET_NAME", None)
 
         # Set this to True to enable the experimental wait-for-job-update code
-        self.enableWatching = True
+        # TODO: Make this an environment variable?
```
DailyDreaming, line 182: Mark Diekhans had a good comment I think:

```
A good approach is to have every config available in three place:
command line, environment, and a config file.

Although environment is mostly for desperation; it is often leads
to "it works for me".

So the general priority hierarchy would be:

  1st: command line
  2nd: environment variables
  3rd: config file
  4th: defaults
```

PR #3180, opened 2020-09-03
Add WDL builtin test 'round'.

w-gao:
Address #3152.

`round()` is not added to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py because we are using the Python built-in function. 

Note that `src/toil/test/wdl/toilwdlTest.py` still contains tests for WDL functions. In the future these tests will be moved to `WdlStandardLibraryFunctionsTest` as we add in the corresponding workflow tests. I'm not moving them all at once because there are tests that depend on the test files in `toilwdlTest.py` and I don't want to mess with those in this PR.


DailyDreaming (approved):
LGTM.

Issue #3149, opened 2020-08-21
stderr() implemented for WDL.


DailyDreaming:
Relies on #3147

Description of WDL function to be added:

 ```
## File stderr()

Returns a `File` reference to the stderr that this task generated.
```
Reference: https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#file-stderr

 - [x] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [x] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-603)
┆Issue Number: TOIL-603

Issue #3148, opened 2020-08-21
stdout() implemented for WDL.


DailyDreaming:
Relies on #3147

Description of WDL function to be added:

 ```
## File stdout()

Returns a `File` reference to the stdout that this task generated.
```
Reference: https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#file-stdout

 - [x] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [x] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-602)
┆Issue Number: TOIL-602

Issue #3147, opened 2020-08-21
Create a stdout and stderr file for each WDL task.


DailyDreaming:
This is to support the cromwell built-ins `stdout()` and `stderr()`.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-601)
┆Issue Number: TOIL-601

PR #3181, opened 2020-09-04
Create a stdout and stderr file for each WDL task.

w-gao:
My initial attempt to address #3147, #3148, and #3149. 

It'll generate a stdout (as well as stderr) file locally after the subprocess/or docker call, and `stdout()` (and `stderr()`) WDL calls are replaced to the respective file paths in the compiled Python script. 

I'm not sure if this is the best approach and it could use more testing with complex WDL workflows.


src/toil/wdl/wdl_functions.py
```diff
@@ -348,6 +348,37 @@ def process_and_read_file(f, tempDir, fileStore, docker=False):
...
+    # Cromwell generates a folder for each task so the file is simply named stdout and lives in
+    # the task execution folder.
+
+    # the generated file would have a name like 'ten6l91dw_stdout'
+    prefix = tempDir.split('/')[-1]
+    path = os.path.join(tempDir, 'execution', '{}_{}'.format(prefix, 'stderr' if stderr else 'stdout'))
+    with open(path, 'ab+') as f:
+        f.write(stdout)
+
+    # should we import to filestore?
```
DailyDreaming, line 375: We would need to in order to work on any kind of distributed system.  Once the workflow completes, the file would need to be exported from the filestore to obtain the final output path, wherever that may be.

Our WDL support currently only works with a local jobstore however, and this is not necessary for the current local-only functionality.  We would have to open a PR supporting AWS or Google jobstore support in WDL, in which case this would become necessary.

adamnovak: >Our WDL support currently only works with a local jobstore however

What kind of crimes is our generated WDL code doing to make this happen? Is it reaching inside the job store instead of restricting itself to the FileStore API?

w-gao: > Our WDL support currently only works with a local jobstore however

I could be wrong but does this refers to how the generated WDL code writes temporary files to  `fileStore.getLocalTempDir()`? This is essentially what I’m doing here as well, since stdout and stderr files are temporary if not used ( unless we want to preserve them?)

```wdl
task task1 {
  command {
    python do_stuff.py
  }
  output {
    File results = stdout()
  }
}
```
In the case like this, we would then copy the temp stdout/stderr file to the specified output directory, along with other files defined in the output block. I can see that the output files are exported to the local jobstore here, but I suppose we could export to other jobstores as well? 

https://github.com/DataBiosphere/toil/blob/b2b09b1d912fa68052104f1bd22311d46bd03ed1/src/toil/wdl/wdl_functions.py#L263

> ... and this is not necessary for the current local-only functionality.

It'd be nice if I could get some clarification since I'm not sure if I quite understand this part.

DailyDreaming: @adamnovak It uses the filestore API, but it has been hardcoded to use a local jobstore and the default args I believe.  I suspect it wouldn't be a lot of work to convert to running on AWS/Google, but I'd have to investigate it, and since it's never been attempted, there may or may not be blockers to doing so.  The toil args are handed down as is, so in theory it should know what to do once that's changed.  I'll try and tinker with it to see if I can get it to work.

@w-gao That sounds right to me and should work.  The temp files only need to live for the life of the worker before export.

DailyDreaming:
@w-gao I understand where you're coming from with the explanation of 451063c.  Some things to consider.  stdout() is a file with the record of the command run, so if you run the command:

`echo ${read_lines(stdout())} > output.txt`

I would expect `output.txt` to be empty, as the content for `stdout()` hasn't been completely written yet (even if the prior line may have generated a line of `stdout`).

I think that calling `stdout()` within a workflow command should probably be avoided, as it's kind of a chicken and egg situation, where, let's say the user does something with the file and writes to it, but after the command is run, that file is overwritten.  Or worse, a command pipes something to it that changes its permissions or deletes it, and now the workflow can no longer write to it.

Something to keep in mind too... `stdout_file` and `stderr_file` are likely names that users might pick for arbitrary variables in their own workflows.  We should probably add a static UUID to each (probably to the stdout and stderr variables too, and that was an oversight).  Cromwell openly declares their built-ins, and for very common names, we should avoid making them potentially incompatible with cromwell workflows.

Overall, this might shift to an interpreted model, rather than the compiled model it's currently now in.  The really helpful bit is the testing framework, as once that's in place, development is far faster (and more reliable).

w-gao:
This should work other than stderr for docker call (#3184). In addition, it might be better if we generate a directory for each WDL task like Cromwell (to avoid duplicate and overriding of output files).

adamnovak:
@w-gao Is this done being a draft?

w-gao:
@adamnovak This should work other than getting the standard error from docker calls (#3184). I suppose it's ready for review and it would be helpful if I could get some feedback from you or @DailyDreaming.

DailyDreaming:
@w-gao This looks good to me so far, though the `stderr` portion is a stub and we should wait for that portion to be done before merging this.

src/toil/wdl/wdl_analysis.py
```diff
@@ -903,7 +903,11 @@ def parse_declaration_expressn_fncall(self, name, params, es):
                 if name.source_string == 'length':
                     es = es + 'len('
                 elif name.source_string == 'stdout':
-                    return es + 'stdout'
+                    # stdout is a str or bytes object. Since stdout() expects a File return
```
adamnovak, line 906: I'm concerned that we don't know whether stdout is str or bytes. Are we sure the generated code is always going to know and use the right one?

src/toil/wdl/wdl_functions.py
```diff
@@ -348,6 +348,34 @@ def process_and_read_file(f, tempDir, fileStore, docker=False):
...
 
 
+def generate_stdout_file(stdout, tempDir, stderr=False, fileStore=None):
+    """
+    Create a stdout (or stderr) file from a string or bytes object.
+    Returns a file path.
+
+    :param stdout: str|bytes
+    :param tempDir: str directory
+    :param stderr: bool
```
adamnovak, line 358: I think these are supposed to be `:param type varname: description of meaning`, and not `:param varname: type`.

src/toil/wdl/wdl_synthesis.py
```diff
@@ -665,6 +666,9 @@ def run(self, fileStore):
                                  fileStore.logToMaster("{jobname}")
                                  tempDir = fileStore.getLocalTempDir()
                                  
+                                 b3a2e464d6a4464cb138952a7d042595_stdout_file = os.path.join(tempDir, 'stdout')
```
adamnovak, line 669: We should document that this variable is prefixed with a high-entropy string to try and make it unique, so it can't collide with anything else in the genrated code.

Are you sure that's the right approach, rather than something human-readable, underscore-rich, and Toil-y as a prefix, like `_toil_wdl_internal__stdout_file`? It's not statistically guaranteed to avoid a collision unless someone actually copy-pastes it, but it is more meaningful if someone has to go spelunking through the generated code.

adamnovak: If we are going to use a high-entropy string, rather than copy-paste it everywhere, we should source it from one place in the generator.

adamnovak (changes requested):
I think some of the docstrings aren't quite formatted right (although I'm not sure because I don't use any of the tooling that reads them), and I don't think putting the same high-entropy string everywhere in the code generator is quite the right approach. But otherwise this looks good.

src/toil/wdl/wdl_synthesis.py
w-gao: Good point! A `_toil_wdl_internal__` prefix is definitely more readable than a high-entropy string. I'll switch to that instead. I wonder if I should also change [`asldijoiu23r8u34q89fho934t8u34fcurrentworkingdir`](https://github.com/DataBiosphere/toil/blob/releases/4.2.x/src/toil/wdl/wdl_synthesis.py#L115) since I based the naming on that?

src/toil/wdl/wdl_analysis.py
w-gao: Valid concern - it seems like the captured stdout from both subprocess and the docker API should be in bytes, but one of the methods in the docker API suggests str is returned. I'll make sure to check the type and make the output consistent before writing to the file store.

src/toil/wdl/wdl_synthesis.py
adamnovak: I would change that too.

adamnovak (approved):
This looks good to me now, when it passes tests.

adamnovak:
I'm going to pull this in. I thing we still have to hook up the newly-available Docker output though, right?

src/toil/wdl/wdl_synthesis.py
```diff
@@ -741,18 +745,23 @@ def write_function_dockercall(self, job):
...
-            ''', docker_dict, indent='        ')[1:]
+        # apiDockerCall() with demux=True returns a tuple of bytes objects (stdout, stderr).
+        _toil_wdl_internal__stdout, _toil_wdl_internal__stderr = \\
+            apiDockerCall(self, 
+                          image={docker_image}, 
+                          working_dir=tempDir, 
+                          parameters=[os.path.join(tempDir, "{job_task_reference}_script.sh")], 
+                          entrypoint="/bin/bash", 
+                          user={docker_user}, 
+                          stderr=True,
```
w-gao, line 756: @adamnovak Thanks for the review! The newly-available Docker output has been implemented here.

PR #3183, opened 2020-09-04
Add support for TOIL_CUSTOM_INIT_COMMAND.

### LINKED ISSUE #3182: Add support for running custom init commands before starting appliance
This is similar to #2561, except that the command is run before the appliance starts (i.e. it would be `TOIL_CUSTOM_INIT_COMMAND`). One use case for this is for any custom authentication for pulling the "base" docker images (e.g. we actually use this to pull the Toil docker image itself from a private repo).

I initially hesitated adding this upstream, but given the recent announcement from docker about limiting pulls from anonymous users, I figured it may become useful for other use cases as well (while I realize that we're switching to quay for now, it may still become useful in the future...and also reduces the diffs from our fork :) ).

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-627)
┆Issue Number: TOIL-627

arostamianfar:
Similar to #2561 except that it provides a way to run custom commands before running the appliance itself (e.g. we use this to authenticate to a private docker registry to pull the Toil appliance itself). May be useful for other use cases now that there seems to be more restrictions on doing a lot of pulls from public registries.

Resolves #3182

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 import errno
 import logging
 import os
+import re
 import requests
 import socket
 import sys
@@ -220,14 +221,16 @@ def customDockerInitCmd():
     Returns the custom command (if any) provided through the ``TOIL_CUSTOM_DOCKER_INIT_COMMAND``
     environment variable to run prior to running the workers and/or the primary node's services.
     This can be useful for doing any custom initialization on instances (e.g. authenticating to
-    private docker registries). An empty string is returned if the environment variable is not
-    set.
+    private docker registries). Any single quotes are escaped and the command cannot contain a
+    set of blacklisted chars (newline or tab). An empty string is returned if the environment
+    variable is not set.
 
     :rtype: str
     """
     command = lookupEnvVar(name='user-defined custom docker init command',
                            envName='TOIL_CUSTOM_DOCKER_INIT_COMMAND',
                            defaultValue='')
+    _check_custom_bash_cmd(command)
     return command.replace("'", "'\\''")  # Ensure any single quotes are escaped.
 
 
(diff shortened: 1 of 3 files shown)


src/toil/__init__.py
```diff
@@ -231,6 +231,22 @@ def customDockerInitCmd():
...
 
 
+def customInitCmd():
+    """
+    Returns the custom command (if any) provided through the ``TOIL_CUSTOM_INIT_COMMAND``
+    environment variable to run prior to running Toil appliance itself in workers and/or the
+    primary node (i.e. this is run one stage before ``TOIL_CUSTOM_DOCKER_INIT_COMMAND``).
+    This can be useful for doing any custom initialization on instances (e.g. authenticating to
+    private docker registries). An empty string is returned if the environment variable is not
+    set.
```
adamnovak, line 241: We should specify that it has embedded single quotes escaped, for concatenation with enclosing single quotes.

src/toil/__init__.py
```diff
@@ -231,6 +231,22 @@ def customDockerInitCmd():
...
+    :rtype: str
+    """
+    command = lookupEnvVar(name='user-defined custom init command',
+                           envName='TOIL_CUSTOM_INIT_COMMAND',
+                           defaultValue='')
+    return command.replace("'", "'\\''")  # Ensure any single quotes are escaped.
```
adamnovak, line 247: This eventually goes into a YAML multiline string where indentation is important. We might want to bail out with an error if the environment variable contains newlines or anything else that could corrupt the YAML we're trying to embed this into that we can't safely escape.

adamnovak (changes requested):
This is a good idea. But I'm suspicious of the robustness of the escaping (which also applies to the custom Docker command).

In theory what we want to do is impossible; you can't represent an arbitrary string as a YAML multiline string, which is where this eventually goes. If we wanted to get as close as possible, we'd have to Bash-escape out the newlines and other stuff YAML can't hold, which I don't think anything knows how to do.

But we could at least check for newlines and bail out.

src/toil/__init__.py
arostamianfar: That's a good point. Added a sanity check to env var. In practice, I'm not too worried about the edge cases since folks can't use this for malicious purposes (it's their own cluster) and we can always handle them as they come up.

arostamianfar (commented):
Thanks for the quick review! Added sanity checks for the env vars. PTAL.

adamnovak (approved):

PR #3186, opened 2020-09-04
Update prom/node-exporter to use quay.io

jonathanxu18:


adamnovak (approved):
Looks like this should work.

PR #3187, opened 2020-09-05
Check against joining the current thread during CachingFileStore destructor

### LINKED ISSUE #3176: The CachingFileStore's destructor can run in its commit thread
The CachingFileStore's destructor can run in its commit thread. This results in it attempting to join the current thread, which fails. This can I think cause the `JobFileStoreTest.testJobFileStore` test to fail.

https://travis-ci.org/github/DataBiosphere/toil/builds/723196253#L1105-L1113

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-625)
┆Issue Number: TOIL-625

adamnovak:
This should stop us trying to join the thread that we're running in, and fix #3176.

### DIFF
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1712,12 +1712,17 @@ class CachingFileStore(AbstractFileStore):
 
     def waitForCommit(self):
         # We need to block on the upload thread.
+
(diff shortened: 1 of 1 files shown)


DailyDreaming (approved):
LGTM.

PR #3194, opened 2020-09-08
slurm.py error, undefeind stdout - stdout should be stderr in code

### LINKED ISSUE #3192: slurm.py error, undefeind stdout - stdout should be stderr in code
There's a potential bug in slurm.py that leads to toil-cwl-runner failing. Line 127 declares stderr but line 129 references stdout leading to an error message (NameError: name 'stdout' is not defined)

Line 129 from [slurm.py](https://github.com/DataBiosphere/toil/blob/0aec2884aa829d4e1c9a888a4959b762bca39168/src/toil/batchSystems/slurm.py#L129) should be 

```python
            for line in stderr:
```

not

```python
            for line in stdout:
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-634)
┆Issue Number: TOIL-634

DailyDreaming:
This should fix #3192 and also remove some of the old python2 imports.

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -169,7 +169,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
 
             if mem is not None:
                 # memory passed in is in bytes, but slurm expects megabytes
-                sbatch_line.append(f'--mem={int(mem / 2 ** 20)}')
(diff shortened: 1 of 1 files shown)


adamnovak (approved):
This looks like it should work.

PR #3224, opened 2020-09-11
Add 'stream' and 'demux' options to apiDockerCall

### LINKED ISSUE #3184: Change apiDockerCall to handle stream=True.
Currently, apiDockerCall returns one output of stdout and stderr joined together:

```
>>> cmd = '/bin/sh -c "echo hello stdout ; echo hello stderr >&2"'
>>> res = client.containers.run(cmd, stream=False)
>>> res.output
b'hello stderr\nhello stdout\n'
```

The API inside can be changed with stream=True, which returns an iterator:

```
>>> cmd = '/bin/sh -c "echo hello stdout ; echo hello stderr >&2"'
>>> res = client.containers.run(cmd, stream=True)
>>> next(res.output)
b'hello stdout\n'
>>> next(res.output)
b'hello stderr\n'
>>> next(res.output)
StopIteration
```

Our apiDockerCall would need to change to handle the case where stream=True, so that we can return separate stdout and stderr streams.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-628)
┆Issue Number: TOIL-628

w-gao:
Resolves #3184. 

The `demux` implementation follows [docker-py's docs](https://docker-py.readthedocs.io/en/stable/user_guides/multiplex.html) for `exec_run` and `attach`. It is extremely inconvenient that this isn't supported in `client.containers.run()`. I had to add a modified version of [`_multiplexed_response_stream_helper()`](https://github.com/docker/docker-py/blob/4.3.1-release/docker/api/client.py#L370) to handle different stream types. 

Note: when upgrading docker in the future make sure that the private `_response` attribute in [`CancellableStream`](https://github.com/docker/docker-py/blob/master/docker/types/daemon.py#L11) has not changed.

### DIFF
--- a/src/toil/lib/docker.py
+++ b/src/toil/lib/docker.py
@@ -1,18 +1,14 @@
 from __future__ import absolute_import
 
-try:
-    # In Python 3 we have this quote
-    from shlex import quote
-except ImportError:
-    # But in 2.7 we have this deprecated one
-    from pipes import quote
 
 import docker
 import base64
 import requests
 import logging
 import os
-import sys
+import struct
+from shlex import quote
+from docker.utils.socket import consume_socket_output, demux_adaptor
 from docker.errors import create_api_error_from_http_exception
 from docker.errors import ContainerError
 from docker.errors import ImageNotFound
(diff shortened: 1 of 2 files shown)


adamnovak (approved):
This looks great!

adamnovak:
It looks like the Docker API snippet we're borrowing is the same license as us, so we shouldn't have a license problem.

Issue #3038, opened 2020-04-14
invalid --batchSystem generates a key error instead of a usage message


diekhans:
specifying` --batchSystem=sge`  generates  the error:

```
quanah 2020-04-13 21:02:47,010 MainThread DEBUG toil.lib.bioio: Root logger is at level 'DEBUG', 'toil' logger at level 'DEBUG'.
quanah 2020-04-13 21:02:47,012 MainThread DEBUG toil.lib.bioio: Logging to file 'log.txt'.
quanah 2020-04-13 21:02:47,013 MainThread DEBUG toil.lib.threading: Total machine size: 72 cores
quanah 2020-04-13 21:02:47,013 MainThread DEBUG toil.lib.threading: CPU quota: -1
Traceback (most recent call last):
  File "arf.py", line 11, in <module>
    with Toil(options) as toil:
  File "/lustre/scratch/mdiekhan/cactus/src/toil-debug/src/toil/common.py", line 736, in __enter__
    config.setOptions(self.options)
  File "/lustre/scratch/mdiekhan/cactus/src/toil-debug/src/toil/common.py", line 217, in setOptions
    setBatchOptions(self, setOption)
  File "/lustre/scratch/mdiekhan/cactus/src/toil-debug/src/toil/batchSystems/options.py", line 115, in setOptions
    factory = batchSystemFactoryFor(batchSystem)
  File "/lustre/scratch/mdiekhan/cactus/src/toil-debug/src/toil/batchSystems/registry.py", line 93, in batchSystemFactoryFor
    return _batchSystemRegistry[batchSystem ]
KeyError: 'sge'

```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-540)
┆Issue Number: TOIL-540

DailyDreaming:
@jeffrey856 Think you could add this error handling?  It should tell the user that they've specified an invalid batch system and then list the valid ones.

diekhans:
doable with the choices argparse option

DailyDreaming:
@diekhans You're right, I think adding `choices=unique_names()` should do it right here: https://github.com/DataBiosphere/toil/blob/727b7e1c30938c575717b46f0bc9f88758e3b6e9/src/toil/batchSystems/options.py#L124

PR #3225, opened 2020-09-14
batchSystem generates a key error instead of a usage message 

jeffrey856:
Added choices argparse to list of uniqueNames().

### DIFF
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -117,9 +117,9 @@ the logging module:
 
   --batchSystem BATCHSYSTEM
                         The type of batch system to run the job(s) with,
-                        currently can be one of LSF, Mesos, Slurm, Torque,
-                        HTCondor, singleMachine, parasol, gridEngine'.
-                        (default: singleMachine)
+                        currently can be one of lsf, Mesos, slurm, torque,
+                        htcondor, single_machine, parasol, grid_engine', kubernetes.
+                        (default: single_machine)
   --parasolCommand PARASOLCOMMAND
                         The name or path of the parasol program. Will be
                         looked up on PATH unless it starts with a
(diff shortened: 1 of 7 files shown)


adamnovak (changes requested):
I think this will stop accepting the extra names we defined in _DEFAULT_REGISTRY here:

https://github.com/DataBiosphere/toil/blob/8491a03a620f09f3e477405b5d4a5fabfe92c8bd/src/toil/batchSystems/registry.py#L52-L102

What does the help look like if we add another function to that file (`allAvailableNames()`?) to expose that list, and use that as the choices? Does argparse print out all the choices for us even if we don't want to display everything we accept?

Or should we take the opportunity to get rid of all those extra names and force people to always spell and capitalize each batch system exactly one way (and delete all the duplicates)?

jeffrey856:
@adamnovak If we were to do `choice = allAvailableNames()`, it would output all the choices even if we wouldn't want to. This is the sample that I have ran below. Ideally, I don't think we want to force our users to use all caps to use a certain batchSystem however that is the most straight-forwarded solution. 

Screen Shot 2020-09-14 at 15 19 02

adamnovak:
@DailyDreaming Do you think it would be a problem if we dropped down to accepting only one form of each batch system name? Maintaining the ability to use several makes it hard to use Argparse to make our UI nice.

DailyDreaming:
@adamnovak I don't mind, since accepting multiple cases with capitalization isn't normal, so long as everything is moved to lower-case only.  We'll have to find all cases in the docs, and also bump the major version to 5.0.0 since these will be considered breaking changes.

If we do, I think that we should also consider moving all of the options to `snake_case`.  I think some people get (understandably) hung up on things like `--nodestorage` vs. `--nodeStorage`.

adamnovak (approved):

adamnovak:
I'm going to go ahead and merge this. We can bump the version to 5 as part of the next release process, right?

adamnovak:
This fixed #3038.

PR #3227, opened 2020-09-16
kubeWatch optimization #3174

jeffrey856:
In progress to optimize kube watch.

A/C
Get status of all pods at timeout instead watching each pod at timeout.


jeffrey856:
Wait till this ticket is completed. https://github.com/DataBiosphere/toil/issues/3232

src/toil/batchSystems/kubernetes.py
```diff
@@ -163,7 +163,9 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
 
         # Set this to True to enable the experimental wait-for-job-update code
         # TODO: Make this an environment variable?
-        self.enableWatching = False
+        self.enableWatching = os.environ.get("KUBE_WATCH_ENABLED", False)
+
+        self.runID = 'toil-{}'.format(self.uniqueID)
```
adamnovak, line 168: This line appears to be duplicated.

src/toil/batchSystems/kubernetes.py
```diff
@@ -677,44 +679,55 @@ def getUpdatedBatchJob(self, maxWait):
...
+                # Grab the metadata data, ID and the list of conditions of the current job
+                jobObject = event['object']
+                jobID = int(jobObject.metadata.name[len(self.jobPrefix):])
+                jobObjectListConditions =jobObject.status.conditions
+                # Exit Reason defaults to 'Successfully Finsihed` unless said otherwise 
+                ExitReason = 1
+                ExitCode = 0
+
+                # Check if jobs exists by seeing if condition logs are pressent
+                if jobObjectListConditions is None or len(jobObjectListConditions):
```
adamnovak, line 695: I think you need to check something about the event instead of assuming the job is done if it doesn't have any conditions. For one thing, we might be getting "added" or "deleted" events about new jobs being created or old jobs being successfully destroyed, not just "updated" events. For another, we can get an "updated" event when the job undergoes any change, not just when it finishes or fails.

We should ignore any events that aren't "updated" events, and we should ignore any events where the job isn't actually done running. I think you might be seeing jobs without `jobObject.status.conditions` set because the job is not really done yet.

You should check `jobObject.status.finished` and `jobObject.status.failed`. A job that is actually finished will have a nonzero value for one of those. Pending and running jobs will have 0 for both.

src/toil/batchSystems/kubernetes.py
```diff
@@ -677,44 +679,55 @@ def getUpdatedBatchJob(self, maxWait):
...
+                    logger.debug("No job container statuses for job %s" % (jobObject.metadata.name))
+                    return UpdatedBatchJobInfo(jobID=jobID, exitStatus=EXIT_STATUS_UNAVAILABLE_VALUE, wallTime=0, exitReason=3)
+                
+                # Check if there are any active pods
+                if jobObject.status.acitve > 0:
+                    logger.info("%s has %d pods running" % jobObject.metadata.name, jobObject.status.active)
+                    continue
+                else:
+                    # No more active pods in the current job ; must be finished
```
adamnovak, line 704: Is this true? What happens if the pod is pending because it can't run right now because the cluster is full? Does the job still show it as "active"? What if the job has been created but Kubernetes hasn't yet created a pod for it?

I think you should use `jobObject.status.finished` and `jobObject.status.failed` instead. if either of those is 1, a pod has definitely finished or failed for the job. But there are a variety of reasons a pod might not be "actively running", even though it hasn't finished or failed yet.

adamnovak (changes requested):
This looks almost right.

I think the first thing you have to do when you get a job event is make sure it's an update (and not a creation or a deletion). Then you have to make sure that the job has a `finished` or a `failed` pod; otherwise it's not really done. After that I think you want to continue on with the code you have now.

src/toil/batchSystems/kubernetes.py
```diff
@@ -677,44 +677,56 @@ def getUpdatedBatchJob(self, maxWait):
...
+            for event in self._try_kubernetes_stream(self._api('batch').list_namespaced_job, self.namespace, 
+                                                        label_selector="toil_run={}".format(self.runID),
+                                                        timeout_seconds=maxWait):
+                # Grab the metadata data, ID, the list of conditions of the current job, and the total pods
+                jobObject = event['object']
+                jobID = int(jobObject.metadata.name[len(self.jobPrefix):])
+                jobObjectListConditions =jobObject.status.conditions
+                totalPods = jobObject.status.active + jobObject.status.finished + jobObject.status.failed
+                # Exit Reason defaults to 'Successfully Finsihed` unless said otherwise 
+                ExitReason = 1
```
adamnovak, line 690: I think we want to import BatchJobExitReason and use e.g. `BatchJobExitReason.FINISHED` instead of just `1`, in case we reassign the numbers, and so we can tell what they mean.

Also, variables probably shouldn't be capitalized like this; Github is highlighting them like classes.

adamnovak (approved):
This looks like it should do the right thing. I have a couple style complaints, but I think I can just fix them.

jeffrey856:
@adamnovak  forgot to update the documentation. We should add  `os.environ.get("KUBE_WATCH_ENABLED", False)` to toil environmental documentation.

adamnovak:
Oh good idea @jeffrey856. Can you make a PR for that?

PR #3228, opened 2020-09-16
Remove Python 2 imports in WDL

w-gao:
Removes old Python 2 imports + PEP 8 changes.

Edit: revert part that breaks tests.

### DIFF
--- a/src/toil/wdl/toilwdl.py
+++ b/src/toil/wdl/toilwdl.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 UCSC Computational Genomics Lab
+# Copyright (C) 2018-2020 UCSC Computational Genomics Lab
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


w-gao:
Marked as draft because I see the purpose of the try block now - although I feel like there should be a better way?

DailyDreaming:
@w-gao You're not wrong, and there are better ways of implementing it.  There was some time pressure in writing this, so it was written to work on the workflows we needed it to, after which point manpower was cut back, and which is why we don't have most of the WDL built-in functions, or properly support a portion of the WDL spec.  This is the first significant effort to shore up WDL functionality in a while.

The overall design of the WDL implementation should probably be reworked.  Compiling into a written toil script was what was implemented at the time, but it's likely not the best method.  Shifting to an interpreted model is likely the cleaner approach.

The manual WDL tests are currently turned off, or they would have failed when the try/except was removed.  These tests shoulder quite a bit of the functionality and I'll add them back in to the main test infra.

DailyDreaming (approved):
LGTM.  Will merge after tests pass.

PR #3233, opened 2020-09-21
Adding Labels to Pods/Jobs

jeffrey856:
A/C:
Add labels to pods and jobs.
Refactor existing search to work with the labels. 

Issue Ticket:
https://github.com/DataBiosphere/toil/issues/3232

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -964,7 +964,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
                 pass
             try:
                 logger.debug('Cleaning up pod at shutdown: %s', str(pod))
-                respone = self._api('core').delete_namespaced_pod,  pod.metadata.name,
+                respone = self._try_kubernetes_expecting_gone(self._api('core').delete_namespaced_pod,  pod.metadata.name,
                                     self.namespace, 
                                     propagation_policy='Background')
             except ApiException as e:
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -468,7 +468,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
             
             return jobID
     
-    def _ourJobObject(self, onlySucceeded=False, limit=None):
+    def _ourJobObject(self, onlySucceeded=False):
         """
         Yield Kubernetes V1Job objects that we are responsible for that the
         cluster knows about.
@@ -503,10 +503,10 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
 
             if onlySucceeded: 
                 results =  self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
-                                                label_selector="toil_run={}".format(self.runID), field_selector="status.successful==1", **kwargs)
+                                                label_selector="toil_run={}".format(self.runID), field_selector="status.successful==1", **kwargs).items
             else:
                 results = self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
-                                                label_selector="toil_run={}".format(self.runID), **kwargs)
+                                                label_selector="toil_run={}".format(self.runID), **kwargs).items
             yield results
 
             # Remember the continuation token, if any
@@ -534,7 +534,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
             if token is not None:
                 kwargs['_continue'] = token
 
-            results = self._try_kubernetes(self._api('core').list_namespaced_pod, self.namespace, label_selector="toil_run={}".format(self.runID), **kwargs)
+            results = self._try_kubernetes(self._api('core').list_namespaced_pod, self.namespace, label_selector="toil_run={}".format(self.runID), **kwargs).items
             
             yield results
             # Remember the continuation token, if any
@@ -771,7 +771,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         # Put 'done', 'failed', or 'stuck' here
         chosenFor = ''
 
-        for j in self._ourJobObject(onlySucceeded=True, limit=1):
+        for j in self._ourJobObject(onlySucceeded=True):
             # Look for succeeded jobs because that's the only filter Kubernetes has
             jobObject = j
             chosenFor = 'done'
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -937,13 +937,11 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         # Shutdown local processes first
         self.shutdownLocal()
        
-        # Kill jobs whether they succeeded or failed and clean up pods
+        # Kill jobs whether they succeeded or failed and clean up pods that are associated with those jobs
         try:
-            # Delete with background poilicy so we can quickly issue lots of commands
             self._try_kubernetes_expecting_gone(self._api('batch').delete_collection_namespaced_job, 
                                                             self.namespace, 
-                                                            label_selector="toil_run={}".format(self.runID),
-                                                            propagation_policy='Background')
+                                                            label_selector="toil_run={}".format(self.runID))
             logger.debug('Killed jobs with delete_collection_namespaced_job; cleaned up')
         except ApiException as e:
             if e.status != 404:
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -465,21 +465,6 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
             logger.debug('Launched job: %s', jobName)
             
             return jobID
-            
-            
-    def _isJobOurs(self, jobObject):
-        """
-        Determine if a Kubernetes job belongs to us.
-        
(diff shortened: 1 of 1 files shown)


src/toil/batchSystems/kubernetes.py
```diff
@@ -441,13 +445,16 @@ def issueBatchJob(self, jobNode):
             pod_spec = kubernetes.client.V1PodSpec(containers=[container],
                                                    volumes=volumes,
                                                    restart_policy="Never")
+            # Make metadata to labl the job/pod with info.
+            metadata = kubernetes.client.V1ObjectMeta(name=jobName
+                                                    ,labels={"toil_run": self.runID})
```
adamnovak, line 450: This comma really wants to be on the line above.

src/toil/batchSystems/kubernetes.py
```diff
@@ -441,13 +445,16 @@ def issueBatchJob(self, jobNode):
             pod_spec = kubernetes.client.V1PodSpec(containers=[container],
                                                    volumes=volumes,
                                                    restart_policy="Never")
+            # Make metadata to labl the job/pod with info.
```
adamnovak, line 448: `s/labl/label/`

src/toil/batchSystems/kubernetes.py
```diff
@@ -441,13 +445,16 @@ def issueBatchJob(self, jobNode):
...
+            
             # Make another spec for the job, asking to run the template with no backoff
             job_spec = kubernetes.client.V1JobSpec(template=template, backoff_limit=0)
-            # Make metadata to tag the job with info.
-            # We use generate_name to ensure a unique name
-            metadata = kubernetes.client.V1ObjectMeta(name=jobName)
+            
             # And make the actual job
             job = kubernetes.client.V1Job(spec=job_spec,
                                           metadata=metadata,
```
adamnovak, line 460: It seems a bit weird to reuse the same V1ObjectMeta instance for the job and the pod, but it seems to be working.

src/toil/batchSystems/kubernetes.py
```diff
@@ -497,47 +489,19 @@ def _ourJobObjects(self, onlySucceeded=False, limit=None):
...
-            token = getattr(results.metadata, 'continue', None)
+        if onlySucceeded: 
+            return self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
+                                            label_selector="toil_run={}".format(self.runID), field_selector="status.successful==1").items
+ 
+        return self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
+                                            label_selector="toil_run={}".format(self.runID)).items
+       
+    
+    def _getOurPodObjectList(self):
```
adamnovak, line 501: This ought to have a docstring. Also, why does this one have `get` in it when `_ourJobObjectList` doesn't? They should match.

src/toil/batchSystems/kubernetes.py
```diff
@@ -497,47 +489,19 @@ def _ourJobObjects(self, onlySucceeded=False, limit=None):
...
-                    # This job belongs to us
-                    yield job
-                    
-                    # Don't go over the limit
-                    seen += 1
-                    if limit is not None and seen >= limit:
-                        return
-                    
-            # Remember the continuation token, if any
-            token = getattr(results.metadata, 'continue', None)
```
adamnovak, line 535: With the label, you can rip out the `if self._isJobOurs(job):` logic, because the returned jobs will always be ours. But I don't think you can safely get rid of all this token and `_continue` logic. When you call `list_namespaced_job`, even with a `label_selector`, you don't always get all the jobs in the response, because there could be too many of them to send all at once. Instead, you get the first page of jobs (like Google search results), and you need to make another call, providing the "continuation token" from the first page, in order to get the second page, and so on until the server runs out of pages to send you.

If you want to be able to retrieve jobs and also pods, maybe you can factor out the get-all-the-pages logic in such a way that it can wrap calls to either `list_namespaced_job` or `list_namespaced_pod`, but you need to be able to get all the pages. We could be in a situation where there's a pod that's stuck on the very last page and we need to clean it up before anything else can make progress.

The reason it was an iterator before was so you could work on the pages as they arrive. I guess you could collect all the pages into a single list if you really want to, but it might be a huge list with thousands of items, and using an iterator could be more efficient. The downside of using an iterator is that the continuation tokens eventually expire, so if you spend too much time working on one page, you won't be able to get the next page. So it might indeed be safer to compose the whole list first.

src/toil/batchSystems/kubernetes.py
```diff
@@ -763,13 +727,17 @@ def _getUpdatedBatchJobImmediately(self):
         jobObject = None
         # Put 'done', 'failed', or 'stuck' here
         chosenFor = ''
-        for j in self._ourJobObjects(onlySucceeded=True, limit=1):
+        # Grab our job objects via toil_run label
+        ourJobObjects  = self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
+                                            label_selector="toil_run={}".format(self.runID))
```
adamnovak, line 732: The result of this call is never used.

adamnovak (changes requested):
This mostly looks good. But you've ripped out some fetch-all-the-pages logic that I think we need to keep, along with the check-if-we-own-the-job logic that we are getting rid of.

src/toil/batchSystems/kubernetes.py
jeffrey856: Should we use different metadata and labels for job and pods? I don't think it is necessary to since it is serving one purpose which is for easy aggregation. However I do see the need if we want to log out details for each jobs and pods.

interfect: I think it's fine unless we catch it causing a bug.

src/toil/batchSystems/kubernetes.py
```diff
@@ -489,17 +489,59 @@ def _ourJobObjectList(self, onlySucceeded=False, limit=None):
...
+            if token is not None:
+                kwargs['_continue'] = token
+
+            if onlySucceeded: 
+                results =  self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
+                                                label_selector="toil_run={}".format(self.runID), field_selector="status.successful==1", **kwargs).items
+            else:
+                results = self._try_kubernetes(self._api('batch').list_namespaced_job, self.namespace, 
+                                                label_selector="toil_run={}".format(self.runID), **kwargs).items
+            yield results
```
adamnovak, line 510: This is going to maker the iterator yield a series of lists, instead of all the items from the lists.

src/toil/batchSystems/kubernetes.py
```diff
@@ -489,17 +489,59 @@ def _ourJobObjectList(self, onlySucceeded=False, limit=None):
...
+        while True:
+            # We can't just pass e.g. a None continue token when there isn't
+            # one, because the Kubernetes module reads its kwargs dict and
+            # cares about presence/absence. So we build a dict to send.
+            kwargs = {}
+            
+            if token is not None:
+                kwargs['_continue'] = token
+
+            yield self._try_kubernetes(self._api('core').list_namespaced_pod, self.namespace, label_selector="toil_run={}".format(self.runID), **kwargs).items
```
adamnovak, line 537: Here also we're yielding a list of objects, rather than the objects themselves as specified in the docstring.

adamnovak (changes requested):
I think you just need to go through and yield the objects from each page, instead of the page lists themselves, and this will be good. Although you might want to remove the `List` suffix from the function names, because they don't return lists anymore and are back to being generators.

adamnovak:
Looks like it failed a test:

```
 _____________________ KubernetesBatchSystemTest.testSetEnv _____________________
  File "/builds/databiosphere/toil/src/toil/test/batchSystems/batchSystemTest.py", line 216, in testSetEnv
    jobUpdateInfo = self.batchSystem.getUpdatedBatchJob(maxWait=1000)
  File "/builds/databiosphere/toil/src/toil/batchSystems/kubernetes.py", line 688, in getUpdatedBatchJob
    result = self._getUpdatedBatchJobImmediately()
  File "/builds/databiosphere/toil/src/toil/batchSystems/kubernetes.py", line 843, in _getUpdatedBatchJobImmediately
    jobID = int(jobObject.metadata.name[len(self.jobPrefix):])
AttributeError: 'V1ListMeta' object has no attribute 'name'
```

src/toil/batchSystems/kubernetes.py
```diff
@@ -460,25 +467,10 @@ def issueBatchJob(self, jobNode):
...
-        :return: True if the job is our responsibility, and false otherwise.
-        :rtype: bool
-        """
-        
-        return jobObject.metadata.name.startswith(self.jobPrefix)
-        
-        
     
-    def _ourJobObjects(self, onlySucceeded=False, limit=None):
+    def _ourJobObjectList(self, onlySucceeded=False, limit=None):
```
adamnovak: It looks like we no longer actually implement a limit, so you should remove the limit option here.

src/toil/batchSystems/kubernetes.py
adamnovak: You need [this loop](https://github.com/DataBiosphere/toil/pull/3233/files#diff-183d11f3af830162fadf362f8c641fdfL524-L527) back, except without the `_isJobOurs` check.

adamnovak (approved):
This looks right to me.

PR #3234, opened 2020-09-21
Move checkForDeadlocks() in the leader

jonathanxu18:


jonathanxu18:
This will fix #3104.

DailyDreaming (approved):
@adamnovak Is there a good unit test for this, or should it be run manually and called good?

adamnovak (approved):
Looks like it should work to me.

adamnovak:
I'm happy to believe @jonathanxu18's testing results, and the unit tests saying we can still detect deadlocks. Testing to see if Kubernetes job collection is "fast enough" would be nice to automate, but also is probably a lot of work, and I don't think we need it to block merging this.

PR #3243, opened 2020-09-25
Fix boto3 migration bugs.

arostamianfar:
Just fixing a couple of small bugs in PR #3145 that I noticed after rebasing to head.


adamnovak (approved):
This looks right to me.

DailyDreaming:
Thanks for catching those!

Issue #3161, opened 2020-08-26
read_boolean() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Boolean read_boolean(String|File)

The `read_boolean()` function takes a file path which is expected to contain 1 line with 1 Boolean value (either "true" or "false" on it).  This function returns that Boolean value.

If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. Examples of failure include but are not limited to not having access to the file, resource limitations (e.g. memory) when reading the file, and implementation imposed file size limits.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#boolean-read_booleanstringfile

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-614)
┆Issue Number: TOIL-614
┆Epic: WDL Standard Library Implementation

Issue #3159, opened 2020-08-26
read_string() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## String read_string(String|File)

The `read_string()` function takes a file path which is expected to contain 1 line with 1 string on it.  This function returns that string.

No trailing newline characters should be included

If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. Examples of failure include but are not limited to not having access to the file, resource limitations (e.g. memory) when reading the file, and implementation imposed file size limits.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#string-read_stringstringfile

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
    - Partially implemented here: https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py#L463
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-612)
┆Issue Number: TOIL-612
┆Epic: WDL Standard Library Implementation

w-gao:
I'm not sure if I understand the current implementation of `read_string()` (as well as `read_int()`, `read_float()`, etc) in `wdl_functions`: https://github.com/DataBiosphere/toil/blob/releases/4.2.x/src/toil/wdl/wdl_functions.py#L463-L478

```python
def read_string(inputstring):
    if isinstance(inputstring, tuple):
        inputstring = inputstring[0]
    return str(inputstring)
```

If I understand correctly, this function parses the input _path_, instead of reading the content from the path which is what the function is supposed to do. Are these functions used for another purpose?

DailyDreaming:
@w-gao You're right, these are incorrectly implemented.  Please remove the current functions and add the correct functions which read from a file.

Issue #3158, opened 2020-08-26
read_int() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Int read_int(String|File)

The `read_int()` function takes a file path which is expected to contain 1 line with 1 integer on it.  This function returns that integer.

If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#int-read_intstringfile

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
    - Partially implemented here: https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py#L475
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-611)
┆Issue Number: TOIL-611
┆Epic: WDL Standard Library Implementation

PR #3244, opened 2020-09-25
Add WDL read functions and builtin tests

w-gao:
This PR fixes the current implementation (https://github.com/DataBiosphere/toil/issues/3159#issuecomment-696434782) and adds their built-in tests. 

This completes #3158, #3159, #3160, #3161. There are four more `read_*` WDL functions that I will add (either here or in a separate PR) once this is good.


src/toil/test/wdl/builtinTest.py
```diff
@@ -58,6 +62,46 @@ def testFn_Floor(self):
...
 
+    def testFn_ReadInt(self):
+        """Test the wdl built-in functional equivalent of 'read_int()'."""
+        num = 10
+        path = self._write_temp_file('read_int', content=str(num))
+        self.assertEqual(num, read_int(path))
+
+        num = 10.5
+        path = self._write_temp_file('read_int', content=str(num))
+        self.assertRaises(ValueError, read_int, path)
```
DailyDreaming, line 73: Is this Cromwell's behavior, and is it the same for `10.0`?

We should probably implement what Cromwell does in this case, since if they implement this sort of duck typing for floats/ints would mean that "reasonable" workflows will run without error on Cromwell, but will error on Toil.  If not, ignore my comment.  If so, we should include a comment about why we diverge from the spec.

src/toil/test/wdl/builtinTest.py
```diff
@@ -58,6 +62,46 @@ def testFn_Floor(self):
...
+
+    def testFn_ReadString(self):
+        """Test the wdl built-in functional equivalent of 'read_string()'."""
+        some_str = 'some string'
+        path = self._write_temp_file('read_string', content=some_str)
+        self.assertEqual(some_str, read_string(path))
+
+        # with new line characters
+        path = self._write_temp_file('read_string', content=some_str + '\n\n')
```
DailyDreaming, line 82: The WDL spec says that Cromwell only cuts off trailing newlines:

```No trailing newline characters should be included```

But `read_string()` seems to `strip()` from the front and the end.  Is this the same as Cromwell's behavior?

Preceding newlines should be added as a test based on Cromwell's behavior.

**Review by DailyDreaming — COMMENTED** [review]
Overall, this looks good to me, but please see comments.  This is probably ready for merging after that.

File: src/toil/test/wdl/builtinTest.py (hunk 1 above)
Comment (w-gao) [reply src/toil/test/wdl/builtinTest.py:73]: Anything that is not an integer would not work, even `10.0`. Cromwell does a similar thing where it throws the following errors when passing `10.0` and `10.5` to `read_int()`: 

```
Caused by: common.exception.AggregatedMessageException: Error(s):
Could not evaluate expression: read_int(in_file): For input string: "10.0”
        at common.validation.Validation$ValidationTry$.toTry$extension1(Validation.scala:89)
…
```
```
Caused by: common.exception.AggregatedMessageException: Error(s):
Could not evaluate expression: read_int(in_file): For input string: "10.5”
        at common.validation.Validation$ValidationTry$.toTry$extension1(Validation.scala:89)
…
```

So, we should be good in these cases. This is a good point to bring up though, since I didn’t think about testing `10.0`.

File: src/toil/test/wdl/builtinTest.py (hunk 2 above)
Comment (w-gao) [reply src/toil/test/wdl/builtinTest.py:82]: Good catch! I did not realize `strip()` removes preceding newlines.. Although, it seems like Cromwell removes preceding newlines as well. For reference, I am using `cromwell-52.jar` if that makes a difference at all. 

Either Cromwell isn't following their own spec, or, perhaps, "`[the file] is expected to contain 1 line with 1 string on it`" implies that there shouldn't be any preceding newlines to begin with?

I'll update the test as soon as possible. Thanks for the review!

**Comment by w-gao** [comment]
@DailyDreaming I added the rest of the `read` functions since you last reviewed. Please check if I missed any of the cases in the tests when you have time. Thanks!

File: src/toil/test/wdl/builtinTest.py [hunk 3]
```diff
@@ -58,6 +66,101 @@ def testFn_Floor(self):
...
+        path = self._write_temp_file('read_tsv', tsv_str)
+        self.assertEqual(tsv, read_tsv(path))
+
+    def testFn_ReadJson(self):
+        """Test the wdl built-in functional equivalent of 'read_json()'."""
+        json_obj = {'str': 'some string', 'num': 3.14, 'bool': True, 'null': None, 'arr': ['test']}
+        json_arr = ['1', '2']
+        json_num = 3.14
+
+        import json
```
DailyDreaming, line 95: Nit: Please import at the top of the file.

DailyDreaming (approved):
@w-gao The functions and the testing look great!  Though this might be getting unwieldy to review.  We haven't always followed best practice, but we should probably keep one PR per issue, and try to keep those PRs slimmer.  Smaller PRs are easier to review, and therefore easier to find bugs in.

Only one nit, otherwise this looks good to merge.  I'll approve and let you merge this but please address the comment before doing so first.

w-gao:
@DailyDreaming Thanks for the review, Lon. I'll keep that in mind in the future!

PR #3250, opened 2020-09-30
Job Concept Unification

### LINKED ISSUE #3215: Refactor JobGraph and JobNode -> JobDescription
We should have one class for the info about a job needed for scheduling.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-655)
┆Issue Number: TOIL-655

adamnovak:
This fixes #3215 by eliminating JobGraph and JobNode in favor of a single JobDescription class, which stores the official single copy of a job's requirements and service and child relationships, all by ID reference.

This should make #3216 much easier by providing a sensible type to use.


adamnovak:
This is starting to pass a lot of the tests, so I think it's about ready for review. There are a few fixes for other issues that came up in testing, and some py2->3 fixes I made along the way, but I'm not sure it makes sense to try and unmake them here and turn them into their own PRs.

The major changes are:
* JobGraph and JobNode are no more. There is now only JobDescription, which contains all the scheduling information.
* JobDescriptions are not comparable for equality. We want to move towards a system where only one JobDescription for a particular job can exist in memory at a time.
* Instead of duplicating the fields for ID, name, predecessors/successors, and requirements, Job now contains a JobDescription.
* On serialization, the JobDescription is persisted separately from the Job, so the leader can load the description only.
* Jobs storing other Jobs for insertion along with them into the job store has been unified. Instead of keeping a list of child Jobs, and a list of follow-on Jobs, or whatever, Jobs use ID references (via JobDescription) to describe relationships, and all jobs not yet in the job store live in a "registry" dict by ID, shared among all jobs in a connected component of the graph of job relationships.
* When JobDescriptions are first created, they use temporary fake IDs to facilitate ID references. The JobStore now assigns final IDs to all the JobDescriptions in a graph before we serialize any of them.
* Services are no longer themselves Jobs, since they were never scheduled as Jobs.
* I created some service start-up bugs and had to refactor the ServiceManager slightly to understand them. The ServiceManager now has a concept of a job's services having failed to start, and will log accordingly.
* JobDescriptions internally distinguish child and follow-on relationships. The `stack` member is now a property that synthesizes a stack from the child and follow-on relationships, instead of a directly mutable list. Instead of splicing the stack, there are now methods on JobDescription to drop completed successors.
* Because the stack can't be freely spliced anymore, I couldn't work out how to chain to a child, and then go on to chain to a follow-on of the original job. Now, when a job has both children and follow-ons, chaining stops.
* I created promise serialization order bugs, and to debug them I improved reporting of promise passing mistakes. Error messages should now describe the jobs that the offending promise was passed between.

adamnovak:
:tada: It's alive! Tests pass :100:

@DailyDreaming This should be properly ready for review now.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+                jobStore.update(self)
+        return successorsDeleted
+
+class Job:
     """
     Class represents a unit of work in toil.
     """
     def __init__(self, memory=None, cores=None, disk=None, preemptable=None,
-                       unitName=None, checkpoint=False, displayName=None):
+                       unitName='', checkpoint=False, displayName='',
```
DailyDreaming, line 891: Could typing be added here?

DailyDreaming (commented):
Some partial review comments for now.  Will finish the review later this afternoon.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+        
+    def __eq__(self, other):
+        return isinstance(other, FakeID) and self._value == other._value
+        
+    def __ne__(self, other):
+        return not isinstance(other, FakeID) or self._value != other._value
 
-    @property
-    def cores(self):
+class ConfigClient:
```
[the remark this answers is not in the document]
adamnovak: I suppose it might make sense to roll this into `Requirer`. I think I made `ConfigClient` first.

We can't initialize with a config in `__init__` because the flow we need to support is deserializing a JobDescription and then attaching it to the *current* config for this run of the workflow, whatever that is.

We could just make it a field that gets set, especially if `assignConfig` doesn't *actually* do any checking to catch multiple calls or configs having been serialized anything. But I wanted to make the semantics clear that when you ahve one of these things you *need* to give it a config before you really use it. The best I dea I came up with to do that was having a method.

I suppose I could change it around so that there's a property that inheriting classes use, and if `assignConfig` hasn't been called we throw when the property is read.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+            elif isinstance(value, bool):
+                return value
+            else:
+                raise TypeError(f"The '{name}' requirement does not accept values that are of type {type(value)}")
+        else:
+            # Anything else we just pass along without opinons
             return value
-        elif isinstance(value, float) and name == 'cores':
+        
+    def _fetchRequirement(self, requirement):
```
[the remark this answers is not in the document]
adamnovak: What do you mean by "typing"? Documentation that `requirement` is `str` and the return type can be `int`, `float`, or `bool`, depending on the particular requirement queried and its value?

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+    Subclassed into variants for checkpoint jobs and service jobs that have
+    their specific parameters.
     """
-    def __init__(self, requirements, jobName, unitName, jobStoreID,
-                 command, displayName=None, predecessorNumber=1):
-        super().__init__(requirements=requirements, displayName=displayName, unitName=unitName, jobName=jobName)
-        self.jobStoreID = jobStoreID
-        self.predecessorNumber = predecessorNumber
+    
+    def __init__(self, requirements=None, unitName='', displayName='', jobName='', command=None):
```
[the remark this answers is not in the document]
adamnovak: The tests create a bunch of mock-ish JobDescriptions. They do seem to go out of their way to fill in all these (formerly required) fields, but I'm not sure we should require them to.

I can try making some of these required. We probably can't do much with a JobDescription with no `requirements`. But we shouldn't necessarily need a name, and `command` is often empty anyway.

src/toil/job.py
adamnovak: I think these all have types documented already.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
-                   unitName=job.unitName,
-                   displayName=job.displayName,
-                   predecessorNumber=predecessorNumber)
-
-class Job(BaseJob):
+                # Update again to commit the removal of successors.
+                jobStore.update(self)
+        return successorsDeleted
+
+class Job:
```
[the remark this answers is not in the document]
adamnovak: The way it works is that both `Job` and `JobDescription` have `cores` and friends as properties. `JobDescription` stores the actual values, and `Job` fulfills all reads and writes by sending them along to the associated `JobDescription`.

The values come in initially through the `__init__` of Job, and they're passed along to the JobDescription when it is created.

Note that descriptionClass is a class, *not* a JobDescription instance. Job is responsible for creating its own JobDescription. It's just that some Job descendants need to use JobDescription descendants as their descriptions, and the best way I thought of to do that was to pass the type along to Job's `__init__`. Otherwise all the name determining and requirements munging would have to move to methods that all the Job descendants would have to call in their `__init__` methods before constructing their descriptions.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+            super().__init__(f"Job {promisingJob.description} cannot promise its return value to a job that is not its successor")
         else:
-            raise AttributeError("Default value for 'disk' cannot be determined")
+            # Write a full error message
+            super().__init__(f"Job {promisingJob.description} cannot promise its return value to non-successor {recipientJob.description}")
+            
 
-    @property
-    def memory(self):
+class FakeID:
```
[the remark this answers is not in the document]
adamnovak: I'm going to call it a `TemporaryID` now.

adamnovak:
OK @DailyDreaming, I addressed your comments so far, I think.

adamnovak:
OK, making the job name mandatory introduced a bunch of errors, and making `assignConfig` actually do some checking found a bunch of places where we were not really associating configs exactly once, so I fixed all that up.

I changed associating configs to be the job of the JobStore when it loads (so the leader and worker no longer need to cvare about it), and I made the config contagious to newly-constructed config-less jobs so that chaining to newly-constructed jobs works properly.

I think it should all be working now.

src/toil/job.py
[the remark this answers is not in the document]
DailyDreaming: Specifically: `def _fetchRequirement(self, requirement: str):`

See PEP484: https://www.python.org/dev/peps/pep-0484/#type-definition-syntax

Also, I just personally prefer this form of type hinting, which is probably bias I admit, but I'd like to start using it where possible in Toil going forward.  Type hints in docstrings are also pretty nice, but I still want to encourage PEP484 type hints.

src/toil/job.py
```diff
@@ -59,246 +59,897 @@
...
+        return dict(self._requirementOverrides)
+    
+    @property
+    def disk(self):
+        """
+        The maximum number of bytes of disk required.
+        """
+        return self._fetchRequirement('disk')
+    @disk.setter
+    def disk(self, val):
```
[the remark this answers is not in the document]
DailyDreaming: *Same PEP484 type hinting as previous comment(s).

src/toil/job.py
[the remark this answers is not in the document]
DailyDreaming: Sounds good to me.

src/toil/leader.py
```diff
@@ -282,195 +283,234 @@ def create_status_sentinel_file(self, fail):
...
             # Remove the successor job from the cache
-            self.toilState.jobsToBeScheduledWithMultiplePredecessors.pop(successorJobStoreID)
+            self.toilState.jobsToBeScheduledWithMultiplePredecessors.pop(successor.jobStoreID)
             return True
 
-    def _makeJobSuccessorReadyToRun(self, jobGraph, jobNode):
-        """make a successor job ready to run, returning False if they should
-        not yet be run"""
-        successorJobStoreID = jobNode.jobStoreID
+    def _makeJobSuccessorReadyToRun(self, successor, predecessor):
```
DailyDreaming, line 360: This makes things far clearer.

src/toil/jobStores/googleJobStore.py
```diff
@@ -199,7 +199,9 @@ def load(self, jobStoreID):
             jobString = self._readContents(jobStoreID)
         except NoSuchFileException:
             raise NoSuchJobException(jobStoreID)
-        return pickle.loads(jobString)  # UPDATE bz2.decompress(
+        job = pickle.loads(jobString) 
+        job.assignConfig(self.config)
```
DailyDreaming, line 203: Could a comment be added as to what `job.assignConfig` is doing here?

src/toil/jobStores/googleJobStore.py
```diff
@@ -165,20 +165,20 @@ def destroy(self):
...
-            self._writeString(jobStoreID, pickle.dumps(job, protocol=pickle.HIGHEST_PROTOCOL))  # UPDATE: bz2.compress(
-        return job
+        log.debug("Assigning ID to job %s for '%s'",
+                  jobStoreID, '<no command>' if jobDescription.command is None else jobDescription.command)
+        jobDescription.jobStoreID = jobStoreID
 
-    def _newJobID(self):
-        return "job"+str(uuid.uuid4())
+    def create(self, jobDescription):
+        # TODO: we don't implement batching, but we probably should.
```
DailyDreaming, line 179: Can an issue be created for this?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -667,30 +688,42 @@ def discardStream(stream):
...
+        """
+        raise NotImplementedError()
 
     @contextmanager
     def batch(self):
         """
-        All calls to create() with this context manager active will be performed in a batch
-        after the context manager is released.
-
+        If supporeted by the batch system, calls to create() with this context
```
DailyDreaming, line 706: `supporeted`

src/toil/jobStores/abstractJobStore.py
```diff
@@ -476,102 +488,113 @@ def getJob(jobId):
                 return self.load(jobId)
 
         def haveJob(jobId):
+            assert len(jobId) > 1, "Job ID {} too short; is a string being used as a list?".format(jobId)
```
DailyDreaming, line 491: Should this check type?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -108,6 +108,17 @@ def __init__(self, locator):
 class AbstractJobStore(with_metaclass(ABCMeta, object)):
     """
     Represents the physical storage for the jobs and files in a Toil workflow.
+    
+    JobStores are responsible for storing :class:`toil.job.JobDescription`s
```
DailyDreaming, line 112: Thanks for adding this, this is great.

src/toil/jobGraph.py
```diff
@@ -1,235 +0,0 @@
-# Copyright (C) 2015-2018 Regents of the University of California
```
DailyDreaming, line 1: The sound of taps played on a bugle is heard far off in the distance...

src/toil/job.py
```diff
@@ -1402,41 +2351,61 @@ def _executor(self, jobGraph, stats, fileStore):
...
-        :param class jobGraph: Instance of a jobGraph object
         :param class jobStore: Instance of the job store
-        :param toil.fileStores.abstractFileStore.AbstractFileStore fileStore: Instance of a cached
-               or uncached filestore
+        :param toil.fileStores.abstractFileStore.AbstractFileStore fileStore: Instance
+               of a cached or uncached filestore
         :param defer: Function yielded by open() context
                manager of :class:`toil.DeferredFunctionManager`, which is called to
                register deferred functions.
+        :param kwargs: Catch-all to accept superfluous arguments passed by old
```
DailyDreaming, line 2374: Can we create an issue in Cactus for this (and a subsequent toil issue to remove this after this PR is merged)?  I really dislike having this and it may cause unexpected errors on cactus's end if they're expecting certain behavior from their placebo kwargs.

src/toil/job.py
```diff
@@ -1402,41 +2351,61 @@ def _executor(self, jobGraph, stats, fileStore):
                 )
             )
 
-    def _runner(self, jobGraph, jobStore, fileStore, defer):
+    def _runner(self, jobStore=None, fileStore=None, defer=None, **kwargs):
         """
         This method actually runs the job, and serialises the next jobs.
+        
+        It marks the job as completed (by clearing its command) and creates the
```
DailyDreaming, line 2358: Thanks again for the clear, well-explained docstring additions.

src/toil/job.py
```diff
@@ -1323,49 +2207,113 @@ def _serialiseFirstJob(self, jobStore):
...
+        May be (but currently is not) overridden by specialized Toil-internal jobs.
 
-    def _run(self, jobGraph, fileStore):
+        Should not be overridden by non-Toil code!
+
+        Despite this, it has been overridden by non-Toil code, so we keep it
+        around and use a hardened kwargs-based interface to try and tolerate
+        bad behavior by workflows (e.g. Cactus).
+
+        When everyone has upgraded to a sufficiently new Cactus, we can remove
```
DailyDreaming, line 2299: :tada:

src/toil/job.py
```diff
@@ -1323,49 +2207,113 @@ def _serialiseFirstJob(self, jobStore):
...
+
+                # TODO: We ought to just unpickle straight from a streaming read
+        finally:
+            # Clean up the file
+            if os.path.exists(filename):
+                os.unlink(filename)
+            # Clean up the directory we put it in
+            # TODO: we assume nobody else put anything in the directory
+            if os.path.exists(directory):
+                os.rmdir(directory)
```
DailyDreaming, line 2283:
```
            # TODO: we assume nobody else put anything in the directory
            if os.path.exists(directory):
                os.rmdir(directory)
```

I think it might be worth preemptively using `shutil.rmtree` here as it might (or might not) prevent this code's spirit from reincarnating later as a bug and I see no downside to doing so.

src/toil/job.py
```diff
@@ -1323,49 +2207,113 @@ def _serialiseFirstJob(self, jobStore):
...
+                job = cls._unpickle(userModule, fileHandle, requireInstanceOf=Job)
+                # Fill in the current description
+                job._description = jobDescription
+                
+                # Set up the registry again, so children and follow-ons can be added on the worker
+                job._registry = {job.jobStoreID: job}
+                
+                return job
+
+                # TODO: We ought to just unpickle straight from a streaming read
```
DailyDreaming, line 2275: Could a toil issue be made for this?

src/toil/job.py
```diff
@@ -1045,276 +1926,279 @@ def _collectAllSuccessors(self, visited):
...
+        # Prohibit cycles and multiple roots
         self.checkJobGraphForDeadlocks()
-
-        #Create the jobGraphs for followOns/children
-        with jobStore.batch():
-            jobsToJobGraphs = self._makeJobGraphs(jobGraph, jobStore)
-        #Get an ordering on the jobs which we use for pickling the jobs in the
-        #correct order to ensure the promises are properly established
+        
+        # Make sure everybody in the registry is registrered with the job store
```
DailyDreaming, line 2124: `registrered`

src/toil/job.py
```diff
@@ -1045,276 +1926,279 @@ def _collectAllSuccessors(self, visited):
...
+        Used to save the initial job graph containing the root job of the workflow.
+        
+        :param toil.jobStores.abstractJobStore.AbstractJobStore jobStore: The job store
+            to save the jobs into.
+        :param bool saveSelf: Set to True to save this job along with its children,
+            follow-ons, and services, or False to just save the children, follow-ons,
+            and services and to populate the return value.
+        :param returnValues: The collection of values returned when executing
+            the job (or starting the service the job is hosting). If saveSelf
+            is not set, will be used to fulfil the job's return value promises.
```
DailyDreaming, line 2118: `fulfil` -> `fulfill`

src/toil/job.py
```diff
@@ -1045,276 +1926,279 @@ def _collectAllSuccessors(self, visited):
...
+                logger.debug("Visit job %s", job.jobStoreID)
+                
+                for otherID in itertools.chain(job.description.followOnIDs, job.description.childIDs):
+                    if otherID in self._registry:
+                        # Stack up descendants so we process children and then follow-ons.
+                        # So stack up follow-ons deeper
+                        todo.append(self._registry[otherID])
+                        logger.debug("Queue successor %s", otherID)
+                    else:
+                        logger.debug("Unregistered successor %s", otherID)
```
DailyDreaming, line 1985: These log messages might be noisy... ?  I'd consider removing them.  Though I have not experimented with this to know whether they are prohibitively noisy, after https://github.com/DataBiosphere/toil/issues/3274 , I think we should carefully consider logs that do not explicitly notify the user of either unexpected behavior, or a sort of throttled regular status message, since almost all log hunting seems to be searching for unexpected behavior.

Maybe better would be just a log message every ten minutes with the number of successors queued/unregistered in the last ten minutes and the total overall running/queued?  Though that might be worth making a separate PR/issue rather than continuing to add here.

DailyDreaming (commented):
I believe I've finished the other half of this review.

Overall, this looks great and is a vast improvement over what we had before, both in terms of readability and how things are structured.

Adding type hinting would improve things quite a bit, but would also require a large amount of effort added to an already large PR, so I think we should probably just make an issue for it for now.

Please see the comments, and once they're addressed, I'll do one more pass over.  Thanks a ton for doing this, this really is awesome work.

src/toil/worker.py
```diff
@@ -346,37 +363,51 @@ def workerScript(jobStore, config, jobName, jobStoreID, redirectOutputToLogFile=
...
-                            job._runner(jobGraph=jobGraph, jobStore=jobStore, fileStore=fileStore, defer=defer)
+                            
+                            # Run the job, save new successors, and set up
+                            # locally (but don't commit) successor
+                            # relationships and job completion.
+                            # Pass everything as name=value because Cactus
+                            # likes to override _runner when it shouldn't and
+                            # it needs some hope of finding the arguments it
+                            # wants across multiple Toil versions. We also
+                            # still pass a jobGraph argument to placate old
```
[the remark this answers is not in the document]
adamnovak: I have a relevant PR at https://github.com/ComparativeGenomicsToolkit/cactus/pull/327/files. But I opened https://github.com/ComparativeGenomicsToolkit/cactus/issues/336 just now too.

src/toil/serviceManager.py
```diff
@@ -166,81 +196,135 @@ def shutdown(self):
...
+                            # Send the service JobDescription off to be started
+                            logger.debug('Service manager is starting service job: %s, start ID: %s', serviceJobDesc, serviceJobDesc.startJobStoreID)
+                            serviceJobsToStart.put(serviceJobDesc)
                 except Empty:
                     # No new jobs that need services scheduled.
                     pass
+                    
+                pendingServiceCount = len(servicesThatAreStarting)
+                if pendingServiceCount > 0 and logLimiter.throttle(False):
+                    logger.info('%d services are starting...', pendingServiceCount)
```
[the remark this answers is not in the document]
adamnovak: Sure, sorry.

src/toil/resource.py
```diff
@@ -539,7 +539,7 @@ def globalize(self):
         except IOError as e:
             if e.errno == errno.ENOENT:
                 if self._runningOnWorker():
-                    log.warning("Can't globalize module %r.", self)
+                    log.debug("Can't globalize module %r.", self)
```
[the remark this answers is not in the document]
adamnovak: I'll just drop it altogether. If we ever find a problem we can put it back.

src/toil/serviceManager.py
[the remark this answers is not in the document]
DailyDreaming: No worries!  :sweat_smile:

src/toil/lib/docker.py
```diff
@@ -382,10 +383,13 @@ def containerIsRunning(container_name, timeout=365 * 24 * 60 * 60):
...
 def getContainerName(job):
-    """Create a random string including the job name, and return it."""
-    return '--'.join([str(job),
-                      base64.b64encode(os.urandom(9), b'-_').decode('utf-8')]) \
-        .replace("'", '').replace('"', '').replace('_', '')
+    """
+    Create a random string including the job name, and return it. Name will
+    match [a-zA-Z0-9][a-zA-Z0-9_.-]
+    """
+    parts = ['toil', str(job.description), base64.b64encode(os.urandom(9), b'-_').decode('utf-8')]
```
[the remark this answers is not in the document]
adamnovak: `job.description` is a `JobDescription`. We could truncate its stringification, or just pick just one of its various stored job names to represent it here. The stringified jobs can get pretty long for CWL where we include entire paths. Is there a length limit we need to obey?

src/toil/jobStores/googleJobStore.py
adamnovak: Sure.

src/toil/jobStores/googleJobStore.py
adamnovak: I opened https://github.com/DataBiosphere/toil/issues/3301

src/toil/jobStores/abstractJobStore.py
adamnovak: When I caused the kind of problem I'm trying to check against here, `haveJob` was being called with single-character strings, because elsewhere I had used a single job ID string where I meant to have used a list of them. So a type check here in `haveJob` wouldn't help, and if I was smart enough to put a type check where the problem was, I wouldn't have caused it.

adamnovak: Python weirdly doesn't distinguish between characters and strings as types.

src/toil/jobStores/abstractJobStore.py
adamnovak: The part about "the information from" was from a different design, so I am fixing that. `loadJob` wants the whole JobDescription. When I wrote this I had it taking just the command.

src/toil/job.py
adamnovak: I have the Cactus PR and issue already I think, but I opened https://github.com/DataBiosphere/toil/issues/3302 here for the cleanup after Cactus changes and people upgrade.

Cactus doesn't do anything with the args right now; they just insist on having them to hand back to us.

src/toil/job.py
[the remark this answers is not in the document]
adamnovak: Converting all our docstring types to PEP484 might be A Project, but I've never really fully understood the docstring type syntax anyway, because I don't use an editor that speaks it. PEP484 looks like it actually comes with tooling that checks everything.

We should probably change our contributing doc and open an issue to move over. Maybe there's a tool that can convert the properly-formatted docstring type annotations?

src/toil/job.py
adamnovak: I write these because I'm totally lost otherwise.

src/toil/job.py
adamnovak: OK, I'll do that.

src/toil/job.py
adamnovak: I made https://github.com/DataBiosphere/toil/issues/3303

src/toil/job.py
adamnovak: Oh yeah, I think I left a bunch of debugging in here that we don't really need. I'll pull it out.

I don't think handling these as events to report on really makes sense; they just tell you about Toil's traversal of the graphs of jobs you are adding, and I only added them to make sure the traversal ported over properly to the `JobDescription` class and the ID references.

adamnovak:
OK @DailyDreaming, I think I addressed your comments.

I'm not volunteering to add PEP484 type hinting as part of this, but I do think we should learn it and adopt it overall.

src/toil/lib/docker.py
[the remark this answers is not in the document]
DailyDreaming: Well, that's potentially `f"'{self.jobName}' {self.unitName} {self.jobStoreID}"`, so I was thinking maybe the standard 255 char limit for a filename?

DailyDreaming (approved):
LGTM.  Thanks for addressing all of the comments!  I agree PEP484 type hinting is a weighty task and should be handled separately.

Issue #3167, opened 2020-08-26
sub() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## String sub(String, String, String)

Given 3 String parameters `input`, `pattern`, `replace`, this function will replace any occurrence matching `pattern` in `input` by `replace`.
`pattern` is expected to be a regular expression. Details of regex evaluation will depend on the execution engine running the WDL.

Example:

  String chocolike = "I like chocolate when it's late"
  String chocolove = sub(chocolike, "like", "love") # I love chocolate when it's late
  String chocoearly = sub(chocolike, "late", "early") # I like chocoearly when it's early
  String chocolate = sub(chocolike, "late$", "early") # I like chocolate when it's early

The sub function will also accept `input` and `replace` parameters that can be coerced to a String (e.g. File). This can be useful to swap the extension of a filename for example.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#string-substring-string-string

 - [x] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
    - Implemented here: https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py#L226
 - [x] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
task example {
  input {
    File input_file = "my_input_file.bam"
    String output_file_name = sub(input_file, "\\.bam$", ".index") # my_input_file.index
  }
  command {
    echo "I want an index instead" > ${output_file_name}
  }

  output {
    File outputFile = output_file_name
  }
}
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-620)
┆Issue Number: TOIL-620
┆Epic: WDL Standard Library Implementation

PR #3252, opened 2020-10-01
Add WDL builtin test 'sub'.

w-gao:
#3167.

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import shutil
 import uuid
+from toil.wdl.wdl_functions import sub
 from toil.wdl.wdl_functions import ceil
 from toil.wdl.wdl_functions import floor
 from toil.wdl.wdl_functions import read_lines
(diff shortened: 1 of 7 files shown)


src/toil/test/wdl/builtinTest.py
```diff
@@ -234,6 +243,14 @@ def check_function(self, function_name, cases, expected_result):
                 self.assertEqual(result, expected_result)
                 shutil.rmtree(output_dir)
 
+    def test_sub(self):
+        # this workflow swaps the extension of a TSV file to CSV, with String and File inputs.
+        self.check_function('sub', cases=['as_input'], expected_result='src/toil/test/wdl/test.csv')
+
+        # NOTE: the result differs from Cromwell since we copy the file to the file store without
+        # preserving the path. Cromwell would return 'src/toil/test/wdl/test.csv' instead.
+        self.check_function('sub', cases=['as_input_with_file'], expected_result='test.csv')
```
w-gao, line 252: I don't think we can produce the same behavior with our current implementation?

src/toil/wdl/wdl_functions.py
```diff
@@ -228,15 +229,24 @@ def process_infile(f, fileStore):
         raise RuntimeError('Error processing file: '.format(str(f)))
 
 
-def sub(a, b, c):
-    if isinstance(a, tuple):
-        a = a[1]
-    if isinstance(a, tuple):
```
w-gao, line 234: I assume `a` is a typo?

w-gao (commented):
Apparently I'm reviewing my own PR.. I added some questions that came up during development.

src/toil/wdl/wdl_functions.py
DailyDreaming: Yes, it is.

src/toil/test/wdl/builtinTest.py
DailyDreaming: Oh, interesting, you're right.  I don't think we need to address this immediately, and the comment should be enough for now, especially since `sub()` is still doing exactly what it should.

DailyDreaming (approved):
LGTM.  I don't think that we can account for Toil's path mangling right now.  Thanks for pointing it out and adding a test appropriate to the edge case!

Issue #3166, opened 2020-08-26
size() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Float size(File, [String])

Given a `File` and a `String` (optional), returns the size of the file in Bytes or in the unit specified by the second argument.

Supported units are KiloByte ("K", "KB"), MegaByte ("M", "MB"), GigaByte ("G", "GB"), TeraByte ("T", "TB") as well as their [binary version](https://en.wikipedia.org/wiki/Binary_prefix) "Ki" ("KiB"), "Mi" ("MiB"), "Gi" ("GiB"), "Ti" ("TiB").
Default unit is Bytes ("B").

### Acceptable compound input types

Varieties of the `size` function also exist for the following compound types. The `String` unit is always treated the same as above. Note that to avoid numerical overflow, very long arrays of files should probably favor larger units.

* `Float size(File?, [String])`: Returns the size of the file, if specified, or 0.0 otherwise.
* `Float size(Array[File], [String])`: Returns the sum of sizes of the files in the array.
* `Float size(Array[File?], [String])`: Returns the sum of sizes of all specified files in the array.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#float-sizefile-string

 - [x] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
    - Implemented here: https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py#L443
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
task example {
  input {
    File input_file
  }

  command {
    echo "this file is 22 bytes" > created_file
  }

  output {
    Float input_file_size = size(input_file)
    Float created_file_size = size("created_file") # 22.0
    Float created_file_size_in_KB = size("created_file", "K") # 0.022
  }
}
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-619)
┆Issue Number: TOIL-619
┆Epic: WDL Standard Library Implementation

PR #3255, opened 2020-10-05
Add WDL builtin test 'size'. 

w-gao:
#3166.

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -254,8 +254,11 @@ class WdlStandardLibraryWorkflowsTest(ToilTest):
     def test_size(self):
         self.check_function('size', cases=['as_command'], expected_result='19.0')
 
-        # the output file contains 22 bytes in bytes, K, and Ki separated with spaces.
-        # NOTE: Cromwell returns '22.0 0.021484375 0.021484375' where the binary version is returned for K.
+        # this workflow outputs the size of a 22-byte file in 'B', 'K', and 'Ki' separated with spaces.
+
+        # NOTE: Cromwell treats the decimal and binary units (e.g.: 'K' and 'Ki') the same, which differs from
+        # the spec (https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#float-sizefile-string).
+        # The correct output should be '22.0 0.022 0.021484375' not '22.0 0.021484375 0.021484375'
         self.check_function('size', cases=['as_output'], expected_result='22.0 0.022 0.021484375')
 
     def test_ceil(self):
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -251,6 +251,13 @@ class WdlStandardLibraryWorkflowsTest(ToilTest):
         # preserving the path. Cromwell would return 'src/toil/test/wdl/test.csv' instead.
         self.check_function('sub', cases=['as_input_with_file'], expected_result='test.csv')
 
+    def test_size(self):
+        self.check_function('size', cases=['as_command'], expected_result='19.0')
+
+        # the output file contains 22 bytes in bytes, K, and Ki separated with spaces.
+        # NOTE: Cromwell returns '22.0 0.021484375 0.021484375' where the binary version is returned for K.
+        self.check_function('size', cases=['as_output'], expected_result='22.0 0.022 0.021484375')
+
     def test_ceil(self):
         self.check_function('ceil', cases=['as_input', 'as_command'], expected_result='12')
 
--- a/src/toil/wdl/wdl_functions.py
+++ b/src/toil/wdl/wdl_functions.py
@@ -486,7 +486,7 @@ def is_number(s):
         return False
 
 
-def size(f: Union[tuple, List[tuple]] = None,
+def size(f: Optional[Union[tuple, List[tuple]]] = None,
          unit: Optional[str] = 'B',
          fileStore: Optional[AbstractFileStore] = None) -> float:
     """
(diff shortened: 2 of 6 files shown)


w-gao:
I'm not sure if I should move [the function test](https://github.com/DataBiosphere/toil/blob/master/src/toil/test/wdl/toilwdlTest.py#L135-L151) to `builtinTest.py`, since it isn't quite a standalone function (it requires a file store file in order to work).

DailyDreaming:
@w-gao It's fine to leave the `testFn_Size()` test where it is since it's not really a unit test.

w-gao:
@DailyDreaming Should I be adding a unit test then? If so, it would have to set up a Toil workflow as well, so that I can test the varieties of the function with a file store.

DailyDreaming:
@w-gao I don't think it needs it, so long as the current unit test passes.  This looks good so far.  I'm actually going to ask a coworker of mine ( @xbrianh ) who's volunteered to help with some light PR reviews if he could take a look and see what he thinks.  I think that you write good code, so this will be an easier introduction for him to the toil code base.

src/toil/test/wdl/builtinTest.py
```diff
@@ -251,6 +251,13 @@ def test_sub(self):
         # preserving the path. Cromwell would return 'src/toil/test/wdl/test.csv' instead.
         self.check_function('sub', cases=['as_input_with_file'], expected_result='test.csv')
 
+    def test_size(self):
+        self.check_function('size', cases=['as_command'], expected_result='19.0')
+
+        # the output file contains 22 bytes in bytes, K, and Ki separated with spaces.
+        # NOTE: Cromwell returns '22.0 0.021484375 0.021484375' where the binary version is returned for K.
```
xbrianh, line 258: This comment is difficult to understand. Can context be added for newbies (or people that haven't looked at the code for a long time)?

src/toil/wdl/wdl_functions.py
```diff
@@ -486,15 +486,32 @@ def is_number(s):
         return False
 
 
-def size(f, unit='B', fileStore=None):
+def size(f: Union[tuple, List[tuple]] = None,
```
xbrianh, line 489: Since `f` is initialized to None, the correct type hint is:
```
Optional[Union[tuple, List[tuple]]]
```

src/toil/test/wdl/builtinTest.py
xbrianh, line 258: This comment is difficult to understand. Can context be added for newbies (or people that haven't looked at the code for a long time)?

src/toil/wdl/wdl_functions.py
```diff
@@ -486,15 +486,32 @@ def is_number(s):
...
-    Returns the size of a file in bytes.
+    Given a `File` and a `String` (optional), returns the size of the file in Bytes
+    or in the unit specified by the second argument.
 
-    :param f: Filename
-    :param unit: Return the byte size in these units (gigabytes, etc.).
-    :return:
+    Supported units are KiloByte ("K", "KB"), MegaByte ("M", "MB"), GigaByte
+    ("G", "GB"), TeraByte ("T", "TB") as well as their binary version "Ki" ("KiB"),
+    "Mi" ("MiB"), "Gi" ("GiB"), "Ti" ("TiB"). Default unit is Bytes ("B").
```
xbrianh, line 498: Please consider including a brief discussion of "decimal" and "binary" in this context. For instance, the decimal units KB, MB and GB are powers of 1000. Binary units, KiB, MiB, and GiB are powers of 1024.

src/toil/wdl/wdl_functions.py
```diff
@@ -486,15 +486,32 @@ def is_number(s):
...
+
+    WDL syntax: Float size(File, [String])
+    Varieties:  Float size(File?, [String])
+                Float size(Array[File], [String])
+                Float size(Array[File?], [String])
     """
+    if not f:
```
xbrianh, line 505: Is it OK that `size(False)` evaluates to 0?

src/toil/wdl/wdl_functions.py
```diff
@@ -486,15 +486,32 @@ def is_number(s):
...
+        return 0
+
     divisor = return_bytes(unit)
+
+    if isinstance(f, list):
+        total_size = sum(file[0].size for file in f if isinstance(file, tuple))
```
xbrianh, line 511: Which members of `f` are expected to be dropped from the size calculation?

xbrianh (commented):
Are `src/toil/test/wdl/standard_library/size.json` and `src/toil/test/wdl/standard_library/size_as_command.wdl` related to the other changes in this diff? If not consider moving them to a separate PR.

Otherwise this looks reasonable to me, but see comments.

src/toil/wdl/wdl_functions.py
w-gao: I intended this to filter out anything that is not a File. Maybe I should let it error instead?

xbrianh: My naive assumption is that something has gone wrong if a mixture of files and not-files is passed into this method, so letting it error is the correct action.

@DailyDreaming should probably pipe in here.

w-gao: Right, I modified the code a bit to let `process_infile` to validate and process the input.

w-gao:
@xbrianh `src/toil/test/wdl/standard_library/size.json` and `src/toil/test/wdl/standard_library/size_as_command.wdl` are part of the standard library test. They are used in [`test_size()`](https://github.com/DataBiosphere/toil/blob/35ff582e6b78fc50cd3d0ae52c01890976eb0f20/src/toil/test/wdl/builtinTest.py#L255).

src/toil/wdl/wdl_functions.py
DailyDreaming: That seems like a sane check.  I agree with Brian.

src/toil/wdl/wdl_functions.py
DailyDreaming: That depends on Cromwell's behavior.  I agree that we should probably checking if this is a String/File/Tuple/List.

xbrianh (approved):
lgtm

DailyDreaming (approved):

PR #3259, opened 2020-10-07
Make kubernetes & botocore packages fully optional

### LINKED ISSUE #3251: Kubernetes import in retry module needs to be able to fail
https://github.com/DataBiosphere/toil/blob/79c4c43c1f76484c99ec550aee609c6579a6a77c/src/toil/lib/retry.py#L133

This will break the retry module when the `[kubernetes]` extra isn't installed.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-678)
┆Issue Number: TOIL-678

mr-c:
Fixes #3251

### DIFF
--- a/src/toil/lib/retry.py
+++ b/src/toil/lib/retry.py
@@ -130,19 +130,25 @@ import http.client
 import urllib.error
 import urllib3.exceptions
 import sqlite3
-import kubernetes.client.rest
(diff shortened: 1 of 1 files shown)


DailyDreaming (approved):
LGTM.  Thanks for the fix!

PR #3262, opened 2020-10-07
Make promises indexable with more path components to fix #1695

### LINKED ISSUE #1695: Can't access multiple values from a service job
In a slight inconsistency with the `addChild` and `addFollowOn` methods, the `addService` method returns a Promise rather than a Job. Correct me if I'm wrong, but I think this makes it impossible, using only the public API, to get a specific return value with `.rv(path)` from a service.

Obviously changing the return value to Job would be a pretty annoying breaking change for everything that uses services. Maybe it'd be possible to add a new method on Promise to create a new Promise given a path, much like `.rv()`?

Currently I'm using private Job members to work around this, which gets the job done, but doesn't seem like the cleanest way of doing things.

adamnovak:
This lets a Promise produce another Promise with `[]` indexing. So if a job returns a tuple and you have a Promise for the whole thing, you can do:

```
itemPromise = wholeValuePromise[1]
```

This will fix #1695 and let us remove one of Cactus's hard-to-support hacks where it reaches inside `Job._services`.

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -1822,6 +1822,18 @@ class Promise(object):
         self.job = job
         self.path = path
 
+    def __getitem__(self, path_tuple):
+        """
+        Given a Promise for a certain path, get a Promise for that path
+        extended with the additional path elements from path_tuple.
+
+        For example, if you have a job `j` that returns a tuple of three
+        elements, and you have a promise p = j.rv() for the whole tuple, you
+        can do p2 = p[1] to get a new promise p2 for element 1 of the tuple.
+        """
+
+        return Promise(self.job, tuple(path + path_tuple))
+
     def __reduce__(self):
         """
         Called during pickling when a promise (an instance of this class) is about to be be


adamnovak:
This probably needs a test.

src/toil/job.py
```diff
@@ -1822,6 +1822,18 @@ def __init__(self, job, path):
...
+        """
+        Given a Promise for a certain path, get a Promise for that path
+        extended with the additional path elements from path_tuple.
+
+        For example, if you have a job `j` that returns a tuple of three
+        elements, and you have a promise p = j.rv() for the whole tuple, you
+        can do p2 = p[1] to get a new promise p2 for element 1 of the tuple.
+        """
+
+        return Promise(self.job, tuple(path + path_tuple))
```
DailyDreaming, line 1835: Should `path` be `self.path`?

adamnovak: Indeed it should. This is why we need a test.

adamnovak:
I'm going to close this on the theory that I'm not going to finish it.

PR #3272, opened 2020-10-13
Fix None check when parsing WDL JSON file

w-gao:
In WDL, I noticed that `0` in the JSON file is interpreted as `None` when I was testing `range()`. 

I don't think this would break anything from before (?), but  `''`, `false`, `0`, and `[]` in the JSON file would be parsed to the correct value in the compiled Python script.


adamnovak (approved):
This looks like a good idea to me.

Issue #3269, opened 2020-10-12
Double Mem functionality with batch schedulers toil-cwl-runner


drkennetz:
Referencing #2042 here. This is less of an issue and more of an enhancement. On batch systems such as HPC environments, it would be incredibly helpful to have a `--doublemem` flag for workflows. 

In short, the doublemem flag would be passed at runtime on the first instantiation of the workflow. This flag would tell toil to capture failure reasons of jobs, and if the failure reason corresponded to a TERM_MEMLIMIT or TERM_ENOMEM, the specific step that failed would be retried with double the initial requested memory. This would be helpful for genomics workflows with specific steps that did not have well-defined memory profiles (I think most commonly seen with Structural Variants or even STAR).

An example of this would be something like:
```
steps:
  bwa:
    requirements:
      ResourceRequirement:
        ramMax: 5000
...
```
This would likely fail on about 50% of bwa jobs because bwa can use anywhere from 4-10G depending on samples / version of bwa. by running this with `--doublemem`, this would capture failures due to TERM_MEMLIMIT (this might be LSF specific), so for your 50% of failed jobs, they would automatically be restarted with 10G, making the section look like:
```
steps:
  bwa:
    requirements:
      ResourceRequirement:
        ramMax: 10000
...
```
Do you guys have thoughts on this?

Thanks,
Dennis

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-689)
┆Issue Number: TOIL-689

drkennetz:
I would be happy to work on this myself as it is something that I want implemented... haha.

adamnovak:
The place to look for this might be here (although it's moving in #3250):
https://github.com/DataBiosphere/toil/blob/cf2984e9019e8c59528e2f462236cf628dc0bf10/src/toil/jobGraph.py#L135

We already have code to increase memory to `--defaultMemory` when a job fails.

The trouble might be getting the information that a job ran out of memory specifically into that function, so that it can do more aggressive memory scaling without wasting time when the problem is not memory. We might need to define a more specific `BatchJobExitReason` [here](https://github.com/DataBiosphere/toil/blob/cf2984e9019e8c59528e2f462236cf628dc0bf10/src/toil/batchSystems/abstractBatchSystem.py#L51-L56) and implement it in the batch systems that can recognize an out-of-memory error.

drkennetz:
[...]

PR #3276, opened 2020-10-15
Added doubleMem functionality for lsf jobs that die due to reaching imposed memory limit

drkennetz:
Often in genomics pipelines, especially for post-processing, memory limits of programs are not well-defined and can be dependent on data. This PR will add a memory doubling functionality if the flag `--doubleMem` is included on the command line for the LSF scheduler which will double the memory of and resubmit jobs that die due to reaching the LSF imposed memory limit. This functionality is incredibly useful for on-prem clusters. This addresses the issue opened by me: #3269

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -107,6 +107,7 @@ class Config(object):
         # Retrying/rescuing jobs
         self.retryCount = 1
         self.enableUnlimitedPreemptableRetries = False
+        self.doubleMem = False
         self.maxJobDuration = sys.maxsize
         self.rescueJobsFrequency = 3600
 
@@ -272,6 +273,7 @@ class Config(object):
         # Retrying/rescuing jobs
         setOption("retryCount", int, iC(1))
         setOption("enableUnlimitedPreemptableRetries")
+        setOption("doubleMem")
         setOption("maxJobDuration", int, iC(1))
         setOption("rescueJobsFrequency", int, iC(1))
 
@@ -554,6 +556,10 @@ def _addOptions(addGroupFn, config):
                 help=("If set, preemptable failures (or any failure due to an instance getting "
                       "unexpectedly terminated) would not count towards job failures and "
                       "--retryCount."))
+    addOptionFn("--doubleMem", dest="doubleMem", action='store_true', default=False,
+                help=("If set, batch jobs which die to reaching memory limit on batch schedulers "
+                      "will have their memory doubled and they will be retried. The remaining "
+                      "retry count will be reduced by 1. Currently supported by LSF."))
     addOptionFn("--maxJobDuration", dest="maxJobDuration", default=None,
                 help=("Maximum runtime of a job (in seconds) before we kill it "
                       "(this is a lower bound, and the actual time before killing "
--- a/src/toil/jobGraph.py
+++ b/src/toil/jobGraph.py
@@ -152,7 +152,7 @@ class JobGraph(JobNode):
             self._memory = self.memory * 2
             logger.warning("We have doubled the memory of the failed job %s to %s bytes due to doubleMem flag",
                            str(self), str(self.memory))
-        if self.memory < config.defaultMemory:
+        if self.memory < config.defaultMemory and not config.doubleMem:
             self._memory = config.defaultMemory
             logger.warning("We have increased the default memory of the failed job %s to %s bytes",
                            self, self.memory)
--- a/src/toil/jobGraph.py
+++ b/src/toil/jobGraph.py
@@ -148,6 +148,10 @@ class JobGraph(JobNode):
         # Set the default memory to be at least as large as the default, in
         # case this was a malloc failure (we do this because of the combined
         # batch system)
+        if exitReason == BatchJobExitReason.MEMLIMIT and config.doubleMem:
+            self._memory = self.memory * 2
+            logger.warning("We have doubled the memory of the failed job %s to %s bytes due to doubleMem flag",
+                           str(self), str(self.memory))
         if self.memory < config.defaultMemory:
             self._memory = config.defaultMemory
             logger.warning("We have increased the default memory of the failed job %s to %s bytes",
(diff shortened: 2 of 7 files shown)


adamnovak:
I don't think using a magic exit status and then rewriting the exit reason in the leader is the right way to do this. When the job runs out of memory, the LSF batch system should return an actually-correct `UpdatedBatchJobInfo` from `getUpdatedBatchJob` with the right exit reason set already.

I think what we want to do is change [the semantics defined for `getJobExitCode`](https://github.com/DataBiosphere/toil/blob/9a2cee472d0e2da2fb391314b1d1a9d666ad1425/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L281), so that it can return both the code and the reason (or maybe so that it can return either the code or an instance of `BatchJobExitReason`, which I think we could tell apart with `isinstance`), and then update [here](https://github.com/DataBiosphere/toil/blob/9a2cee472d0e2da2fb391314b1d1a9d666ad1425/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L197) to fill in the reason if it is available, and (if adding more return values) fix the other `getJobExitCode` implementations for the other grid engine batch systems to still work.

src/toil/common.py
```diff
@@ -554,6 +556,10 @@ def _addOptions(addGroupFn, config):
                 help=("If set, preemptable failures (or any failure due to an instance getting "
                       "unexpectedly terminated) would not count towards job failures and "
                       "--retryCount."))
+    addOptionFn("--doubleMem", dest="doubleMem", action='store_true', default=False,
```
adamnovak, line 559: Since this adds another option it should also update https://github.com/DataBiosphere/toil/blob/c526b6fc20a81b9e04bad46d14840511666d8c3e/docs/running/cliOptions.rst

src/toil/jobGraph.py
```diff
@@ -148,6 +148,10 @@ def setupJobAfterFailure(self, config, exitReason=None):
         # Set the default memory to be at least as large as the default, in
         # case this was a malloc failure (we do this because of the combined
         # batch system)
+        if exitReason == BatchJobExitReason.MEMLIMIT and config.doubleMem:
+            self._memory = self.memory * 2
+            logger.warning("We have doubled the memory of the failed job %s to %s bytes due to doubleMem flag",
+                           str(self), str(self.memory))
         if self.memory < config.defaultMemory:
```
adamnovak, line 155: It looks like memory can be both doubled and upped to the default in the same call. That might confuse people; it should do only one or the other.

adamnovak (changes requested):
In addition to changing how the reason is arrived at, I found a couple other things I think we want to change before taking this.

src/toil/jobGraph.py
drkennetz: This is true, if the default is greater than double the step's memory, it will be bumped to the default's memory. I actually thought that behavior seemed okay. What will end up happening if the memory is like 500M and the default is 2G and a job fails due to memory:

1a. 500M -> 1G from the memory double
1b. 1G -> 2G due to default
 Job Fails again due to memory:
2a. 2G -> 4G
Job Fails again:
3a. 4G -> 8G
...
I can make these exclusive conditions if you think that would be better.

drkennetz:
@adamnovak yeah I didn't really think this was the ideal way to go about this, but I think I was still missing some pieces of the puzzle when trying to implement. I figured with the PR you might be able to give me some valuable feedback to finish the changes, and I was right! Thanks for the review. I'll go ahead and make the recommended changes and request a review.

drkennetz:
@adamnovak I went with the logic of returning an instance of `BatchJobExitReason` in lsf then adding some additional logic rather than increasing the number of return values. This gets rid of the magic exit code and doesn't really change the flow. I added:
```
if exitStatus == 0 and exitReason == None:
```
because the exitReason is being captured now. This isn't entirely necessary because I assign an exitCode of 1 if the return type isinstance BatchJobExitReason simply because I am no longer returning the exit code by lsf and I did want to maintain the fact that there was actually an exit code. Let me know what you think.

adamnovak (changes requested):
This looks great! I think it still needs a little documentation here on the new getJobExitCode semantics, though:

https://github.com/DataBiosphere/toil/blob/37d5e395c2a518d86f4b91b274cd693774d6ae09/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L285-L291

We should say `:rtype: int|toil.batchSystems.abstractBatchSystem.BatchJobExitReason`, and mention in the text that it gets either the exit code, or a BatchJobExitReason if something else happened to the job other than exiting. Since that new behavior isn't really suggested by the function name, it's important to document it.

adamnovak:
I've pulled this in for CI testing.

drkennetz:
@adamnovak That's awesome! Hoping for the best. Should I wait for the CI to finish before updating the docs? I think you make a good point. Not sure how pushing while CI is running will affect things.

drkennetz:
@adamnovak, I have updated the docs for `getJobExitCode`.

adamnovak (approved):
This all looks good to me!

DailyDreaming (approved):
LGTM.  Thanks for adding this!

drkennetz:
@DailyDreaming no problem! I think it could be a useful feature for all GridEngines and relatively easy to implement for other schedulers if they report failure due to imposed mem limits.

Do you guys resolve merge conflicts? I'm seeing that we have one now. Unlikely that they touch the same part of the file but maybe.

mr-c:
Worse case one of the committers can resolve the conflict, but if you don't mind doing it yourself, it would be appreciated!

mr-c:
I'm guessing https://github.com/DataBiosphere/toil/pull/3250 was the culprit :-P

drkennetz:
Since this PR is lsf specific and I know I can test it, I'll resolve it myself. It might take a second because I have to understand what was changed by converging these classes... haha. It shouldn't be too big of a problem though.

drkennetz:
Would it be easier for us all if I just closed this PR and opened a new one?

PR #3277, opened 2020-10-16
Add WDL builtin test 'range'.

### LINKED ISSUE #3196: range() implemented for WDL.

```
## Array[Int] range(Int)

Given an integer argument, the `range` function creates an array of integers of length equal to the given argument. For example `range(3)` provides the array: `(0, 1, 2)`.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#arrayint-rangeint

 - [x] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
    - Implemented as the built-in, so nothing to add.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-636)
┆Issue Number: TOIL-636

w-gao:
Closes #3196. 

`range()` is a Python built-in function but the return type is `range`, so I'm adding a new function to make sure the correct type is returned.

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -321,6 +321,9 @@ class WdlStandardLibraryWorkflowsTest(ToilTest):
         self.check_function('write_map', cases=['as_command'],
                             expected_result='key1\tvalue1\nkey2\tvalue2')
 
+    def test_range(self):
+        self.check_function('range', cases=['as_input'], expected_result='0\n1\n2\n3\n4\n5\n6\n7')
+
     def test_transpose(self):
         # this workflow writes a transposed 2-dimensional array as a TSV file.
         self.check_function('transpose', cases=['as_input'], expected_result='0\t3\n1\t4\n2\t5')
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -4,6 +4,8 @@ import os
 import subprocess
 import shutil
 import uuid
+from typing import Optional, List
+
 from toil.wdl.wdl_functions import sub
 from toil.wdl.wdl_functions import ceil
 from toil.wdl.wdl_functions import floor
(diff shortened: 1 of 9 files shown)


src/toil/wdl/wdl_functions.py
```diff
@@ -839,6 +839,19 @@ def write_map(in_map: Dict[str, str],
...
 
+def wdl_range(num: int) -> List[int]:
+    """
+    Given an integer argument, the range function creates an array of integers of
+    length equal to the given argument.
+
+    WDL syntax: Array[Int] range(Int)
+    """
+    assert isinstance(num, int) and num >= 0, \
+        f'range() requires an integer greater than or equal to 0 (but got {num})'
```
DailyDreaming, line 850: Nice!  Just tested this and I see you've implemented the exact same message that Cromwell gives.  Great attention to detail.

src/toil/test/wdl/builtinTest.py
```diff
@@ -321,7 +323,13 @@ def test_write(self):
         self.check_function('write_map', cases=['as_command'],
                             expected_result='key1\tvalue1\nkey2\tvalue2')
 
+    def test_range(self):
+        # NOTE: this test depends on write_lines().
+        self.check_function('range', cases=['as_input'], expected_result='0\n1\n2\n3\n4\n5\n6\n7')
```
DailyDreaming, line 328: Consider adding a test for `0` and `-1` as well.

For -1, `with self.assertRaises(AssertionError):` is useful.  `AssertionError` is fine, though if you want, it might be worth making it a `RuntimeError` or a WDL specific error class.

DailyDreaming (approved):
One suggestion, see below.  Otherwise looks good to me, great work as usual.

src/toil/test/wdl/builtinTest.py
w-gao: I added `WDLRuntimeError` for WDL-related runtime errors. I should probably replace the other asserts at some point. 

Also, it turns out `with self.assertRaises(...)` doesn't work. We're checking the output via subprocess, so `subprocess.CalledProcessError` is thrown when the command fails. Although, the subprocess exception seems to be consumed by the unittest `TestCase` so we end up with nothing to catch. 

I'll see if there's anything else I can do to add the test for `-1`.

DailyDreaming (approved):
This looks great.  Thanks for implementing the additional tests!

PR #3279, opened 2020-10-19
Automatically infer singularity for kubernetes

### LINKED ISSUE #3223: Kubernetes batch system doesn't provide Docker for CWL workflows that use Docker
On Gitter, it was reported that if you take our Kubernetes batch system and try to use it to run a CWL workflow that uses a DcokerRequirement, it will fail.

The Kubernetes batch system does not privilege and start up Docker inside the pod, because we haven't usually needed it: at UCSC we use user-mode Singularity to run containers when on Kubernetes. But `cwlrunner` insists on using Docker, and the Singularity code isn't part of Toil itself while Docker-running code is.

We should make it possible for the Kubernetes batch system to launch all workers as privileged, if the user asks for it, and to start up a Docker daemon before running jobs. Then, we can make the CWL execution code ask for this, either when it sees that the workflow is going to need Docker, all the time unless instructed not to, or maybe just when instructed to do so, depending on what seems most user-friendly.

The right way to do this would be a needs-docker requirement on the jobs, I think, but we don't really have the infrastructure to support one-off annotation requirements like that yet. So I think it has to be all or nothing.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-662)
┆Issue Number: TOIL-662

jonathanxu18:
Should fix #3223 once https://github.com/common-workflow-language/cwltool/pull/1350 is added to new cwltool version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ def runSetup():
     gcs = 'google-cloud-storage==1.6.0'
     gcs_oauth2_boto_plugin = 'gcs_oauth2_boto_plugin==1.14'
     apacheLibcloud = 'apache-libcloud==2.2.1'
-    cwltool = 'cwltool==3.0.20200807132242'
+    cwltool = 'cwltool==3.0.20201017180608'
     galaxyToolUtil = 'galaxy-tool-util'
     htcondor = 'htcondor>=8.6.0'
     kubernetes = 'kubernetes>=10, <11'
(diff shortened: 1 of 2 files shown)


mr-c:
@jonathanxu18 I've made a new release https://pypi.org/project/cwltool/3.0.20201017180608/ , so you'll want to update `setup.py` to refer to this

adamnovak (approved):
This looks like it should work!

PR #3280, opened 2020-10-20
CWL Test #236: Inplace update has side effect on directory content.

### LINKED ISSUE #3256: CWL conformance test 236 fails
┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-682)
┆Epic: CWL v1.2 support
┆Issue Number: TOIL-682

DailyDreaming:
This PR adds full support for CWL1.1 and allows us to pass conformance test 236.

This test has three jobs and toil trips at two points:

**First job**:

`mkdir stepDir1`

which is then sym-linked into the staging dir of the next job as `inp -> stepDir1`

**Second job**:

`touch inp/blurb`

So, problem one was `inp` didn't exist.  This was because CWL has a `_:` prefix for file paths which it uses to designate directories that do not exist yet, but should be created in place rather than copied over because they are empty.  We were failing to create the symlink because it only makes the symlink if the folder exists and the `_:` prefixed uri was still a token instead of a real directory.

We now resolve `_:` prefixes during the resolution of promises, though only for `_:file://` paths, as I believe that this will not work for other URIs without changes to `cwltool` itself, which creates the symlink here: https://github.com/common-workflow-language/cwltool/blob/main/cwltool/job.py#L169

I'm currently trying to decide if toil really should resolve these or if there's a better way to allow cwltool to handle this.

Finally, the **third job** just returns a one level deep directory listing of the output from step 1 and the output from step 2.

That leads to problem two, where the step 1 output directory is recorded as a token with attached metadata stating that the directory contains nothing (`cwljob['d1']['listing'] == []`).  In step 2, the directories are symlinked, so both get the `blurb` file but only the step 2 output now (correctly) has metadata telling toil that it contains the blurb file, but the output from step 1 still has the original empty listing, and so in step 3, even though the directory now actually has the file, the metadata still contains an empty listing and doesn't reevaluate.

So the last step reevaluates listings used as inputs to jobs which may have had alterations at some point prior.  This currently only reevaluates empty listings, which allows the test to pass, but I'm trying to come up with a more robust re-evaluation.  It also seems fairly expensive to get an arbitrarily deep directory listing twice, so we should probably only do this in one place if possible.

#236 now passes, but only without caching (which we allow to `xfail`, so tests should "pass").  So I guess that's a third issue to address.  I assume it's because I'm not exporting directories to the jobstore after I make them.

Closes #3256 .


DailyDreaming:
@adamnovak This appears to be passing, but fails with caching turned on (which we currently allow to fail with `xfail`).

My first thought is that caching might not handle symlinks very well?  I'll take a deeper look into getting caching working, but if you have any thoughts based on the PR description above off of the top of your head, I'd be grateful.

mr-c:
Congratulations @DailyDreaming on the progress here!

mr-c:
Is this ready for review?

DailyDreaming:
@mr-c Not yet.  I think it might need to be a bit more involved unfortunately (also, this is currently a hack to test where it breaks).  Peter originally introduced a `MutationManager` to avoid concurrency issues, which we neglect to provide, so I'm try to figure out the best way to work with it currently: https://github.com/common-workflow-language/cwltool/pull/317/files

DailyDreaming:
This now relies on: https://github.com/common-workflow-language/cwltool/pull/1376

src/toil/cwl/cwltoil.py
```diff
@@ -814,6 +882,18 @@ def writeGlobalFileWrapper(file_store: AbstractFileStore, fileuri: str) -> str:
     return file_store.writeGlobalFile(schema_salad.ref_resolver.uri_file_path(fileuri))
 
 
+def remove_empty_listings(rec: CWLObjectType) -> None:
+    if rec.get("class") != "Directory":
+        finddirs = []  # type: List[CWLObjectType]
+        visit_class(rec, ("Directory",), finddirs.append)
+        for f in finddirs:
+            remove_empty_listings(f)
```
mr-c, line 890: Interesting. Is it `cwltool` code or `cwltoil` code that judges a `class: Directory` based upon if there is a `listing` or not; and can't handle an empty `listing`?

setup.py
```diff
@@ -30,7 +30,7 @@ def runSetup():
     gcs = 'google-cloud-storage==1.6.0'
     gcs_oauth2_boto_plugin = 'gcs_oauth2_boto_plugin==1.14'
     apacheLibcloud = 'apache-libcloud==2.2.1'
-    cwltool = 'cwltool==3.0.20201116114821'
+    cwltool = 'cwltool>=3.0.20201116114821'
```
mr-c, line 33:
```suggestion
    cwltool = 'cwltool==3.0.20201121085451'
```

mr-c (approved):
Looking good! One question, but you could merge with my suggestions and I'd be okay with it.

src/toil/cwl/cwltoil.py
DailyDreaming: I don't think it's necessarily a problem on the cwltool end of things.

DailyDreaming:
@mr-c Thanks for the release and changes made!

Issue #3203, opened 2020-09-09
collect_by_key() implemented as WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Map[X,Array[Y]] collect_by_key(Array[Pair[X,Y]])

Given an Array consisting of Pairs, the `collect_by_key` function returns a Map in which the left elements of the Pairs are the keys and the right elements the values. The left element of the Pairs passed to `collect_by_key` must be a primitive type. The values will be placed in an Array to allow for multiple Pairs to produce the same key. The order of the keys in the Map is the same as the order in the Array based on their first occurence. The order of the elements in the resulting Arrays is the same as their occurence in the given Array of Pairs.
```

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
Array[Pair[String,Int]] x = [("a", 1), ("b", 2), ("a", 3)]
Array[Pair[String,Pair[File,File]]] y = [("a", ("a_1.bam", "a_1.bai")), ("b", ("b.bam", "b.bai")), ("a", ("a_2.bam", "a_2.bai"))]

Map[String,Array[Int]] xmap = collect_by_key(x) # {"a": [1, 3], "b": [2]}
Map[String,Array[Pair[File,File]]] ymap = collect_by_key(y) # {"a": [("a_1.bam", "a_1.bai"), ("a_2.bam", "a_2.bai")], "b": [("b.bam", "b.bai")]}
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-643)
┆Epic: WDL Standard Library Implementation
┆Issue Number: TOIL-643

Issue #3201, opened 2020-09-09
as_map() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Map[X,Y] as_map(Array[Pair[X,Y]])

Given an Array consisting of Pairs, the `as_map` function returns a Map in which the left elements of the Pairs are the keys and the right elements the values. The left element of the Pairs passed to `as_map` must be a primitive type. The order of the key/value pairs in the resulting Map is the same as the order of the Pairs in the Array.

In cases where multiple Pairs would produce the same key, the workflow will fail.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#mapxy-as_maparraypairxy

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
Array[Pair[String,Int]] x = [("a", 1), ("b", 2), ("c", 3)]
Array[Pair[String,Pair[File,File]]] y = [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]

Map[String,Int] xmap = as_map(x) # {"a": 1, "b": 2, "c": 3}
Map[String,Pair[File,File]] ymap = as_map(y) # {"a": ("a.bam", "a.bai"), "b": ("b.bam", "b.bai")}
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-641)
┆Epic: WDL Standard Library Implementation
┆Issue Number: TOIL-641

Issue #3200, opened 2020-09-09
as_pairs() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Array[Pair[X,Y]] as_pairs(Map[X,Y])

Given a Map, the `as_pairs` function returns an Array containing each element in the form of a Pair. The key will be the left element of the Pair and the value the right element. The order of the the Pairs in the resulting Array is the same as the order of the key/value pairs in the Map.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#arraypairxy-as_pairsmapxy

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
Map[String, Int] x = {"a": 1, "b": 2, "c": 3}
Map[String,Pair[File,File]] y = {"a": ("a.bam", "a.bai"), "b": ("b.bam", "b.bai")}

Array[Pair[String,Int]] xpairs = as_pairs(x) # [("a", 1), ("b", 2), ("c", 3)]
Array[Pair[String,Pair[File,File]]] ypairs = as_pairs(y) # [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-640)
┆Epic: WDL Standard Library Implementation
┆Issue Number: TOIL-640

Issue #3199, opened 2020-09-09
cross() implemented for WDL.


DailyDreaming:
Description of WDL function to be added:

```
## Array[Pair[X,Y]] cross(Array[X], Array[Y])

Return the cross product of the two arrays. Array[Y][1] appears before
Array[X][1] in the output.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#arraypairxy-crossarrayx-arrayy

 - [ ] Add the following WDL function to https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py
 - [ ] Add a test that runs a wdl/json with toil and then checks against cromwell to verify that toil and cromwell have the same behavior
       See: https://github.com/DailyDreaming/wdl-conformance-tests/tree/master/tests/standard_library

For example:

```wdl
Array[Int] xs = [ 1, 2, 3 ]
Array[String] ys = [ "a", "b" ]

Array[Pair[Int, String]] crossed = cross(xs, ys) # i.e. crossed = [ (1, "a"), (1, "b"), (2, "a"), (2, "b") ]
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-639)
┆Epic: WDL Standard Library Implementation
┆Issue Number: TOIL-639

PR #3304, opened 2020-10-26
Implement WDL Pair type

### LINKED ISSUE #3312: Implement WDL pair type
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-719)
┆Issue Number: TOIL-719

w-gao:
Closes #3312.

Needed before #3198, #3199, #3200, #3201, and #3203. 

This PR would
- add (full) support to WDL Pair defined here: https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#pair-literals
- also add support for file imports for Pair and Map compound types.


w-gao:
I’m sure there is room for improvement but this should work for the most part.

src/toil/wdl/wdl_analysis.py
```diff
@@ -620,18 +655,35 @@ def parse_declaration_type(self, typeAST):
...
+            if isinstance(subtype, wdl_parser.AstList):
+                # we're looking at a compound type
+                name = self.parse_declaration_type(typeAST.attr('name'))
+                elements = [self.parse_declaration_type(element) for element in subtype]
+
+                if name == 'Pair':
+                    return WDLPairType(*elements)
+                elif name == 'Map':
+                    return WDLMapType(*elements)
+
```
adamnovak, line 684: Shouldn't we throw an error here in case we encounter an unsupported WDL type?

src/toil/wdl/wdl_analysis.py
```diff
@@ -21,6 +21,41 @@
 wdllogger = logging.getLogger(__name__)
 
 
+class WDLType(str):
+    """
+    Represents a primitive or compound WDL type:
+
+    https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#types
+    """
```
adamnovak, line 29: Can you explain a bit about why this has to extend `str`? I don't quite get it.

If we need to change our internal representation of WDL types over to actual classes to make `Pair` and `Map` work, why not change all of them and have `WDLIntType`, etc.? Then I think we wouldn't need to extend `str` here.

src/toil/wdl/wdl_analysis.py
```diff
@@ -620,18 +655,35 @@ def parse_declaration_type(self, typeAST):
         which Toil needs to import, so we recursively travel down to the innermost
         type which will tell us if the variables are files that need importing.
 
+        For Pair and Map compound types, we recursively travel down the subtypes and
+        store them as attributes of a `WDLType` string. This way, the type structure is
+        preserved, which will allow us to import files appropriately.
+
```
adamnovak, line 661: The `Currently supported` section above should be updated with `Pair` and also I think `Map`?

src/toil/wdl/wdl_analysis.py
```diff
@@ -620,18 +655,35 @@ def parse_declaration_type(self, typeAST):
         which Toil needs to import, so we recursively travel down to the innermost
         type which will tell us if the variables are files that need importing.
 
+        For Pair and Map compound types, we recursively travel down the subtypes and
```
adamnovak, line 658: Shouldn't we use this same recursive traversal for `Array`? And optional types would then want to be, internally, `Optional[subtype]`, which just gets parsed as a `?` suffix.

src/toil/wdl/wdl_analysis.py
```diff
@@ -776,16 +828,21 @@ def parse_declaration_expressn_memberaccess(self, lhsAST, rhsAST, es):
         if isinstance(lhsAST, wdl_parser.Terminal):
             es = es + lhsAST.source_string
         elif isinstance(lhsAST, wdl_parser.Ast):
-            raise NotImplementedError
+            es = es + self.parse_declaration_expressn(lhsAST, es)
         elif isinstance(lhsAST, wdl_parser.AstList):
             raise NotImplementedError
 
-        es = es + '_'
+        # hack-y way to make sure pair.left and pair.right are parsed correctly.
```
adamnovak, line 835: Can you explain a bit more about why this hack is like it is? What's the underscore separator doing in the first place? Have we elsewhere turned all the hierarchical return types into flat underscore-separated keys?

src/toil/wdl/wdl_functions.py
```diff
@@ -36,10 +36,40 @@
...
     def __init__(self, message):
         super(WDLRuntimeError, self).__init__(message)
 
 
+class WDLPair:
+    """
+    Represent a WDL Pair literal defined at
+    https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#pair-literals
+    """
+    # TODO: figure out placement for these classes.
```
adamnovak, line 49: Can `WDLPair` and `WDLPairType` go together somewhere? maybe in a `wdl_types.py`? If they're far apart it can be confusing that we have one to represent the type and one to represent the instance.

If we were really cool, we would figure out how to set things up so that `WDLPairType(WDLIntType, WDLStringType)` was itself a type, and a WDL pair object would be an instance of it, so we wouldn't need a separate `WDLPair`. We could say `WDLPairType(WDLIntType, WDLStringType)(5, 'cows')` for example. I think this would involve metaclasses.  But that might be over-engineered, so don't do it unless the spirit moves you.

src/toil/wdl/wdl_functions.py
```diff
@@ -235,6 +265,68 @@ def process_infile(f, fileStore):
...
+                          var_type: str,
+                          read_in_file: bool = False,
+                          file_store: Optional[AbstractFileStore] = None,
+                          **kwargs):
+    """
+    Calls at runtime. This function parses and validates input from its type. File
+    import is also handled.
+
+    For values set in a task block, set `read_in_file` to True to process and read
+    all encountered files. This requires `cwd`, `temp_dir`, and `docker` to be
```
adamnovak, line 278: What are the semantics of these arguments that are required if `read_in_file` is set? They should probably be their own keyword arguments, rather than just being in `**kwargs`.

src/toil/wdl/wdl_functions.py
```diff
@@ -235,6 +265,68 @@ def process_infile(f, fileStore):
...
+    passed into this function.
+    """
+
+    def validate(val: bool, msg: str):
+        if not val:
+            raise WDLRuntimeError(f'Invalid input: {msg}')
+
+    if not in_data:  # optional type?
+        return in_data
```
adamnovak, line 287: We should throw an error here if the type is not in fact an optional.

adamnovak (changes requested):
This looks like it should work, but I'm worried that it leaves our internal representation of WDL types a bit half-refactored: we represent pair types and map types with Python objects whose structure mirrors the structure of the types, but we don't do the same for optionals and arrays, and all the terminal types are still strings.

I think all WDL types ought to be represented as `WDLType` subclasses, even terminal ones, and `WDLType` shouldn't extend `str`.

(If we were really clever we could make the `WDLType` subclasses metaclasses, so they could be the types of the objects we pass around at runtime, but that might be impossible because, among other problems, calling `WDLArrayType(WDLIntType)` is going to want to produce a new, distinct type every time you write it. So we probably will have to stick with using `list`, `dict`, and a couple special classes like `WDLPair`.)

I also don't understand the `.` vs `_` hack and why it's needed or why it works.

src/toil/wdl/wdl_analysis.py
w-gao: Extending `str` is certainly not the best approach, but it requires the least amount of refactoring since the rest of the implementation expects type to be a `str`. 

My intention with this PR is not to refactor the entire internal representation of WDL types, since @DailyDreaming has suggested that we should probably rework the overall design of the WDL implementation. 

If this is something we would want to have for future implementations, then I can adjust my tickets and focus on this instead.

src/toil/wdl/wdl_analysis.py
w-gao: The underscore separator is for task outputs, I believe.

```wdl
workflow {
  call t
  call t2 { input: in=t.out }
}
```
For this workflow, `t.out` is changed to `t_out = t.rv("out")`. See: https://github.com/DataBiosphere/toil/blob/releases/4.2.x/src/toil/wdl/wdl_synthesis.py#L376

In this case, we want to access `left` and `right` inside WDLPair. 

This can potentially cause problems if someone has an output variable named left or right, but this is the best way I could think of without passing the type into `parse_declaration_expressn()` everywhere.

w-gao:
Review comments will be addressed by #3330.

src/toil/wdl/wdl_analysis.py
adamnovak: I think this needs to be rethought overall; any acceptable WDL variable name needs to be allowed in user code, not just the ones that don't upset our parser. But it looks like it might be a bigger issue than just this ticket, and need some kind of unified `escape_user_variable_name()` function.

adamnovak (approved):
I think we've broken off most of my complaints into a new ticket.

DailyDreaming (approved):
This looks good to me at a glance.  I left a couple of minor comments.  Mostly I'm realizing that I'd like to have more testing to be confident, but that can be handled in the future.  This looks like really good work, thanks for putting so much thought into it.

src/toil/wdl/wdl_analysis.py
```diff
@@ -613,25 +614,51 @@ def parse_declaration_type(self, typeAST):
         Required.
 
         Currently supported:
-        Types are: Boolean, Float, Int, File, String, and Array[subtype].
-        OptionalTypes are: Boolean?, Float?, Int?, File?, String?, and Array[subtype]?.
+        Types are: Boolean, Float, Int, File, String, Array[subtype],
+                    Pair[subtype, subtype], and Map[subtype, subtype].
+        OptionalTypes are: Boolean?, Float?, Int?, File?, String?, Array[subtype]?,
+                            Pair[subtype, subtype]?, and Map[subtype, subtype].
```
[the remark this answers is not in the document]
DailyDreaming: Ignoring this for now as it's pretty minor.

PR #3307, opened 2020-10-28
Add length() WDL function

### LINKED ISSUE #3204: length() implemented as WDL.

```
## Int length(Array[X])

Given an Array, the `length` function returns the number of elements in the Array as an Int.
```
https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#int-lengtharrayx

For example:

```wdl
Array[Int] xs = [ 1, 2, 3 ]
Array[String] ys = [ "a", "b", "c" ]
Array[String] zs = [ ]

Int xlen = length(xs) # 3
Int ylen = length(ys) # 3
Int zlen = length(zs) # 0
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-644)
┆Issue Number: TOIL-644

w-gao:
Closes #3204. 

\- `length()` was supported but I'm making it part of `wdl_functions` for completeness, and to validate input.

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -232,7 +232,7 @@ class WdlStandardLibraryFunctionsTest(ToilTest):
         self.assertRaises(AssertionError, transpose, [[0, 1, 2], [3, 4, 5, 6]])
 
     def testFn_Length(self):
-        """Test the wdl built-in functional equivalent of 'length()'."""
+        """Test the WDL 'length()' built-in."""
         self.assertEqual(3, length([1, 2, 3]))
         self.assertEqual(3, length(['a', 'b', 'c']))
         self.assertEqual(0, length([]))
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -389,6 +389,10 @@ class WdlStandardLibraryWorkflowsTest(ToilTest):
                             json_file_name='length_invalid',
                             expected_exception='WDLRuntimeError')
 
+        # length() should not work with Map[X, Y].
+        self.check_function('length', cases=['as_input_with_map'],
+                            expected_exception='WDLRuntimeError')
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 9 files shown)


src/toil/test/wdl/builtinTest.py
```diff
@@ -230,6 +231,12 @@ def testFn_Transpose(self):
         self.assertEqual([[0]], transpose([[0]]))
         self.assertRaises(AssertionError, transpose, [[0, 1, 2], [3, 4, 5, 6]])
 
+    def testFn_Length(self):
+        """Test the wdl built-in functional equivalent of 'length()'."""
```
adamnovak, line 235: Isn't it just the "WDL 'length()' built-in"? It's not a functional equivalent of anything else, right?

src/toil/wdl/wdl_functions.py
```diff
@@ -873,3 +873,15 @@ def transpose(in_array: List[List[Any]]) -> List[List[Any]]:
...
     return [list(i) for i in zip(*in_array)]
+
+
+def length(in_array: List[Any]) -> int:
+    """
+    Given an Array, the `length` function returns the number of elements in the Array
+    as an Int.
+    """
+    if not isinstance(in_array, list):
+        # Cromwell throws an exception for anything other than a WDL Array
```
adamnovak, line 884: Does WDL have a dict-y type? If so, can you maybe add a test that enforces that we don't allow its length to be taken?

adamnovak (approved):
I think this is probably good as-is, but I pointed out some possible improvements.

src/toil/test/wdl/builtinTest.py
w-gao: Are you suggesting that I should change it to something like `"""Test the WDL 'length()' built-in"""`? That makes sense but I was following the docstring from the other tests to keep them consistent.

adamnovak: Yeah, that would make more sense to me.

adamnovak (approved):
I like it!

PR #3309, opened 2020-10-28
Add reporting for accessed files of failed jobs

### LINKED ISSUE #2649: Report jobs' accessed files when they fail
We should log each job's accesses to the file store. If the job fails, we should report the files that it downloaded and how to get at them, and (if available) what names they were downloaded as.

This would help a lot with debugging workflows, because you could then pull the files yourself and try any shell commands that failed on them.

This is part of #1855.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-2)
┆Issue Number: TOIL-2

adamnovak:
This should fix #2649.

When a job fails, its log will now include reporting of all global files downloaded/streamed via the FileStore:

```
[2020-10-28T15:07:50-0700] [MainThread] [W] [toil.fileStores.accessReportingWrapper] Failed job accessed files:
[2020-10-28T15:07:50-0700] [MainThread] [W] [toil.fileStores.accessReportingWrapper] Downloaded file 'files/for-job/kind-_accessAndFail/instance-4y_o2opj/file-cdc426fe38cd4cccb382d8211965d482/stream' to path '/scratch/tmp/toil-test-toil.test.src.fileStoreTest.NonCachingFileStoreTestWithFileJobStore-testFileStoreLogging-ta8ejk5l/node-bb28a5ba-cf48-4d48-8131-72ff00b5bbcf-11c2d78d2bcc48f28d74bd1f7dc0a45f/tmp7g6lpkc3/398f24bd-e074-49e0-8dde-eecf7390ec65/t9ow6ylpp/cats.txt'
[2020-10-28T15:07:50-0700] [MainThread] [W] [toil.fileStores.accessReportingWrapper] Streamed file 'files/for-job/kind-_accessAndFail/instance-4y_o2opj/file-cdc426fe38cd4cccb382d8211965d482/stream'
```

I'm not quite sure about my wrapper approach being quite the best way to achieve this. I didn't want to duplicate code between file store implementations, and the AbstractFileStore doesn't actually get a chance to run its own code in `open()`/`readGlobalFile()`/`readGlobalFileStream()`, so I wrote a wrapper class that intercepts those calls and delegates everything else, and put it in front of the actual file stores. This means that `isinstance(job.fileStore, AbstractFileStore)` wouldn't be true, and also it involves a lot of double-underscores.

A bit of code duplication might actually be better than that.


src/toil/fileStores/accessReportingWrapper.py
```diff
@@ -0,0 +1,118 @@
...
+import os
+from typing import Any, Union
+
+from toil.fileStores import FileID
+from toil.fileStores.abstractFileStore import AbstractFileStore
+
+
+logger = logging.getLogger(__name__)
+
+class AccessReportingWrapper:
```
mr-c, line 26: Why not make this a subclass of AbstractFileStore ?

adamnovak: If it is a subclass of AbstractFileStore, it gets all the methods of AbstractFileStore, which are stubs that raise `NotImplementedError`. I *think* those would have higher priority than `__getattr__` when someone goes to call a method, so we'd end up calling those stubs and not the methods of the wrapped FileStore.

I haven't tested it that way, though.

I could also extend AbstractFileStore and implement all those methods with overrides that do the call on the wrapped file store. That would be more code, but probably simpler code.

adamnovak:
Cactus weirdly hung for 8 hours in the integration test, and it looks like it's less code this way, so I've changed approaches and I'm now implementing the logging by requiring file store implementations to call the appropriate methods on the AbstractFileStore base class in the `readGlobalFile` and `readGlobalFileStream` and `open` implementations. It's not as *cool*, but I think it's probably clearer.

DailyDreaming (approved):
LGTM, provided tests pass (and cactus doesn't spin for significantly longer this time around).

Issue #3107, opened 2020-07-08
cwltool fails for secondary files stored on S3 jobstore


arostamianfar:
CWL workflows that contain a secondary file that is stored on the S3 jobstore fail with the new `cwltool` version with the exception below:

```
traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/cwltool/executors.py", line 233, in run_jobs
    for job in jobiter:
  File "/usr/local/lib/python3.6/dist-packages/cwltool/command_line_tool.py", line 591, in job
    builder = self._init_job(job_order, runtimeContext)
  File "/usr/local/lib/python3.6/dist-packages/cwltool/process.py", line 893, in _init_job
    discover_secondaryFiles=getdefault(runtime_context.toplevel, False),
  File "/usr/local/lib/python3.6/dist-packages/cwltool/builder.py", line 343, in bind_input
    discover_secondaryFiles=discover_secondaryFiles,
  File "/usr/local/lib/python3.6/dist-packages/cwltool/builder.py", line 410, in bind_input
    0 : cast(str, datum["location"]).rindex("/") + 1
ValueError: substring not found
```

The reason is that the `location` for S3 Toil objects is actually like `"location": "toilfs:17008:0b6b9de2-ef48-4bce-834f-fa94a1cd068e",`, which doesn't contain a `/`.
 
@mr-c this looks like a CWL bug rather than a Toil bug, but I figured to file it here first since apparently it only happens on Toil. Please file another issue on the CWL side if desired. Thanks!

P.S. This didn't happen in previous CWL versions as that code only ran [if `basename` didn't exist](https://github.com/common-workflow-language/cwltool/blob/1e5ad10c6b0d1c5f531737d12ef64062a00baef2/cwltool/builder.py#L394), which in Toil, it always exists.

mr-c:
Thanks for catching this @arostamianfar . Maybe you can help develop a toil or cwltool test so this doesn't happen again?

Here's a simple attempt at fixing this issue: https://github.com/common-workflow-language/cwltool/pull/1319/files

arostamianfar:
Thank you for the quick fix, @mr-c ! I've verified that your fix indeed works.

And great point about adding tests to cover these. I filed https://github.com/DataBiosphere/toil/issues/3109

arostamianfar:
Will keep this issue open on the Toil side until a new version of cwltool with the fix has been released and Toil is updated to use that version. @mr-c any ideas when the next release will be?

[...]

PR #3311, opened 2020-11-02
Test CWL secondary file from s3.

### LINKED ISSUE #3109: Add S3 support for CWL conformance tests
We currently run CWL conformance tests using different batch systems, but not with different job stores. This can result in bugs like https://github.com/DataBiosphere/toil/issues/3107

We should at least add S3 jobstore support to the CWL conformance tests.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-583)
┆Issue Number: TOIL-583

DailyDreaming:
Resolves #3109 .


DailyDreaming:
@arostamianfar Is this what you had in mind for #3109 / #3107?

mr-c:
@DailyDreaming Is there a test coverage report?

DailyDreaming:
@mr-c Hmmm... reworking the Makefile and the gitlab config to include those again.

DailyDreaming:
@mr-c For `cwltoil.py`, it looks like we have:

```
src/toil/cwl/cwltoil.py                                    873    136    84%
```

When running the CWL 1.0 tests at least.  Coverage is split up though, so each CI job only reports the coverage for the subset of tests that it runs.  We'd have to collect all of the coverage results in one place and somehow sum them up overall, or run the gamut of `make test` on occasion if we want a comprehensive coverage report.

arostamianfar:
Thanks, @DailyDreaming ! this is great! Although, the test I had in mind was to run `cwltoil` with `--jobStore` option that's on S3. I believe the current test ends up downloading the file into a local job store, so it may not necessarily catch the original bug.
Not sure how feasible it is to add such an option?

Regardless, this is a great test to have! Thanks so much!

DailyDreaming:
@arostamianfar Ah, that makes sense.  I just amended it to run with an s3 jobstore.  It passed for me locally so hopefully it runs through gitlab with no problems.

I ended up with a bunch of refactor cruft tacked on after re-enabling coverage.  I think I'm going to break that off into a separate PR and merge it separately.

mr-c:
@DailyDreaming Does the test result in these lines of `cwltool` being executed?

https://github.com/common-workflow-language/cwltool/pull/1319/files#diff-237bb277b731e18286d6ab18769c0991ad2fc6aa0ef4fb64a5f9adeb3034f319R425

DailyDreaming:
@mr-c I just tested it and it does.

src/toil/batchSystems/options.py
```diff
@@ -1,4 +1,3 @@
-from __future__ import absolute_import
 # Copyright (C) 2015-2016 Regents of the University of California
```
mr-c, line 1: no 2017-2020 ?

mr-c (approved):

src/toil/batchSystems/options.py
DailyDreaming: Ha, will update, thanks!

PR #3313, opened 2020-11-03
added doubleMem functionality for LSF for lsf jobs that die due to reaching imposed memory limit

drkennetz:
Often in genomics pipelines, especially for post-processing, memory limits of programs are not well-defined and can be dependent on data. This PR will add a memory doubling functionality if the flag --doubleMem is included on the command line for the LSF scheduler which will double the memory of and resubmit jobs that die due to reaching the LSF imposed memory limit. This functionality is incredibly useful for on-prem clusters. This addresses the issue opened by me: #3269

### DIFF
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -224,6 +224,11 @@ the logging module:
   --retryCount RETRYCOUNT
                         Number of times to retry a failing job before giving
                         up and labeling job failed. default=1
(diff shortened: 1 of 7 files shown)


mr-c (approved):

drkennetz:
These updates fix the merge conflicts from #3276 . Figured it would be easier in a new PR.

PR #3317, opened 2020-11-05
CWL v1.2 conformance tests 330 and 331.

DailyDreaming:
Tests 330 and 331 were failing because some CWL v1.2 test fixtures were deleted by prior tests.  Restoring the CWL v1.2 fixtures prior to running allows them to pass.

Adds a slight refactor and also removes some of the `xfail` decorators that allow CWL caching tests to fail.

### DIFF
--- a/src/toil/test/cwl/cwlTest.py
+++ b/src/toil/test/cwl/cwlTest.py
@@ -22,6 +22,7 @@ import shutil
 import zipfile
 import pytest
 import uuid
+import subprocess
 from urllib.request import urlretrieve
 from io import StringIO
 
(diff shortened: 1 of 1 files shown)


src/toil/test/cwl/cwlTest.py
```diff
@@ -414,46 +401,27 @@ def tearDown(self):
             shutil.rmtree(self.outDir)
         unittest.TestCase.tearDown(self)
 
+    def refresh_deleted_fixtures(self):
+        # TODO: Prevent Toil from deleting these?
```
mr-c, line 405: Great catch, thanks! Can we get an issue to track why this is needed?

src/toil/test/cwl/cwlTest.py
```diff
@@ -414,46 +401,27 @@ def tearDown(self):
...
-                m = p.search(line)
-                if m:
-                    if int(m.group("failures")) == 0 and int(m.group("unsupported")) > 0:
-                        only_unsupported = True
-                        break
-            if not only_unsupported:
-                print(error_log)
-                raise e
+        # TODO: we do not currently pass tests: 237 (offset from other versions), 307, 309, 310, 311, 332
+        for selected_tests in ['1-236,238-306,308,312-329,333-336', '330', '331']:
```
mr-c, line 421: Is running tests 330 and 331 last needed? If so, the comment should document why.

mr-c (changes requested):
Looks good! I'll approve once CI is done and that comment is clarified.

src/toil/test/cwl/cwlTest.py
DailyDreaming: Definitely!  I've created: https://github.com/DataBiosphere/toil/issues/3318

src/toil/test/cwl/cwlTest.py
DailyDreaming: Yes, each time the fixtures have to be recreated.  Will add a comment!

DailyDreaming: Added.

mr-c (changes requested):
CWL v1.2 Test 330 still fails https://ucsc-ci.com/databiosphere/toil/-/jobs/64956#L939

DailyDreaming:
@mr-c Looks like it's the caching test only again.  I took out the tolerance to `xfail` the caching tests, so it magically passed earlier.  I'll put a hold on this until I can look at the caching.  Sorry for the false alarm, and thanks for the comments.

DailyDreaming:
Closing in favor of #3336 .

PR #3321, opened 2020-11-05
Expose --statePollingWait param through toil

### LINKED ISSUE #3320: Polling interval for LSF and SGE batch systems is too low
We would like to be able to raise the polling interval for our batch systems without having to change the default in the code (which is how we've previously solved this issue)

I've made this issue to be linked to a PR I will push so we can raise this interval through the command line.

Additionally, I'd like to point out a slightly confusing aspect of the Config object. This line:

https://github.com/DataBiosphere/toil/blob/33f30b34b7a6cfb74be9d7863d08ce2f7be58475/src/toil/batchSystems/options.py#L165

Actually has no effect, because it is called before a line in `common.py` that resets the value to `1`:

https://github.com/DataBiosphere/toil/blob/33f30b34b7a6cfb74be9d7863d08ce2f7be58475/src/toil/common.py#L73
https://github.com/DataBiosphere/toil/blob/33f30b34b7a6cfb74be9d7863d08ce2f7be58475/src/toil/common.py#L95

Therefore the comment in `options.py` is misleading, and `1` will always be used.

It may be sensible to have an additional PR to remove the `self.statePollingWait = 1`, or switch its order with the `setDefaultBatchOptions(self)` call

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-724)
┆Issue Number: TOIL-724

ionox0:
resolves #3320

### DIFF
--- a/src/toil/batchSystems/options.py
+++ b/src/toil/batchSystems/options.py
@@ -151,7 +151,7 @@ def setDefaultOptions(config):
     config.batchSystem = "single_machine"
     config.disableAutoDeployment = False
     config.environment = {}
-    config.statePollingWait = None  # if not set, will default to seconds in getWaitDuration()
+    config.statePollingWait = None
     config.maxLocalJobs = cpu_count()
     config.manualMemArgs = False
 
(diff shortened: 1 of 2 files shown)


ionox0:
My change here was to move the `--statePollingWait` param out of the section limited by `if not config.cwl:`

My thinking is that it was put there along with all of the service related parameters (which CWL does not support), but that this one actually can be used in CWL workflows

I've tested it on an example workflow and works for controlling the `bjobs` polling frequency on our system. Please let me know if this will cause issues though

DailyDreaming (approved):
This looks fine to me.  Thanks for the addition!  Merging.

PR #3322, opened 2020-11-10
Preserve file permissions on imported/exported files

jonathanxu18:


src/toil/test/src/fileStoreTest.py
```diff
@@ -23,10 +23,11 @@
 from struct import pack, unpack
 from uuid import uuid4
 
+from toil.common import Toil
 from toil.job import Job
 from toil.fileStores import FileID
 from toil.fileStores.cachingFileStore import IllegalDeletionCacheError, CacheUnbalancedError, CachingFileStore
-from toil.test import ToilTest, needs_aws_ec2, needs_google, slow, travis_test
+from toil.test import ToilTest, needs_aws_ec2, needs_google, slow, travis_test, src
```
adamnovak, line 29: I don't think we need this `src` here.

src/toil/test/src/fileStoreTest.py
```diff
@@ -250,6 +252,40 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
 
+        def testWriteReadGlobalFile(self):
+            workdir = self._createTempDir(purpose='nonLocalDir')
+            for i in True,False:
+                A = Job.wrapJobFn(self._testWriteReadGlobalFilePermissions,
+                                isLocalFile=False, executable=i, nonLocalDir=workdir)
+                Job.Runner.startToil(A, self.options)
+
+        @staticmethod
+        def _testWriteReadGlobalFilePermissions(job, isLocalFile, executable, nonLocalDir=None):
```
adamnovak, line 262: You should drop `isLocalFile` and `nonLocalDir` from this function, because I don't think we care about testing the exact source directory from which we upload the file for this test.

Just do `workDir = job.fileStore.getLocalTempDir()` always, and always make the file in `workDir`.

src/toil/test/src/fileStoreTest.py
```diff
@@ -250,6 +252,40 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
+                f.write('Hello')
+
+            if executable:
+                os.chmod(srcFile, os.stat(srcFile).st_mode | stat.S_IXUSR)
+
+            initialPermissions = os.stat(srcFile).st_mode
+            fileID = job.fileStore.writeGlobalFile(srcFile)
+            dstFile = job.fileStore.getLocalTempFile() + str(uuid4())
+            dstFile = job.fileStore.readGlobalFile(fileID, userPath=dstFile, mutable=True)
+            currentPermissions = os.stat(dstFile).st_mode
```
adamnovak, line 284: You might want to test a few combinations of `mutable` and `symlink` arguments here, to make sure that when we use hard links or symlinks into the cache we still see the file with the right executability.

src/toil/test/src/fileStoreTest.py
```diff
@@ -250,6 +252,40 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
+
+            assert initialPermissions == currentPermissions
```
adamnovak, line 286: You might not want to compare the whole mode, just the `stat.S_IXUSR` bit (`assert initialPermissions & stat.S_IXUSR == currentPermissions & stat.S_IXUSR`)

src/toil/test/src/importExportFileTest.py
```diff
@@ -98,6 +99,33 @@ def testImportSharedFileName(self):
...
+
+            # Add file owner's execute permissions
+            os.chmod(srcFile, permissionBits | stat.S_IXUSR)
+
+            fileID = toil.importFile('file://' + srcFile)
+            toil.exportFile(fileID, 'file://' + self.dstFile)
+
+            permissionBits = os.stat(self.dstFile).st_mode
+            # Nonzero value indicates executable flag is preserved
+            assert permissionBits & stat.S_IXUSR != 0
```
adamnovak, line 127: You might want to try this test with a non-exacutable file as well, to make sure that it isn't executable when it comes out.

src/toil/fileStores/abstractFileStore.py
```diff
@@ -296,7 +299,8 @@ def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=False,
         Makes the file associated with fileStoreID available locally. If mutable is True,
         then a copy of the file will be created locally so that the original is not modified
         and does not change the file for other jobs. If mutable is False, then a link can
-        be created to the file, saving disk resources.
+        be created to the file, saving disk resources. The file that is downloaded will be 
+        executable if it was originally uploaded from an executable file on the local filesystem.
```
adamnovak, line 303: Since I'm also insisting that non-executable files stay non-executable, maybe we should write "if and only if" here and in the import/export docs.

adamnovak:
OK, I made some not-actually-a-review comments.

The only other thing I have to say is that it would be nice to have a test to make sure that read/writeGlobalFile and import/exportFile are compatible with each other. You could cram it into one of the existing text cases, but it would be good to see that if you e.g. create an executable file in a job, it will still be executable when you export it form the leader.

adamnovak:
This looks pretty good to me so far! It just wants the cross-tests to make sure import/export and read/write work together, and then the actual implementation.

adamnovak:
This still looks good to me. I do see that you are starting Toil multiple times within a single `with` block, with a single `toil` object, which I wouldn't think to do, but I also don't see a good reason why that wouldn't be expected to work.

src/toil/test/src/fileStoreTest.py
```diff
@@ -12,6 +12,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
+from __future__ import print_function
```
DailyDreaming, line 16: Many of these old python2 imports have been removed or reordered.  You might have to do a pull from master to bring things up to date.

src/toil/fileStores/cachingFileStore.py
```diff
@@ -1035,8 +1036,10 @@ def writeGlobalFile(self, localFileName, cleanup=False):
         # Create an empty file to get an ID.
         # Make sure to pass along the file basename.
         # TODO: this empty file could leak if we die now...
+        print("TESTING writeGlobalFile CACHING")
+        permissions = os.stat(localFileName).st_mode & stat.S_IXUSR
+        print("permissions BEFORE: {permissions}".format(permissions=permissions))
```
DailyDreaming, line 1041: Saving these for later?  >.>

DailyDreaming (commented):
Looks great so far.  I made some minor comments (see below).

Would approve, but I think there's a minor error:

```
	  File "/builds/databiosphere/toil/src/toil/job.py", line 2358, in _runner
	  File "/builds/databiosphere/toil/src/toil/job.py", line 2279, in _run
	    return self.run(fileStore)
	  File "/builds/databiosphere/toil/src/toil/job.py", line 2502, in run
	    rValue = userFunction(*((self,) + tuple(self._args)), **self._kwargs)
	  File "/builds/databiosphere/toil/src/toil/test/src/fileStoreTest.py", line 964, in _returnFileTestFn
	    fsId = cls._writeFileToJobStoreWithAsserts(job, isLocalFile=True, fileMB=writeFileSize)
	  File "/builds/databiosphere/toil/src/toil/test/src/fileStoreTest.py", line 528, in _writeFileToJobStoreWithAsserts
	    fsID, testFile = cls._writeFileToJobStore(job, isLocalFile, nonLocalDir, fileMB)
	ValueError: too many values to unpack (expected 2)
```

Which looks like you might have just missed removing a leftover `testFile` (check to make sure that's the only spot).

Otherwise, as I said, this looks great and I'd be happy to merge it once tests are passing.

src/toil/fileStores/__init__.py
```diff
@@ -28,32 +29,35 @@ class FileID(str):
...
     @classmethod
     def unpack(cls, packedFileStoreID):
         """
         Unpack the result of pack() into a FileID object.
         """
 
-        # Find the separating character (first instance)
-        sepPos = packedFileStoreID.find(':')
+        # Separate each part of fileID
+        vals = packedFileStoreID.split(':')
```
adamnovak, line 57: The reason we were using find before is that we can't guarantee that there are no colons in the File ID itself. If you want to use split, you need to make sure to grab the first and last items and paste the rest back together with colons, or else encode/escape internal colons and then decode/unescape them here.

I don't think this will fail in practice, because hardly anybody is going to use colons in filenames anywhere, and all the current file store implementations might avoid them, but it doesn't work in theory.

src/toil/fileStores/nonCachingFileStore.py
```diff
@@ -101,6 +102,8 @@ def readGlobalFile(self, fileStoreID, userPath=None, cache=True, mutable=False,
 
         self.jobStore.readFile(fileStoreID, localFilePath, symlink=symlink)
         self.localFileMap[fileStoreID].append(localFilePath)
+        if fileStoreID.executable:
```
adamnovak, line 105: This is one of the places where we need to handle the user sending us `str` instead of `FileID`. The `pack()`/`unpack()` stuff above is only used by CWL; Cactus isn't smart enough to use it (although it ought to be updated to).

If `fileStoreID` is a `str`, it won't have an `executable` attribute. You might want to replace this line with something that will supply a default value if the attribute is missing:

```
if getattr(fileStoreID, 'executable', False):
```

src/toil/jobStores/fileJobStore.py
```diff
@@ -313,6 +315,8 @@ def _exportFile(self, otherCls, jobStoreFileID, url):
                 self._move_and_linkback(srcPath, destPath)
             else:
                 atomic_copy(srcPath, destPath)
+            if jobStoreFileID.executable:
```
adamnovak, line 318: Here again you need to handle the case where `jobStoreFileID` is a `str` and is missing the attribute.

src/toil/fileStores/__init__.py
adamnovak: Another option would be to just ban colons in the file ID with an `assert ':' not in self, 'Colons not allowed in ' + self` in `__init__`, and a suitable comment. Then it would be up to the job store implementations to make sure colons are properly escaped/unescaped when making/interpreting FileIDs.

adamnovak (changes requested):
This looks pretty good, and I had forgotten we even had `pack()` and `unpack()` that need to carry the permission bit through. But you're still not handling the case where someone takes `str(theFileID)` and then hands that back, like Cactus does. I've pointed out the lines where it looks to me like it will fail in that case.

Fix that, the possibility of colons in the ID itself, and the tests and I think this will be ready.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -413,17 +424,24 @@ def _importFile(self, otherCls, url, sharedFileName=None, hardlink=False):
...
     def _exportFile(self, otherCls, jobStoreFileID, url):
+        print("IN AWS _EXPORTFILE")
+        print(f"otherCls: {otherCls}")
+        print(f"AWS url: {url}")
         if issubclass(otherCls, AWSJobStore):
             dstKey = self._getKeyForUrl(url)
+            print(f"AWS jobStoreFileID.executable: {jobStoreFileID.executable}")
             try:
                 info = self.FileInfo.loadOrFail(jobStoreFileID)
                 info.copyTo(dstKey)
```
jonathanxu18, line 443: @adamnovak @DailyDreaming When I was trying to take the executable attribute from a `FileID` to an AWS key during `_exportFile()`, I was running into issues getting access to it. I saw that the `AWS jobstore` uses the FileInfo class to preserve the `FileID` attributes which gets copied over to the new AWS key, so I was wondering if I should maybe add an executable attribute on to that class so it gets copied over too.

FileInfo class: https://github.com/DataBiosphere/toil/blob/a03e24f7e8afbfbd6416c17c6f906b6fdf6435a7/src/toil/jobStores/aws/jobStore.py#L828 

Sorry for all the extra print statements, I just wanted to keep them there for reference until I finish debugging this part.

adamnovak: If you did work out a way to store the executable bit in the AWS FileInfo, then you would have the AWSJobStore be responsible for keeping track of that bit, probably instead of making the FileID do it.

That would let you set the executable bit properly when people give you strings instead if FileIDs when they download files, but it would be a different design than the one you've been implementing so far, where the FileID keeps track of executability, so I wouldn't do it.

These `_importFIle` and `_exportFile` overrides in `AWSJobStore` have one purpose, which is to make copying more efficient when you are importing from something on AWS to an `AWSJobStore`, or exporting from an `AWSJobStore` to somewhere else on AWS. Instead of downloading the file from AWS to the local machine and then uploading it elsewhere on AWS, they detect AWS-to-AWS operations and ask AWS to just copy the data internally.

Since files on AWS have no concept of executability, there's no execute bit to read when you import from AWS, and no execute bit to set when you export to AWS. So you shouldn't need to touch the AWS `_importFile` and `_exportFile` overrides in any substantial way; they should just not set the execute bit in the FileIDs they produce for imports from AWS, and not do anything with the execute bit in the file IDs they get to export to AWS.

If you are having trouble with imports and exports between AWS and local files, you might need to look at `_importFile` and `_exportFile` as defined in the `AbstractJobStore` (which is being called from here when the other side of the copy is not on AWS), or in the `FileJobStore`.

Does that help?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -376,6 +381,8 @@ def _defaultExportFile(self, otherCls, jobStoreFileID, url):
         """
         with self.readFileStream(jobStoreFileID) as readable:
             otherCls._writeToUrl(readable, url)
+        if getattr(jobStoreFileID, 'executable', False):
```
jonathanxu18, line 384: @adamnovak I was also thinking of doing this inside of `writeToUrl` instead by passing in the executable bit as an arg. Do you think that would be a better approach compared to what I currently have? I'm just wondering since we talked about keeping the implementation inside `writeToUrl` and `readToUrl`.

adamnovak: Yeah, you probably have to keep this inside the `_writeToUrl` that actually writes local files.

Here, you don't know that the URL even is a `file://` URL. So if someone tries to export to `s3://bucket/some/place/in/a/bucket`, you are going to grab the `url.path` (`some/place/in/a/bucket`) and try to chmod that executable as if it were a local filename. That will probably throw an error, if you're lucky. If you're unlucky, that path will happen to exist and you'll be setting the executable bit on whatever that file happens to be on the local machine.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -400,7 +407,7 @@ def _readFromUrl(cls, url, writable):
 
         :param writable: a writable stream
 
-        :return int: returns the size of the file in bytes
+        :return int, bool: the size of the file in bytes, the user executable permission bit
```
adamnovak, line 410: Maybe `from typing import Tuple` and call this `Tuple[int, bool]`? Or just make the docstring say `:return:` and put the type as ` def _readFromUrl(cls, url, writable) -> Tuple[int, bool]:`.

I've never quite understood what language we're using to express the argument and return types, and the way you write things with the typing module is at least [documented](https://docs.python.org/3/library/typing.html#typing.Tuple).

src/toil/fileStores/__init__.py
```diff
@@ -33,32 +34,35 @@ class FileID(str):
...
```
DailyDreaming, line 62: Adam's comment on this still needs to be addressed I think: https://github.com/DataBiosphere/toil/pull/3322/files#r534339567

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1152,4 +1160,4 @@ def count(l):
 
             # Do the download
             shutil.copyfileobj(readable, counter)
-            return size[0]
+            return size[0], False
```
DailyDreaming, line 1163: Should we check the permissions of the file copied with `shutil.copyfileobj` or is this always guaranteed to be `False`?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,10 +450,10 @@ def _readFromUrl(cls, url, writable):
             srcKey.get_contents_to_file(writable)
         finally:
             srcKey.bucket.connection.close()
-        return srcKey.size
+        return srcKey.size, False
```
DailyDreaming, line 453: Same here.  Should we check the permissions of the file copied with `shutil.copyfileobj` or is this always guaranteed to be `False`?

src/toil/jobStores/fileJobStore.py
```diff
@@ -351,6 +356,9 @@ def _writeToUrl(cls, readable, url):
         """
         # we use a ~10Mb buffer to improve speed
         atomic_copyobj(readable, cls._extractPathFromUrl(url), length=cls.BUFFER_SIZE)
+        if executable:
+            os.chmod(url.path, os.stat(url.path).st_mode | stat.S_IXUSR)
```
DailyDreaming, line 360: Consider putting this functionality into `atomic_copyobj` and `atomic_copy`.  I'd intuitively expect a copy function to know how to preserve permissions while copying.

src/toil/test/src/fileStoreTest.py
```diff
@@ -233,6 +235,97 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
+            maintains its executability after being read.
+            """
+            for executable in True,False:
+                A = Job.wrapJobFn(self._testWriteReadGlobalFilePermissions, executable=executable)
+                Job.Runner.startToil(A, self.options)
+
+        @staticmethod
+        def _testWriteReadGlobalFilePermissions(job, executable):
+            workDir = job.fileStore.getLocalTempDir()
+            srcFile = '%s/%s%s' % (workDir, 'in', str(uuid4()))
```
DailyDreaming, line 251: Nit: f-strings are easier to read (for me at least): `f'{workdir}/in{uuid4()}'`

DailyDreaming (commented):
The tests for this look great.  Comments are mostly minor.

src/toil/jobStores/aws/jobStore.py
jonathanxu18: I was talking with @adamnovak and I think since AWS objects don't have executability we decided to set it always to False.

src/toil/jobStores/abstractJobStore.py
jonathanxu18: @adamnovak also mentioned that it isn't possible to read an executable bit over HTTP or HTTPS and doing it over FTP would involve tweaking other commands so it would need to be a separate issue.

src/toil/fileStores/__init__.py
```diff
@@ -33,32 +34,34 @@ class FileID(str):
...
     @classmethod
     def unpack(cls, packedFileStoreID):
         """
         Unpack the result of pack() into a FileID object.
         """
-
-        # Find the separating character (first instance)
-        sepPos = packedFileStoreID.find(':')
+        # Only separate twice in case the FileID itself has colons in it
+        vals = packedFileStoreID.split(':', 2)
```
jonathanxu18, line 61: @adamnovak I think this should fix the issue with colons being in the `FileID` itself but it isn't one of the ways you recommended to go about it. Let me know if I'm not accounting for something.

adamnovak: Yeah, that should work, I think.

src/toil/jobStores/abstractJobStore.py
adamnovak: `copyfileobj` just ships bytes between file *objects*; it doesn't know anything about permissions.

src/toil/lib/misc.py
```diff
@@ -118,16 +119,21 @@ def AtomicFileCreate(final_path, keep=False):
                 pass
         raise
 
-def atomic_copy(src_path, dest_path):
+def atomic_copy(src_path, dest_path, executable=False):
     """Copy a file using posix atomic creations semantics."""
     with AtomicFileCreate(dest_path) as dest_path_tmp:
         shutil.copyfile(src_path, dest_path_tmp)
+    if executable:
```
adamnovak, line 126: I think this should be inside the `with`, which provides the protection that makes this actually atomic. If we leave it out here, we can put the file in place where other people can see it before we fix the permissions.

Inside the `with` we'd have to operate on `dest_path_tmp` instead of `dest_path`; `AtomicFileCreate` moves the one to the other when the `with` ends.

src/toil/lib/misc.py
```diff
@@ -118,16 +119,21 @@ def AtomicFileCreate(final_path, keep=False):
...
+        os.chmod(dest_path, os.stat(dest_path).st_mode | stat.S_IXUSR)
 
-def atomic_copyobj(src_fh, dest_path, length=16384):
+def atomic_copyobj(src_fh, dest_path, length=16384, executable=False):
     """Copy an open file using posix atomic creations semantics."""
     with AtomicFileCreate(dest_path) as dest_path_tmp:
         with open(dest_path_tmp, 'wb') as dest_path_fh:
             shutil.copyfileobj(src_fh, dest_path_fh, length=length)
+    if executable:
```
adamnovak, line 134: As with the function above, we should set the permissions inside of the `with AtomicFileCreate` block, operating on `dest_path_tmp`.

src/toil/test/src/fileStoreTest.py
```diff
@@ -233,6 +235,97 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
+            if executable:
+                os.chmod(srcFile, os.stat(srcFile).st_mode | stat.S_IXUSR)
+
+            # Initial file owner execute permissions
+            initialPermissions = os.stat(srcFile).st_mode & stat.S_IXUSR
+            fileID = job.fileStore.writeGlobalFile(srcFile)
+
+            for mutable in True,False:
+                for symlink in True,False:
+                    #with self.subTest(f'Now testing readGlobalFileWith: mutable={mutable} symlink={symlink}'):
```
adamnovak, line 264: If this isn't in use we should drop it and not comment it out.

src/toil/test/src/fileStoreTest.py
```diff
@@ -233,6 +235,97 @@ def _testFileStoreOperations(job, nonLocalDir, numIters=100):
...
+                    dstFile = job.fileStore.getLocalTempFile() + str(uuid4())
```
adamnovak, line 265: If you just want a filename in a temporary directory, you want to use [`getLocalTempDir()`](https://toil.readthedocs.io/en/releases-5.1.0/developingWorkflows/toilAPIFilestore.html#toil.fileStores.abstractFileStore.AbstractFileStore.getLocalTempDir) or [`getLocalTempFileName()`](https://toil.readthedocs.io/en/releases-5.1.0/developingWorkflows/toilAPIFilestore.html#toil.fileStores.abstractFileStore.AbstractFileStore.getLocalTempFileName) and not [`getLocalTempFile()`](https://toil.readthedocs.io/en/releases-5.1.0/developingWorkflows/toilAPIFilestore.html#toil.fileStores.abstractFileStore.AbstractFileStore.getLocalTempFile) which creates a file for you.

adamnovak (changes requested):
This looks like it should work. I just have a couple things I found about the atomic-ness of the changes to the atomic copy functions, and some non-idiomatic and commented out stuff in the tests.

src/toil/lib/misc.py
```diff
@@ -123,16 +123,16 @@ def atomic_copy(src_path, dest_path, executable=False):
     """Copy a file using posix atomic creations semantics."""
     with AtomicFileCreate(dest_path) as dest_path_tmp:
         shutil.copyfile(src_path, dest_path_tmp)
-    if executable:
-        os.chmod(dest_path, os.stat(dest_path).st_mode | stat.S_IXUSR)
+        if executable:
+            os.chmod(dest_path, os.stat(dest_path).st_mode | stat.S_IXUSR)
```
adamnovak: This has to operate on `dest_path_tmp` here, not `dest_path`. When you leave the `with` block, any file at `dest_path` (there might not have been one) is replaced with the one at `dest_path_tmp`.

src/toil/test/src/fileStoreTest.py
```diff
@@ -261,8 +261,7 @@ def _testWriteReadGlobalFilePermissions(job, executable):
 
             for mutable in True,False:
                 for symlink in True,False:
-                    #with self.subTest(f'Now testing readGlobalFileWith: mutable={mutable} symlink={symlink}'):
-                    dstFile = job.fileStore.getLocalTempFile() + str(uuid4())
+                    dstFile = job.fileStore.getLocalTempDir() + str(uuid4())
```
adamnovak: This might need to be an `os.path.join()` call instead of a `+`? Or does `getLocalTempDir()` always returns a path ending in `/`?

src/toil/test/src/fileStoreTest.py
adamnovak: This is still the version that creates an extraneous file. But it should still *work*, so it might be good enough.

adamnovak (approved):

mr-c:
This is huge, thank you @jonathanxu18 and everyone else who assisted!

PR #3323, opened 2020-11-11
Run CWL conformance tests on Kubernetes.

DailyDreaming:
This steals Adam's kubernetes test setup for cactus in order to run the CWL conformance tests on kubernetes.

I suspect some of the lines may be cactus specific, so Adam, please let me know if there's something I should tweak/remove!


mr-c:
Any progress here?

DailyDreaming:
@mr-c

@adamnovak might be able to provide an update.  We're currently working on this though.

adamnovak:
Bad news: the CWL 1.0 tests on Kubernetes take about 4 hours to run, with the failing tests in. I'm skipping those now but I don't think it's going to be much faster. I suspect the problem is that we aren't running the CWL tests in parallel because we're trying to use a single named job store. We're doing that because the CWL runner defaults to a local job store even with the Kubernetes batch system, based off the CWL work directory.

So I think the next step here is to fix Toil's CWL runner to default to a randomized AWS jobstore when running against the Kubernetes batch system, so that when the CWL conformance tester runs it it will do something sensible. We might want to (eventually?) turn that into an option for a job store type or prefix or something.

adamnovak:
Now the tests are failing because Singularity is going around our Docker Hub mirror and running out of pulls. We need to work out how to interpose a mirror in e.g. `singularity pull --force --name debian:stretch-slim.sif docker://debian:stretch-slim` without modifying the command line, or else patch the CWL runner to pass mirror flags, or just use a different registry name in the pull.

adamnovak:
OK, the reason this has taken forever is that we have a large number of CWL conformance tests that fail intermittently when run on Kubernetes. I think we're missing some kind of scheduling constraint in a way that doesn't manifest on single machine; I'm not sure we've been testing CWL conformance on Mesos.

I've set the CWL Kubernetes tests to run but expect failure; we can make new issues to make them (or a subset of them) actually pass, or to resolve particular kinds of exceptions that cause failures.

adamnovak:
I think in #3325 we might have wanted to have a subset of passing tests identified; I don't think that's practical to actually do with the level of intermittent failures we are seeing.

.gitlab-ci.yml
```diff
@@ -45,7 +45,7 @@ py36_main:
     - /^never$/
   script:
     - pwd
-    - virtualenv -p python3.6 venv && . venv/bin/activate && make prepare && make develop extras=[all] && pip install htcondor
+    - virtualenv -p python3.6 venv && . venv/bin/activate && pip install -U pip && make prepare && make develop extras=[all] && pip install htcondor
```
mr-c, line 47: Maybe `pip install -U pip wheel` ? Most venv's done contain `wheel` and that can make a difference

adamnovak: I can add wheel. I don't think not having is is causing any failures though.

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
         )
     return load_listing
 
+class NoAvailableJobStoreException(Exception):
+    """Indicates that no job store name is available."""
+    pass
+
+def generate_default_job_store(batch_system_name: Optional[str], provisioner_name: Optional[str],
```
DailyDreaming, line 1813: Should either of these args default to `None`?

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
...
+                               local_directory: str) -> str:
+    """
+    Choose a default job store appropriate to the requested batch system and
+    porvisioner, and installed modules. Raises an error if no good default is
+    available and the user must choose manually.
+
+    :param batch_system_name: Registry name of the batch system the user has
+           requested, if any.
+    :param provisioner_name: Name of the provisioner the user has requested,
+           if any, May be 'aws' or 'gce'.
```
DailyDreaming, line 1823: Type hinting is `Optional[str]`, should this include `None`?

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
...
+    :return str: Job store specifier for a usable job store.
+    """
+
+    # Apply default batch system
+    batch_system_name = batch_system_name or DEFAULT_BATCH_SYSTEM
+
+    # Work out how to describe where we are
+    situation = f"the '{batch_system_name}' batch system"
+    if provisioner_name:
+        situation += f"with the '{provisioner_name}' provisioner"
```
DailyDreaming, line 1836: I think this might need to be prefixed with a space.

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
...
+            from toil.provisioners.aws import get_current_aws_region
+            region = get_current_aws_region()
+
+            if not region:
+                # We can't generate an AWS job store without a region
+                situation += ' running outside AWS with no TOIL_AWS_ZONE set'
+                raise NoAvailableJobStoreException()
+
+            # Roll a random name
+            return f'aws:{region}:toil-cwl-{str(uuid.uuid4())}'
```
DailyDreaming, line 1868: Should an `else` error with an unrecognized batch system if not `aws`/`gce`/`None`?

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
         )
     return load_listing
 
+class NoAvailableJobStoreException(Exception):
+    """Indicates that no job store name is available."""
+    pass
+
+def generate_default_job_store(batch_system_name: Optional[str], provisioner_name: Optional[str],
+                               local_directory: str) -> str:
```
DailyDreaming, line 1814: Does a `local` directory input make sense when asking for a jobstore for `aws` or `gcp`?

src/toil/cwl/cwltoil.py
```diff
@@ -1805,6 +1806,77 @@ def determine_load_listing(tool: ToilCommandLineTool):
...
+    """
+    Choose a default job store appropriate to the requested batch system and
+    porvisioner, and installed modules. Raises an error if no good default is
```
DailyDreaming, line 1817: `porvisioner`

DailyDreaming (commented):
First pass comments.  Looks good to me overall, but will revisit tomorrow.

src/toil/cwl/cwltoil.py
adamnovak: They can't have defaults with this argument order, I don't think; `local_directory` is a required argument.

The function always needs access to the relevant option values, even if those happen to be None.

src/toil/cwl/cwltoil.py
adamnovak: Yes, I will note that None is allowed.

src/toil/cwl/cwltoil.py
adamnovak: I'll tell it to demand an explicit job store if it sees a provisioner that it doesn't know about.

src/toil/cwl/cwltoil.py
adamnovak: I need the local directory input because the logic to determine if we're asking for a job store for AWS or GCP, or whether it would be appropriate to actually use the local directory, lives inside the function; I wouldn't know whether to pass it along or not without duplicating that logic.

All the logic to determine what `workdir` should be is in `main`, and it's complicated and hard to move safely. I think that, when we're inventing a local job store, we have to have it equal to `workdir`. I don't really want to make the function `generate_default_job_store_and_determine_workdir()`.

DailyDreaming (commented):
LGTM.

PR #3333, opened 2020-11-16
CWL: Reformat w/ Black and better support for tmpdir_prefix with suffix.

mr-c:
For `cwltool` `tmpdir_prefix` is not just a path to a directory, it can also include a suffix after the last path separator that become the prefix of the generated temporary directory.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -18,52 +18,45 @@
 # For an overview of how this all works, see discussion in
 # docs/architecture.rst
 import argparse
-import os
-import tempfile
-import json
-import sys
-import stat
-import logging
 import copy
-import functools
-import uuid
-import urllib
 import datetime
-from typing import (Text,
-                    Mapping,
-                    MutableSequence,
-                    MutableMapping,
-                    List,
-                    Dict,
-                    Union,
-                    Any,
-                    Iterator,
-                    TextIO,
-                    Set,
-                    Tuple,
-                    cast)
-
+import functools
+import json
+import logging
+import os
+import stat
+import sys
+import tempfile
+import urllib
+import uuid
+from typing import (
+    Any,
+    Dict,
+    Iterator,
+    List,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+    Set,
+    Text,
+    TextIO,
+    Tuple,
+    Union,
+    cast,
+)
 from urllib import parse as urlparse
 
-from schema_salad import validate
-from schema_salad.schema import Names
-from schema_salad.sourceline import SourceLine
-import schema_salad.ref_resolver
-
+import cwltool.builder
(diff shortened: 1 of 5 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -1625,8 +1855,8 @@ def main(args: Union[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
 
     # if tmpdir_prefix is not the default value, set workDir if unset, and move
     # workdir and the job store under it
-    if options.tmpdir_prefix != 'tmp':
-        workdir = tempfile.mkdtemp(dir=options.tmpdir_prefix)
+    if options.tmpdir_prefix != "tmp":
+        workdir = cwltool.utils.create_tmp_dir(options.tmpdir_prefix)
         os.rmdir(workdir)
```
mr-c, line 1860: Actual change 1/2 (the rest is just from reformatting with `black`)

src/toil/cwl/cwltoil.py
```diff
@@ -1637,14 +1867,15 @@ def main(args: Union[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
             # --tmpdir-prefix.
             #
             # If set, workDir needs to exist, so we directly use the prefix
-            options.workDir = options.tmpdir_prefix
+            options.workDir = cwltool.utils.create_tmp_dir(options.tmpdir_prefix)
```
mr-c, line 1870: Actual change 2/2 (the rest is just from reformatting with `black`)

DailyDreaming (approved):
This also seems to update the cwltool version and run CWL tests in parallel now according to the CPU count of the PC it's running on.  Approving, but please wait for tests to pass!

I haven't really given `Black` a chance before, but perhaps my disagreements with its decisions will be won over by the time spent not worrying about linting so much.  Since this is just in the CWL files, I'd say it's a good experiment to try.  I just wish we didn't have to smother the commit history.

src/toil/cwl/cwltoil.py
```diff
@@ -1825,19 +2156,15 @@ def set_secondary(fileobj):
...
-        toilStageFiles(
-            toil,
-            outobj,
-            outdir,
-            destBucket=options.destBucket)
+        toilStageFiles(toil, outobj, outdir, destBucket=options.destBucket)
 
         if runtime_context.research_obj is not None:
-            runtime_context.research_obj.create_job(outobj, None, True)
+            runtime_context.research_obj.create_job(outobj, True)
```
mr-c, line 2162: Another changed cwltool function call

mr-c:
Thanks @DailyDreaming I didn't know that merge commits were turned off. I do recommend a single pass of `black` in its own PR. Active PRs can then do their own `black` formatting prior to merging in, to remove merge conflicts.

PR #3335, opened 2020-11-23
Refactor WDL types

### LINKED ISSUE #3330: Refactor typing for WDL based on review
reference toil-719

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-728)
┆Issue Number: TOIL-728

w-gao:
Closes #3330. This is a continuation of #3304. 

I decided to remove `parse_value_from_type()` since I didn’t like how I made it so that file import and input validation are handled together inside one function. Hopefully this is a cleaner approach. 

Summary of changes:
- All WDL types are now represented as `WDLType` subclasses rather than extending `str`. 
- WDL inputs are properly checked during runtime. Required inputs that are not set are also checked (https://github.com/DataBiosphere/toil/pull/3304#discussion_r517591580). 
- Unread files are changed to `WDLFile` instances so we can differentiate them with strings. 
- File output w/ compound types. 

## Changelog Entry

 * Refactor WDL types to support Pair and Map types with file imports.


src/toil/wdl/wdl_types.py
```diff
@@ -1,41 +1,211 @@
...
+                raise WDLRuntimeError(f"Required input for '{self.name}' type not specified.")
+
+        return self._create(value)
+
+    def _create(self, value: Any) -> Any:
+        raise NotImplementedError
+
+    def __eq__(self, other):
+        # TODO: remove after refactor
+        return self.name.__eq__(other)
```
adamnovak, line 55: Is this line still needed, or is the refactor done enough to drop it?

src/toil/wdl/wdl_types.py
```diff
@@ -1,41 +1,211 @@
...
+    def default_value(self):
+        return ''
+
+    def _create(self, value: Any) -> Any:
+        if isinstance(value, tuple):
+            return value
+
+        # Unprocessed and unread files are represented as a tuple: (file_path, preserved_file_name).
+        # Unprocessed files (not imported to the fileStore) are tuples in the form of (file_path, None).
+        # Once the file is read (via readGlobalFile), it will change to the absolute path as a string.
```
adamnovak, line 136: This seems a bit complicated; maybe there should be a `WDLFile` class that can represent these states and give you named methods to ask about them.

Then instead of trying to guess why we care about `isinstance(x, tuple)` and `if x[1]:` we would be able to see that we are really doing `isinstance(x, WDLFile)` and `if x.is_processed()`. Otherwise, you kind of want comments at every point of use to explain stuff, like "Since it's a tuple, it represents a file".

I'd also quibble with "processed" as the word for having saved the file to the file store; it's too generic. Maybe you can call them "written" and "read" to match the methods on the file store?

Also, what actually is the lifecycle of a file? What are all the possible states and the transitions between them? It can't be that an unprocessed file is always `(file_path, None)` if the unprocessed files that are also unread are `(file_path, preserved_file_name)`. And how are processed but not yet read files represented?

Making this into its own class instead of a particular language of tuples might make it easier to explain the underlying concepts. If you don't do that, I think you should expand this description to more fully explain what kinds of files there are, and how each is represented as a tuple, and how and when one becomes another.

adamnovak (changes requested):
This looks pretty good. I'm happy with the type refactor; if you want to leave optionality as a flag like this instead of its own node in the type tree, that's fine.

I'm not convinced by the file tuple representation, because I don't understand it. I made a refactoring suggestion that I think could make it clearer, but some more explanation of how it works now would also satisfy me.

src/toil/wdl/wdl_types.py
w-gao: Using a `WDLFile` class would indeed be better. Thanks for the suggestion! I will update my PR.

src/toil/wdl/wdl_types.py
w-gao: I decided to keep this since it makes sense to compare two types. All the types are fully refactored away from str, however.

adamnovak (approved):
This looks great now!

DailyDreaming (approved):
I agree with Adam, this looks great.  Merging this.

PR #3340, opened 2020-11-24
Re-enable running WDL GATK Tutorial Tests.

DailyDreaming:
Making these non-manual now.  GATK has changed quite a bit and it looks like the WDL tutorials contain commands that don't exist in GATK anymore (they don't exist in version 4 and the version 3 in our bucket seems to have imports that break it): https://support.terra.bio/hc/en-us/sections/360007347652-WDL-Tutorials

The scripts and data have been updated and uploaded back into the s3 test data bucket and the tests have been re-enabled to help prevent future breakage.  Tutorials 1-3 seem to still work.  Tutorial 4 is breaking though and the test is skipped until the issue can be addressed: https://github.com/DataBiosphere/toil/issues/3339

Not sure how java is setup on gitlab currently, so the test environment may need further tweaking.

## Changelog Entry

 * Re-enabled running WDL GATK Tutorial Tests 1-3.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,9 +83,10 @@ py37_wdl:
   stage: main_tests
   script:
     - pwd
-    - apt update && DEBIAN_FRONTEND=noninteractive apt install -y tzdata jq python3.7 python3.7-dev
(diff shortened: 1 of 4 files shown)


DailyDreaming:
Merging this.  I changed some of the test fixtures and it broke `toilwdlintegrationtest.testjson` on other builds, so this should fix things.

PR #3341, opened 2020-11-24
Add a retry logic for BucketNotEmpty error.

### LINKED ISSUE #3263: Rare `BucketNotEmpty` exception when cleaning S3 jobstore
I've noticed that we rarely get `BucketNotEmpty` exception (shown below) when trying to clean the s3 jobstore (likely due to some rare s3 API issues/inconsistencies). Would be good to include this as a retriable error in [`_delete_bucket`](https://github.com/DataBiosphere/toil/blob/6ad3e5eb0e0aa7067fe71a028ebb0f76de34f51e/src/toil/jobStores/aws/jobStore.py#L1562).

```
botocore.exceptions.ClientError: An error occurred (BucketNotEmpty) when calling the DeleteBucket operation: The bucket you tried to delete is not empty. You must delete all versions in the bucket.
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-685)
┆Issue Number: TOIL-685

arostamianfar:
Fixes #3263 

While ideally this should only be added to `_delete_bucket`, this exception is very specific and isn't expected to happen elsewhere, so didn't seem like it was worth the extra complexity, but lemme know if you feel strongly about only having it there!

### DIFF
--- a/src/toil/jobStores/aws/utils.py
+++ b/src/toil/jobStores/aws/utils.py
@@ -33,6 +33,7 @@ from boto.exception import (SDBResponseError,
                             S3CreateError,
                             S3CopyError)
 import boto3
+from botocore.exceptions import ClientError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


DailyDreaming (approved):
LGTM.  Thanks for adding this!

DailyDreaming:
Merging.  The failing test has nothing to do with this PR.

Issue #307, opened 2015-08-26
"make test" breaks if Mesos isn't installed


hannes-ucsc:
Reported by @joelarmstrong in #296.

hannes-ucsc:
I think we need to use the technique from #296 in order to exclude certain classes from test discovery.

PR #3350, opened 2020-11-30
CWL Test #307: Confirm CommandInputParameter expression can receive a File object.

DailyDreaming:
Currently depends on: https://github.com/common-workflow-language/cwltool/pull/1380

We already remove unprocessed secondaryFiles from a `List` of job params, but did not account for a `Dict` of job params.  This now filters appropriately on the additional job params `Dict` type.

There is also a second issue.

In `cwltool`, it swaps:

toilfs:0:files/no-job/**file-ccead684521d44f98ff81b118f37159b**/secondary_file_test.txt.accessory

for

toilfs:0:files/no-job/**file-1120f3b05d62444083a9ff5c60c7f772**/secondary_file_test.txt.accessory

The bolded portion is the same basedir as the primary file, which creates a path to a file that does not exist.

The input we provide is below, so this should not be happening:

```
{
    "inputWithSecondary": {
        "class": "File",
        "location": "toilfs:0:files/no-job/file-1120f3b05d62444083a9ff5c60c7f772/secondary_file_test.txt",
        "size": 0,
        "basename": "secondary_file_test.txt",
        "nameroot": "secondary_file_test",
        "nameext": ".txt",
        "secondaryFiles": [
            {
                "location": "toilfs:0:files/no-job/file-ccead684521d44f98ff81b118f37159b/secondary_file_test.txt.accessory",
                "class": "File",
                "basename": "secondary_file_test.accessory",
                "nameroot": "secondary_file_test",
                "nameext": ".accessory",
                "size": 0
            }
        ]
    }
}
```

This seems to be because cwltool:
1. Assumes that primary and secondary files are in the same directory and thus have the same interchangeable basedir in the `toilfs:` schema.
1. Does not check to see if secondaryfiles have been resolved yet.  (the list has resolved secondary files in toil because we process with a builder already before it runs a builder again in cwltool)

https://github.com/common-workflow-language/cwltool/blob/78fe9d41ee5a44f8725dfbd7028e4a5ee42949cf/cwltool/builder.py#L428

Two possible proposed solutions:
 - The easier solution (imho) is to check for an already processed secondary file before fabricating a new (and potentially overwriting the old) secondary file's entry.
 - The alternative solution might be to force toil to place secondary files at the same basedir as the primary file if the first solution isn't preferable.

## Changelog Entry

 * Support CWL Test #307: Confirm CommandInputParameter expression can receive a File object.


DailyDreaming:
@mr-c @adamnovak I'm fine with adding this as-is, if either of you could review this :pray: , since it adds clean up that may avoid other bugs, and I'm fine with a separate PR to add/enable test 307 once either https://github.com/common-workflow-language/cwltool/pull/1380 has been reviewed or a different tact taken.

mr-c (approved):

DailyDreaming:
@mr-c Thanks!

PR #3356, opened 2020-12-01
Replace abssympath w/ builtin.

DailyDreaming:
Attempting to reduce the amount of needless complexity.

## Changelog Entry

 * Replace abssympath w/ builtin.

### DIFF
--- a/src/toil/jobStores/fileJobStore.py
+++ b/src/toil/jobStores/fileJobStore.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2016 Regents of the University of California
+# Copyright (C) 2015-2020 Regents of the University of California
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/toil/jobStores/fileJobStore.py
```diff
@@ -70,7 +66,7 @@ def __init__(self, path, fanOut=1000):
                            subdirectories
         """
         super(FileJobStore, self).__init__()
-        self.jobStoreDir = absSymPath(path)
+        self.jobStoreDir = os.path.realpath(path)  # doesn't dereference symlinks
```
xbrianh, line 69: This comment is confusing. `os.path.realpath` replaces symlinks with [the actual path](https://docs.python.org/3/library/os.path.html#os.path.realpath).  Is this the desired behavior?

xbrianh (approved):
See comment, otherwise lgtm.

src/toil/jobStores/fileJobStore.py
DailyDreaming: *sigh*  It's not the desired behavior. I'll adjust and include a test for this.  Thanks for the review, really appreciate it.

PR #3362, opened 2020-12-08
Add a better error message for insufficient disk.

### LINKED ISSUE #3359: InsufficientSystemResources error message is confusing
> toil.batchSystems.abstractBatchSystem.InsufficientSystemResources: The job CWLWorkflow is requesting 2147483648 bytes of disk, more than the maximum of 1474707456 bytes of disk that SingleMachineBatchSystem was configured with. Scale is set to 1.

This error doesn't explain where "the maximum of 1474707456 bytes of disk that SingleMachineBatchSystem was configured with" comes from.  This makes troubleshooting a simple problem unnecessarily difficult.

What was actually happening was that it was checking the free space on `/tmp`.  I have a separate `/tmp` partition which is only GiB in size and so it correctly couldn't reserve 2 GiB of space.  However, the error message doesn't say which disk does not have sufficient space, nor does it explain exactly what the disk space is for.

Suggest something like:

> toil.batchSystems.abstractBatchSystem.InsufficientSystemResources: The job CWLWorkflow is requesting 2147483648 bytes of disk for temporary space, more than the 1474707456 bytes of free space on /tmp that SingleMachineBatchSystem was configured with. Try changing the base temporary directory by setting TMPDIR.  Scale is set to 1.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-739)
┆Issue Number: TOIL-739

DailyDreaming:
Resolves #3359 .  Also refactored a little, somewhat due to making it easier to pass the `dir` name needed for the error message.

## Changelog Entry

 * Better error message for insufficient disk.

### DIFF
--- a/src/toil/__init__.py
+++ b/src/toil/__init__.py
@@ -144,7 +144,7 @@ def resolveEntryPoint(entryPoint):
 
 
 @memoize
-def physicalMemory():
+def physicalMemory() -> int:
     """
     >>> n = physicalMemory()
     >>> n > 0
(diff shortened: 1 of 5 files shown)


src/toil/__init__.py
```diff
@@ -144,7 +144,7 @@ def resolveEntryPoint(entryPoint):
 
 
 @memoize
-def physicalMemory():
+def physicalMemory() -> int:
```
mr-c, line 147: Do you want to expand the mypy CI checks?

mr-c (approved):

src/toil/__init__.py
DailyDreaming: Good point!  I will add it in a separate PR.

adamnovak (approved):

PR #3364, opened 2020-12-09
Add WDL as_pairs() function + tests.

### LINKED ISSUE #3200: as_pairs() implemented for WDL.

w-gao:
Closes #3200. 

## Changelog Entry

 * Add as_pairs() WDL function.


src/toil/test/wdl/standard_library/as_pairs_as_input.wdl
```diff
@@ -0,0 +1,18 @@
+workflow asPairsWorkflow {
```
w-gao, line 1: @DailyDreaming It looks like `as_pairs()` is added in the development version so this workflow fails in Cromwell. The following workflow works in Cromwell but doesn't work with our runner since our parser is on the `draft-2` version, I think?

```wdl
version development

workflow asPairsWorkflow {
  input {
    Map[String, Int] in_map
  }

  call copy_output {input: in_pairs=as_pairs(in_map)}
}

task copy_output {
  input {
    Array[Pair[String, Int]] in_pairs
  }

  command {
    cp ~{write_json(in_pairs)} output.txt
  }

  output {
    File the_output = 'output.txt'
  }
}
```

DailyDreaming: @w-gao You can find pregenerated development versions here:

Hermes (an updated version of what we currently use): https://github.com/openwdl/wdl/blob/main/versions/development/parsers/hermes/python/wdl_parser.py

And a less direct antlr4 python3 lexer here: https://github.com/openwdl/wdl/tree/main/versions/development/parsers/antlr4/Python

The Hermes lexer should just work if swapped out.  The antlr4 lexer appears to need to be generated.  If you want to try it, the antlr4 lexer looks like the better path forward, but the old Hermes parser would be just fine too.

w-gao: Great! I’ll see if I can get the antlr4 lexer to work first. 

Should I finish the rest of the draft-2 functions before I update our WDL version?

DailyDreaming: Up to you, though I think that updating the lexer/WDL version first probably makes sense.

DailyDreaming (approved):
LGTM.

PR #3370, opened 2020-12-14
get boto S3 args from environment variables

stevekm:
## Changelog Entry

fix for https://github.com/DataBiosphere/toil/issues/3365 in order to get S3 endpoint configuration for non-AWS S3 endpoints, such as MinIO. 

I ran this test and it passed;

```
BOTO_HOST=127.0.0.1 BOTO_PORT=9010 BOTO_SECURE=False BOTO_CALLING_FORMAT=OrdinaryCallingFormat python -m pytest src/toil/test/cwl/cwlTest.py::CWLv10Test::test_run_s3
```
after setting up a demo MinIO instance hosting the same S3 files as documented [here](https://github.com/stevekm/minio-workflow-demo/blob/1b73b9b86d9b797ba324520fa696d4bd480d5d26/Makefile#L168).

Would be nice to have these options specified from the command line for the Toil CWL runner as well.


mr-c:
Very exciting! Would be great to have that test integrated.

DailyDreaming (commented):
This looks great, thank you for this feature!  :slightly_smiling_face: 

We should have tests for this to ensure it maintains consistent behavior, as well as adding  clear documentation to the readthedocs.

I wonder, could the s3 args be set to one environment variable to be parsed as json?  Something like:

```
S3_BOTO_ARGS='{"host": "host", "port": "port", "is_secure": false, "calling_format": "ordinarycallingformat"}'
```

This has the advantage of parsing the `bool` args for you in a consistent way, though `calling_format` would still need to be swapped out.

Thanks again, this looks great.

DailyDreaming:
Looking at this further, it looks like we create and interact with buckets using `boto`, but delete them with `boto3`.  The `boto3` incantation for this appears to be something like:

```
s3client = boto3.client('s3', endpoint_url='http://X.X.x.X:8080/')
```

Have there been any issues in deleting the buckets?  I know our boto and boto3 clients overlap, so it might be able to, but we should test it and be doubly sure.  The plan is also to move completely over to `boto3` at some point so we should definitely test this with a `boto3` client.

stevekm:
yeah, if you use `boto3` you only need to pass a single arg, the `endpoint_url`, but I did not want to get involved with swapping `boto` for `boto3` just to get this feature implemented. 

The docs for deleting dont make it sound much different between the two, not sure if it would be a big deal in practice

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/migrations3.html#deleting-a-bucket

DailyDreaming:
@stevekm I'll add a test, if you don't mind.  

Hmmm... it looks like `is_secure` has a new look in boto3:

**boto**:
```
is_secure = False
https_validate_certificates = False
```
http://boto.cloudhackers.com/en/latest/boto_config_tut.html#boto

**boto3**:
```
use_ssl = False
verify = False
```
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client

Some proposed changes... would you consider renaming these to:

BOTO_HOST -> TOIL_S3_HOST
BOTO_PORT -> TOIL_S3_PORT
BOTO_SECURE -> TOIL_S3_USE_SSL

...and possibly omitting BOTO_CALLING_FORMAT and always using the default of `boto.s3.connection.OrdinaryCallingFormat()`?

stevekm:
I will make those changes :+1:

as for the CI tests, I am actually not sure how you would go about that since it requires that there's an S3-compatible object store running at the listed IP & port. It might be a lot of overhead to e.g. include a MinIO service inside your testing infrastructure.

stevekm:
The equivalent in `boto3` is this;

```
s3 = boto3.resource('s3', endpoint_url = endpoint_url, aws_access_key_id = aws_access_key_id, aws_secret_access_key = aws_secret_access_key)
```

so you do not need to specify the `is_secure` or connection type, at least I didn't. If you planned to eventually port this section over to `boto3` you might just omit those args, unless you want them to still be exposed to the end user.

stevekm:
tested this time with:

```
$ TOIL_S3_HOST=127.0.0.1 TOIL_S3_PORT=9010 TOIL_S3_USE_SSL=False python -m pytest src/toil/test/cwl/cwlTest.py::CWLv10Test::test_run_s3
```

DailyDreaming:
@stevekm Pretty cool.  I was able to at least manually run this.  I can add a test separately later on for this based on your Makefile setup.  Would you mind adding a short section of documentation about running Toil with Minio either here: https://github.com/DataBiosphere/toil/blob/master/docs/running/cloud/amazon.rst

Or in a new section under Cloud entirely?

And also the env vars to our env var list: https://github.com/DataBiosphere/toil/blob/master/docs/appendices/environment_vars.rst

I think after some documentation explaining this addition, I'd be fine with merging.

Thanks for adding this!

stevekm:
Sure. Are we going to stick with this env var config method? That is what I will describe in the docs then. Unless you wanted to also do the CLI arg methods.

DailyDreaming:
@stevekm Yes, I think that this env var config method makes sense and we can still use it when we refactor over to boto3.  I think that having CLI options in addition to the env vars would make sense (at some point we can get around to having both for all arguments), but it's also something that can be added later.  If you want to add it now, it's up to you (though if you do, the CLI arguments should have a higher priority).

Thanks for adding the documentation!

stevekm:
Ok I added a small section in the docs, let me know if more info is needed. Thanks.

DailyDreaming (approved):
@stevekm Looks great, thanks for the additions!  I think that this is good to merge...

PR #3373, opened 2020-12-17
Cleanup script for buckets, sdb domains, instance profiles, and roles.

DailyDreaming:
We let buckets build up to the point that tests errored this past week because we had hit the limit of s3 buckets we're allowed to have in the account.  I cleaned them up earlier but wanted to add a script to do it in the future.  I've also noticed that we have cruft in other regions (that is currently not cleaned up), so this now inspects all regions across the account.

Manually requires the user to inspect and submit "yes" or "y" to delete buckets/domains.

We mostly have test clean up I think(?), but if I hit cancel job in gitlab, or a test is wonky, it can leave behind its bucket and/or sdb domain, so this is a convenience script that we can run whenever that happens (or just occasionally).  Using `--view-only` allows you to just take a look at what's in the account.  This has a built-in filter so that it actively looks for buckets/domains matching our test patterns; this can be disabled with `--include-all`(but there's still a short whitelist of buckets to never view and never delete).

## Changelog Entry

 * Add an admin clean up script for simpledb domains and s3 buckets.


contrib/admin/cleanup_aws_resources.py
```diff
@@ -0,0 +1,325 @@
...
+                    print('       - ' + f'\n       - '.join(associated_roles) + '\n')
+                else:
+                    print('       - No roles found to be associated.\n')
+                instance_profiles[new_profile] = region
+        except Exception as e:
+            print(f'    An error occurred in this region: {e}')
+    return instance_profiles
+
+
+def find_buckets_in_region(s3_resource, include_all, match):
```
adamnovak, line 161: We have both `find_buckets_to_cleanup` and `find_buckets_in_region`. Deciding whether a bucket that exists look suspiciously test-like and needs cleaning up happens... in `find_buckets_in_region`, confusingly. `find_buckets_to_cleanup` just loops over the regions and prints all the buckets. And there aren't docstrings to explain this.

src/toil/lib/aws/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Caches all boto3 clients we instantiate."""
+from functools import lru_cache
+
+
+@lru_cache(maxsize=None)
+def client(*args, **kwargs):
+    import boto3
+    return boto3.client(*args, **kwargs)
```
adamnovak, line 21: I'm going to need to glue this onto the logic I've been using to establish a properly credential-cached Boto3 session for the AWS provisioner work I've been doing.

src/toil/lib/aws/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    s3_resource.Bucket(bucket).delete()
+    printq(f'\n * Deleted s3 bucket successfully: {bucket}\n\n', quiet)
+
+
+def bucket_in_use():
+    pass
+
+
+def iam_instance_profile_in_use():
+    pass
```
adamnovak, line 88: Did you forget about these or are they meant to be empty? What are they supposed to do?

adamnovak (changes requested):
I spent longer than I should have wandering through the cleanup script trying to satisfy myself that it wasn't just going to delete everything; the checks for what looks deletable are way at the bottom of the call tree and not in the functions with names that suggest that they would filter, and there aren't any docstrings to explain this.

Can you help me not get lost?

src/toil/lib/aws/utils.py
DailyDreaming: I just haven't taken the time to fill them in yet.  I wanted to implement them after your last comment and just put in a reminder for myself.

src/toil/lib/aws/__init__.py
DailyDreaming: That sounds good to me.

contrib/admin/cleanup_aws_resources.py
DailyDreaming: Will do some renaming.

DailyDreaming:
@adamnovak Will try to make the naming clearer and reduce/refactor the code a bit.  I've still to add the checks that ensure resources are not currently being used.  Thanks for looking over it.

adamnovak:
@DailyDreaming This PR didn't come up when we were talking about the next sprint, I don't think. Are you going to do the design changes I suggested? If you can't it might make sense to proceed without them; we kind of want to have these cleanup scripts.

DailyDreaming:
@adamnovak I should be able to.  Will investigate, and if not, will comment.

DailyDreaming:
@adamnovak We should probably proceed without them.  There isn't really a great way to tell if things are in use right now (we can infer... but we might be wrong), but it could easily be added in the form of flags within the bucket, and then the bucket would become the central authority on whether a workflow has fallen off the wagon or is still on its journey in the cloud.

We'll just have to be cautious using it in the meantime.  Killing a test isn't the end of the world and known specially protected buckets are skipped.

src/toil/lib/aws/__init__.py
adamnovak: I feel like this is going to create a default session to get this client, while we now have https://github.com/DataBiosphere/toil/blob/65699cd653d67e9717fdd9a35946550ea3ea21db/src/toil/lib/ec2.py#L74-L80 which is supposed to be the one place to get sessions.

We also might want to think about how/whether things should be split between `toil.lib.ec2` and `toil.lib.aws`.

src/toil/lib/misc.py
```diff
@@ -1,12 +1,18 @@
...
 import os
 import random
+import shutil
 import subprocess
 import sys
 
 logger = logging.getLogger(__name__)
 
 
+def printq(msg, quiet: bool):
```
adamnovak, line 11: This would also want to be `-> None`

adamnovak:
There's also a Travis test failure in `toil.test.provisioners.aws.awsProvisionerTest.AWSProvisionerBenchTest:testAMIFinding`, but I think that that test just can't run on Travis because it needs AWS credentials. I will make an issue.

adamnovak:
I think this is the next PR we want to merge. We should be able to merge it in like 5 hours, after all the re-testing except MyPy passes.

adamnovak:
This is up to date, and everything I expect to pass passes, so I am going to merge it.

Issue #333, opened 2015-08-31
GridEngine still failing under generic batch system tests


hannes-ucsc:
GridEngine was supposedly fixed by #213. In #331 I am in the process of adding GridEngine to the generic batch system tests and they are failing. In #282, @adderan is working on making the batch system tests work against Parasol.

hannes-ucsc:
I should add that this is on CI. So this might be a versioning issue. We should try running the tests on  podk.

hannes-ucsc:
John, if you could work on this issue in the PR #334 request, that be great.

PR #3385, opened 2020-12-22
CWL Test #333: Test input mount locations when container required.

DailyDreaming:
Looks like this was a simple fix.  Our `ToilPathMapper` is very nearly identical to the cwltool `PathMapper`, and so updates there should probably also be effected in Toil.  This could probably be refactored better on either side to reduce the amount of duplicated code (and increase the odds that cwltool side features/fixes will be automatically inherited by Toil).

## Changelog Entry

 * Support for CWL version 1.2: Passing CWL v1.2 Conformance Test #333.

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -37,6 +37,7 @@ from typing import (
     Mapping,
     MutableMapping,
     MutableSequence,
+    Optional,
     Text,
     TextIO,
     Tuple,
@@ -564,8 +565,12 @@ class ToilPathMapper(PathMapper):
         staged: bool = False,
     ) -> None:
         """Iterate over a CWL object, resolving File and Directory path references."""
+        stagedir = cast(Optional[str], obj.get("dirname")) or stagedir
         tgt = convert_pathsep_to_unix(
-            os.path.join(stagedir, cast(str, obj["basename"]))
+            os.path.join(
+                stagedir,
+                cast(str, obj["basename"]),
+            )
         )
         if obj["location"] in self._pathmap:
             return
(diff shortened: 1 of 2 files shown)


mr-c:
A `PathMapper` refactor would be very welcome!

DailyDreaming:
Docker pull throttling again.  >.<  Will rerun in the morning when this resets.

src/toil/cwl/cwltoil.py
```diff
@@ -579,7 +584,7 @@ def visit(
                 resolved, tgt, "WritableDirectory" if copy else "Directory", staged
             )
 
-            if location.startswith("file://") and not self.stage_listing:
+            if location.startswith("file://"):  # and not self.stage_listing
```
adamnovak, line 587: I take it this condition shouldn't be here anymore?

adamnovak (changes requested):
Looks pretty good, but some commented-out code is still there.

src/toil/cwl/cwltoil.py
DailyDreaming: Yep, changing.  Thanks for the catch!

adamnovak (approved):

xbrianh (approved):
Overall this LGTM, however:

Seemingly the reason to use `typing.cast` is to satisfy static analysis. Using it complexifies your code while preventing static analysis from catching type errors. This is the worst of both worlds. I recommend discouraging the use of `typing.cast` in the future, and understanding how to to remove current usage.

DailyDreaming:
A lot of our `typing.cast` usage is inherited from cwltool as a strong convention there (where typing in general is more complex, having `MutableMap` instead of a `dict` most of the time, and many items cast as CWL Objects).  It is some extra overhead, so I should probably read up on their documentation surrounding their typing system, as you have good points that I don't have a good answer to.

Issue #3384, opened 2020-12-22
Update WDL lexer


w-gao:
The WDL version needs to be updated in order to support the newer standard library functions as discussed in https://github.com/DataBiosphere/toil/pull/3364#discussion_r539558414.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-762)
┆Issue Number: TOIL-762

PR #3391, opened 2020-12-24
Make AnalyzeWDL abstract to allow implementation for different WDL versions.

w-gao:
In preparation for #3384. This contains a lot of changes but they are mostly moving files around, so I'm making a separate PR for this to make review easier.

Also removed `wdl_parser.py` and pulled in https://github.com/DataBiosphere/wdl-parsers where it contains the parsers for all WDL versions.


src/toil/wdl/toilwdl.py
```diff
@@ -17,21 +17,21 @@
...
 def main():
-    '''
+    """
     A program to run WDL input files using native Toil scripts.
 
     Calls two files, described below, wdl_analysis.py and wdl_synthesis.py:
 
-    wdl_analysis reads the wdl, json, and extraneous files and restructures them
-    into 2 intermediate data structures before writing (python dictionaries):
+    wdl_analysis reads the wdl and restructures them into 2 intermediate data
```
DailyDreaming, line 33: Extraneous `tsv` files now break, but `json` is still correct, right?

DailyDreaming (approved):
Overall, this looks pretty good to me.  There are two minor comments, see below.  I like the inheritance of `AnalyzeDraft2WDL(AnalyzeWDL)`, and think it should work well for multiple versions as a place that can be used to consolidate functions.

Approving, but please take a look at the comments!  :slightly_smiling_face:

src/toil/wdl/toilwdl.py
w-gao: Good catch. I just moved that part to the `wdl_synthesis` section below since JSON files were never used in the analysis part. 

I also updated the docs in 3d116586afe756fe4454ee3e3cbc19fbf92f8e77 so that needs a review as well. Thanks!

DailyDreaming (approved):
Looks great!  Thanks for addressing the comments.

PR #3394, opened 2020-12-30
Make building a flattened list of live jobs non-recursive.

### LINKED ISSUE #3393: Toil fails with RecursionError when restarting big workflow
```
  File "//venv/bin/cactus-align", line 8, in <module>
    sys.exit(main())
  File "//venv/lib/python3.6/site-packages/cactus/setup/cactus_align.py", line 139, in main
    results_dict = toil.restart()
  File "/usr/local/lib/python3.6/dist-packages/toil/common.py", line 860, in restart
    rootJobGraph = self._jobStore.clean(jobCache=self._jobCache)
  File "/usr/local/lib/python3.6/dist-packages/toil/jobStores/abstractJobStore.py", line 526, in clean
    getConnectedJobs(self.loadRootJob())
  File "/usr/local/lib/python3.6/dist-packages/toil/jobStores/abstractJobStore.py", line 517, in getConnectedJobs
    getConnectedJobs(getJob(successorJobStoreID))
  File "/usr/local/lib/python3.6/dist-packages/toil/jobStores/abstractJobStore.py", line 517, in getConnectedJobs
    getConnectedJobs(getJob(successorJobStoreID))
  File "/usr/local/lib/python3.6/dist-packages/toil/jobStores/abstractJobStore.py", line 517, in getConnectedJobs
    getConnectedJobs(getJob(successorJobStoreID))
  [Previous line repeated 992 more times]
  File "/usr/local/lib/python3.6/dist-packages/toil/jobStores/abstractJobStore.py", line 514, in getConnectedJobs
    for successorJobStoreID in [x.jobStoreID for x in jobs]:
RecursionError: maximum recursion depth exceeded
```

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-765)
┆Issue Number: TOIL-765

DailyDreaming:
Resolves #3393.

## Changelog Entry

 * Make building a flattened list of live jobs non-recursive.


src/toil/jobStores/abstractJobStore.py
```diff
@@ -520,47 +520,67 @@ def getJobDescriptions():
...
+            def successors(job_descriptions: List[JobDescription]) -> Iterable[JobDescription]:
+                """
+                For a list of job description objects, this will iterate each job description object's stack
+                of connected jobs and yield any that have both an active job and have not yet been added to
+                reachable_from_root.
+                """
+                for job_description in job_descriptions:
+                    for jobs in job_description.stack:
+                        for successor_jobstore_id in jobs:
+                            if successor_jobstore_id not in reachable_from_root and haveJob(successor_jobstore_id):
```
xbrianh, line 538: Since reachable_from_root is a set. What is the purpose of the membership check?

Set lookup is similar complexity to insertion, so you can save half the work by skipping the membership check (which will probably be meaningless to Toil's overall performance).

xbrianh (commented):
This PR involves changes to multiply nested functions. Since such functions live in a large external namespace that I don't understand, I don't feel I can provide a meaningful review.

src/toil/jobStores/abstractJobStore.py
DailyDreaming: Ah, it's that it adds the job id plus all attached service ids if it's not in the set.  I can see about replacing the check for a length check after adding to see if that's more efficient.  This graph can get large (at least enough to break recursion, but possibly in the millions, though much more often less than 100), so gains in efficiency here might be noticeable for particularly large workflows.

Thanks for taking a look!  :slightly_smiling_face:

DailyDreaming:
@xbrianh The comment was helpful.  Thanks for the review!

DailyDreaming:
Interestingly, efficiency actually seems better for the check `in set()` vs. `len(set())` with some naive runs:

```
quokka@qcore 06:16 PM ~$ python3.8 -mtimeit -s "a={i for i in range(10000)};" "55 in a"
20000000 loops, best of 5: 18.4 nsec per loop
quokka@qcore 06:16 PM ~$ python3.8 -mtimeit -s "a={i for i in range(10000)};" "len(a) == 55"
5000000 loops, best of 5: 50.1 nsec per loop
```

I think I'll leave it as is rather than trying to make changes with small gains, and I'm pretty confident that this won't break things, but I'll wait on a review from @adamnovak .

src/toil/jobStores/abstractJobStore.py
```diff
@@ -520,47 +520,67 @@ def getJobDescriptions():
...
+                                yield getJobDescription(successor_jobstore_id)
+
+            def add_to_reachable_from_root(job_description: JobDescription) -> None:
+                """Add this job's jobstore ID, and any dependent service job's jobstore IDs."""
+                reachable_from_root.add(job_description.jobStoreID)
```
adamnovak, line 543: It seems a little weird that we're using this variable scoped to the enclosing function that hasn't been defined yet, and that the name of the variable has made it into the internal function name.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -520,47 +520,67 @@ def getJobDescriptions():
...
+                    new_job_descriptions_to_process.append(successor_job_description)
+                unprocessed_job_descriptions = new_job_descriptions_to_process
+
+            logger.debug(f"{len(reachable_from_root)} jobs reachable from root.")
+            return reachable_from_root
+
+        reachableFromRoot = get_jobs_reachable_from_root()
 
         # Cleanup jobs that are not reachable from the root, and therefore orphaned
+        # TODO: Avoid reiterating reachable_from_root (which may be very large)
```
adamnovak, line 569: Do you mean `reachableFromRoot`?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -520,47 +520,67 @@ def getJobDescriptions():
...
-        logger.debug("Checking job graph connectivity...")
-        getConnectedJobs(self.loadRootJob())
-        logger.debug("%d jobs reachable from root." % len(reachableFromRoot))
+        def get_jobs_reachable_from_root() -> Set[Union[TemporaryID, str]]:
+            """
+            Traverse the job graph from the root job and return a flattened set of all active jobstore IDs.
+
+            Note: Jobs returned by self.jobs(), but not this function, are orphaned, and can be removed as dead jobs.
+            """
+            def successors(job_descriptions: List[JobDescription]) -> Iterable[JobDescription]:
```
adamnovak, line 529: This new function is only called once.

adamnovak (approved):
I think this should work. I'm a little confused by the order in which things are explained; you have a bunch of functions inside a function inside a method, which all talk to the one `reachable_from_root` variable, not to be confused with `reachableFromRoot`.

DailyDreaming:
@adamnovak I was hoping that the functions would break it up and make it clearer, but I see it did not have the intended effect.  I'll refactor back into one continuous function.  Thanks for the review!

PR #3396, opened 2021-01-04
Remove unused --defaultPreemptable option.

DailyDreaming:
This option had no `help` message and went I went to look at what it did, it seemed completely unused.

## Changelog Entry

 * Remove unused --defaultPreemptable option.


adamnovak:
This has broken the AWS provisioner tests I am trying to reenable, and we need the option because it is *supposed* to make jobs be marked as preemptable in their requirements by default. Only preemptable jobs are allowed to run on spot nodes.

It looks like this commit pulls it out of the tests, but f483b819ef8f7c954d2926f2e8540d94af1892fe which actually merged only pulls it out of the command line parser, so the tests still want to use it.

I'm going to need to put it back as part of my big Kubernetes PR that re-enables the provisioner tests, I think.

adamnovak:
I think it is pulled out by some special string lookup when we determine resource requirements for jobs that don't set them, which is why it looks unused.

PR #3400, opened 2021-01-04
Move AWS job store uploads/downloads to boto3 (reopened)

w-gao:
## Changelog Entry

 * Move AWS job store uploads/downloads to boto3

### DIFF
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -28,9 +28,7 @@ from io import BytesIO
 from typing import Optional
 
 import boto3
-import boto.s3
 import boto.sdb
-import boto.s3.connection
 import botocore.credentials
 import botocore.session
 from boto.exception import SDBResponseError
@@ -511,8 +509,8 @@ class AWSJobStore(AbstractJobStore):
         except ClientError as e:
             if e.response.get('ResponseMetadata', {}).get('HTTPStatusCode') == 404:
                 objExists = False
-        else:
-            raise
+            else:
+                raise
         if existing is True and not objExists:
             raise RuntimeError(f"Key '{keyName}' does not exist in bucket '{bucketName}'.")
         elif existing is False and objExists:
--- a/src/toil/jobStores/aws/jobStore.py
+++ b/src/toil/jobStores/aws/jobStore.py
@@ -24,8 +24,9 @@ import urllib.error
 import urllib.parse
 import urllib.request
 import uuid
-from contextlib import closing, contextmanager
+from contextlib import contextmanager
 from io import BytesIO
+from typing import Optional
 
 import boto3
 import boto.s3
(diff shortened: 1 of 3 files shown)


DailyDreaming:
@w-gao Thanks for reopening this!

DailyDreaming:
@w-gao Is this ready for review?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -407,141 +421,104 @@ def getEmptyFileStoreID(self, jobStoreID=None, cleanup=False, basename=None):
...
+        if os.environ.get('TOIL_S3_USE_SSL', True) == 'False':
+            protocol = 'http'
+        if host:
+            botoargs['endpoint_url'] = f'{protocol}://{host}' + f':{port}' if port else ''
 
-        if botoargs:
-            botoargs['calling_format'] = boto.s3.connection.OrdinaryCallingFormat()
+        # TODO: OrdinaryCallingFormat equivalent in boto3?
+        # if botoargs:
+        #     botoargs['calling_format'] = boto.s3.connection.OrdinaryCallingFormat()
```
w-gao, line 499: I wasn't able to find `calling_format` in boto3 and test the endpoint configuration added in #3370, but this should be properly tested.

w-gao (commented):
@DailyDreaming I think I addressed most of the review comments in #3153. I’ll leave the replacement of old retries for another ticket since I feel like it would involve breaking the boto3 operations into its own functions in order to use the new retry decorator. 

I noted a few parts that I wasn’t sure about below, but I think this should be ready for review. Thanks again for the thorough review in #3153!

src/toil/jobStores/aws/utils.py
```diff
@@ -365,15 +403,19 @@ def retry_sdb(delays=default_delays, timeout=default_timeout, predicate=retryabl
...
-            or (isinstance(e, ClientError) and 'BucketNotEmpty' in str(e)))
+            # boto3 errors
+            # TODO: switch to @retry decorators
+            or (isinstance(e, ClientError) and
+                ('BucketNotEmpty' in str(e)
+                 # TODO: test these!
+                 or (e.response.get('ResponseMetadata', {}).get('HTTPStatusCode') == 409
+                     and 'try again' in str(e))
+                 or e.response.get('ResponseMetadata', {}).get('HTTPStatusCode') in (
+                     404, 500, 502, 503, 504))))
```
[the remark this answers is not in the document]
DailyDreaming: These look fine to me.

src/toil/jobStores/aws/jobStore.py
DailyDreaming: There isn't one I believe, so we can probably just drop it.  I'm fine with doing that in a subsequent PR and testing myself.  You'd need to download and install minion (which is kind of neat actually).  You're right, it needs proper testing, but that's for a future sprint I think.

DailyDreaming (approved):
This looks great @w-gao .  Thank you for your quick work on this!

PR #3409, opened 2021-01-07
Gridengine mem error

douglowe:
The check for maxMEM fails for 5.1.0 for the grid engine interface - this change corrects the test, so it compares like with like.

## Changelog Entry

 * gridengine maxMEM check now uses MemoryString


douglowe:
I'm going to tidy up my branch - base it on the master instead of the 5.1.0 tag, and resubmit this PR

PR #3410, opened 2021-01-07
Corrected gridengine maxMEM check to use memorystring, not a plain int 0

douglowe:
Using v5.1.0 with SGE kicked out an error about a missing byte subclass for the int object - so I've corrected this check to use the same MemoryString construct that maxMEM is initialised with.

## Changelog Entry

 * bugfix for gridengine maxMEM check


DailyDreaming (approved):
LGTM.  Thanks for the fix!

PR #3416, opened 2021-01-09
Update dictionary structure in AnalyzeWDL.

w-gao:
A few internal structural changes to `self.workflows_dictionary` in AnalyzeWDL to make integration with the other versions easier. In preparation for #3384.

Summary
* draft-2 implementation changes
  - Use `parse_declaration()` and return a `tuple=(name, type, expr)` for both workflow and task declarations
  - Use the same `parse_workflow_body()` function to parse all inner workflow elements (declarations, calls, scatters, and conditionals) 
    * declarations are treated the same as calls, scatters, and conditionals instead of declaring them at the top, so the order will be preserved in cases like [this example workflow](https://github.com/openwdl/wdl/blob/main/versions/draft-2/SPEC.md#declarations):
```wdl
workflow wf {
  call test as x {input: var="x"}
  call test as y {input: var="y"}
  Array[String] strs = [x.value, y.value]  # declaration should come after the calls in this case
  call test2 as z {input: array=strs}
}
```

* Add `get_analyzer()` function to detect WDL version and return the correct AnalyzeWDL instance
* Misc
  - Some docstring/import formatting
  - Move `write_mappings()` to `wdl/utils.py`
  - Handle cases where `if`, `scatter`, or `command` sections are empty

## Changelog Entry

 * Internal update to the dictionary structures in AnalyzeWDL

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -8,27 +8,29 @@ from typing import List, Optional
 
 from toil.test import ToilTest
 from toil.version import exactPython
-from toil.wdl.wdl_functions import (WDLPair,
-                                    WDLRuntimeError,
-                                    ceil,
-                                    cross,
-                                    floor,
-                                    length,
-                                    read_boolean,
-                                    read_float,
-                                    read_int,
-                                    read_json,
-                                    read_lines,
-                                    read_map,
-                                    read_string,
-                                    read_tsv,
-                                    sub,
-                                    transpose,
-                                    wdl_zip,
-                                    write_json,
-                                    write_lines,
-                                    write_map,
-                                    write_tsv)
+from toil.wdl.wdl_functions import (
+    WDLPair,
+    WDLRuntimeError,
+    ceil,
+    cross,
+    floor,
+    length,
+    read_boolean,
+    read_float,
+    read_int,
+    read_json,
+    read_lines,
+    read_map,
+    read_string,
+    read_tsv,
+    sub,
+    transpose,
+    wdl_zip,
+    write_json,
+    write_lines,
+    write_map,
+    write_tsv
+)
 
 
 class WdlStandardLibraryFunctionsTest(ToilTest):
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -8,29 +8,27 @@ from typing import List, Optional
 
 from toil.test import ToilTest
 from toil.version import exactPython
-from toil.wdl.wdl_functions import (
-    WDLPair,
-    WDLRuntimeError,
-    ceil,
-    cross,
-    floor,
-    length,
-    read_boolean,
-    read_float,
-    read_int,
-    read_json,
-    read_lines,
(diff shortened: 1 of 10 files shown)


src/toil/wdl/utils.py
```diff
@@ -32,3 +77,79 @@ def dict_from_JSON(JSON_file):
...
             json_dict[d] = data[d]
     return json_dict
+
+
+def write_mappings(parser: AnalyzeWDL) -> None:
+    """
+    Intended to take an AnalyzeWDL instance and prints the final task dict,
+    and workflow dict.
+
+    Does not work by default with toil since Toil actively suppresses stdout
```
adamnovak, line 87: We shouldn't merge a thing that says it "does not work". Maybe make the filename a parameter or make it take a stream, and document it as writing to that filename or stream?

src/toil/wdl/wdl_synthesis.py
```diff
@@ -300,8 +305,14 @@ def fetch_ignoredifs_chain(self, assignments, breaking_assignment):
         return False
 
     def write_main_jobwrappers_if(self, if_statement):
+        # check for empty if statement
+        if not if_statement:
+            return self.indent_docstring('        pass')
```
adamnovak, line 310: This is not a docstring. Should the method just be called `indent` or something?

adamnovak (changes requested):
There's a few places where I want to see the documentation or naming of things improved, but overall I like it.

src/toil/test/wdl/builtinTest.py
```diff
@@ -8,27 +8,29 @@
...
+    read_string,
+    read_tsv,
+    sub,
+    transpose,
+    wdl_zip,
+    write_json,
+    write_lines,
+    write_map,
+    write_tsv
+)
```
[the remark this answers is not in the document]
DailyDreaming: This is controlled with `isort`: https://github.com/DataBiosphere/toil/blob/master/Makefile#L232

I'm not opposed to the new syntax, though the change should be made to the `Makefile` and then run. I think that the option would be the `Vertical Hanging Indent` though it leaves an extra comma.

src/toil/wdl/utils.py
DailyDreaming: This one is my fault.  It's an old function I used for debugging and developing the WDL parser.  It prints the intermediate data structures so that you can look at what it's doing before creating the final file.  It's useful to leave in, but it needs the `--debugWorker` flag if I remember right to print stdout normally.

w-gao: It looks like this function no longer prints stuff to stdout, so perhaps the docstring is just outdated. 

I'll make the filename a parameter and update the docstring. I think this function is useful to keep - I use it a lot for debugging as well.

src/toil/test/wdl/builtinTest.py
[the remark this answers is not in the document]
w-gao: Oops, my bad. I'll revert this part to keep the styling consistent.

adamnovak (approved):

PR #3423, opened 2021-01-21
Check google storage import as evidence of google extras installed.

DailyDreaming:
I'm not sure why we were checking boto for our tests.

## Changelog Entry

 * Better Google check when testing.


adamnovak (approved):
This makes sense.

I think we were checking for `boto` because (at one time?) we used the `boto` module to access both AWS and Google storage.

PR #3424, opened 2021-01-21
Update dependencies.

DailyDreaming:
In attempting to address #2230 (whose major concern seems to be solved already), I've updated some of our boto  and dev dependencies.

## Changelog Entry

 * Update dependencies (boto, boto3, futures) to be more recent and slightly more forgiving.  Update dev dependencies (pytest, pytest-cov, setuptools).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,9 @@
 mock==4.0.3
-pytest==6.2.0
-pytest-cov==2.10.1
+pytest==6.2.1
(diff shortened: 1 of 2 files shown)


adamnovak (approved):
This also looks like a good idea.

Are we using the new constraint-solving `pip` resolver in our CI currently?

DailyDreaming:
@adamnovak I don't believe so, but I can look into it.

PR #3428, opened 2021-01-23
Allow text/encode options for read and write functions in file store

### LINKED ISSUE #2918: file store read/write open function should  probably allow binary/encode for PY3
I suspect we need to add binary/text, encode, and error options similar to open() to the job store read and write functions for python 3.

For instance,` fileHandle.write("failed once\n")`
in toil/test/src/resumabilityTest.py will fail
with `TypeError: a bytes-like object is required, not 'str'`

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-484)
┆Issue Number: TOIL-484

jonathanxu18:
Should fix #2918.


src/toil/jobStores/abstractJobStore.py
```diff
@@ -903,6 +904,15 @@ def writeFileStream(self, jobStoreID=None, cleanup=False, basename=None):
                file basename so that when searching the job store with a query
                matching that basename, the file will be detected.
 
+        :param str mode: an optional string that specifies the mode in which the file is opened.
+                It defaults to 'b' which means open for reading in binary mode.
```
adamnovak, line 908: This needs to define the supported mode characters. Do I pass `'text'` or `'t'` for text mode? Or just an empty string?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -903,6 +904,15 @@ def writeFileStream(self, jobStoreID=None, cleanup=False, basename=None):
...
+
+        :param str encoding: the name of the encoding used to decode or encode the file.
+                This should only be used in text mode.
```
adamnovak, line 911: This should say which encodings are supported; maybe something like "Encodings are the same as for `str.encode()`".

Also, what encoding is the default in text mode?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -903,6 +904,15 @@ def writeFileStream(self, jobStoreID=None, cleanup=False, basename=None):
...
+
+        :param str errors: an optional string that specifies how encoding and decoding errors
+                are to be handled. This cannot be used in binary mode.
```
adamnovak, line 914: Similarly, this should explain what values are acceptable, or where the list of acceptable values is. Are all the ones for `str.encode()` supported? And what is the default?

src/toil/jobStores/fileJobStore.py
```diff
@@ -409,16 +409,27 @@ def writeFile(self, localFilePath, jobStoreID=None, cleanup=False):
...
+                # to clean ourselves up, somehow, for certain workloads.
+                yield f, relPath
+
+        elif mode == 't':
+            with open(absPath, 'w', encoding=encoding, errors=errors) as f:
+                # Don't yield while holding an open file descriptor to the temp
+                # file. That can result in temp files still being open when we try
+                # to clean ourselves up, somehow, for certain workloads.
+                yield f, relPath
+
```
adamnovak, line 432: What if someone passed `mode='wombats'`? It looks like that won't produce an informative error.

src/toil/jobStores/fileJobStore.py
```diff
@@ -409,16 +409,27 @@ def writeFile(self, localFilePath, jobStoreID=None, cleanup=False):
...
         absPath = self._getUniqueFilePath(basename, jobStoreID, cleanup)
         relPath = self._getFileIdFromPath(absPath)
-        with open(absPath, 'wb') as f:
-            # Don't yield while holding an open file descriptor to the temp
-            # file. That can result in temp files still being open when we try
-            # to clean ourselves up, somehow, for certain workloads.
-            yield f, relPath
+
+        if mode == 'b':
+            with open(absPath, 'wb', errors=errors) as f:
```
adamnovak, line 420: Does it make any sense to pass `errors` to `open()` in binary mode?

src/toil/jobStores/fileJobStore.py
```diff
@@ -565,19 +589,30 @@ def _getSharedFilePath(self, sharedFileName):
...
-                yield f
+            if mode == 'b':
+                with open(self._getSharedFilePath(sharedFileName), 'rb', errors=errors) as f:
+                    yield f
+
+            elif mode == 't':
+                with open(self._getSharedFilePath(sharedFileName), 'r', encoding=encoding,
+                        errors=errors) as f:
+                    yield f
+
```
adamnovak, line 615: You have a lot of copies of this same sort of structure; could you condense this and maybe put the conditional inside the arguments?

```
                with open(self._getSharedFilePath(sharedFileName), 'r' + (mode if mode != 't' else ''), encoding=encoding,
                        errors=errors) as f:
```

Or if `open` will take `t` you could do:

```
                with open(self._getSharedFilePath(sharedFileName), 'r' + mode, encoding=encoding,
                        errors=errors) as f:
```

Then you would be passing `errors` and `encoding` along even in binary mode, but there would be a good reason for it.

adamnovak (changes requested):
This looks like it should mostly work. But it's a bit repetitive, and it will get even more so if you add error checking to reject nonsense modes. I recommend you change it a bit to pass the mode characters along to the backing functions where possible; then they can reject nonsense modes for you, and you won't have 2 copies of the actual opens everywhere.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -903,6 +904,24 @@ def writeFileStream(self, jobStoreID=None, cleanup=False, basename=None):
...
+        :param str mode: an optional string that specifies the mode in which the file is opened.
+                It defaults to 'b' which means open for writing in binary mode.
+
+                Currently supported mode characters:
+
+                    - 'b': open for writing in binary mode
+
+                    - 't': open for writing in text mode
+
+                The default mode is 'b' (open for writing in binary mode, synonym of 'wb').
```
adamnovak, line 916: What does this mean if there's no `'w'`?

adamnovak (approved):
This looks good so far! I think the only missing piece now is the implementations for the other job stores.

diekhans:
Is it going to cause confusing that this different from the python
default, which open in text mode?

src/toil/jobStores/abstractJobStore.py
jonathanxu18: I was thinking that 'w' is implied for both modes in `writeFileStream()` so I wrote the default as 'wb'. Would you recommend leaving it as 'b'  instead to avoid any confusion?

jonathanxu18:
> Is it going to cause confusing that this different from the python default, which open in text mode?

I talked with @adamnovak and we would like to eventually set the default as text mode, but its a change that will break some user workflows so we aren't planning to do it now. We also discussed about possibly giving implicit text mode a DeprecationWarning to make the binary mode default more clear for now.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -903,6 +904,24 @@ def writeFileStream(self, jobStoreID=None, cleanup=False, basename=None):
...
                matching that basename, the file will be detected.
 
+        :param str mode: an optional string that specifies the mode in which the file is opened.
+                It defaults to 'b' which means open for writing in binary mode.
+
+                Currently supported mode characters:
+
+                    - 'b': open for writing in binary mode
+
+                    - 't': open for writing in text mode
```
DailyDreaming, line 914: Hmmm... I dislike going against conventional standards for well established use-cases and `t` isn't a standard.  I think that we should probably remove `mode='b'` and use `bytes` when `encoding=None`, otherwise, use the encoding.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -964,12 +983,29 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
 
     @abstractmethod
     @contextmanager
-    def readFileStream(self, jobStoreFileID):
+    def readFileStream(self, jobStoreFileID, mode='b', encoding=None, errors=None):
```
DailyDreaming, line 986: Same here as above.  Remove `mode` and only use `encoding`.  The default of `None` means no encoding, and just `bytes` will be used.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -565,19 +566,19 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
         info.download(localFilePath, not self.config.disableJobStoreChecksumVerification)
 
     @contextmanager
-    def readFileStream(self, jobStoreFileID):
+    def readFileStream(self, jobStoreFileID, mode='b', encoding=None, errors=None):
```
DailyDreaming, line 569: Same for all of these.

DailyDreaming (changes requested):
My comments basically amount to removing `mode` and the `t` text mode.  `encoding=None` should be sufficient to infer `bytes` when it's `None` (as a default) or something like `utf-8` otherwise.

jonathanxu18:
> My comments basically amount to removing `mode` and the `t` text mode. `encoding=None` should be sufficient to infer `bytes` when it's `None` (as a default) or something like `utf-8` otherwise.

I was thinking to include `mode` and the `t` text mode to try to replicate how the [open()](https://docs.python.org/3/library/functions.html#open) API handles modes as it also includes `t`. Maybe we could discuss this more during standup tomorrow if we have time.

src/toil/test/src/fileStoreTest.py
```diff
@@ -1314,25 +1330,25 @@ class CachingFileStoreTestWithFileJobStore(hidden.AbstractCachingFileStoreTest):
...
 
-@needs_google
+#@needs_google
 class NonCachingFileStoreTestWithGoogleJobStore(hidden.AbstractNonCachingFileStoreTest):
     jobStoreType = 'google'
 
 
 @slow
-@needs_google
+#@needs_google
```
DailyDreaming, line 1351: Leftovers?

DailyDreaming (commented):
This looks great, thanks for making the requested changes.  `@needs_aws_ec2` and `@needs_google` are commented out currently, but I think after those are uncommented and this passes tests, I'll approve and merge.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1150,9 +1150,9 @@ def _get_file_checksum(self, localFilePath, to_match=None):
                 return self._finish_checksum(hasher)
 
         @contextmanager
-        def uploadStream(self, multipart=True, allowInlining=True):
+        def uploadStream(self, multipart=True, allowInlining=True, encoding=None, errors=None):
             """
-            Context manager that gives out a binary-mode upload stream to upload data.
+            Context manager that gives out a binary/text mode upload stream to upload data.
```
adamnovak, line 1155: Maybe this should be `binary or text mode`?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1418,7 +1422,10 @@ def download(self, localFilePath, verifyChecksum=True):
                 assert False
 
         @contextmanager
-        def downloadStream(self, verifyChecksum=True):
+        def downloadStream(self, verifyChecksum=True, encoding=None, errors=None):
+            """
+            Context manager that gives out a binary/text mode download stream to download data.
```
adamnovak, line 1427: I guess you could just say `a download stream`, since binary and text are two opposite modes a stream can have.

src/toil/jobStores/utils.py
```diff
@@ -209,15 +218,24 @@ def _writer(self):
             if e.errno != errno.EPIPE:
                 raise
 
-    def __init__(self):
+    def __init__(self, encoding=None, errors=None):
+        """
+        :param str encoding: the name of the encoding used to encode the file. Encodings are the same
```
adamnovak, line 223: You should make it clear that these apply to whoever reads the pipe, and that the `writeTo` implementation still needs to work on binary streams.

adamnovak: Also I noticed the `writeTo` docstring says `Implement this method to read data from the pipe.`. That's wrong; you implement `wrtieTo` to write data.

src/toil/jobStores/utils.py
```diff
@@ -269,8 +287,8 @@ class ReadableTransformingPipe(ReadablePipe):
     
     """
     
-    def __init__(self, source):
-        super(ReadableTransformingPipe, self).__init__()
+    def __init__(self, source, encoding=None, errors=None):
```
adamnovak, line 290: Previously, the source stream passed here was always assumed to be in binary mode. If you set an encoding on the transforming pipe, should it also expect a text-mode input stream and decode it with that encoding?

If you don't do that, you're going to have trouble swapping a transformation in and out of a pipeline of other text-mode streams; you'd need to ensure all but the last thing in the pipeline are running in binary mode, which I don't think you do for e.g. AWS downloads.

adamnovak (changes requested):
This looks pretty good, but I'm not convinced the AWS hash checking logic still works when you swap files over to text mode. Do your tests really cover it? And if so, can you explain why it still works?

Also I think some of the docstrings need to change with the change to not have an explicit mode argument.

adamnovak (approved):
This looks good to me.

DailyDreaming (approved):
Looks good to me as well.

PR #3437, opened 2021-02-02
Only determine execute permissions on files we copy/move.

### LINKED ISSUE #3433: PermissionError for readGlobalFile()
We are having this issue as of toil 5.2.0:

```
  File "/home/shahr2/tools/system/miniconda/versions/4.6.14/envs/nucleo_test/lib/python3.6/site-packages/toil/cwl/cwltoil.py", line 615, in visit
    deref = self.get_file(path) if self.get_file else ab
  File "/home/shahr2/tools/system/miniconda/versions/4.6.14/envs/nucleo_test/lib/python3.6/site-packages/toil/cwl/cwltoil.py", line 756, in toil_get_file
    src_path = file_store.readGlobalFile(FileID.unpack(file_store_id[7:]))
  File "/home/shahr2/tools/system/miniconda/versions/4.6.14/envs/nucleo_test/lib/python3.6/site-packages/toil/fileStores/nonCachingFileStore.py", line 99, in readGlobalFile
    os.chmod(localFilePath, os.stat(localFilePath).st_mode | stat.S_IXUSR)
PermissionError: [Errno 1] Operation not permitted: '/home/shahr2/bergerlab/test/nucleo_test/test_29Jan2020/toil_1/node-5bbdc7f7-e26f-4958-aff8-2986c4b0b199-2919ee05-ad3b-4b1c-a90e-1d35c9ebd4ab/tmpt5zanbkb/4f6b5fd2-537c-4aed-ac07-87cbb60d665a/tmpwwfjv6k7.tmp'
```

I believe it's related to this change:

https://github.com/DataBiosphere/toil/pull/3322/files#diff-7141faf78b8bb77178d22d78d6af5f62ae090690f86a87f53ee2ca02f2b2fca3R98

Would it be possible to get some info regarding whether this is an intended error, and perhaps we're trying to import a file (using toil-cwl-runner) incorrectly? Or whether there is an issue with how execute permissions are set on imported files?

Let us know if I can provide any additional information, thank you kindly!

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-786)
┆Issue Number: TOIL-786

DailyDreaming:
Resolves #3433 .

@ionox0 Could you please test this to see if it fixes your problem?  I had trouble reproducing the original issue, so if you had a reproducible test workflow that would be appropriate here, I'd be grateful.

I moved setting the execute bit from the filestores to the jobstores. This allows the filestore to attempt, in order, to symlink, hardlink, or finally copy, based on it's ability to `chmod`, which prioritizes the file's executability over the defined linking/copying when using the method.

This also moves some of the I/O in `lib.misc` into a newly created `lib.io`.

## Changelog Entry

 * Prioritize execute permissions when linking, and copy when necessary.


DailyDreaming:
Tests are failing because CI/CD was updated today and needs to be fixed. Once #3440 is merged, these can be rerun.

diekhans:
I am a bit confused by this.   If Toil didn't create the file, it should not be modifying the metadata.  If it did create the file, then it should not get an error modifying the metadata.  So the try/except around setting executable seems wrong.

There is not need to modify metadata on a symlink, as the target file metadata is used.

adamnovak:
@diekhans I think the problem is if Toil symlinks a file that is owned by another user into the job store, and again from there to where a job wants to read it. We then try to mark it user-executable, even if it's already executable by us, because it was user-executable when we imported it, and we end up trying to chmod somebody else's file via a couple layers of symlink.

src/toil/jobStores/fileJobStore.py
```diff
@@ -454,21 +452,25 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
...
                 else:
                     raise
 
+            if executable:
+                try:
+                    os.chmod(localFilePath, os.stat(localFilePath).st_mode | stat.S_IXUSR)
+                    return
+                except PermissionError:
+                    # can't symlink and preserve exec permissions
+                    # prioritize permissions and attempt to hardlink next
```
adamnovak, line 471: If we do it this way, we'll end up copying every linked import we don't own that's executable by its owner.

src/toil/jobStores/googleJobStore.py
```diff
@@ -269,6 +271,8 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
             with open(tmpPath, 'wb') as writeable:
                 blob = self.bucket.get_blob(compat_bytes(jobStoreFileID), encryption_key=self.sseKey)
                 blob.download_to_file(writeable)
+        if getattr(jobStoreFileID, 'executable', False):
```
adamnovak, line 274: There should be an update to AbstractJobStore specifying that implementations of readFile need to respect/provide the executable attribute on FileIDs, since we're changing all the implementations.

diekhans:
It seems that the chmod should happen as part of the copy, not
sometime later, e.g. shutil.copy().

adamnovak (changes requested):
I don't think the fallbacks to hardlinking and copying are quite the right approach. As it it looks like we're going to try and fail to chmod every owner-executable file that is linked in as an import and owned by someone else.

We could only do the chmod if we actually own the ultimate target of any links. And if we don't own the target, and we want it to be executable, we should only make a copy if it isn't actually executable by us already.

It also might be worth checking whether files are executable, when we're uploading them, by getting the effective permission bit for us rather than just whatever the permissions are for the owner. But files that aren't executable to their owner but need to be executable to us are probably rare.

src/toil/jobStores/googleJobStore.py
DailyDreaming: Should be fixed.

src/toil/jobStores/fileJobStore.py
DailyDreaming: I've removed changing the permissions on links and this only changes permissions on files that we either copy or `shutil.move` now (since that doesn't seem to preserve permissions https://bugs.python.org/issue1355826).

DailyDreaming:
@adamnovak This should be ready for re-review if you have a minute.  This now sets the execution bit on everything we copy/move, and shouldn't affect anything that we link.

src/toil/jobStores/googleJobStore.py
```diff
@@ -269,6 +271,8 @@ def readFile(self, jobStoreFileID, localFilePath, symlink=False):
...
+            os.chmod(localFilePath, os.stat(localFilePath).st_mode | stat.S_IXUSR)
```
adamnovak, line 275: It would be nicer if we could tell AtomicFileCreate whether we wanted an executable file or not. What's the point of atomically creating it with the right contents if the metadata isn't also part of that atomic operation?

adamnovak (approved):
I think this is the right approach. I would like to see the executable bit set through `AtomicFileCreate` rather than added on after the atomic creation, though, because I feel like doing otherwise invites weird race conditions.

src/toil/jobStores/googleJobStore.py
DailyDreaming: That's a good point.  I'll look into changing this.

DailyDreaming: I think I'm actually going to merge this and then make an issue to address `AtomicFileCreate` separately.

DailyDreaming: #3467

Issue #3380, opened 2020-12-17
Toil generating too long paths


caballero:
I am running Toil 5.1.0 in an LSF cluster, one of the steps in my workflow (Flye assembly) is failing because it uses socket.bind with Python, the error is "AF_UNIX path too long"

This is the error on Flye step:

```
Traceback (most recent call last):
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/managers.py", line 608, in _run_server
    server = cls._Server(registry, address, authkey, serializer)
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/managers.py", line 154, in __init__
    self.listener = Listener(address=address, backlog=16)
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/connection.py", line 448, in __init__
    self._listener = SocketListener(address, family, backlog)
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/multiprocessing/connection.py", line 592, in __init__
    self._socket.bind(address)
OSError: AF_UNIX path too long
Traceback (most recent call last):
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/bin/flye", line 33, in <module>
    sys.exit(load_entry_point('flye==2.8.2', 'console_scripts', 'flye')())
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/site-packages/flye/main.py", line 797, in main
    _run(args)
  File "/hps/nobackup2/production/metagenomics/jcaballero/miniconda3/envs/mgnify-lr/lib/python3.8/site-packages/flye/main.py", line 576, in _run
    jobs[i].run()
[...]

PR #3438, opened 2021-02-03
Reduce path name lengths.

DailyDreaming:
Somewhat addresses #3380.

Long filepaths have been a problem and this PR attempts to somewhat address this.

This reduces the length of the `getLocalWorkflowDir()` path returned by removing `node_`.

Each filestore creates a working dir with `len(uuid4()) == 36`.  This attempts to make one with length 4, then length 5, up to length 32 (removing slashes).  This is is the base working dir for a job, so reducing its length should be beneficial.

Reduces the size of the log file name returned by `formatStdOutErrPath()`.

This also warns users if the workdir set at the beginning of the run is over 80 chars long.

## Changelog Entry

 * Reduce path name lengths.


DailyDreaming:
Tests are failing because CI/CD was updated today and needs to be fixed.  Once https://github.com/DataBiosphere/toil/pull/3440 is merged, these can be rerun.

mr-c (approved):
LGTM!

src/toil/fileStores/__init__.py
```diff
@@ -23,6 +25,33 @@ def make_public_dir(dirName: str) -> str:
...
+            truncated_uuid = str(uuid4()).replace('-', '')[:i]
+            generated_dir_path = os.path.join(prefix, truncated_uuid)
+            if not os.path.exists(generated_dir_path):
+                return generated_dir_path
+    this_should_never_happen = os.path.join(prefix, str(uuid4()))
+    assert not os.path.exists(this_should_never_happen)
+    return this_should_never_happen
+
+
+def make_unique_public_dir(prefix: str) -> str:
```
adamnovak, line 50: Are multiple processes/threads ever able to run this at the same time?

When we use UUIDs for making directories, we don't have to worry about races, because UUID collisions are effectively impossible due to their length. When our directory names might be as short as 5 characters, collisions are very possible. This would probably manifest as the mkdir in make_public_dir failing because the directory already exists.

[`tempfile.mkdtemp`](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) is the Right Way to make a unique subdirectory. If for some reason we can't do that, I would unify the name-generation and directory-creation into one loop, using a failed mkdir as the test for the directory already existing.

Were we using UUIDs to avoid collisions not just on this machine, but also on other machines?

src/toil/fileStores/__init__.py
```diff
@@ -14,6 +14,8 @@
 import os
 import stat
 
+from uuid import uuid4
+
 
 def make_public_dir(dirName: str) -> str:
```
adamnovak, line 20: Is there a good reason we're making world-writable temp directories for Toil jobs? It looks like this function is used in the file stores as well. Are we expecting some jobs to run under another user?

If the directories do need to be world-writable, we probably need to set the sticky bit like on a real /tmp; I think otherwise anyone can use your write permission on the directory to delete anyone else's files.

adamnovak (changes requested):
If we're not using UUIDs, `tmpfile.mkdtemp` is probably the right way to get a temp directory, and the names it makes are tolerably short.

Also, we should take this opportunity to figure out why our temp directories are all world-writable with no sticky bit.

adamnovak:
Also Lon, did you mean to PR this *into* `issues/3433-exec-bit-symlink-catch`?

src/toil/fileStores/__init__.py
adamnovak: Arjun decided these directories should be public when first implementing caching: https://github.com/DataBiosphere/toil/commit/3974681ac5b903c967ca31b59d6890e20761859d#diff-bdafbc89001da41469f9aa17902123fa92515e69dfe87beb2643e48dbde8f2f0R583

I didn't see any justification there.

src/toil/fileStores/__init__.py
DailyDreaming: You're right, I should have accounted for races.  I think the try/except on a `mkdir` makes sense to me.

`tempfile.mkdtemp` creates a ` directory [that] is readable, writable, and searchable only by the creating user ID.` and while I'm unsure that that's a case we have to account for naturally, I suspect that this might make it difficult to inspect directories by an outside user in the event of a failed workflow trying to go through the files so I'd rather avoid it.

src/toil/fileStores/__init__.py
DailyDreaming: You have some good points.  This is a discussion outside of the context of this PR, however, as this PR is just shortening names, not changing current behavior.

That said, just for discussion, an argument against changing the permissions of the directory that comes to mind to me would be inspection by an outside user on the dir/files of a failed workflow (mentioned above).  I also think that... while it's possible for an outside program to attempt to write to our toil created directory while it's running... in practice, I don't see it happening.

DailyDreaming:
@adamnovak I did mean to PR this into `issues/3433-exec-bit-symlink-catch`, since I accidentally created this branch off of that one and it makes the diff clean for review.  Either `issues/3433-exec-bit-symlink-catch` will be merged prior to this and then the effect on `master` will be equivalent, or I can rebase if this gets approved/merged first.

Sorry, I should have made a comment in the opening comment box.

src/toil/fileStores/__init__.py
mr-c: This PR may be relevant to the historical discussion: https://github.com/DataBiosphere/toil/pull/3229

DailyDreaming:
@adamnovak  This should be done and ready for re-review.  Also... I made comments above, but I'm wary of changing to `tmpfile.mkdtemp`, and feel it should be a separate PR if done.

src/toil/batchSystems/abstractBatchSystem.py
```diff
@@ -360,11 +360,7 @@ def formatStdOutErrPath(self, jobID, batchSystem, batchJobIDfmt, fileDesc):
         batch system job IDs, for ease of debugging job failures.
 
         :param: string jobID : Toil job ID
```
adamnovak, line 362: This parameter's name has changed.

adamnovak: And shouldn't `cluster_job_id` also have a docstring mention?

src/toil/common.py
```diff
@@ -181,6 +181,12 @@ def parse_int_list(s: str):
             self.workDir = os.path.abspath(self.workDir)
             if not os.path.exists(self.workDir):
                 raise RuntimeError(f"The path provided to --workDir ({self.workDir}) does not exist.")
+
+            if len(self.workDir) > 80:
+                logger.warning(f'Length of workDir path "{self.workDir}" is {len(self.workDir)} char long.  '
```
adamnovak, line 186: `Length of XXX is XXXX char long` is a construction that upsets me. Make it `Length of workDir path "{self.workDir}" is {len(self.workDir)} characters.` instead.

src/toil/fileStores/__init__.py
adamnovak, line 20: I don't think that, after this PR, `make_public_dir` is ever called anymore. We should remove it if it is no longer used.

adamnovak (changes requested):
This looks mostly good. There are a couple docstring changes I think are needed, and `make_public_dir` might not be being called anymore.

DailyDreaming:
@adamnovak Comments should all be addressed.

DailyDreaming:
@mr-c The new diff-cov is failing:

```
Found 7479 errors in 168 files (checked 192 source files)

diff-cover --fail-under=100 cobertura.xml

Failure. Coverage is below 100.0%.

-------------

Diff Coverage

Diff: origin/master...HEAD, staged and unstaged changes

-------------

src/toil/batchSystems/abstractBatchSystem.py (50.0%): Missing lines 372,374

src/toil/batchSystems/gridengine.py (0.0%): Missing lines 133-134

src/toil/batchSystems/htcondor.py (0.0%): Missing lines 75-77

src/toil/batchSystems/lsf.py (0.0%): Missing lines 253-254

src/toil/batchSystems/slurm.py (0.0%): Missing lines 180-181

src/toil/batchSystems/torque.py (0.0%): Missing lines 196-197

src/toil/common.py (0.0%): Missing lines 185-186,1093

src/toil/fileStores/__init__.py (100%)

src/toil/fileStores/cachingFileStore.py (50.0%): Missing lines 977

src/toil/fileStores/nonCachingFileStore.py (50.0%): Missing lines 44

-------------

Total:   37 lines

Missing: 18 lines

Coverage: 51%

-------------

Makefile:252: recipe for target 'diff_mypy' failed

make: *** [diff_mypy] Error 1
```

Should we set this to be less strict or simply print a warning rather than fail?  This looks like it's going to fail for most PRs for the foreseeable future.

mr-c:
> Should we set this to be less strict or simply print a warning rather than fail? 

Well, if you are willing to edit a line of code, then you should add/fix the types hints for that line. I don't think it is too much to ask.

DailyDreaming:
@mr-c Hmmm... I'm coming around to type hinting for declarations.  It's still a habit to only type inputs and returns for me.  I'll dig in, as this seems to get into typing any functions or objects called by the referenced lines as well.

DailyDreaming:
Hmmm... I've added a lot more typing and I think that the coverage may have actually gone down as it's adding more files and functions.

@mr-c I think I'd like to add this as is, as the diff is mostly typing additions now.  I may take a stab at a typing only PR at some point, but I'd like to keep the diff small for this PR, if possible.

@adamnovak Have I addressed your earlier review comments?

adamnovak (approved):
My comments have been addressed.

I'm a bit worried about squashing this down to one commit since it's now doing a bunch of type hinting unrelated to the ticket it's closing. But the type hinting also looks right to me, to the extent that I'm qualified to evaluate it.

mr-c:
> Hmmm... I've added a lot more typing and I think that the coverage may have actually gone down as it's adding more files and functions.

If you don't mind, I'll add a commit to finish satisfying diff_mypy

mr-c:
>  I may take a stab at a typing only PR at some point, but I'd like to keep the diff small for this PR, if possible.

I understand, but so many type hints are needed that it will be a bit overwhelming! I personally prefer this method where we take it one chunk at a time because we're always looking at that part of the code

DailyDreaming:
@mr-c An extra commit to satisfy `diff_mypy` would be much appreciated, thank you!  I'll wait on this then.

mr-c:
I merged with the primary branch, which will pibk up https://github.com/DataBiosphere/toil/pull/3495/commits/ba7da1ce9406922af3a49f858ab749f0388d7a84 ; lets see if that is enough!

adamnovak:
Looks like the Gitlab tests hit an unrelated AWS throttling error that we're already working on.

But mypy is still upset. Should we merge this so the type hints added here are usable in other PRs? Or should we keep going around on this until mypy accepts it?

adamnovak:
This passes everything except type checking, and adds a bunch of type info. I am going to merge it.

Issue #3441, opened 2021-02-03
cwl: disregard --defaultMemory for CWL_INTERNAL_JOBS


mr-c:
A user reported issues when they ran `toil-cwl-runner` on their cluster with `--defaultMemory 400G`, causing the locally run `CWLWorkflow` and `ResolveIndirect` jobs to fail for lack of memory

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-790)
┆Issue Number: TOIL-790

PR #3442, opened 2021-02-03
set resource reqs for all CWL_INTERNAL_JOBS

mr-c:
Fixes: #3441

## Changelog Entry

 * toil-cwl-runner: The `CWLWorkflow`, `CWLGather`, `CWLScatter`, and `ResolveIndirect` jobs are no longer affected by the `--defaultMemory`, `--defaultCores`, and `--defaultDisk` command line options. This is useful when running on a batch scheduling system where there are significantly more resources available on the processing nodes compared to where `toil-cwl-runner` is launched and one or more of the aforementioned `--default` options are used.


DailyDreaming:
Thank you @mr-c !

I'll pulled in master, which should allow tests to run.  Once those pass, this LGTM.

mr-c:
@DailyDreaming I turned on automerge, so if you mark this PR as approved then it will automatically squash+merge when the CI is done and all green.

DailyDreaming (approved):

src/toil/batchSystems/lsfHelper.py
```diff
@@ -228,7 +228,7 @@ def convert_mb(kb, unit):
              "TB": 2}
     assert unit in UNITS, ("%s not a valid unit, valid units are %s."
                            % (unit, list(UNITS.keys())))
-    return int(float(kb) // float(math.pow(1024, UNITS[unit])))
+    return float(kb) / float(math.pow(1024, UNITS[unit]))
```
mr-c, line 231:
```suggestion
    return kb / math.pow(1024, UNITS[unit])
```
This should also work; what do you think @DailyDreaming ?

Without this commit, LSF was rounding sub-1 Gigabyte resource requests to 0 (!!!)

DailyDreaming: @mr-c It looks like this ultimately ends up specifying `rusage`, which seems to accept floats: https://github.com/DataBiosphere/toil/blob/4b888115ec331c026df09fdaefe00eecd67ca24f/src/toil/batchSystems/lsf.py#L257

I'm worried that with the old method, 0.1 GB would specify: `0`, which, while odd, is probably still valid, but the new method returns: `9.765625e-05`, which is pretty close to zero, and seems wrong?

My test script:

```
import math

UNITS = {"B": -2,
         "KB": -1,
         "MB": 0,
         "GB": 1,
         "TB": 2}

def convert_mb(kb, unit):
    assert unit in UNITS, ("%s not a valid unit, valid units are %s."
                           % (unit, list(UNITS.keys())))
    return int(float(kb) // float(math.pow(1024, UNITS[unit])))

def convert_mb2(kb, unit):
    assert unit in UNITS, ("%s not a valid unit, valid units are %s."
                           % (unit, list(UNITS.keys())))
    return float(kb) / float(math.pow(1024, UNITS[unit]))

def convert_mb3(kb, unit):
    assert unit in UNITS, ("%s not a valid unit, valid units are %s."
                           % (unit, list(UNITS.keys())))
    return kb / math.pow(1024, UNITS[unit])

for i in (0, 0.1, 0.5, 0.9, 1, 7, 10, 100, 1000):
    for unit in UNITS.keys():
        print(convert_mb(float(i) * 1024, unit))
        print(convert_mb2(float(i) * 1024, unit))
        print(convert_mb3(float(i) * 1024, unit))
        print(f'={i}={unit}=')
```

This seems like something we could and should probably unit test.

DailyDreaming: I would probably leave it as it but maybe have a minimum default value when it's zero?  Maybe 100mb minimum?

mr-c: 0.1 as 0 G breaks LSF submission

>  exit status 255: MEMLIMIT value <0> is not valid. Limit must be a positive integer. You can use the following units for the limit: KB (or K) , MB (or M), GB (or G), TB (or T), PB (or P), EB (or E), ZB (or Z). Job not submitted.

DailyDreaming:
@mr-c I've changed around the function and added a test.  Hopefully this is safer.  Let me know what you think.

src/toil/batchSystems/lsfHelper.py
```diff
@@ -21,22 +21,23 @@
...
-
 import fnmatch
-import math
 import os
 import re
 import subprocess
 
 from packaging import version
 
+from toil.wdl.wdl_functions import return_bytes
```
mr-c, line 31: Yay for code reuse! Shouldn't `return_bytes` now be placed somewhere more generic than the WDL package?

DailyDreaming: Good point!  I've generalized it now.

DailyDreaming:
I think that this is good to merge.

PR #3446, opened 2021-02-05
Ensure that python3 and not python (2) is used

mr-c:
## Changelog Entry

 * src/toil/test/utils/toilDebugTest.py no longer assumes that there is a executable named `python`; fixes tests for systems without Python2 (like the upcoming 'Bullseye' release of Debian, which does not install Python2 by default)


DailyDreaming (approved):
LGTM.

PR #3448, opened 2021-02-08
Add WDL as_map() function.

### LINKED ISSUE #3201: as_map() implemented for WDL.

w-gao:
Closes #3201.

## Changelog Entry

 * Add as_map() WDL function.

### DIFF
--- a/src/toil/test/wdl/builtinTest.py
+++ b/src/toil/test/wdl/builtinTest.py
@@ -484,6 +484,9 @@ class WdlStandardLibraryWorkflowsTest(WdlWorkflowsTest):
                                             '{"left":2,"right":"a"},{"left":2,"right":"b"},'
                                             '{"left":3,"right":"a"},{"left":3,"right":"b"}]')
 
+    def test_as_map(self):
+        self.check_function('as_map', cases=['as_input'], expected_result='{"a":1,"b":2,"c":3}')
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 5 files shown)


DailyDreaming (approved):
LGTM.

PR #3458, opened 2021-02-16
Add mypy checking.

DailyDreaming:
This should make all new files and all currently passing files beholden to mypy, and creates an additional linting step prior to any code test runs, so failure should be fairly quick if mypy complains.

Note that this PR contains an explicit list of ignored files.  Removing the ignored files from the list is progress made towards having the entire project under mypy.  It's a bit of work so I expect it will be done in chunks.  It's also hard to account for files which lack typing entirely (and thus pass).

## Changelog Entry

 * Add mypy checking step to cicd.  Consolidate glob functions into one.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,9 +32,18 @@ after_script:
   - stopdocker || true
 
 stages:
+  - linting
   - basic_tests
   - main_tests
   - integration
+
+
+lint:
+  stage: linting
+  script:
+    - pwd
+    - virtualenv -p python3.6 venv && . venv/bin/activate && make prepare && make develop extras=[all] && pip install htcondor
+    - make mypy
 
 
 # Python3.6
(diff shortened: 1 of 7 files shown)


mr-c (changes requested):
Can we use mypy.ini configuration instead? That way anyone can just type `mypy` and it does the right thing.

Example: https://github.com/common-workflow-language/cwltool/blob/main/mypy.ini

DailyDreaming:
@mr-c It seems that using the config to ignore files requires an `__init__.py` file to be created at `toil/__init__.py` and `toil/src/__init__.py` and for every folder in a paths we explicitly define outside of the main `[mypy]` section.

I'm not sure that this will cause issues(?)... but the script seems like a lower impact change.  I can try this out later on.

mr-c (approved):

PR #3488, opened 2021-03-10
Migrate cloud-config to ignition

jonathanxu18:
The cloud-config should be migrated over to Ignition as per documentation: https://kinvolk.io/docs/flatcar-container-linux/latest/provisioning/ignition/

The documentation recommends starting with a Container Linux Config and converting it over to Ignition using a config transpiler: 
https://github.com/kinvolk/container-linux-config-transpiler/
https://kinvolk.io/docs/flatcar-container-linux/latest/provisioning/cl-config/

However, the benefits of doing so don't apply as much to our situation so I want to attempt directly using Ignition.

### DIFF
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -17,7 +17,7 @@ import os.path
 import subprocess
 import tempfile
 import textwrap
-import yaml
+import json
 from abc import ABC, abstractmethod
 from functools import total_ordering
 from typing import Dict, Optional, Set
@@ -471,28 +471,30 @@ class AbstractProvisioner(ABC):
             self.sshPublicKeys.append("ssh-rsa " + keyData)
 
 
-        def toCloudConfig(self) -> str:
+        def toIgnitionConfig(self) -> str:
             """
-            Return a CloudConfig configuration describing the desired config.
+            Return a Ignition configuration describing the desired config.
             """
 
             # Define the base config
             config = {
-                'write_files': self.files,
-                'coreos': {
-                    'update': {
-                        'reboot-strategy': 'off'
+                'ignition': {
+                    'version': '3.0.0'
+                },
+                'storage': {
+                    'files': self.files
                     },
+                'systemd': {
                     'units': self.units
                 }
             }
 
             if len(self.sshPublicKeys) > 0:
                 # Add SSH keys if needed
-                config['ssh_authorized_keys'] = self.sshPublicKeys
+                config['passwd']['users']['sshAuthorizedKeys'] = self.sshPublicKeys
 
-            # Mark as CloudConfig and serialize as YAML
-            return '#cloud-config\n\n' + yaml.dump(config)
+            # Serialize as JSON
+            return json.dump(config)
 
 
     def getBaseInstanceConfiguration(self) -> InstanceConfiguration:
@@ -1005,7 +1007,7 @@ class AbstractProvisioner(ABC):
             ExecStart=/usr/bin/bash /home/core/join-kubernetes-cluster.sh
             '''))
 
-    def _getCloudConfigUserData(self, role, keyPath=None, preemptable=False):
+    def _getIgnitionConfigUserData(self, role, keyPath=None, preemptable=False):
         """
         Return the text (not bytes) user data to pass to a provisioned node.
 
@@ -1047,7 +1049,7 @@ class AbstractProvisioner(ABC):
                 # anyone in the cloud account.
                 self.addKubernetesWorker(config, self._leaderWorkerAuthentication, preemptable=preemptable)
 
-        # Make it into a string for CloudConfig
-        return config.toCloudConfig()
+        # Make it into a string for Ignition
+        return config.toIgnitionConfig()
 
 
--- a/src/toil/provisioners/aws/awsProvisioner.py
+++ b/src/toil/provisioners/aws/awsProvisioner.py
@@ -458,7 +458,7 @@ class AWSProvisioner(AbstractProvisioner):
         bdm = self._getBoto2BlockDeviceMapping(instanceType, rootVolSize=self._nodeStorageOverrides.get(nodeType, self._nodeStorage))
 
         keyPath = self._sseKey if self._sseKey else None
-        userData = self._getCloudConfigUserData('worker', keyPath, preemptable)
+        userData = self._getIgnitionConfigUserData('worker', keyPath, preemptable)
         if isinstance(userData, str):
             # Spot-market provisioning requires bytes for user data.
             userData = userData.encode('utf-8')
@@ -929,7 +929,7 @@ class AWSProvisioner(AbstractProvisioner):
         bdms = self._getBoto3BlockDeviceMappings(instanceType, rootVolSize=rootVolSize)
 
         keyPath = self._sseKey if self._sseKey else None
-        userData = self._getCloudConfigUserData('worker', keyPath, preemptable)
+        userData = self._getIgnitionConfigUserData('worker', keyPath, preemptable)
 
         # The name has the cluster name in it
         lt_name = f'{self.clusterName}-lt-{nodeType}'
--- a/src/toil/provisioners/gceProvisioner.py
+++ b/src/toil/provisioners/gceProvisioner.py
@@ -139,7 +139,7 @@ class GCEProvisioner(AbstractProvisioner):
             tags.update(kwargs['userTags'])
         self._tags = json.dumps(tags)
 
-        metadata = {'items': [{'key': 'user-data', 'value': self._getCloudConfigUserData('leader')}]}
+        metadata = {'items': [{'key': 'user-data', 'value': self._getIgnitionConfigUserData('leader')}]}
         imageType = 'flatcar-stable'
         sa_scopes = [{'scopes': ['compute', 'storage-full']}]
         disk = {}
@@ -252,7 +252,7 @@ class GCEProvisioner(AbstractProvisioner):
             logger.debug('Launching %s preemptable nodes', numNodes)
 
         #kwargs["subnet_id"] = self.subnetID if self.subnetID else self._getClusterInstance(self.instanceMetaData).subnet_id
-        userData =  self._getCloudConfigUserData('worker', keyPath, preemptable)
+        userData =  self._getIgnitionConfigUserData('worker', keyPath, preemptable)
         metadata = {'items': [{'key': 'user-data', 'value': userData}]}
         imageType = 'flatcar-stable'
         sa_scopes = [{'scopes': ['compute', 'storage-full']}]


src/toil/provisioners/abstractProvisioner.py
```diff
@@ -471,28 +471,30 @@ def addSSHRSAKey(self, keyData: str):
...
-            Return a CloudConfig configuration describing the desired config.
+            Return a Ignition configuration describing the desired config.
             """
 
             # Define the base config
             config = {
-                'write_files': self.files,
-                'coreos': {
-                    'update': {
-                        'reboot-strategy': 'off'
```
adamnovak, line 484: Is there no support for configuring Flatcar's update reboot strategy via Ignition?

adamnovak (approved):
If the units and files really are already in the same format, this looks good to me.

I would like an answer to my question about configuring the reboot strategy though. If we can still keep the nodes from rebooting out from under workflows, we should.

src/toil/provisioners/abstractProvisioner.py
adamnovak: Container Linux Config [still supports this](https://github.com/kinvolk/container-linux-config-transpiler/blob/flatcar-master/doc/configuration.md) as `locksmith.reboot_strategy` being set to `off`, so Ignition must support it. I think https://github.com/kinvolk/container-linux-config-transpiler is just going to turn that into `"locksmith": {"reboot_strategy": "off"}` in the JSON, so you could try adding that? If we really want to be sure that's how Ignition does it, we could build a little CLC file and run the transpiler on it and look at the output.

adamnovak (changes requested):
After further review, I think there is a way to still tell the machines to not reboot (see above) and so we should do it.

adamnovak:
The provisioner integration tests failed with this:

```
  File "/builds/databiosphere/toil/venv/bin/toil", line 33, in <module>
    sys.exit(load_entry_point('toil', 'console_scripts', 'toil')())
  File "/builds/databiosphere/toil/src/toil/utils/toilMain.py", line 30, in main
    module.main()
  File "/builds/databiosphere/toil/src/toil/utils/toilLaunchCluster.py", line 170, in main
    awsEc2ExtraSecurityGroupIds=options.awsEc2ExtraSecurityGroupIds)
  File "/builds/databiosphere/toil/src/toil/provisioners/aws/awsProvisioner.py", line 240, in launchCluster
    userData = self._getIgnitionConfigUserData('leader')
  File "/builds/databiosphere/toil/src/toil/provisioners/abstractProvisioner.py", line 1053, in _getIgnitionConfigUserData
    return config.toIgnitionConfig()
  File "/builds/databiosphere/toil/src/toil/provisioners/abstractProvisioner.py", line 497, in toIgnitionConfig
    return json.dump(config)
TypeError: dump() missing 1 required positional argument: 'fp'
```

I suspect you need `dumps()` and not `dump()`, @jonathanxu18. Also, when I was working on the provisioner I found things went a lot faster when I manually did something like `TOIL_APPLIANCE_SELF=5.3.0a1-8179a0359de7682b8ca528bacfbd1e770b0be486-py3.7 toil launch-cluster --provisioner=aws -z us-west-2a --clusterType mesos --leaderNodeType=t2.medium --keyPairName=yourkey@name.on.aws jon-test-cluster` to start a test cluster, and then `toil destroy-cluster --provisioner=aws -z us-west-2a jon-test-cluster` to destroy it, before sending my PR in for CI. The CI doesn't actually test the provisioner stuff until the very end of the run, which means it will take like 4 hours to report even a trivial issue like this one, while manual testing can be a lot faster.

src/toil/provisioners/abstractProvisioner.py
jonathanxu18: I couldn't find reboot strategy when I was looking at [https://coreos.github.io/ignition/configuration-v3_0/](url) so I thought there may not be support for it in Ignition. But as you mentioned, I think Ignition must support it since CLC still supports it so I'll try adding it in the JSON.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -442,14 +443,16 @@ def __init__(self):
             # Holds strings like "ssh-rsa actualKeyData" for keys to authorize (independently of cloud provider's system)
             self.sshPublicKeys = []
 
-        def addFile(self, path: str, owner: str = 'root', permissions: str = '0755', content: str = ''):
+        def addFile(self, path: str, filesystem: str = 'root', mode: int = 755, contents: str = ''):
```
w-gao, line 446: `'0755'` is in octal right? It looks like Ignition expects decimal values (see `mode` in https://coreos.github.io/ignition/configuration-v3_2).

I'll suggest a change.

jonathanxu18: Yep, I think it should be changed to a decimal value instead.

w-gao (commented):
@adamnovak @jonathanxu18 This should be ready if the tests pass. We might need to find a way to reduce our config size though (see below), because we're getting close to the AWS User Data limit.

src/toil/provisioners/abstractProvisioner.py
w-gao: There's no explicit option to set `reboot_strategy`. We'd just write the settings to `/etc/coreos/update.conf`. 

```yaml
locksmith:
  reboot_strategy: "off"
```

turns into 

```json
{
  ...
  "storage": {
    "files": [
      {
        "filesystem": "root",
        "path": "/etc/coreos/update.conf",
        "contents": {
          "source": "data:,%0AREBOOT_STRATEGY%3D%22off%22",
          "verification": {}
        },
        "mode": 420
      }
    ]
  }
}
```

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -497,30 +501,37 @@ def addSSHRSAKey(self, keyData: str):
...
+                    'users': [
+                        {
+                            'name': 'core',
+                            'sshAuthorizedKeys': self.sshPublicKeys
+                        }
+                    ]
+                }
 
+            # Serialize as JSON
+            return json.dumps(config, separators=(',', ':'))
```
w-gao, line 534: We're getting pretty close to AWS's [16 KB limit](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html) - We're writing the most amount of config when `clusterType == 'kubernetes' and role == 'leader'`, and we had ~14276 bytes previously with YAML. We are ~16244 bytes now (I think it depends on the cluster name too).

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -469,26 +472,27 @@ def __init__(self):
...
 
-        def addFile(self, path: str, owner: str = 'root', permissions: str = '0755', content: str = ''):
+        def addFile(self, path: str, filesystem: str = 'root', mode: int = int("0755", 8), contents: str = ''):
             """
-            Make a file on the instance with the given owner, permissions, and content.
+            Make a file on the instance with the given filesystem, mode (in decimal), and contents.
             """
 
-            self.files.append({'path': path, 'owner': owner, 'permissions': permissions, 'content': content})
+            contents = 'data:text/plain;charset=utf-8;base64,' + b64encode(contents.encode('utf-8')).decode('utf-8')
```
w-gao, line 480: It appear to me that base64 encoding the files actually takes less space. I think it's because the non-base64 encoded data URIs use three bytes to store special characters (e.g.: space is represented as `%20`), and we have a lot of spaces, line breaks, etc. Both are larger than the YAML equivalent though.

We should probably consider switching to the `https` or `s3` schemes and let Ignition fetch them during runtime, instead of writing them here.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -497,30 +501,37 @@ def addSSHRSAKey(self, keyData: str):
...
 
             # Define the base config
             config = {
-                'write_files': self.files,
-                'coreos': {
-                    'update': {
-                        'reboot-strategy': 'off'
-                    },
+                'ignition': {
+                    'version': '2.2.0'
```
adamnovak, line 512: If we want to use compression we might need to bump this version. Also where can we find the version 2.2 docs? Maybe the code should have a link to them?

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -612,14 +629,17 @@ def addVolumesService(self, config: InstanceConfiguration):
             Type=oneshot
             Restart=no
             ExecStart=/usr/bin/bash /home/core/volumes.sh
+
+            [Install]
+            WantedBy=multi-user.target
```
adamnovak, line 634: We need this because Ignition doesn't actually support the old `command` part of a unit, but just enables the unit, and we need it to start, right?

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -1076,9 +1112,12 @@ def addKubernetesWorker(self, config: InstanceConfiguration, authVars: Dict[str,
             Type=oneshot
             Restart=no
             ExecStart=/usr/bin/bash /home/core/join-kubernetes-cluster.sh
+
+            [Install]
+            WantedBy=multi-user.target
             '''))
 
-    def _getCloudConfigUserData(self, role, keyPath=None, preemptable=False):
+    def _getIgnitionConfigUserData(self, role, keyPath=None, preemptable=False):
```
adamnovak, line 1120: Why is there a "Config" in the new name?

adamnovak (changes requested):
This looks like it should work. However, I think the new interface for setting file modes is a bit awkward at the call sites.

Also, I'm not sure how closely we're bumping up against the 16kb user data limit with cluster names of reasonable length, and whether that's a problem we need to solve before we merge this. Is it?

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -469,26 +472,27 @@ def __init__(self):
             # Holds strings like "ssh-rsa actualKeyData" for keys to authorize (independently of cloud provider's system)
             self.sshPublicKeys = []
 
-        def addFile(self, path: str, owner: str = 'root', permissions: str = '0755', content: str = ''):
+        def addFile(self, path: str, filesystem: str = 'root', mode: int = int("0755", 8), contents: str = ''):
             """
-            Make a file on the instance with the given owner, permissions, and content.
+            Make a file on the instance with the given filesystem, mode (in decimal), and contents.
```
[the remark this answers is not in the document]
w-gao: The docs for the `filesystem` field is here: https://github.com/coreos/ignition/blob/spec2x/doc/configuration-v2_2.md. I don't think users can add their own filesystems so we probably don't need this.

And yeah, it would be better to do the conversion inside the function.

src/toil/provisioners/abstractProvisioner.py
w-gao: Correct

src/toil/provisioners/abstractProvisioner.py
w-gao: For posterity, we'll make Ignition fetch from cloud storage if the config is larger than the limit, because Flatcar only supports 2.x config versions, and this is their recommended setup.

adamnovak (approved):
This looks like it should work!

PR #3525, opened 2021-03-31
Add support for MPI with CWL.

### LINKED ISSUE #3517: Support for MPI?
Hello,

We are currently using Toil to run CWL workflows on a Slurm cluster. This is working well in general, but we have some jobs that can only use MPI to distribute the work over multiple nodes. I believe that MPI support was added recently to CWL (see https://github.com/common-workflow-language/cwltool/pull/1276), but I couldn't find any support for it in Toil currently -- is that correct? If so, are there any plans to add support, specifically for use with CWL+Slurm?

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-838)
┆Issue Number: TOIL-838

DailyDreaming:
Resolves #3517.

Based on: https://github.com/common-workflow-language/cwltool/pull/1276 I'm hoping that this is all we need to support MPI.

## Changelog Entry

 * Add cwltool option: --mpi-config-file.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


mr-c:
It is highly likely that this will not be enough as MPI often needs to be coordinated with the scheduler

@rupertnash, what do you think?

rupertnash:
Great to see Toil picking this up! I am able to make a bit of time after Easter to support getting MPI support into Toil, in case that is useful.

Afraid I don't know anything about Toil (except that it's a workflow system), so can't be specific, but few generalities that might be helpful.

In the case that we know we have one batch system job per CWL `CommandLineTool`, this might be sufficient, because each MPI tool invocation is pretty isolated.

If we have multiple MPI tool invocations occurring with a single batch system job, then it's more complicated. Toil needs to account for the resources taken up by the MPI program and ensure that it is launched with the correct options such that it uses the resources allocated by Toil.

adamnovak (changes requested):
I think we need a test for this, at least on the single machine batch system, to show that this actually does whatever is necessary to get cwltool to run the job with MPI available.

mr-c:
> I think we need a test for this, at least on the single machine batch system, to show that this actually does whatever is necessary to get cwltool to run the job with MPI available.

There are tests for this in `cwltool` that can be copy pasted here

DailyDreaming:
@mr-c @adamnovak @rupertnash I added the simplest MPI test and it should be working now after some tweaks (I had to add `--enable-ext` as well).

I also had to shorten some of Toil's directory paths, because this kept failing on path names that were absurdly long.  We now use `mkdtemp()` for our per worker work dirs, which Adam had suggested in an earlier PR.  I think the new directory paths are safe, but I'm not 100% sure, so I'd really appreciate a review (especially @adamnovak ).  A big advantage to this, if everything checks out, is that paths should be significantly shorter after this, especially in CWL workflows.

@mr-c @rupertnash Please let me know if I should add more testing than this.

src/toil/cwl/cwltoil.py
```diff
@@ -1926,6 +1926,12 @@ def main(args: Union[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
         action="store_true",
         help="Enable loading and running development versions of CWL",
     )
+    parser.add_argument(
+        "--enable-ext",
+        action="store_true",
+        help="Enable loading and running 'cwltool:' extensions to the CWL standards.",
+        default=False,
+    )
```
mr-c, line 1934: Oh, good! This will allow for InplaceUpdateRequirement before CWL v1.2.

It also allows for the rest of https://github.com/common-workflow-language/cwltool/blob/main/cwltool/extensions.yml & https://github.com/common-workflow-language/cwltool/blob/main/cwltool/extensions-v1.1.yml ; which by a quick glance should not be a problem.

However, maybe the tests for those should also be copied over into Toil?

DailyDreaming: @mr-c That sounds like a good idea to me!  I think I prefer it in a separate PR.  I'll make an issue for it and follow up with it once MPI support is in.

adamnovak (approved):
I think this looks good.

We won't support any kind of co-scheduling MPI tasks at the same time on multiple nodes or anything, but cwltool is single node and can't do that either, so I think we're fine.

PR #3526, opened 2021-03-31
Sniff raid better

### LINKED ISSUE #3500: Toil tries to include /dev/xvda in the RAID on r5d.24xlarge when it shouldn't, leading to cluster nodes coming up without their ephemeral drives
When Toil makes an AWS `r5d.24xlarge` instance, it seems to see the root volume as both /dev/xda and /dev/nvme0 something. When it greps for `/dev/xvda` in the output of `mount`, it doesn't find it, and so it tries to include it in the RAID.

This fails, but the setup service doesn't notice that creating /dev/md0 fails, and so it goes on to successfully boot the node and present the cluster with only the root volume for storage, and not all the ephemeral disks. This in turn will deadlock a Kubernetes autoscaling workflow, as the autoscaler thinks the nodes available are adequate when they aren't.

We may need to check for symlinks when polling all the drive names, or we may need to come up with a better way then grepping mount to see if a drive or a partition on it is mounted.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-827)
┆Issue Number: TOIL-827

### LINKED ISSUE #3528: Kubernetes Toil clusters can't use ephemeral instance storage
Because Toil isn't moving `/var/lib/kubelet` over to the ephemeral SSD RAID along with `/var/lib/docker`, Kubernetes nodes don't actually see or use that storage, or expose it as allocateable by Kubernetes.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-844)
┆Issue Number: TOIL-844

adamnovak:
This fixes #3500 by allowing instance storage to be mounted again. Flatcar recently(?) started including some symlinks in /dev that confused the old code and made it try to put the root device in the RAID via a symlink.

I also log the AMI used when launching clusters, since we've been having some issues with it not being found properly in some tests sometimes.

This also fixes #3528.

## Changelog Entry

 * Ephemeral instance storage can again be mounted on AWS
 * AMI used to start AWS clusters is now reported
 * Toil Kubernetes clusters can now use ephemeral instance SSDs on AWS

## Reviewer Checklist

        ```
        ```

## Merger Checklist


DailyDreaming (approved):
Assuming tests pass, this looks alright to me.

PR #3529, opened 2021-04-02
Consolidate memory functions

### LINKED ISSUE #3046: Consolidate memory functions.
We have byte translation happening in a few places:

https://github.com/DataBiosphere/toil/blob/master/src/toil/wdl/wdl_functions.py#L350

https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/__init__.py#L46

https://github.com/DataBiosphere/toil/blob/master/src/toil/lib/humanize.py#L23

These should probably all be refactored into one function.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-544)
┆Issue Number: TOIL-544

w-gao:
Consolidate memory functions. Closes #3046.  

## Changelog Entry

* Replace `human2bytes()` and `bytes2human()` functions in `src/toil/lib/humanize.py`.  Please use the respective functions in `toil.lib.conversions` instead.
* Remove `MemoryString` and `self.obtainSystemConstants()` in `toil.batchSystems.abstractGridEngineBatchSystem`. See https://github.com/DataBiosphere/toil/pull/3529#discussion_r611871006 

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/lib/conversions.py
```diff
@@ -1,8 +1,8 @@
 """Conversion utilities for mapping memory, disk, core declarations from strings to numbers and vice versa."""
 # TODO: Consolidate all conversion utilities here to use the same functions:
 #  src/toil/lib/humanize.py (bytes2human; human2bytes)
```
w-gao, line 3: @adamnovak @DailyDreaming I wasn't sure how to consolidate [`src/toil/lib/humanize.py `](https://github.com/DataBiosphere/toil/blob/master/src/toil/lib/humanize.py) here since the entire file was taken elsewhere under a different license.

DailyDreaming: @w-gao `src/toil/lib/humanize.py` can be either deleted/replaced, or become the one true source of conversions.  The same goes for `bytes_in_unit()` and `convert_units()`.  The idea is to simplify the code to make it dirt simple.  We have three ways to turn `10MB` into `10000000` bytes, and we only need one.  The issue is one of design, and it involves choosing one of the three methods we currently use to convert units and bytes: `MemoryString`, `humanize.py`, or the `bytes_in_unit()`/`convert_units()` functions, and deleting the other two, then replacing everything with one way of doing things.

DailyDreaming: Or coming up with a fourth way you think is better and deleting the all of the current three methods.  It's your choice.

DailyDreaming: While I'm not recommending using [src/toil/lib/humanize.py](https://github.com/DataBiosphere/toil/blob/master/src/toil/lib/humanize.py) (variable names like `iec_ext` are not immediately clear) it's under an MIT license, so feel free to modify it (though if you do and we only use this method, be sure to add your name and that you've modified it under the comment with `Giampaolo Rodola`'s name).

w-gao: @DailyDreaming I've rewritten `toil.lib.humanize`'s `human2bytes()` and `bytes2human()` functions to use `bytes_in_unit() / convert_units()`, and moved them to `toil.lib.conversions`. I feel like these functions are still useful to keep, since `bytes_in_unit() / convert_units()` are too basic to be useful.

As for `MemoryString`, I made it to use the `human2bytes()` function instead of using its own logic, but I didn't feel comfortable removing it entirely because I couldn't find where it is used in Toil other than [here](https://github.com/DataBiosphere/toil/blob/master/src/toil/batchSystems/abstractGridEngineBatchSystem.py#L301), but I wasn't sure where `self.maxMEM` is used...

src/toil/lib/conversions.py
```diff
@@ -1,8 +1,9 @@
 """Conversion utilities for mapping memory, disk, core declarations from strings to numbers and vice versa."""
-# TODO: Consolidate all conversion utilities here to use the same functions:
-#  src/toil/lib/humanize.py (bytes2human; human2bytes)
-#  src/toil/batchSystems/__init__.py (MemoryString)
-#
+from functools import total_ordering
+from typing import Optional, SupportsInt
+
+VALID_UNITS = ['b', 'k', 'm', 'g', 't', 'p', 'e', 'kb', 'mb', 'gb', 'tb', 'pb', 'eb',
```
DailyDreaming, line 5: It might be worth making two lists so to reduce redundancy and use one in `bytes2human()`.  Something like:

```
BINARY_PREFIXES = []
BASETEN_PREFIXES = []
VALID_PREFIXES = BINARY_PREFIXES + BASETEN_PREFIXES
```

src/toil/lib/conversions.py
```diff
@@ -24,14 +29,85 @@ def bytes_in_unit(unit: str = 'B') -> int:
...
     """Returns a float representing the converted input in dst_units."""
-    units = ['B', 'KB', 'MB', 'GB', 'TB']
-    assert src_unit in units, f"{src_unit} not a valid unit, valid units are {units}."
-    assert dst_unit in units, f"{dst_unit} not a valid unit, valid units are {units}."
+    assert src_unit.lower() in VALID_UNITS, f"{src_unit} not a valid unit, valid units are {VALID_UNITS}."
+    assert dst_unit.lower() in VALID_UNITS, f"{dst_unit} not a valid unit, valid units are {VALID_UNITS}."
     return (num * bytes_in_unit(src_unit)) / bytes_in_unit(dst_unit)
+
+
+def human2bytes(string: str) -> int:
```
DailyDreaming, line 48: I would split this into two functions.  One to parse and one to convert.  Something like:

```
def parse_memory_string(string: str) -> Tuple[float, str]:
    """Given a string representation of some memory (i.e. '1024 Mib'), return the number and unit."""
    for i, character in enumerate(string):
        # find the first character of the unit
        if character not in '0123456789.-_ ':
            units = string[i:].strip()
            assert units.lower() in VALID_UNITS, f"{units} not a valid unit, valid units are {VALID_UNITS}."
            return float(string[:i]), units
    return float(string), 'b'

def human2bytes(string: str) -> int:
    """Given a string representation of some memory (i.e. '1024 Mib'), return the integer number of bytes."""
    value, unit = parse_memory_string(string)
    return int(convert_units(value, src_unit=unit, dst_unit='b'))
```

src/toil/lib/conversions.py
DailyDreaming: @w-gao  Sadly, I think that's really the only place it's used and so we should probably remove it.

src/toil/batchSystems/torque.py
```diff
@@ -196,25 +195,12 @@ def generateTorqueWrapper(self, command, jobID):
...
-        # See: https://github.com/BD2KGenomics/toil/pull/1617#issuecomment-293525747
-        logger.debug("PBS/Torque does not need obtainSystemConstants to assess global cluster resources.")
-
+            _, tmp_file = tempfile.mkstemp(suffix='.sh', prefix='torque_wrapper')
+            with open(tmp_file, 'w') as f:
+                f.write("#!/bin/sh\n")
+                f.write("#PBS -o {}\n".format(stdoutfile))
+                f.write("#PBS -e {}\n".format(stderrfile))
+                f.write("cd $PBS_O_WORKDIR\n\n")
+                f.write(command + "\n")
```
DailyDreaming, line 204: Is this missing `return tmpFile`?

src/toil/test/lib/test_conversions.py
```diff
@@ -87,68 +87,56 @@ def test_convert(self):
...
+            '11234234 Ki': 11503855616,
+            '11234234 Mi': 11779948150784,
+            '11234234 Gi': 12062666906402816,
+            '11234234 Ti': 12352170912156483584
         }
 
         results = {}
         for i in (0, 0.1, 0.5, 0.9, 1, 7, 7.42423, 10, 100, 1000, 11234234):
-            for src_unit in ['K', 'M', 'G', 'T']:
+            for src_unit in ['Ki', 'Mi', 'Gi', 'Ti']:
```
DailyDreaming, line 140: Can we add `b` and the decimal counterparts?  So: `['b', 'Ki', 'Mi', 'Gi', 'Ti', 'K', 'M', 'G', 'T']`

DailyDreaming (commented):
This looks great so far.  Thank you for incorporating the comments!

Just two new comments, otherwise this looks good to me.

src/toil/batchSystems/torque.py
w-gao: Nope, it's on [line 206](https://github.com/DataBiosphere/toil/pull/3529/files#diff-33bd8eda6e64d400c55cf8043aa23ce884ce8147caf20df1e4a674aa975a9135R206).

w-gao:
@DailyDreaming I think the tests are failing because the old `human2bytes()` function treats everything as the binary version (e.g.: `human2bytes('10M')` and `human2bytes('10Mi')` both give `10485760`, and we're now making `human2bytes('10M')` return `10000000`). 

I'm wondering if we should keep the old behavior (unless there are places where we need the decimal prefix?), because I would imagine it breaking other workflows if people are not aware of this change.

DailyDreaming:
@w-gao Hmmm... that's unfortunate.  Let me think about it.  We should be accurate, but we could also just choose to never use the decimal system.  I'm leaning more towards breaking with the old, inaccurate behavior and just fixing the function, but maybe that's too harsh.

@adamnovak Any strong opinions on this?

adamnovak:
We talked about this on the call today, but my opinion is to change over to support both SI and binary prefixes under their official abbreviations, and also bump the Toil major version.

DailyDreaming (approved):
This looks great.  Thank you @w-gao .

PR #3595, opened 2021-05-05
Handle streaming reads from the cache when the data isn't written back yet

adamnovak:
We were seeing some failures in our file job store tests because the caching file store was actually inconsistent: you wouldn't have read your own writes consistency when writing a file and then trying to stream it back, because the file would be uploaded asynchronously but the streaming read would always come from the backing job store.

This fixes the problem by making sure to do streaming reads from the cache when the file being read is there and in an uploading/uploadable state.

I ran the offending `src/toil/test/src/jobFileStoreTest.py::JobFileStoreTest::testJobFileStore` test in a loop, and this change stopped me being able to reproduce the error:

```
FAILED=0
for TRY_NUM in {1..5} ; do
    for ITER in {1..32} ; do
        make test_offline cov="" tests=src/toil/test/src/jobFileStoreTest.py::JobFileStoreTest::testJobFileStore >iter${ITER}.log 2>&1 &
    done
    
    for ITER in {1..32} ; do
        wait
    done
    
    for ITER in {1..32} ; do
        grep "failed job" "iter${ITER}.log"
        if [[ "${?}" == "0" ]] ; then
            echo "Failure observed in iter${ITER}.log!"
            FAILED=1
            break
        fi
    done
    
    if [[ "${FAILED}" == "1" ]] ; then
        break
    fi
    echo "No failures observed in try ${TRY_NUM}"
done
```

## Changelog Entry

 * Streaming reads with the caching file store can now read your own writes consistently

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/fileStores/cachingFileStore.py
+++ b/src/toil/fileStores/cachingFileStore.py
@@ -1140,60 +1140,59 @@ class CachingFileStore(AbstractFileStore):
         # read a file that is 'uploadable' or 'uploading' and hasn't hit
         # the backing job store yet.
 
-        # Try and make a 'copying' reference to such a file
-        self._write([('INSERT INTO refs SELECT ?, id, ?, ? FROM files WHERE id = ? AND (state = ? OR state = ?)',
-            (localFilePath, readerID, 'copying', fileStoreID, 'uploadable', 'uploading'))])
+        with self._with_copying_reference_to_upload(fileStoreID, readerID, localFilePath) as ref_path:
+            if ref_path is not None:
+                # We got a copying reference, so the file is being uploaded and
+                # must be read from the cache for consistency. And it will
+                # stick around until the end of the block.
 
-        # See if we got it
-        have_reference = False
-        for row in self.cur.execute('SELECT COUNT(*) FROM refs WHERE path = ? and file_id = ?', (localFilePath, fileStoreID)):
-            have_reference = row[0] > 0
+                assert ref_path == localFilePath
 
-        if have_reference:
-            # If we succeed, copy the file. We know the job has space for it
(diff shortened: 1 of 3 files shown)


src/toil/fileStores/cachingFileStore.py
```diff
@@ -1589,16 +1588,83 @@ def _readGlobalFileWithCache(self, fileStoreID, localFilePath, symlink, readerID
...
+        :param str reader_id: Job ID of the job reading the file
+        :param str local_file_path: absolute destination path, if a particular one is desired
+        """
+
+        if not local_file_path:
+            # Generate a file path for the reference if one is not provided.
+            local_file_path = self.getLocalTempFileName()
+
+        # Try and make a 'copying' reference to such a file
+        self._write([('INSERT INTO refs SELECT ?, id, ?, ? FROM files WHERE id = ? AND (state = ? OR state = ?)',
```
DailyDreaming, line 1614: I'm not entirely familiar with the syntax here, so please ignore if this doesn't make sense, but could:

```FROM files WHERE id = ? AND (state = ? OR state = ?)```

be just:

```FROM files WHERE id = ? AND state = ?```

?

adamnovak: No, I need to match either `uploadable` or `uploading`, which are the two distinct `state` values for which there's data that isn't fully written back yet, so I need an `OR` in the expression. Maybe if the state was really a set of flags we could make the condition shorter, but right now it isn't.

DailyDreaming (approved):
LGTM.  Thanks for all of the commented explanations.

PR #3599, opened 2021-05-12
Add type hints toilMain.py

### LINKED ISSUE #3598: Add type hints to toilMain.py
Add type hints to src/toil/utils/toilMain.py so it can be checked under mypy during linting.

Refers to #3568.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-891)
┆Issue Number: TOIL-891

jonathanxu18:
Adds type hints to src/toil/utils/toilMain.py so it can be checked under mypy during linting.

Fixes #3598. 

Refers to #3568.

## Changelog Entry

Type hints added to:
 * `main()`
 * `loadModules()`
 * `printHelp()`
 * `printVersion()`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/utils/toilMain.py
```diff
@@ -29,7 +30,7 @@ def main():
     del sys.argv[1]
     module.main()
 
-def loadModules():
+def loadModules() -> Dict[str, Any]:
```
mr-c, line 33: Can you be more specific here?

adamnovak: I think these are Python modules in the dict; they don't really have a nice base type, do they?

adamnovak (approved):
This all looks good to me.

src/toil/utils/toilMain.py
adamnovak: Yeah I don't think there's a nice way to get ahold of the real type, and even if we had it it would tell us nothing useful about the members:

```
[anovak@octagon ~]$ python3 -i
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import collections
>>> type(collections)
<class 'module'>
>>> module
NameError: name 'module' is not defined
>>> import typing
>>> typing.Module
AttributeError: module 'typing' has no attribute 'Module'
>>> typing.module
AttributeError: module 'typing' has no attribute 'module'
>>> 
```

mr-c: https://docs.python.org/3/library/types.html#types.ModuleType ??

mr-c:
```suggestion
def loadModules() -> Dict[str, types.ModuleType]:
```

and add `import types` at the top of the file

mr-c (changes requested):

src/toil/utils/toilMain.py
jonathanxu18: When I tried adding in `types.ModuleType` I was running into this:

```
src/toil/utils/toilMain.py:32: error: Module has no attribute "main"
src/toil/utils/toilMain.py:51: error: Item "None" of "Optional[str]" has no attribute "strip"
```

Since ModuleType doesn't have attribute main I'm not sure if we can specify this without producing an error.

adamnovak:
@mr-c, it looks like your suggested change doesn't quite work alone; we get some type checking errors that don't show up with `Any`:

https://ucsc-ci.com/databiosphere/toil/-/jobs/84792#L1459

One of them is [here, where we call `.main()` on a module](https://github.com/DataBiosphere/toil/blob/f3d662fa62f80b76d19daef8c07871f5aa682c21/src/toil/utils/toilMain.py#L31); not *all* modules have a `.main()`, but all the ones that hold Toil subcommands do.

I think the second one is [here, where we access `.__doc__`](https://github.com/DataBiosphere/toil/blob/f3d662fa62f80b76d19daef8c07871f5aa682c21/src/toil/utils/toilMain.py#L50), which is `None` if the module lacks a docstring. We know all the Toil subcommand modules are going to have docstrings, but MyPy doesn't.

It looks like [we can add `# type: ignore` comments on the offending lines](https://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker) to shut the typechecker up, because we know those particular operations should work. We could also replace the direct accesses to module members with code that either uses `getattr` and handles missing stuff (for `.main()`) or just tolerates None (for `__doc__`).

@mr-c Is there a way to sketch out the contract for these modules in a way that MyPy can understand, in the absence of an actual shared base type? We're only ever working with a fixed set of our own modules, which implement a common interface, but we don't have a way for them to extend a base class like if they were objects. Type checking based on the interface of all modules is not what we want; that's not the contract our code is meant to work with.

mr-c:
The easy way would be to use getters and give a nice error message if it falls.

Perhaps a fancy type can be written, but then you'd need to cast to that type. Which means it would be polite to check before casting.

So if these modules are only used once, then I recommend the getters approach. If they are used many times, then the special type may be worth it.

adamnovak:
OK @mr-c I have added some getter logic to @jonathanxu18's PR here.

mr-c (approved):

adamnovak:
All the real tests passed; the extra Travis test is from the .com migration. I shall merge this.

PR #3628, opened 2021-05-20
Turn on more CWL tests

### LINKED ISSUE #3538: Fix CWL 1.2 conformance test 35 to reliably succeed on Kubernetes
Sometimes this test fails when run on Kubernetes. It should always pass, and it should be made a non-optional part of CI.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-851)
┆Epic: CWL v1.2 support
┆Issue Number: TOIL-851

### LINKED ISSUE #3539: Fix CWL 1.2 conformance test 39 to reliably succeed on Kubernetes
Sometimes this test fails when run on Kubernetes. It should always pass, and it should be made a non-optional part of CI.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-852)
┆Epic: CWL v1.2 support
┆Issue Number: TOIL-852

### LINKED ISSUE #3540: Fix CWL 1.2 conformance test 42 to reliably succeed on Kubernetes
Sometimes this test fails when run on Kubernetes. It should always pass, and it should be made a non-optional part of CI.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-853)
┆Epic: CWL v1.2 support
┆Issue Number: TOIL-853

### LINKED ISSUE #3541: Fix CWL 1.2 conformance test 56 to reliably succeed on Kubernetes
Sometimes this test fails when run on Kubernetes. It should always pass, and it should be made a non-optional part of CI.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-854)
┆Epic: CWL v1.2 support
┆Issue Number: TOIL-854

adamnovak:
## Changelog Entry

 * CWL 1.2 tests 35, 39, 42, and 56 now run (#3538, #3539, #3540, #3541)
 * Kubernetes batch system no longer runs the root job an extra time on the cluster if it is a CWL local job
 * Job descriptions track the version number of the description, for debugging job description update logic
 * Debug logging for promise tracking has been expanded
 * CWL tools which use input files where the paths are specified in the tool definition will now have their input files imported properly

Closes #3538
Closes #3539
Closes #3540
Closes #3541

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak:
This is a lie; I've reproduced some of these failing with complaints about the promise-fiddling job at the end of the workflow running too soon.

adamnovak:
This can reproduce the issue with test 39:

```
FAILED=0
for TRY_NUM in {1..5} ; do
    for ITER in {1..32} ; do
        OMP_NUM_THREADS=1 TOIL_AWS_ZONE=us-west-2a TOIL_AWS_SECRET_NAME=shared-s3-credentials TOIL_APPLIANCE_SELF=quay.io/ucsc_cgl/toil:5.4.0a1-1a69dc2607c73f3b1eb707a1bab22693071c07ad-py3.6 'cwltest' '--verbose' '--tool=toil-cwl-runner' '--test=/public/home/anovak/build/toil/src/toil/test/cwl/spec_v12/conformance_tests.yaml' '--timeout=2400' '--basedir=/public/home/anovak/build/toil/src/toil/test/cwl/spec_v12' '-n=39' '-j8' '--' '--disableCaching=False' '--clean=always' '--logDebug' --setEnv OMP_NUM_THREADS '--batchSystem=kubernetes' >iter${ITER}.log 2>&1 &
    done
    
    for ITER in {1..32} ; do
        wait
        echo "Status: $?"
    done
    
    for ITER in {1..32} ; do
        grep "All tests passed" "iter${ITER}.log"
        if [[ "${?}" == "1" ]] ; then
            echo "Failure observed in iter${ITER}.log!"
            FAILED=1
            break
        fi
    done
    
    if [[ "${FAILED}" == "1" ]] ; then
        break
    fi
    echo "No failures observed in try ${TRY_NUM}"
done
```

You may want to hack the `--quiet` option in cwltoil so it actually logs at debug level.

adamnovak:
The failures are happening because the root CWLWorkflow job is managing to get its body run twice, which messes up all the cross-references to what promises ought to be filled in when, I think.

adamnovak:
I think the problem is that we have the same job (`CWLWorkflow`) running at the same time on Kubernetes and locally, and they are racing and generally causing trouble.

adamnovak:
OK, the real problem was that the Kubernetes batch system was mistaking local job 0 for a non-local job and also running it on Kubernetes, leading to races and maybe the whole-workflow-runs-twice issue?

adamnovak:
I've added a fix for test 56 to this; we weren't quite recursing right when exploring the CWL workflow looking for tool definitions and `File` objects, so we were forgetting to import files that came along as part of tool definitions instead of workflow inputs.

adamnovak:
OK, I noticed that I had broken a test on single machine, and digging into that I noticed that the way we had been shipping Directory CWL objects around and dealing with them could never work between machines.

I didn't quite understand it to start with, but I've now changed it so we have two kinds of location paths:
* `toilfs:` for actual files, which have a FileID encoded in there
* `_:` for directories, which have a URL-safe base64-encoded JSON description of the directory structure, with files represented by their `toilfs:` paths.

When a job finishes or when we upload the initial workflow/order, we go through, get full CWL `listing` values with all files recursively uploaded, then pack those for the top-level Directories into the weird encoded `location` format. Then `ToilFsAccess` and `toil_get_file` have enough information to actually find a file, given its path relative to a directory, or to e.g. glob in a directory.

Right now `ToilFsAccess` just downloads the whole directory into a local cache as soon as someone wants to touch it, but it seems like its whole point is to present a virtual filesystem, so we could make that more efficient.

If we don't like encoding whole directory structures, we could probably write the directory structure to a file in the filestore and point to that instead. What we can't do is what we've been doing where we just use the local path corresponding to the CWL Directory's original path, and see if there's any stuff there or not.

adamnovak:
CWL specifies that tasks be presented with a configurable-depth `listing` for each directory, and it insists on computing this when it goes to run the task. So instead of precomputing listings when we put the job in, we make sure the `listing` field is empty and only top-level Directories exist in the CWL objects, and the tasks compute the levels of listings they need via the `ToilFsAccess`.

adamnovak:
This might not quite be ready to merge yet; I have some extraneous debugging hooks I want to rip out if it somehow passes.

src/toil/cwl/cwltoil.py
```diff
@@ -1757,8 +2406,29 @@ def filtered_secondary_files(unfiltered_secondary_files: dict) -> list:
...
 
+def scan_for_unsupported_requirements(tool: Process) -> None:
+    """
+    Scan the given CWL tool for any unsupported optional features.
+
+    If it has them, raise an informative UnsupportedRequirement.
+
+    """
+
+    UNSUPPORTED_REQUIREMENTS = ["InplaceUpdateRequirement"]
```
mr-c, line 2417: What happened? `toil-cwl-runner` used to support this!

mr-c: This definitely worked for single node execution and via batch schedulers when using a shared filesystem. In other scenarios (non shared filesystems for the job store) I agree that supporting `InplaceUpdateRequirement` doesn't make sense.

src/toil/worker.py
```diff
@@ -597,7 +618,7 @@ def workerScript(jobStore, config, jobName, jobStoreID, redirectOutputToLogFile=
         jobStore.delete(jobDesc.jobStoreID)
 
     if jobAttemptFailed:
-        return 1
+        return fail_exit_code
```
mr-c, line 621: > UnboundLocalError: local variable 'fail_exit_code' referenced before assignment

src/toil/worker.py
```diff
@@ -492,9 +499,13 @@ def workerScript(jobStore, config, jobName, jobStoreID, redirectOutputToLogFile=
...
     #Trapping where worker goes wrong
     ##########################################
-    except: #Case that something goes wrong in worker
+    except Exception as e: #Case that something goes wrong in worker
         traceback.print_exc()
         logger.error("Exiting the worker because of a failed job on host %s", socket.gethostname())
+        if isinstance(e, CWL_UNSUPPORTED_REQUIREMENT_EXCEPTION):
+            # We need to inform the leader that this is a CWL workflow problem
+            # and it needs to inform its caller.
+            fail_exit_code = CWL_UNSUPPORTED_REQUIREMENT_EXIT_CODE
```
mr-c, line 508:
```suggestion
            failure_exit_code = CWL_UNSUPPORTED_REQUIREMENT_EXIT_CODE
```

src/toil/worker.py
mr-c:
```suggestion
        return failure_exit_code
```

src/toil/cwl/cwltoil.py
adamnovak: I wanted to turn this off because I'm reworking Toil to use logic that works for multiple nodes, which ends up actually keeping Directories' contents in Toil instead of on the filesystem. Right now in the implementation I have, the Directory contents are basically packed into strings that Toil passes around, meaning they can't be updated from elsewhere.

I could optimize it to check if the `single_machine` batch system is being used, or any batch system with a file job store, and then find a place for CWL to keep all its stuff in a local temp directory and support InplaceUpdateRequirement, but that seems awkward. I don't *think* the FileJobStore technically presents an interface to get a shared temp directory right now; I'm not really sure I understand how we were ensuring that files or Directories identified by paths weren't deleted when jobs ended before.

What I am hoping we might be able to do eventually is keep Directory contents in mutable files, which could be replaced from other jobs, allowing for side-effect updates from any machine to be visible on any other, with all the race conditions that you would expect in that scenario being exposed to the workflow. But I'm not sure that that semantically matches what `InplaceUpdateRequirement` is supposed to mean; Lon said it was added because people had big files that were too huge to copy, and allowing the side effects without actually being able to prevent the copy is no help to anyone.

adamnovak:
OK, I fixed the last remaining bug where we need to skip missing files to make optional `secondaryFiles` work, and this is ready for one last review of those changes and then to be merged.

mr-c (dismissed):
Epic PR!

I would really appreciate it if the existing support for `InplaceUpdateRequirement` on single 
machines, or with batch system backends was not removed.

src/toil/cwl/cwltoil.py
```diff
@@ -109,6 +115,79 @@
...
+        else:
+            # Look inside it instead
+            for key in rec:
+                visit_top_cwl_class(rec[key], classes, op)
+    elif isinstance(rec, MutableSequence):
+        # This item is actually a list of things, so look at all of them.
+        for key in rec:
+            visit_top_cwl_class(key, classes, op)
+
+# TODO: we need some type constraints here that I don't know how to spell.
```
mr-c, line 151: Check out https://github.com/common-workflow-language/cwltool/blob/dc5381e04fa2a9d18de5c78e0f8c2907919983ee/cwltool/utils.py#L85 and friends

adamnovak: I guess `CWLObjectType` could be used in place of `MutableMapping`, but this comment really refers to how the `Any` returned from `op_down` needs to be the same `Any` accepted by `op_up`, for a given call to this function. I don't know if Python's type system supports that kind of constraint.

mr-c: Try `OpType= TypeVar('OpType')` and use that

adamnovak:
Reimplementing in place update when the job store is file-backed is going to require a bunch of code changes at places that might not currently know what job store type is in use (to bypass the Directory storage system), and possibly some changes to the AbstractJobStore or FileJobStore API to give them a concept of a shared directory. I personally would want to develop it as its own ticket, so the changes can go through their own review, but I can try and pile them on top of this if we'd rather review them that way.

I think we'll still need a new ticket for me to say I'm working on though, which would then block CWL Kubernetes support until it's done.

Do we have consensus on the relative priority of CWL Kubernetes support and reimplementation of InplaceUpdateRequirement @mr-c and @DailyDreaming?

mr-c:
As long as the in place update is restored before the next Toil release, then I'm okay.

DailyDreaming:
@adamnovak I'm going to do a release today/tomorrow, before these are merged.  We can then work on `InplaceUpdateRequirement` prior to the release that follows.

src/toil/cwl/cwltoil.py
```diff
@@ -792,25 +1193,217 @@ def write_file(writeFunc: Any, index: dict, existing: dict, file_uri: str) -> st
...
+    :param cwl_object: CWL tool (or workflow order) we are importing files for
+
+    :param skip_broken: If True, when files can't be imported because they e.g.
+    don't exist, leave their locations alone rather than failing with an error.
+    """
+
+    try:
+        tool_id = cwl_object['id']
+    except:
+        tool_id = str(cwl_object)
```
DailyDreaming, line 1249: Suggestion:

```
    try:
        tool_id = cwl_object['id']
    except:
        tool_id = str(cwl_object)
```

->

```
tool_id = cwl_object.get('id', str(cwl_object))
```

src/toil/cwl/cwltoil.py
```diff
@@ -792,25 +1193,217 @@ def write_file(writeFunc: Any, index: dict, existing: dict, file_uri: str) -> st
...
+
+    logger.debug('Importing files for %s', tool_id)
+
+    # We need to upload all files to the Toil filestore, and encode structure
+    # recursively into all Directories' locations. But we cannot safely alter
+    # the listing fields of Directory objects, because the handling required by
+    # the 1.2 conformance tests does not actually match the spec, and we need
+    # to pass the conformance tests, and only cwltool really knows when to
+    # make/destroy the listings in order to do that.
```
DailyDreaming, line 1258: This sounds like it should be made into a cwltool issue and that issue linked in the comment so that future coder-lings will know if that state has changed.  I'm not sure where in the spec this is referencing and if the spec changes to address this in the future, I really won't know.

src/toil/cwl/cwltoil.py
```diff
@@ -937,35 +1539,64 @@ def _collectDirEntries(
...
-                ):  # TODO: CreateFile for buckets is not under testing
-                    local_file_path = os.path.join(
-                        file_store.getLocalTempDir(), baseName
-                    )
-                    with open(local_file_path, "wb") as n:
-                        n.write(p.resolved.encode("utf-8"))
+    # TODO: I think we need to recursively populate listings here!
+
+    # Now we need to save all the output files and directories.
+    # We could use a ToilPathMapper, but that contains stuff to work with Toil directories as encoded in Locations, while we can assume that we have all the Files and
```
DailyDreaming, line 1548: and... ?  XD

src/toil/cwl/cwltoil.py
```diff
@@ -937,35 +1539,64 @@ def _collectDirEntries(
...
 
-                if (
-                    p.type == "CreateFile"
-                ):  # TODO: CreateFile for buckets is not under testing
-                    local_file_path = os.path.join(
-                        file_store.getLocalTempDir(), baseName
-                    )
-                    with open(local_file_path, "wb") as n:
-                        n.write(p.resolved.encode("utf-8"))
+    # TODO: I think we need to recursively populate listings here!
```
DailyDreaming, line 1545: Is this not done with the next `pm = ToilPathMapper`call?  Or old listings are not up-to-date?

src/toil/cwl/cwltoil.py
```diff
@@ -937,35 +1539,64 @@ def _collectDirEntries(
...
+                            f.write(file_id_or_contents.encode('utf-8'))
+                            f.close()
+                            # Import it and pack up the file ID so we can turn around and export it.
+                            file_id_or_contents = 'toilfs:' + toil.importFile(f.name).pack()
+
+                    if file_id_or_contents.startswith('toilfs:'):
+                        # This is something we can export
+                        destUrl = "/".join(s.strip("/") for s in [destBucket, baseName])
+                        toil.exportFile(FileID.unpack(file_id_or_contents[len('toilfs:'):]), destUrl)
+                    # TODO: can a toildir: "file" get here?
```
DailyDreaming, line 1585: I think if the type is a variant of `File` but the id is `toildir:` then it should error here.  That _should_ never happen.

src/toil/cwl/cwltoil.py
```diff
@@ -1207,15 +1844,15 @@ def run(self, file_store: AbstractFileStore) -> Any:
...
-        # runtime_context.builder = self.builder
-
         process_uuid = uuid.uuid4()  # noqa F841
         started_at = datetime.datetime.now()  # noqa F841
 
-        logger.debug("Running CWL job: %s", cwljob)
+        original = cwltool.command_line_tool.check_adjust
+        def wrapper(builder, file_o):
+            original(builder, file_o)
+        cwltool.command_line_tool.check_adjust = wrapper
```
DailyDreaming, line 1853: This `original`/`wrapper` incantation should probably have a comment explaining itself.

src/toil/cwl/cwltoil.py
```diff
@@ -1752,13 +2413,37 @@ def filtered_secondary_files(unfiltered_secondary_files: dict) -> list:
...
+    If it has them, raise an informative UnsupportedRequirement.
+
+    """
+
+    UNSUPPORTED_REQUIREMENTS = ["InplaceUpdateRequirement"]
+
+    # TODO: Can we change this to a set of supported requirements instead? Otherwise when new ones show up we assume we have them even when we don't.
+
+    for req_name in UNSUPPORTED_REQUIREMENTS:
+        # Grab each reqirement that might be a problem
```
DailyDreaming, line 2437: `reqirement`

src/toil/cwl/cwltoil.py
```diff
@@ -2353,37 +3053,42 @@ def set_secondary(fileobj):
...
-            # changed to 'toilfs:'
-            import_files(initialized_job_order)
-
-            visitSteps(tool, import_files)
-
-            for job_name, job_params in initialized_job_order.items():
-                rm_unprocessed_secondary_files(job_params)
+            for param_name, param_value in initialized_job_order.items():
+                # Loop through all the parameters for the workflow overall.
+                # Drop any files we couyldn't import; they will cause an error
```
DailyDreaming, line 3089: `couyldn't`

src/toil/cwl/cwltoil.py
```diff
@@ -2353,37 +3053,42 @@ def set_secondary(fileobj):
...
 
-            # files with the 'file://' uri are imported into the jobstore and
-            # changed to 'toilfs:'
-            import_files(initialized_job_order)
-
-            visitSteps(tool, import_files)
-
-            for job_name, job_params in initialized_job_order.items():
-                rm_unprocessed_secondary_files(job_params)
+            for param_name, param_value in initialized_job_order.items():
```
DailyDreaming, line 3087: Hmmm... why was this changed from `job_name, job_params` to something more ambiguous?  Are the `initialized_job_order.items()` not all jobs?

src/toil/job.py
```diff
@@ -472,6 +472,10 @@ def makeString(x):
         # the logging has been captured to be reported on the leader.
         self.logJobStoreFileID = None
 
+        # Every time we update a job description in place in the job store, we
+        # increment this.
```
DailyDreaming, line 476: Is this only for debugging?  The purpose of this isn't clear to me.

src/toil/cwl/cwltoil.py
```diff
@@ -2345,6 +3044,7 @@ def set_secondary(fileobj):
             # make sure this doesn't add listing items; if shallow_listing is
             # selected, it will discover dirs one deep and then again later on
             # (producing 2+ deep listings instead of only 1)
+            # TODO: when is later on?
```
DailyDreaming, line 3047: When the builder operates on this in the `job.run()` step if I remember correctly.  Though I may be misremembering.

DailyDreaming (commented):
The directory implementation makes sense to me and I think brings us more closely into sync with cwltool, so I'm pretty excited to have this in.

Some of the functions have some overlapping functionality that I feel could be reduced (like the traversing up and down and both and normal visit() functions), but that's probably more time than it's worth trying to reduce and more of a nit.

A few suggestions:

- Could `toilfs:` be `toilfile:` now that we have `toildir:`?
- Could `cwltoil.py` be broken into two or more files now that we're exceeding 3000 lines of code?
- I believe I commented on the relevant functions, but some of the utility functions might benefit from adding unit tests.

Note to future code archaelogists:  This PR drops support for `InplaceUpdateRequirement`, but will be added back in https://github.com/DataBiosphere/toil/pull/3652 so Toil will avoid doing a release without this support.

src/toil/cwl/cwltoil.py
adamnovak: OK, I'll define a TypeVar for the down operation return type.

src/toil/job.py
```diff
@@ -472,6 +472,10 @@ def makeString(x):
...
+        self._generation = 0
```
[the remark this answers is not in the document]
adamnovak: I'll rename this.

src/toil/job.py
```diff
@@ -823,6 +831,16 @@ def __str__(self):
...
         return '%s( **%r )' % (self.__class__.__name__, self.__dict__)
 
+    def pre_update_hook(self) -> None:
+        """
+        Called by the job store before pickling and saving a created or updated
+        version of a job.
+        """
+
+        self._generation += 1
+        logger.debug("New generation: %s", self)
```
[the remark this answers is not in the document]
adamnovak: Let's go with "New job version".

src/toil/cwl/cwltoil.py
adamnovak: Yeah, this will work as long as the type constraints are followed. I think I meant to also maybe support lists?

adamnovak: Although it will always evaluate `str()`

src/toil/cwl/cwltoil.py
adamnovak: I think we don't actually need to do it after all. The path mapper and whatever stages using it is clever enough to work even without content listings.

adamnovak: I think we're using our own custom code that just makes a directory to "stage" the directory, instead of using e.g. ToilFsAccess to list its files and copy them. Seems like it might be wrong, but also there's no conformance test here that we're failing.

adamnovak: It's not that listings won't be up to date so much as it's that they might not exist or be recursive; the child Directory and File objects are only sometimes created at the CWL level, but all the contents are supposed to come along all the time.

src/toil/cwl/cwltoil.py
adamnovak: I think it can happen when you take a `toildir:` directory and compose a path to a file within it by joining subdirectories/filenames onto its URL with `/`. That happens in some places, and I'm not sure if it can happen here or not. It might be safest to actually implement `toildir:`.

src/toil/cwl/cwltoil.py
adamnovak: I've dropped this; I just had it so I could cram debugging in there.

src/toil/cwl/cwltoil.py
adamnovak: Nope! This isn't an "order" as in a sequence of jobs; it's an "order" as in instructions to run the jobs with certain arguments. I changed it to stop being misleading.

src/toil/job.py
adamnovak: I added this for debugging because I thought we might have been forgetting to save an update or running the wrong version of a job somewhere; it turned out that wasn't the case and we were giving the same version of the job to two batch systems, but I left this in because I want to use it later when poking around the leader looking for race conditions with its weird job update system.

src/toil/cwl/cwltoil.py
```diff
@@ -109,6 +115,79 @@
...
+# parse workflow files see an unsupported feature?
+CWL_UNSUPPORTED_REQUIREMENT_EXIT_CODE = 33
+
+# And what error will make the worker exit with that code
+CWL_UNSUPPORTED_REQUIREMENT_EXCEPTION = UnsupportedRequirement
+
+
+# Customized CWL utilities
+
+def visit_top_cwl_class(
```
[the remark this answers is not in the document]
adamnovak: I'll write one.

src/toil/cwl/cwltoil.py
```diff
@@ -109,6 +115,79 @@
...
+            # Look inside it instead
+            for key in rec:
+                visit_top_cwl_class(rec[key], classes, op)
+    elif isinstance(rec, MutableSequence):
+        # This item is actually a list of things, so look at all of them.
+        for key in rec:
+            visit_top_cwl_class(key, classes, op)
+
+# TODO: we need some type constraints here that I don't know how to spell.
+def visit_cwl_class_and_reduce(
```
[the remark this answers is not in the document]
adamnovak: I'm writing one for this too.

src/toil/cwl/cwltoil.py
```diff
@@ -792,25 +1193,217 @@ def write_file(writeFunc: Any, index: dict, existing: dict, file_uri: str) -> st
                 raise
         return index[file_uri]
 
+def path_to_loc(obj: Dict) -> None:
```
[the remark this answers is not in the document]
adamnovak: I tried calling `path_to_loc` on the already imported `cwltool.main` and it didn't exist; maybe it's not/hasn't been/isn't staying part of the API?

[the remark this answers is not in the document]
adamnovak: It looks like it's an internal function of `init_job_order` so we can't get at it.

src/toil/cwl/cwltoil.py
```diff
@@ -682,83 +907,259 @@ def toil_make_tool(
...
+            f'Cannot obtain file {file_store_id} from host '
+            f'{socket.gethostname()}; all imports must happen on the '
+            f'leader!'
         )
-    src_path = file_store.readGlobalFile(FileID.unpack(file_store_id[7:]), symlink=True)
-    index[src_path] = file_store_id
-    existing[file_store_id] = src_path
-    return schema_salad.ref_resolver.file_uri(src_path)
+
+def download_structure(
```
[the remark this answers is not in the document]
adamnovak: It could, but I don't think I'll have time to write it before I leave. You might have to do it, or I can do it when I get back. This one won't be as simple because we have to make up a file store and find it a place to download to. We might actually want to redesign this a bit if we really intend to test it.

adamnovak:
OK @DailyDreaming, I've addressed most of your comments and synced this up so it's the same commit as #3652.

I split out a couple of the utility functions, but I don't think it made much of a dent in the bigness of the file. We might want to pull out ToilPathMapper, ToilFsAccess, etc. into their own files.

I also still don't have unit tests for `download_structure`, and probably won't for the next 2 weeks. If we want to ship this sooner and we need them, someone else might need to write them. Otherwise I can write them when I get back.

adamnovak:
OK, I've fixed the problem that failed the tests on the last merge with the main development line, and added a test for `download_structure`. This *should* pass and is ready for re-review.

adamnovak:
Actually since Lon is out, I think I might just merge this when it passes, since I have addressed the comments.

mr-c:
Can this be closed now that #3652 is merged?

PR #3629, opened 2021-05-24
Clean up docker containers by default

### LINKED ISSUE #3505: Exited Docker containers are not cleaned up
I ran this command a bunch in `toil-vg` commit b3c290f7143376978e35e98ed524ca613941de01 or so:

```
make test tests=src/toil_vg/test/test_vg.py::VGCGLTest::test_10_construct_multiple_contigs
```

Now I have a lot of "Exited" Docker containers on my machine:

```
$ docker ps -a | wc -l
1950
```

It looks like when Toil runs a Docker container, it does not delete the finished container. It could be because `toil-vg` always uses `detach=True` with `apiDockerCall`.

Toil should delete the job's containers after the job is finished, or, better yet, always do the equivalent of `docker run --rm` when running Docker containers.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-830)
┆Issue Number: TOIL-830

w-gao:
Make sure Toil removes the docker containers after they are finished running, so we don't have a bunch of "Exited" Docker containers. 

Closes #3505.

## Changelog Entry

 * Clean up docker containers by default.  Set `remove=False` in `apiDockerCall()` to leave the containers untouched and running.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/lib/docker.py
```diff
@@ -222,19 +222,16 @@ def toil_job(job):
     client = docker.from_env(version='auto', timeout=timeout)
 
     if deferParam == STOP:
-        job.defer(dockerStop, containerName)
-
-    if deferParam == FORGO:
-        remove = False
+        job.defer(dockerStop, containerName, remove)
+    elif deferParam == FORGO:
```
w-gao, line 226: I hope I didn't change the behavior here... If I understand correctly, we had `if` here so it is possible to remove after stop if `deferParam == STOP`. But now I've moved the remove option to `dockerStop / dockerKill`.

src/toil/lib/docker.py
```diff
@@ -59,8 +59,8 @@ def apiDockerCall(job,
                   entrypoint=None,
                   detach=False,
                   log_config=None,
-                  auto_remove=None,
-                  remove=False,
+                  auto_remove=False,
+                  remove=True,
```
adamnovak, line 63: Actually, a lot of these kwargs are borrowed from the Docker API, and we sort of refer to it in the docstring as useful documentation for their semantics. If we're now differing more substantially from the default behavior of the Docker API, maybe we need to write something about differences from the normal Docker API in our docstring.

src/toil/lib/docker.py
adamnovak: It looks like the original behavior maybe didn't match the docstring. `RM` is supposed to be the default if no `deferParam` is passed, but both now and before it looks like for a `deferParam` of `None`, no deferred cleanup is done (unless `remove` is set).

adamnovak (changes requested):
I have some small complaints, but mostly I'm now suspicious of the whole `deferParam` system. I think we need to at least re-document it so it makes sense along with the `remove` flag. As is it's not clear what all the combinations are supposed to do.

A much simpler fix for #3505 might be possible: I don't think the documented default `deferParam` behavior of `RM` has been working, which would account for all the left-behind containers from toil-vg.

Do we have any unit tests for the cleanup here? If not, the first step should really be to add some. Then we can be certain of what the combinations of parameters are meant to do, because they will be tested.

adamnovak:
@w-gao I've added a bunch of complaints here, but I don't think it would make sense to go through and address all of them. I think we want to do one of:

1. Not actually change the old `remove` flag at all and instead test the `RM` `deferParam` and make sure it works as documented to clean up containers after the job when not explicitly set.
2. Refactor `remove` to actually work through the deferred cleanup system like you have done, give up on `apiDockerCall` looking quite like the Python Docker API, and then start revising its API to make more sense (maybe deprecating `deferParam` in favor of `remove` to control cleanup).

w-gao:
Thanks @adamnovak for the review! 

I'm leaning towards fixing the default `deferParams` behavior for clean up ([your first suggestion](https://github.com/DataBiosphere/toil/pull/3629#issuecomment-848196668)), because I do want `apiDockerCall()` to look like the Python Docker API, and I think it would make sense to not change the old `remove` flag so users can control the underlying `client.containers.run()` call if needed. 


As for unit tests, we actually have these [here](https://github.com/DataBiosphere/toil/blob/releases/5.3.x/src/toil/test/lib/dockerTest.py#L54) that test the different combinations of options. Although, I don't think they are tested on the CI. I'll look into the tests and see if the default `deferParam` behavior of `RM` is being tested.

src/toil/lib/docker.py
```diff
@@ -135,7 +135,7 @@ def toil_job(job):
                         stderr and/or stdout are True. Defaults to "output.log".
     :param dict log_config: Specify the logs to return from the container.  See:
                       https://docker-py.readthedocs.io/en/stable/containers.html
-    :param bool remove: Remove the container on exit or not.
+    :param bool remove: Remove the container on exit or not. (default: True).
```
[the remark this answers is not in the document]
w-gao: Ah, thanks for catching the problem with conflicting options. That was something that I didn't think about. 

And yes, I chose to use the `deferParam` system because `remove`/`auto_remove` doesn't work well with `detach=True`. So for example: 

```python
def docker_call(job):
    container = apiDockerCall(job,
                              image="ubuntu:latest",
                              parameters=["ls", "-lha", "/"],
                              detach=True,
                              remove=True,
                              )

    stdout = container.logs(stderr=False, stdout=True)  # 404 error!
    logger.warning(stdout)
```

Here, you'd get a `docker.errors.NotFound: 404 Client Error: Not Found ("No such container: "...")` error, because I think the container is removed before the `logs()` call. I'd be curious to know if there's a way to solve this? 

Using the `deferParams=RM` system, however, would call `docker kill` and `docker rm` after users are done with the container, so I think that would give the desired behavior? I originally planned to use the `remove` flag for clean up when `detach=False`, and defer remove when `detach=True`, but that seemed to complicate things more. 

I'm leaning towards fixing the default `deferParams` behavior, but let me know what you think.

src/toil/lib/docker.py
w-gao: > It looks like the original behavior maybe didn't match the docstring. 

Was there a reason for this? I don't want to make `RM` to be the default and break something else.

adamnovak: If there's a reason for it to be the way it is now that was important enough to us that we wanted to remember it later, we would have written it down in the documentation, or a comment. Or *maybe* a commit message.

I think the reason the behavior doesn't match the documentation is just a mistake.

src/toil/lib/docker.py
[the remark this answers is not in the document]
adamnovak: Hm, that is a good reason not to lean on `remove`/`auto_remove` for all our cleanup.

Fixing the default `deferParams` sounds like the right approach to me.

adamnovak:
It does look like the `dockerTest.py` file might not be getting run on CI; maybe we can add it to the Gitlab tests next to [other `lib` tests](https://github.com/DataBiosphere/toil/blob/1f0930b7b9ecc31ca556d15ab07ff836ba85eb23/.gitlab-ci.yml#L246)?

adamnovak (approved):
This is definitely simpler now, and I think it should work. @w-gao Are you still adding tests?

w-gao:
Yeah @adamnovak, I'm trying to make sure all combinations work as intended, so this is not quite ready yet.

w-gao:
@adamnovak This should be ready now! I changed a few things since your last review, so this may need a re-review. 

For some reason the [file permission check](https://github.com/DataBiosphere/toil/blob/master/src/toil/test/lib/dockerTest.py#L106-L107) doesn't pass on the CI, but works fine locally. I'll take a deeper look into that but that's probably a different issue.

adamnovak (approved):
I'm happy with this.

adamnovak:
I think the file permission test fails because we actually are running as root on CI, whereas the test expects us to be running as someone other than root, who it expects to see owning the relevant files.

PR #3652, opened 2021-06-16
Allow filestore bypass

adamnovak:
## Changelog Entry

 * `--bypass-file-store` lets `toil-cwl-runner` support `inplaceUpdateRequirement`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak:
OK @mr-c, I have it so that if you want to have `InplaceUpdateRequirement` you can pass `--bypass-file-store`, and instead of putting files in the file store, it will leave them in place in the temp output directories.

Can you have a look at how the temp output directories work? I think we are just creating them and never cleaning them up. We used to use a default prefix of `tmp` and we were dumping them all in the current directory, but now I changed it to basically`$TMPDIR/tmp` so they will live under Python's configured temporary directory.

How does `cwltool` manage its intermediate files? Does it move them all over to the output directory? Does it clean them up somehow at the end of the workflow?

docs/running/cwl.rst
```diff
@@ -118,6 +118,18 @@ samples inputs, it could look something like:
...
+---------------------------------------------------
+
+Some CWL workflows use the ``InplaceUpdateRequirement`` feature, which requires
+that operations on files have visible side effects that Toil's file store
+cannot support. If you need to run a workflow like this, you can make sure that
+all of your worker nodes have a shared filesystem, and use the
+``--bypass-file-store`` option to ``toil-cwl-runner``. This will make it leave
+all CWL intermediate files on disk and share them between jobs using file
+paths, instead of storing them in the file store and downloading them when jobs
+need them.
```
mr-c, line 131: Wouldn't that be faster for all users with a shared filesystem?

mr-c:
> Does it clean them up somehow at the end of the workflow?

Yes

https://github.com/common-workflow-language/cwltool/blob/main/cwltool/job.py#L429

docs/running/cwl.rst
adamnovak: Maybe in the case where you're also using a file job store on the shared filesystem, and symlinking everything, because you don't need to take the time to make those symlinks. And in the case where you have a lot of files, because with the file store you're reconstructing the directory structure for all the jobs that want it, instead of sharing it. But I could see cases where using S3 would be faster than bypassing the filestore through a shared filesystem, if your cluster is in AWS and has a pretty slow shared filesystem for home directories or something.

In general I see bypassing the filestore as a big hack to support some weird workflows that need a weird feature, and not really what we want most people to be doing most of the time. It definitely opens up new ways to break things: we don't really have a way for jobs that *don't* have `InplaceUpdateRequirement` to *not* see them, when we do it this way, so jobs might end up seeing files they shouldn't have, and then not work right when run on another cluster where there isn't a shared filesystem and the file store actually has to be used.

src/toil/cwl/cwltoil.py
```diff
@@ -1754,17 +2524,59 @@ def filtered_secondary_files(unfiltered_secondary_files: dict) -> list:
...
+    # don't.
+
+    for req_name in unsupported:
+        # Grab each reqirement that might be a problem
+        req, is_mandatory = tool.get_requirement(req_name)
+        if req and is_mandatory:
+            # The tool actualy uses this one, and it isn't just a hint.
+            # Complain.
+            raise UnsupportedRequirement(
+                f"Toil cannot support {req_name} in the current configuration"
```
DailyDreaming, line 2576: Since we only have one unsupported feature, we should add a special message to the user of which option to use to enable `InplaceUpdateRequirement` and that is must be a shared file system.

DailyDreaming (changes requested):
This makes sense to me, though it throws away the filestore and opens us up to some potential errors, I think it's probably in the spirit of the original feature (not spending time copying very very large files).

I feel like we could be smarter about detecting if a system is shared at the start and giving the appropriate warnings, but I think it's already trying to make it clear how and why one would use this and it's alright for now to lean on the user to get it right.

I made one comment about adding a message about how to use `InplaceUpdateRequirement` when it errors that I feel should be addressed.

adamnovak:
OK @DailyDreaming, I've written a better message that explains what you should need to get `InplaceUpdateRequirement` working.

DailyDreaming (approved):
@adamnovak Thanks.  This looks good to me, provided the comments in the PR this is based on are addressed (https://github.com/DataBiosphere/toil/pull/3628).

adamnovak:
When this is finally done it will probably fix #3667.

PR #3686, opened 2021-06-29
Warn users if a different cwltool version is installed

### LINKED ISSUE #2555: toil-cwl-runner sort of works without 'cwl' extra installed
I neglected to install the 'cwl' extra and was trying to run toil-cwl-runner. I had previously installed cwltool. I got an error from toil code trying to import from cwltool:
```
  File "/cluster/home/barberc/software/anaconda/envs/py3/bin/toil-cwl-runner", line 6, in <module>
    from toil.cwl.cwltoil import main
  File "/cluster/home/barberc/software/anaconda/envs/py3/lib/python3.6/site-packages/toil/cwl/cwltoil.py", line 60, in <module>
    from cwltool.process import (shortname, fill_in_defaults, compute_checksums,
ImportError: cannot import name 'collectFilesAndDirs'
```
This symbol at one time existed in cwltool. In fact, I went so far as identifying and installing cwltool==1.0.20180819175200 which was close enough to the required version for toil 3.18.0 (cwltool==1.0.20180820141117) that everything seemed to work.

Apologies if there is not a practical way to do this with setup.py extras, but I am opening this in hopes that there is a simple way I haven't thought of by which to address this. Ideally the cwl-related commands would not be present, or not be on the PATH, or a warning would be issued when running. Another line of argumentation might be: users may be trying toil after experimenting with cwltool, so there is a good chance they have it already installed (likely the wrong version), and the cwl extras are currently not so heavy of dependencies that they couldn't just be moved out of extras and into a base `pip install toil`. I'm not sure if those points are true, but it could be a way around this.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-45)
┆Issue Number: TOIL-45

w-gao:
Attempts to close #2555. 

## Changelog Entry

 * Warn users if a different cwltool version is installed. 

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/version_template.py
+++ b/version_template.py
@@ -88,7 +88,7 @@ def dockerTag():
     """The primary tag of the Docker image for the appliance. This uniquely identifies the appliance image."""
     return version() + _pythonVersionSuffix()
 
-  
+
 def currentCommit():
     import os
     from subprocess import check_output
@@ -127,12 +127,15 @@ def dirty():
         return False  # In case the git call fails.
 
 
-def expand_(name=None):
+def expand_(name=None, others=None):
     variables = {k: v for k, v in globals().items()
                  if not k.startswith('_') and not k.endswith('_')}
 
+    if others is not None:
+        variables.update(others)
+
     def resolve(k):
-        v = variables[k]
+        v = variables.get(k, None)
         if callable(v):
             v = v()
         return v
--- a/version_template.py
+++ b/version_template.py
@@ -128,6 +128,14 @@ def dirty():
 
 
 def expand_(name=None, others=None):
+    """
+    Returns a string of all the globals and additional variables passed as the
+    others keyword argument.
+
+    :param str name: If set, only the value of the given symbol is returned.
+    :param dict others: A dictionary of additional variables to be included in
+                        the return value.
+    """
     variables = {k: v for k, v in globals().items()
                  if not k.startswith('_') and not k.endswith('_')}
 
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,9 @@ from tempfile import NamedTemporaryFile
 from setuptools import find_packages, setup
 
 
+cwltool_version = '3.1.20210616134059'
+
+
 def run_setup():
     """
     Calls setup(). This function exists so the setup() invocation preceded more internal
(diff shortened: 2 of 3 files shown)


src/toil/cwl/cwltoil.py
```diff
@@ -48,6 +48,14 @@
 )
 from urllib import parse as urlparse
 
+
+# Warn user if the wrong cwltool version is installed.
+try:
+    from toil.cwl.utils import check_cwltool_version
+    check_cwltool_version()
+except ImportError:
+    raise
```
DailyDreaming, line 57: Hmmm... this try/except looks like a no op.

version_template.py
```diff
@@ -31,6 +31,8 @@
 baseVersion = '5.5.0a1'
 cgcloudVersion = '1.6.0a1.dev393'
 
+cwltoolVersion = '3.0.20201203173111'
```
DailyDreaming, line 34: New variables should be snake case.

Makefile
```diff
@@ -93,7 +93,8 @@ red=\033[0;31m
 cyan=\033[0;36m
 
 develop: check_venv
-	pip install -e .$(extras) $(packages)
+	- rm src/toil/version.py
```
DailyDreaming, line 96: Why is this removing `src/toil/version.py`?

DailyDreaming (changes requested):

Makefile
w-gao: This is probably not the best way to do it, but if I put the cwltool version in `version_template.py`, we might need to regenerate `src/toil/version.py` when someone wants to upgrade locally. I could also make it so that `src/toil/version.py` is always regenerated in `setup.py`? 

Suggestions would be appreciated. I don't like how I moved the cwltool version outside of `setup.py`, but I couldn't find another way to access the version inside Toil.

src/toil/cwl/cwltoil.py
w-gao: Ah... this can stop the linter from complaining about `PEP 8: E402 module level import not at top of file`. I think we have to do this before we import cwltool. Perhaps I should move this to `src/toil/cwl/__init__.py`?

mr-c (approved):

version_template.py
```diff
@@ -127,12 +127,15 @@ def dirty():
         return False  # In case the git call fails.
 
 
-def expand_(name=None):
+def expand_(name=None, others=None):
```
adamnovak, line 130: It would be good to explain the new argument here.

adamnovak (changes requested):
This looks pretty good, but I think the new hole we punched to pull the variable through needs some documentation.

version_template.py
```diff
@@ -127,12 +127,23 @@ def dirty():
...
+    """
+    Returns a string of all the globals and additional variables passed as the
+    others keyword argument.
+
+    :param str name: If set, only the value of the given symbol is returned.
+    :param dict others: A dictionary of additional variables to be included in
```
adamnovak, line 136: We've been moving towards just `:param whatever:` here and the type hints in the method name. But we can clean this up if/when we actually MyPy the file, I think.

adamnovak (approved):
I don't think we *quite* check all the type checking boxes, but I missed it before and I think it's worth it to merge this anyway.

Issue #3706, opened 2021-07-13
Add type hints to toilwdl.py


jonathanxu18:
Add type hints to `src/toil/wdl/toilwdl.py` so it can be checked under mypy during linting.

Refers to #3568.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-956)
┆Issue Number: TOIL-956

PR #3707, opened 2021-07-14
Add type hints to wdl files

### LINKED ISSUE #3703: Add type hints to wdl_functions.py
Add type hints to `src/toil/wdl/wdl_functions.py` so it can be checked under mypy during linting.

Refers to #3568.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-954)
┆friendlyId: TOIL-954

jonathanxu18:
Adds type hints to:
 *  `src/toil/wdl/wdl_functions.py`
 *  `src/toil/wdl/toilwdl.py`

Fixes #3703, #3706

Refers to #3568.

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/wdl/wdl_functions.py
+++ b/src/toil/wdl/wdl_functions.py
@@ -239,13 +239,13 @@ def sub(input_str: str, pattern: str, replace: str) -> str:
     """
 
     if isinstance(input_str, WDLFile):
-        input_str = input_str.file_name
+        input_str_name = input_str.file_name
     if isinstance(pattern, WDLFile):
-        pattern = pattern.file_name
+        pattern_name = pattern.file_name
     if isinstance(replace, WDLFile):
-        replace = replace.file_name
+        replace_name = replace.file_name
 
-    return re.sub(pattern=str(pattern), repl=str(replace), string=str(input_str))
+    return re.sub(pattern=str(pattern_name), repl=str(replace_name), string=str(input_str_name))
 
 
 def defined(i: Any) -> bool:
@@ -274,8 +274,8 @@ def process_single_outfile(wdl_file: WDLFile, fileStore: AbstractFileStore, work
         for std_file in ('stdout', 'stderr'):
             std_file = os.path.join(workDir, 'execution', std_file)
             if os.path.exists(std_file):
-                with open(std_file, 'rb') as f:
-                    logger.info(f.read())
+                with open(std_file, 'rb') as fh:
+                    logger.info(fh.read())
 
         raise RuntimeError('OUTPUT FILE: {} was not found in {}!\n'
                            '{}\n\n'
@@ -344,9 +344,9 @@ def abspath_file(f: Any, cwd: str) -> Any:
 def read_single_file(f: WDLFile, tempDir: str, fileStore: AbstractFileStore, docker: bool = False) -> str:
     import os
     try:
-        fpath = fileStore.readGlobalFile(f.file_path, userPath=os.path.join(tempDir, f.file_name))
+        fpath = fileStore.readGlobalFile(f.file_path, userPath=os.path.join(tempDir, str(f.file_name)))
     except:
-        fpath = os.path.join(tempDir, f.file_name)
+        fpath = os.path.join(tempDir, str(f.file_name))
     return str(fpath)
 
 
@@ -517,10 +517,11 @@ def size(f: Optional[Union[str, WDLFile, List[Union[str, WDLFile]]]] = None,
     divisor = bytes_in_unit(unit)
 
     if isinstance(f, list):
-        total_size = sum(file.file_path.size for file in f) # type: ignore
+        total_size = sum(file.file_path.size for file in f if isinstance(file, WDLFile))
         return total_size / divisor
 
-    fileID = f.file_path
+    if isinstance(f, WDLFile):
+        fileID = f.file_path
     return float(fileID.size / divisor)
 
 
@@ -679,9 +680,9 @@ def read_int(path: Union[str, WDLFile]) -> int:
     WDL syntax: Int read_int(String|File)
     """
     if isinstance(path, WDLFile):
-        path = path.file_path
+        file_path = path.file_path
 
-    with open(path, 'r') as f:
+    with open(file_path, 'r') as f:
         return int(f.read().strip())
 
 
@@ -693,9 +694,9 @@ def read_string(path: Union[str, WDLFile]) -> str:
     WDL syntax: String read_string(String|File)
     """
     if isinstance(path, WDLFile):
-        path = path.file_path
+        file_path = path.file_path
 
-    with open(path, 'r') as f:
+    with open(file_path, 'r') as f:
         return str(f.read().strip())
 
 
@@ -707,9 +708,9 @@ def read_float(path: Union[str, WDLFile]) -> float:
     WDL syntax: Float read_float(String|File)
     """
     if isinstance(path, WDLFile):
-        path = path.file_path
+        file_path = path.file_path
 
-    with open(path, 'r') as f:
+    with open(file_path, 'r') as f:
         return float(f.read().strip())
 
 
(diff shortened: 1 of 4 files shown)


src/toil/wdl/wdl_functions.py
```diff
@@ -226,7 +225,7 @@ def process_infile(f: Any, fileStore: AbstractFileStore):
     elif isinstance(f, (int, str, bool, float)):
         return f
     else:
-        raise WDLRuntimeError('Error processing file: '.format(str(f)))
+        raise WDLRuntimeError('Error processing file: {0}'.format(str(f)))
 
 
 def sub(input_str: str, pattern: str, replace: str) -> str:
```
mr-c, line 231:
```suggestion
def sub(input_str: Union[str, WDLFile], pattern: Union[str, WDLFile] replace: Union[str, WDLFile]) -> str:
```

From https://ucsc-ci.com/databiosphere/toil/-/jobs/89507#L1492

src/toil/wdl/wdl_functions.py
```diff
@@ -20,20 +20,19 @@
...
 import textwrap
 import uuid
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Any, Optional, Tuple, Union
 
 from toil.fileStores.abstractFileStore import AbstractFileStore
-from toil.wdl.wdl_types import WDLFile, WDLPair
+from toil.wdl.wdl_types import WDLType, WDLFile, WDLPair
 from toil.lib.conversions import bytes_in_unit
-from toil.lib.resources import glob  # type: ignore
```
adamnovak, line 28: It looks like this function was imported to be re-exported and dropping this broke the tests. Maybe the import needs a comment to that effect?

src/toil/wdl/wdl_functions.py
```diff
@@ -511,26 +512,28 @@ def size(f: Optional[Union[str, WDLFile, List[Union[str, WDLFile]]]] = None,
...
     # validate the input. fileStore is only required if the input is not processed.
-    f = process_infile(f, fileStore)
+    if fileStore:
+        f = process_infile(f, fileStore)
 
     divisor = bytes_in_unit(unit)
 
     if isinstance(f, list):
-        total_size = sum(file.file_path.size for file in f)
+        total_size = sum(file.file_path.size for file in f if isinstance(file, WDLFile))
```
adamnovak, line 521: Aren't they all WDLFiles by now because f was replaced above in `elif isinstance(f, list):`?

src/toil/wdl/wdl_types.py
```diff
@@ -210,7 +211,7 @@ class WDLFile:
     """
     Represents a WDL File.
     """
-    def __init__(self, file_path: str, file_name: Optional[str] = None, imported: bool = False):
+    def __init__(self, file_path: FileID, file_name: Optional[str] = None, imported: bool = False):
```
adamnovak, line 214: If you need a `FileID` to make a `WDLFile`, how is `size()` going to run correctly when passes a `str` that is not a `FileID`? Won;t it end up passing it to the `WDLFile` constructor?

adamnovak (commented):
I didn't see this the first time, but it looks like WDL `size()` isn't going to be type correct when passed a generic `str`. Am I missing something?

src/toil/wdl/wdl_functions.py
jonathanxu18: They would all be WDLFiles there but mypy was throwing the error `src/toil/wdl/wdl_functions.py:521: error: Item "str" of "Union[str, WDLFile]" has no attribute "file_path"` since it isn't able to infer that. I don't think we can prevent the error without specifying they are WDLFiles again so maybe we could include `# type: ignore` here.

src/toil/wdl/wdl_types.py
jonathanxu18: Yeah, looking at it again I think it should probably stay as `str` actually. I was just a bit confused from this error mypy was throwing `src/toil/wdl/wdl_functions.py:525: error: Item "str" of "Union[Any, str]" has no attribute "size"` but I think it was because Union checks through all possible types each time. I think we also might need to put `type: ignore` here for same reason as above.

src/toil/wdl/wdl_functions.py
mr-c: If the WDL tests need `glob`, why don't they import it directly?

src/toil/wdl/wdl_functions.py
mr-c: Why wasn't my suggestion kept? Below we clearly see that all three parameters allowed to be of type `WDLFile`..

src/toil/wdl/wdl_functions.py
```diff
@@ -287,7 +287,7 @@ def process_single_outfile(wdl_file: WDLFile, fileStore, workDir, outDir) -> WDL
     return WDLFile(file_path=output_file, file_name=preserveThisFilename, imported=True)
 
 
-def process_outfile(f, fileStore, workDir, outDir):
+def process_outfile(f: Any, fileStore: AbstractFileStore, workDir: str, outDir: str) -> Any:
```
mr-c, line 290:
```suggestion
def process_outfile(f: Union [WDLFile, List[Any], WDLPair, Dict[Any, Any], int, str, bool, float], fileStore: AbstractFileStore, workDir: str, outDir: str) -> Union[WDLFile, List[Any], WDLPair, Dict[Any, Any], int, str, bool, float]:
```

src/toil/wdl/wdl_functions.py
```diff
@@ -313,7 +313,7 @@ def abspath_single_file(f: WDLFile, cwd: str) -> WDLFile:
     return f
 
 
-def abspath_file(f: Any, cwd: str):
+def abspath_file(f: Any, cwd: str) -> Any:
```
mr-c, line 316: I think the `Any`s here could also be more specific.

src/toil/wdl/wdl_functions.py
```diff
@@ -342,16 +342,16 @@ def abspath_file(f: Any, cwd: str):
...
-def read_single_file(f: WDLFile, tempDir, fileStore, docker=False) -> str:
+def read_single_file(f: WDLFile, tempDir: str, fileStore: AbstractFileStore, docker: bool = False) -> str:
     import os
     try:
-        fpath = fileStore.readGlobalFile(f.file_path, userPath=os.path.join(tempDir, f.file_name))
+        fpath = fileStore.readGlobalFile(f.file_path, userPath=os.path.join(tempDir, cast(str, f.file_name)))
     except:
-        fpath = os.path.join(tempDir, f.file_name)
-    return fpath
+        fpath = os.path.join(tempDir, cast(str, f.file_name))
```
mr-c, line 350: Better to add the types of the attributes to class WDLFile than to repetitively cast in multiple places.

src/toil/wdl/wdl_functions.py
```diff
@@ -342,16 +342,16 @@ def abspath_file(f: Any, cwd: str):
...
+    return cast(str, fpath)
 
 
-def read_file(f: Any, tempDir: str, fileStore: AbstractFileStore, docker: bool = False):
+def read_file(f: Any, tempDir: str, fileStore: AbstractFileStore, docker: bool = False) -> Any:
```
mr-c, line 354: Again, this `Any` can be more specific.

mr-c (changes requested):
Still too many `Any`s and castings.

The lack of recursive type checking is annoying, but going down one or two levels in your type hints makes a big difference to code quality and understability. 

For inspiration, see what we did for `cwltool` https://github.com/common-workflow-language/cwltool/blob/7094ede917c2d5b16d11f9231fe0c05260b51be6/cwltool/utils.py#L85

src/toil/wdl/wdl_functions.py
mr-c: The way to fix this is to do less variable overwriting. Switching or consolidating types? Make a new variable 😎

src/toil/wdl/wdl_functions.py
adamnovak: Maybe that giant union should be a named type variable? Why are all those types and those types specifically allowed here?

PR #3708, opened 2021-07-14
Issues/3702 time data format

Hexotical:
## Changelog Entry

Modifies documentation in conversions.py to clarify file is now for general conversion functions as well
Adds a new function hms_to_seconds that takes a string input hh:mm:ss and outputs equivalent seconds value
Adds new tester function in test_conversions.py to confirm this functionality
Edits torque.py, replaces time.strptime with hms_to_seconds to fix case where strptime would error on values greater than 24 hours

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/lib/conversions.py
+++ b/src/toil/lib/conversions.py
@@ -89,25 +89,20 @@ def bytes2human(n: SupportsInt) -> str:
 
 #General Conversions
 
-def hms_to_seconds(hms: str) -> float:
-    #60 seconds to minute, 60 minutes to hour
+def hms_to_seconds(hms: str) -> float: 
+    """
+    Parses a given time string in hours:minutes:seconds, 
+    returns an equivalent total seconds value
+    """
     vals_to_convert = hms.split(':')
     seconds = 0.0
     
-    for val in vals_to_convert: #error check for nonegative time
+    for val in vals_to_convert:
         if(float(val) < 0): 
             raise ValueError("Invalid Time")
 
-    seconds += float(vals_to_convert[0]) * 60 * 60 #hour conversion
-    seconds += float(vals_to_convert[1]) * 60 # minute conversion
-    seconds += float(vals_to_convert[2]) # second conversion
+    seconds += float(vals_to_convert[0]) * 60 * 60 
+    seconds += float(vals_to_convert[1]) * 60 
+    seconds += float(vals_to_convert[2]) 
 
     return seconds
-    """
-    Parse the given time string, in hours:minutes:seconds, and get the
-total seconds
-    of time it represents.
-
-    :param hms: Time string hin housr:minutes:seconds format.
-    :returns: Total tiome in seconds.
-    """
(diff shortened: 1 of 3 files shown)


diekhans:
I would suggest making in clear that hms is a duration, not a 24 hour clock.

maybe hms_duration_to_seconds ???

src/toil/lib/conversions.py
```diff
@@ -1,4 +1,6 @@
+"""Conversions.py is now for general conversion functions as well as of July 13 2021"""
```
mr-c, line 1: Should be an extension of the original doc string, or comment.

mr-c (changes requested):

src/toil/lib/conversions.py
Hexotical: This has been updated in new commit

src/toil/lib/conversions.py
```diff
@@ -1,4 +1,8 @@
-"""Conversion utilities for mapping memory, disk, core declarations from strings to numbers and vice versa."""
+"""
+Conversion utilities for mapping memory, disk, core declarations from strings to numbers and vice versa.
+Conversions.py is now for general conversion functions as of July 13 2021
```
adamnovak, line 3: Are you sure this sort of log-like documentation is more useful than just editing the comment to describe what is now true?

src/toil/lib/conversions.py
```diff
@@ -84,3 +88,23 @@ def bytes2human(n: SupportsInt) -> str:
...
+    """
+    Parses a given time string in hours:minutes:seconds, 
+    returns an equivalent total seconds value
+    """
+    vals_to_convert = hms.split(':')
+    seconds = 0.0
+    
+    for val in vals_to_convert:
+        if(float(val) < 0): 
+            raise ValueError("Invalid Time")
```
adamnovak, line 104: Why did you choose to check for negative components but not e.g. the wrong number of colons?

adamnovak (approved):
This looks pretty good to me. I added some comments about things I think are weird.

For the commit messages, you usually want to write them in the imperative, rather than the present tense. So `Adds initial conversion function for hours minutes seconds to seconds` would want to be `Add initial conversion function for hours minutes seconds to seconds`. But I don't think it's really necessary to try and go fix them.

src/toil/lib/conversions.py
Hexotical: No, that's more just a habit from class assignments, where they say document everything and then I don't document it in a seperate thing so i go back and see when I changed things cause some of the journals were meant to be chronological about issues, so in this scenario probably not

src/toil/lib/conversions.py
Hexotical: Mainly due to the use case, in torque.py we're provided with time in the format of "hh:mm:ss", I think that was probably more because I was approaching the specific case seen in the issue, In retrospect, given that isn't written specifically for this case a error check should be implmented for that, I'll add that in the next commit

mr-c (approved):

PR #3710, opened 2021-07-15
workaround for S3 in us-east-1

### LINKED ISSUE #3709: Make AWSJobStore work with us-east-1
As noted in https://github.com/boto/boto3/issues/125, the AWS API has a long-standing bug where a bucket cannot be told to be in the `us-east-1` region in the normal way.

Instead, when a bucket is to be placed in `us-east-1`, the whole `CreateBucketConfiguration`, which we unconditionally pass [here](https://github.com/DataBiosphere/toil/blob/70359d1fe4deed2edc60fb76a6d162bf319beec1/src/toil/jobStores/aws/jobStore.py#L748), must be omitted.

As is, Toil lets you request that a job store be created in `us-east-1`, but will fail to create the bucket with:

```
botocore.exceptions.ClientError: An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid
```

We could also revisit whether having manually specified locations for buckets makes sense anymore; I think that since the system was designed, S3 was more or less globalized.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-957)
┆Issue Number: TOIL-957

mr-c:
Fixes #3709 

- [x] Add a test

## Changelog Entry

 * Workaround for https://github.com/boto/boto3/issues/125 when the job store is S3 based and the region is "us-east-1". 

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/jobStores/aws/jobStore.py
```diff
@@ -743,9 +743,13 @@ def bucket_creation_pending(error):
                         if create:
                             logger.debug("Creating bucket '%s'.", bucket_name)
                             location = region_to_bucket_location(self.region)
```
w-gao, line 745: We probably don't need [`region_to_bucket_location()`](https://github.com/DataBiosphere/toil/blob/releases/5.4.x/src/toil/jobStores/aws/utils.py#L439) anymore, or location will never be `"us-east-1"`?

mr-c:
Huh, just running the integration tests in `us-east-1` is not sufficient to reproduce the bug. I'm guessing we need to force testing without creating the bucket first to trigger the code path in question.

mr-c:
Huzzah, finally able to reproduce the error "InvalidLocationConstraint" (in causes all the AWSJobStore tests to fail)

https://ucsc-ci.com/databiosphere/toil/-/jobs/89756

src/toil/test/__init__.py
```diff
@@ -98,10 +99,15 @@ def tearDown(self):
...
-        Use us-west-2 unless running on EC2, in which case use the region in which
-        the instance is located
+        If TOIL_AWS_ZONE is set, use that.
+        Else use "us-west-2" unless running on EC2, in which case
+        use the region in which the instance is located
         """
-        return cls._region() if running_on_ec2() else 'us-west-2'
+        return zone_to_region(
+            os.environ.get(
+                "TOIL_AWS_ZONE", cls._region() if running_on_ec2() else "us-west-2"
```
adamnovak, line 108: This isn't nested right; we're passing some things that are already regions to `zone_to_region()`.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -744,11 +744,25 @@ def bucket_creation_pending(error):
...
+                                bucket = self.s3_resource.create_bucket(
+                                    Bucket=bucket_name
+                                )
+                            else:
+                                bucket = self.s3_resource.create_bucket(
+                                    Bucket=bucket_name,
+                                    CreateBucketConfiguration={
+                                        "LocationConstraint": location
+                                    },
+                                )
```
adamnovak, line 765: Since @w-gao is also going to be creating buckets, should this be pulled out into a function (`toil.lib.aws.utils.create_s3_bucket()`)?

Then we could add a unit test for that and avoid running all the AWS job store tests again in a second region.

.gitlab-ci.yml
```diff
@@ -309,3 +309,17 @@ cactus_integration:
...
+    - pwd
+    - virtualenv -p ${MAIN_PYTHON_PKG} venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[all] packages='htcondor awscli'
+    - export TOIL_TEST_INTEGRATIVE=True
+    - export TOIL_AWS_KEYNAME=id_rsa
+    - export TOIL_AWS_ZONE=us-east-1a
+    # This reads GITLAB_SECRET_FILE_SSH_KEYS
+    - python setup_gitlab_ssh.py
+    - chmod 400 /root/.ssh/id_rsa
+    - make test tests=src/toil/test/jobStores/jobStoreTest.py::AWSJobStoreTest
+    - make test tests=src/toil/test/jobStores/jobStoreTest.py::EncryptedAWSJobStoreTest
```
adamnovak, line 325: This is going to run all the job store tests a second time in a new region, and add a new Kubernetes job for every CI run. Can't we *just* test bucket creation in that region? Maybe if we factor out the bucket creation function?

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1408,14 +1408,19 @@ def _hashTestFile(self, url: str) -> str:
...
         from toil.jobStores.aws.utils import retry_s3
-        from toil.jobStores.aws.utils import region_to_bucket_location
         from toil.jobStores.aws.jobStore import establish_boto3_session
         resource = establish_boto3_session().resource('s3', region_name=self.awsRegion())
         bucket = resource.Bucket('import-export-test-%s' % uuid.uuid4())
+        location = self.awsRegion()
 
         for attempt in retry_s3():
             with attempt:
-                bucket.create(CreateBucketConfiguration={'LocationConstraint': region_to_bucket_location(self.awsRegion())})
```
adamnovak, line 1418: This ought to just call the One True Bucket Creation Function.

adamnovak (changes requested):
This looks like it should mostly work. But I think we'd end up passing regions where we mean to pass zones, sometimes, as written.

Also, rather than running another copy of the AWS job store tests, can't we just add a single unit test to an existing CI job to test a single bucket creation utility function? We already have a slightly unmanageable number of distinct things that can go wrong in each CI run; it would be good to minimize the number of times AWS can fail on us for things that aren't our fault.

mr-c:
> Also, rather than running another copy of the AWS job store tests, can't we just add a single unit test to an existing CI job to test a single bucket creation utility function?

Sure. I ran the entire test suite on `us-east-1` to confirm that there were no other easily detected surprises with that region. It did uncover https://github.com/DataBiosphere/toil/pull/3710/files#diff-2a277145725ad8637547dd4492cc349b942cc67a4a2d003755bd2f6ddeb11ad9R444

The testing can be scaled back to just a single bucket creation utility function, yes

src/toil/test/__init__.py
mr-c: I'll remove this once there is a targeted test for just `us-east-1`.

mr-c:
@adamnovak Thanks for the review. I believe I have addressed all of your comments.

adamnovak (approved):
I think this looks good. I added a missing docstring and hooked the new tests file up to a CI job so that CI will run it.

.gitlab-ci.yml
```diff
@@ -271,6 +271,7 @@ py37_main:
     - make test tests=src/toil/test/src
     - make test tests=src/toil/test/utils
     - make test tests=src/toil/test/lib/test_ec2.py
+    - make test tests=src/toil/test/lib/aws
```
mr-c, line 274: The new test is already be run as part of the `quick_test_offline `

https://ucsc-ci.com/databiosphere/toil/-/jobs/90265/raw
```
src/toil/test/lib/aws/test_s3.py::S3Test::test_create_bucket 
-------------------------------- live log call ---------------------------------
INFO     toil.test:__init__.py:91 Setting up toil.test.lib.aws.test_s3.S3Test.test_create_bucket ...
INFO     toil.test:__init__.py:96 Tore down toil.test.lib.aws.test_s3.S3Test.test_create_bucket
PASSED 
```

adamnovak: But isn't it an online test? Shouldn't it be pulled out of there? `make test_offline` isn't supposed to need Internet access according to the Makefile.

mr-c: I agree, it is an online test. Lots of the `needs_aws`  tests run in the `quick_test_offline`, including the entire `AWSJobStoreTest` suite!

Can we merge this and fix the test distribution in a separate PR?

mr-c: I opened https://github.com/DataBiosphere/toil/issues/3717

adamnovak: We might have already had an issue for that; I guess we need a new decorator or something. We probably don't need to fix it here.

PR #3727, opened 2021-07-22
Add AWS provisioner storage system

### LINKED ISSUE #3714: Deal with large configurations for Toil managed clusters with cloud storage
Ignition configs are bigger than cloud config configs and we're starting to hit the user data limit.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-961)
┆Issue Number: TOIL-961

w-gao:
Closes #3714. 

## Changelog Entry

 * Add storage system to the AWS provisioner

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -187,20 +187,18 @@ class AbstractProvisioner(ABC):
         """
         raise NotImplementedError
 
-    def _writeGlobalFile(self, key: str, contents: bytes) -> str:
+    # TODO: think of a better name
+    def _write_file_to_cloud(self, key: str, contents: bytes) -> str:
         """
         Write a file to a physical storage system that is accessible to the
-        leader and all nodes during the life of the cluster.
-
-        Not to be confused with the `writeGlobalFile()` method in the file
-        store, this is an internal system to store files for the cluster.
-        Additional resources should be cleaned up in `self.destroyCluster()`.
+        leader and all nodes during the life of the cluster. Additional
+        resources should be cleaned up in `self.destroyCluster()`.
 
         :return: A public URL that can be used to retrieve the file.
         """
         raise NotImplementedError
 
-    def _readGlobalFile(self, key: str) -> bytes:
+    def _read_file_from_cloud(self, key: str) -> bytes:
         """
         Return the contents of the file written by `self._writeGlobalFile()`.
         """
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -187,7 +187,6 @@ class AbstractProvisioner(ABC):
         """
         raise NotImplementedError
 
-    # TODO: think of a better name
     def _write_file_to_cloud(self, key: str, contents: bytes) -> str:
         """
         Write a file to a physical storage system that is accessible to the
--- a/src/toil/provisioners/abstractProvisioner.py
+++ b/src/toil/provisioners/abstractProvisioner.py
@@ -29,6 +29,7 @@ from toil.provisioners.node import Node
 a_short_time = 5
 logger = logging.getLogger(__name__)
 
+
 class ManagedNodesNotSupportedException(RuntimeError):
     """
     Raised when attempting to add managed nodes (which autoscale up and down by
@@ -39,6 +40,7 @@ class ManagedNodesNotSupportedException(RuntimeError):
     are available from a provisioner.
     """
     pass
+
 
 @total_ordering
 class Shape(object):
@@ -52,6 +54,7 @@ class Shape(object):
     The memory and disk attributes store the number of bytes required by a job (or provided by a
     node) in RAM or on disk (SSD or HDD), respectively.
     """
+
     def __init__(self, wallTime, memory, cores, disk, preemptable):
         self.wallTime = wallTime
         self.memory = memory
@@ -138,7 +141,7 @@ class AbstractProvisioner(ABC):
 
         if self.clusterType not in self.supportedClusterTypes():
             # This isn't actually a cluster type we can do
-            ClusterTypeNotSupportedException(type(self), clusterType)
+            raise ClusterTypeNotSupportedException(type(self), clusterType)
 
         self._zone = zone
         self._nodeStorage = nodeStorage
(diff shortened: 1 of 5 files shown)


src/toil/provisioners/gceProvisioner.py
```diff
@@ -109,9 +110,11 @@ def _readCredentials(self):
         self._projectId = self.googleConnectionParams['project_id']
         self._clientEmail = self.googleConnectionParams['client_email']
         self._credentialsPath = self._googleJson
-        self._clearLeaderWorkerAuthentication() # TODO: Why are we doing this?
+        self._clearLeaderWorkerAuthentication()  # TODO: Why are we doing this?
         self._gceDriver = self._getDriver()
 
+    def _writeGlobalFile(self, key: str, contents: bytes) -> str:
```
w-gao, line 116: We don't really need cloud storage for the GCE provisioner for #3488 for two reasons: 
* the gceProvisioner doesn't support kubernetes clusters so our config size will be small, and
* It looks like the GCE user data limit is [256 KB](https://cloud.google.com/compute/docs/metadata/setting-custom-metadata#limitations)

Although, I can implement this after review for completeness, and in case we need this for things other than Ignition configs.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -184,6 +187,25 @@ def readClusterSettings(self):
         """
         raise NotImplementedError
 
+    def _writeGlobalFile(self, key: str, contents: bytes) -> str:
+        """
+        Write a file to a physical storage system that is accessible to the
+        leader and all nodes during the life of the cluster.
+
+        Not to be confused with the `writeGlobalFile()` method in the file
```
adamnovak, line 195: Wouldn't it be easier to avoid confusion by naming this concept something other than a "global file"? Why did you pick that name?

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -184,6 +187,25 @@ def readClusterSettings(self):
         """
         raise NotImplementedError
 
+    def _writeGlobalFile(self, key: str, contents: bytes) -> str:
```
adamnovak, line 190: This should be snake_case.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -184,6 +187,25 @@ def readClusterSettings(self):
...
+
+        Not to be confused with the `writeGlobalFile()` method in the file
+        store, this is an internal system to store files for the cluster.
+        Additional resources should be cleaned up in `self.destroyCluster()`.
+
+        :return: A public URL that can be used to retrieve the file.
+        """
+        raise NotImplementedError
+
+    def _readGlobalFile(self, key: str) -> bytes:
```
adamnovak, line 203: This should also be snake_case.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -140,20 +148,24 @@ def __init__(self, clusterName, clusterType, zone, nodeStorage, nodeStorageOverr
...
+        self.s3_resource = self.session.resource('s3')
+        self.s3_client = self.session.client('s3')
 
         # Call base class constructor, which will call createClusterSettings()
         # or readClusterSettings()
         super(AWSProvisioner, self).__init__(clusterName, clusterType, zone, nodeStorage, nodeStorageOverrides)
 
-
+        # After self.clusterName is set, generate a valid name for the S3 bucket associated with this cluster
+        max_bucket_name_len = 63
```
adamnovak, line 166: Maybe this should also be a constant next to `_S3_INTERNAL_BUCKET_SUFFIX`?

src/toil/test/provisioners/aws/awsProvisionerTest.py
```diff
@@ -50,6 +51,26 @@ def testAMIFinding(self):
             # Make sure we got an AMI and it looks plausible
             assert(ami.startswith('ami-'))
 
+    @needs_aws_ec2
+    def testReadWriteGlobalFiles(self):
```
adamnovak, line 55: This should probably be snake_case too.

src/toil/provisioners/gceProvisioner.py
adamnovak: I think we want the GCE provisioner to support Kubernetes clusters. Does it really not support the statically-sized ones where `toil launch-cluster` creates all the nodes?

Also, how will the AbstractProvisioner methods tell if the config needs to be stored in a file or not? We're going to need some way for the AbstractProvisioner implementations to expose a config size limit, so we know when we don't need to call `_writeGlobalFile()`. Is that going to come in the next PR?

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -184,6 +187,25 @@ def readClusterSettings(self):
...
+        """
+        Return the contents of the file written by `self._writeGlobalFile()`.
```
adamnovak, line 205: Is this something we ever need to do? Or is it only here for testing?

adamnovak (changes requested):
I think some things are not the right naming convention, and I'm worried that calling these things "global files" again just invites confusion. But it looks like it ought to work.

src/toil/provisioners/abstractProvisioner.py
w-gao: Yes, I agree. The only reason for the name is that I couldn’t think of anything else. Any suggestions for a better name?

src/toil/provisioners/gceProvisioner.py
w-gao: Oh maybe it does? I think I assumed it doesn’t because `kubernetes` isn’t one of the [`supportedClusterTypes()`](https://github.com/DataBiosphere/toil/blob/releases/5.4.x/src/toil/provisioners/gceProvisioner.py#L51). 

And I was originally planning to check that in the subclasses by overriding the `_getIgnitionUserData()` method in #3488, but I think it’ll make more sense to keep that logic in AbstractProvisioner. In that case, I can probably include the method to expose a config size limit in this PR.

src/toil/provisioners/abstractProvisioner.py
w-gao: Yeah this is only for testing now, since we don't need this for Ignition.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -184,6 +187,30 @@ def readClusterSettings(self):
         """
         raise NotImplementedError
 
+    # TODO: think of a better name
+    def _write_file_to_cloud(self, key: str, contents: bytes) -> str:
```
w-gao, line 191: Would appreciate if there are suggestions for better names for these methods.

adamnovak (approved):
This looks good to me.

PR #3740, opened 2021-08-04
Issues/3301 google jobstore batching

Hexotical:
## Changelog Entry

Implements batch function based on abstract job store in google job store, implements batching for storage in destroy function, other functions if batching doesn't break them

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/jobStores/googleJobStore.py
+++ b/src/toil/jobStores/googleJobStore.py
@@ -179,13 +179,19 @@ class GoogleJobStore(AbstractJobStore):
                   jobStoreID, '<no command>' if jobDescription.command is None else jobDescription.command)
         jobDescription.jobStoreID = jobStoreID
 
+    jobsPerBatchInsert = 200
+
     @contextmanager
     def batch(self):
         self._batchedUpdates = []
         yield
-        for jobDescription in self._batchedUpdates:
-            jobDescription.pre_update_hook()
-            self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
+
+        batches = [self._batchedUpdates[i:i + self.jobsPerBatchInsert] for i in
+                   range(0, len(self._batchedUpdates), self.jobsPerBatchInsert)]
+        for batch in batches:
+            for jobDescription in batch:
+                jobDescription.pre_update_hook()
+                self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
         self._batchedUpdates = None
         
 
--- a/src/toil/jobStores/googleJobStore.py
+++ b/src/toil/jobStores/googleJobStore.py
@@ -190,11 +190,8 @@ class GoogleJobStore(AbstractJobStore):
         
 
     def create(self, jobDescription):
-        if hasattr(self, "_batchedUpdates") and self._batchedUpdates is not None:
-            self._batchedUpdates.append(jobDescription)
-        else:
-            jobDescription.pre_update_hook()
-            self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
+        jobDescription.pre_update_hook()
+        self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
         return jobDescription
 
     @googleRetry
--- a/src/toil/jobStores/googleJobStore.py
+++ b/src/toil/jobStores/googleJobStore.py
@@ -158,9 +158,14 @@ class GoogleJobStore(AbstractJobStore):
         except exceptions.NotFound:
             # just return if not connect to physical storage. Needed for idempotency
             return
-        with self.storageClient.batch():
-            for blob in self.bucket.list_blobs():
-                blob.delete()
+        
+        try:
+            self.bucket.delete(force=True)
+        except ValueError:
+            with self.storageClient.batch():
+                for blob in self.bucket.list_blobs():
+                    blob.delete()
+            self.bucket.delete()
         self.bucket = None
         # try:
         #     self.bucket.delete(force=True)
@@ -243,8 +248,9 @@ class GoogleJobStore(AbstractJobStore):
         self._delete(jobStoreID)
 
         # best effort delete associated files
-        for blob in self.bucket.list_blobs(prefix=compat_bytes(jobStoreID)):
-            self._delete(blob.name)
+        with self.storageClient.batch():
+            for blob in self.bucket.list_blobs(prefix=compat_bytes(jobStoreID)):
+                self._delete(blob.name)
 
     def getEnv(self):
         """
--- a/src/toil/jobStores/googleJobStore.py
+++ b/src/toil/jobStores/googleJobStore.py
@@ -185,7 +185,7 @@ class GoogleJobStore(AbstractJobStore):
 
     @contextmanager
     def batch(self):
-        # not implemented, google could storage does not support batching for uploading
+        # not implemented, google could storage does not support batching for uploading or downloading
         yield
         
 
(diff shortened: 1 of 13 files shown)


src/toil/jobStores/googleJobStore.py
```diff
@@ -158,17 +158,26 @@ def destroy(self):
         except exceptions.NotFound:
             # just return if not connect to physical storage. Needed for idempotency
             return
+        
         try:
             self.bucket.delete(force=True)
-            # throws ValueError if bucket has more than 256 objects. Then we must delete manually
         except ValueError:
-            self.bucket.delete_blobs(self.bucket.list_blobs())
```
adamnovak, line 165: Is composing the batch yourself here going to be more efficient than letting `delete_blobs` do it? Is `delete_blobs` documented not to use a batch internally, or do you have some other evidence that it doesn't?

If so, I'd add a comment here pointing to the evidence that led you to choose this approach over the more obvious single `delete_blobs` call.

src/toil/jobStores/googleJobStore.py
```diff
@@ -158,17 +158,26 @@ def destroy(self):
...
             self.bucket.delete()
-            # if ^ throws a google.cloud.exceptions.Conflict, then we should have a deletion retry mechanism.
+        self.bucket = None
+        # try:
+        #     self.bucket.delete(force=True)
+        #     # throws ValueError if bucket has more than 256 objects. Then we must delete manually
+        # except ValueError:
+        #     self.bucket.delete_blobs(self.bucket.list_blobs())
+        #     self.bucket.delete()
+        #     # if ^ throws a google.cloud.exceptions.Conflict, then we should have a deletion retry mechanism.
```
adamnovak, line 176: This old code should just be removed; we can always go back and get the old version from the Git history if we want it.

src/toil/jobStores/googleJobStore.py
```diff
@@ -158,17 +158,26 @@ def destroy(self):
...
 
         # google freaks out if we call delete multiple times on the bucket obj, so after success
         # just set to None.
-        self.bucket = None
+        #self.bucket = None
```
adamnovak, line 180: Is this line no longer needed? If not, it and its comment should be removed.

src/toil/jobStores/googleJobStore.py
```diff
@@ -179,10 +188,28 @@ def assignID(self, jobDescription):
                   jobStoreID, '<no command>' if jobDescription.command is None else jobDescription.command)
         jobDescription.jobStoreID = jobStoreID
 
+    jobsPerBatchInsert = 200
```
adamnovak, line 191: We want all our new variables (and eventually all our old ones) to be `snake_case`. So this should be called `jobs_per_batch_insert` instead.

src/toil/jobStores/googleJobStore.py
```diff
@@ -158,17 +158,26 @@ def destroy(self):
...
             # just return if not connect to physical storage. Needed for idempotency
             return
+        
         try:
             self.bucket.delete(force=True)
-            # throws ValueError if bucket has more than 256 objects. Then we must delete manually
         except ValueError:
-            self.bucket.delete_blobs(self.bucket.list_blobs())
+            with self.storageClient.batch():
+                for blob in self.bucket.list_blobs():
```
adamnovak, line 166: It looks like there might be [a limit of 1000 total items](https://medium.com/analytics-vidhya/pagination-batching-restricting-result-data-with-google-cloud-storage-python-client-library-b87bc50393d) on the size of a batch. If there are more than 1000 items to delete, this might not work.

I did some research and it looks like [`delete_blobs` is documented to just call `delete_blob`, with none of its own batching](https://googleapis.dev/python/storage/latest/buckets.html?highlight=delete_blobs#google.cloud.storage.bucket.Bucket.delete_blobs).

I think overall what you should probably do is break the list of blobs up into groups of `MAX_BATCH_SIZE` or less, and for each group, make a  `self.storageClient.batch()` and insidfe it a single call to `self.storageClient.delete_blobs()` with all the blobs in the group. Then you don't need the inner loop over individual blobs.

The most important thing, though, is to explain in comments why you've decided to do it the way you're doing it. Otherwise someone is going to "simplify" away all the work you did to make it fast, and replace it with a single `delete_blobs()` again.

src/toil/jobStores/googleJobStore.py
adamnovak: And if it's just going to be a constant, it should be `JOBS_PER_BATCH_INSERT` and probably live at the top of the module under the imports.

And despite its informative variable name, it should have a comment explaining what it controls, and with some justification as to why it ought to be 200 and not, say, 20 or 2000.

src/toil/jobStores/googleJobStore.py
```diff
@@ -179,10 +188,28 @@ def assignID(self, jobDescription):
...
+    def batch(self):
+        self._batchedUpdates = []
+        yield
+
+        batches = [self._batchedUpdates[i:i + self.jobsPerBatchInsert] for i in
+                   range(0, len(self._batchedUpdates), self.jobsPerBatchInsert)]
+        for batch in batches:
+            for jobDescription in batch:
+                jobDescription.pre_update_hook()
+                self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
```
adamnovak, line 203: I don't think this quite addresses the thrust of the issue.

In the AWS job store, when we batch up the updates, we collect all the updates that need to be made and then we make [one call into the AWS Python module](https://github.com/DataBiosphere/toil/blob/de3ecf6d22dea55752d1487ac9957527fc03aa27/src/toil/jobStores/aws/jobStore.py#L315) with the whole batch. The point of this is that the AWS module can more efficiently send the updates to the cloud when it has all of them at once, versus sending each one at a time.

Here, you've implemented the machinery to defer sending the updates to Google until you have them all collected, but then you're still looping over them and sending them to the Google Python module one at a time. Like you did with the deletion, you should be using `with self.storageClient.batch():` here around each batch.

src/toil/jobStores/googleJobStore.py
```diff
@@ -179,10 +188,28 @@ def assignID(self, jobDescription):
...
 
+    jobsPerBatchInsert = 200
+
+    @contextmanager
+    def batch(self):
+        self._batchedUpdates = []
+        yield
+
+        batches = [self._batchedUpdates[i:i + self.jobsPerBatchInsert] for i in
+                   range(0, len(self._batchedUpdates), self.jobsPerBatchInsert)]
```
adamnovak, line 199: You should explain why you are breaking up the Toil-level batch into smaller sub-batches. If you were using a Google batch behind here, it would make sense, because I think those do have a maximum size. But as is you are just immediately looping over all these sub-batches and then each `JobDescription` in each of them, with no actual code running between the loops, so this re-shuffling  doesn't accomplish anything as is.

adamnovak (changes requested):
This looks like it's pretty close, but it's missing the actual Google-side batching on the updates, and on the deletes I think it needs some machinery to limit batch size to what Google will accept.

Also, the changelog entry should be rewritten to be more user-facing. The current ones from other PRs are here: https://github.com/DataBiosphere/toil/wiki/Draft-Changelog

The PR description can explain what is changing and why, but the bulleted changelog entry for the changelog would be something more like:

* Use batching to make Google job store more efficient

src/toil/jobStores/googleJobStore.py
```diff
@@ -179,10 +183,18 @@ def assignID(self, jobDescription):
...
+        yield
+        
+
     def create(self, jobDescription):
-        # TODO: we don't implement batching, but we probably should.
-        jobDescription.pre_update_hook()
-        self._writeBytes(jobDescription.jobStoreID, pickle.dumps(jobDescription, protocol=pickle.HIGHEST_PROTOCOL))
+        if hasattr(self, "_batchedUpdates") and self._batchedUpdates is not None:
+            self._batchedUpdates.append(jobDescription)
+        else:
```
adamnovak, line 195: This part can't happen anymore, right?

adamnovak (changes requested):
This looks pretty good to me now, but I think I spotted some dead code.

adamnovak (approved):
This looks good now!

PR #3742, opened 2021-08-05
issues/3741 AttributeError

ionox0:
## Changelog Entry

 * Fixed `chainedJobs` AttributeError in leader.py

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/leader.py
```diff
@@ -1035,7 +1035,9 @@ def processFinishedJob(self, batchSystemID, resultStatus, wallTime=None, exitRea
                                                                   message='The batch system left a non-empty file %s:' % batchSystemFile)
                                 if self.config.writeLogs or self.config.writeLogsGzip:
                                     batchSystemFileRoot, _ = os.path.splitext(os.path.basename(batchSystemFile))
-                                    jobNames = replacementJob.chainedJobs
+                                    jobNames = None
+                                    if hasattr(replacementJob, 'chainedJobs'):
+                                        jobNames = replacementJob.chainedJobs
```
mr-c, line 1040:
```suggestion
                                    jobNames = getattr(replacementJob, 'chainedJobs', None)
```

mr-c (dismissed):
Might as well simplify this.

I would like to know in which circumstances the attribute is missing. And a test that triggers that would also be appreciated.

adamnovak:
I think writing a test to trigger the issue would be pretty hard. The underlying problem is that we don't know what we're doing with `chainedJobs`. [Only `CheckpointJobDescription` has it to start](https://github.com/DataBiosphere/toil/blob/4f973e4e13a9758eb77633f7730cec9a4b372537/src/toil/job.py#L903-L905); for all other `JobDescription` types it's absent by default.

When a job fails (and worker output is going to a file), the worker will [set it](https://github.com/DataBiosphere/toil/blob/aa1cca8bb6d959ec1f1a340ecdc0ed8e819577c8/src/toil/worker.py#L570) and [save it back](https://github.com/DataBiosphere/toil/blob/aa1cca8bb6d959ec1f1a340ecdc0ed8e819577c8/src/toil/worker.py#L581)

The leader then [reads it back](https://github.com/DataBiosphere/toil/blob/f194cbfc206f12b8016009205428ba10db2d95b0/src/toil/leader.py#L984) and goes looking for the attribute [here](https://github.com/DataBiosphere/toil/blob/f194cbfc206f12b8016009205428ba10db2d95b0/src/toil/leader.py#L1007) and [also here](https://github.com/DataBiosphere/toil/blob/f194cbfc206f12b8016009205428ba10db2d95b0/src/toil/leader.py#L1038-L1040). The first access is protected by a check for `logJobStoreFileID`, which the worker also sets at the same time; we apparently rely on that behavior. The second access, which this PR changes, is not.

I think really we're probably seeing a stale read from the job store; the worker is replacing the job description and saying it failed, and the leader is reading the job description and getting the old version of it without the fields it expects to see. What job store are you using @ionox0? Is it a file job store on a not-very-consistent network filesystem?

We still ought to not crash (or at least not *accidentally* crash) when we get a stale read, though.

What we probably should do is move the `chainedJobs` field initialization to the base `JobDescription`, so that all job descriptions always have the attribute. Then we don't need any none or hasattr/getattr checks and just always have access to the field as a (possibly empty) list.

ionox0:
Yes I was just looking into the difference between `CheckpointJobDescription` and a regular `Job` and I see that the `Job` class has no `chainedJobs` attribute. We are using the FileJobStore on our GPFS filesystem which should be consistent, so I would be surprised if that was the cause.

ionox0:
I've moved the `chainedJobs` to the parent class, but I can look into the underlying cause of this issue. Is it not possible that a `CheckpointJobDescription` can be replaced with a `JobDescription`? That would also lead to this error no?

ionox0:
But I might need to better understand what is meant by "consistency" in this case

Which of the following does toil require:
1. That a write to a file will always finish before reads from other nodes
2. That two reads from different nodes will always return the same result

It may be that we have (2) and we are calling our filesystem consistent, but Toil requires (1) in its definition of consistency

adamnovak (approved):
I think this ought to work to fix the missing member problem.

adamnovak:
The race condition here is that the node obviously writes the job update before the worker returns failure, but the leader can see the failure result from the worker before it can see the updated job on disk. The two messages are being sent through two different distributed systems that don't talk to each other to avoid this.

What we need from the job store is that, at the real-time time when one node's write to the job store comes back successful, all real-time subsequent reads from any node will see that write (or some later write, if there is one). I think this is most like your number 2, but slightly weaker. It's fine if different nodes see different values at the same time sometimes (for example, during the time between when a node initiates a write and when that write operation returns), but after the return of a write operation, we can't allow any reads to see the value that was overwritten.

Some filesystems don't guarantee this; they'll let you see a slightly old view of the filesystem, and fix things up when you go to write yourself in the case that your operation would conflict with something that has already happen (like moving a file that has already been moved by someone else). This is fine as long as you are only communicating via the filesystem, and the filesystem is self-consistent. Your process has no way of knowing that it was lied to; the file very well might have been moved just before your process went to move it, or your process might have actually done its read before the deletion happened. But since Toil processes also communicate through the scheduler, they are able to see through the performance-enhancing lies the filesystem is telling. We don't just need self-consistency on the filesystem; we need ordering relationships between events to make sense for both the filesystem and the scheduler together. In practice that means we usually want a filesystem that is consistent with real time.

PR #3746, opened 2021-08-12
Multi-zone balancing within regions for AWS autoscaling groups

### LINKED ISSUE #3527: Balance across spot pools in multiple AWS zones in Kubernetes ASGs
This is related to #3415. Instead of only balancing across pools in a single zone like #3490 does, we also want to be able to balance across multiple zones.
┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-843)
┆Epic: Scalable kubernetes 
┆Issue Number: TOIL-843

adamnovak:
This PR fixes #3527 and several of its subtasks on the Jira side.

We now ask AWS for all the subnets that look like the subnet the leader is in, and then use all those subnets as options for the autoscaling groups we use for Kubernetes autoscaling. This means that the instances deployed can be in any availability zone in the region; Amazon is supposed to place them to minimize the likelihood of being preempted.

Non-ASG-managed non-spot nodes should still be in the same subnet as the leader, and the non-ASG spot nodes have an in-Toil spot bid optimizer that I'm not sure is actually under test.

One caveat here is that I haven't added a way to turn this off, and network traffic between AZs isn't free. Toil should be sending all the bulk traffic to and from S3, but if you are running a workflow that involves nodes chatting to each other through some kind of database (like maybe Cactus @glennhickey), and using an auto-scaling Kubernetes cluster, this could increase costs.

## Changelog Entry

 * AWS autoscaling groups for Toil Kubernetes clusters now balance across the zones in an AWS region

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -126,28 +132,88 @@ def awsFilterImpairedNodes(nodes, ec2):
...
     pass
 
+class AWSConnectionManager:
+    """
+    Class that represents a connection to AWS. Caches Boto 3 and Boto 2 objects
+    by region.
+
+    Access to any kind of item goes through the particular method for the thing
+    you want (session, resource, service, Boto2 Context), and then you pass the
+    zone you want to work in, and possibly the type of thing you want, as arguments.
```
adamnovak, line 142: I keyed this all by zone in the external API, but it turns out all AWS connections are scoped to region and not zone. Maybe we really need a `_region` on the provisioner so we can have it handy to use the API here?

glennhickey:
Cactus no longer uses databases.  All interprocess communication is strictly through the jobstore now.

adamnovak:
This failed `KubernetesBatchSystemTest.testRunJobs`, which is against the shared Kubernetes and not anything deployed on AWS by Toil. It looks like the job pods didn't start in time for the test; I am going to retry it.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -169,29 +241,38 @@ def __init__(self, clusterName, clusterType, zone, nodeStorage, nodeStorageOverr
...
         # The cluster name is the same as the name of the leader.
         self.clusterName = str(instance.tags["Name"])
-        self._buildContext()
-        # This is where we will put the workers.
-        # See also: self._vpcSubnet
-        self._subnetID = instance.subnet_id
+        # Determine what subnet we, the leader, are in
+        self._leader_subnet = instance.subnet_id
+        # Determine where to deploy workers.
+        self._worker_subnets_by_zone = self._get_good_subnets_like(self._leader_subnet)
```
DailyDreaming, line 274: I support good naming conventions.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -137,29 +145,93 @@ def awsFilterImpairedNodes(nodes, ec2):
 class InvalidClusterStateException(Exception):
     pass
 
+class AWSConnectionManager:
```
DailyDreaming, line 148: I really like this.  I'll have to figure out how to work it into some of the simpledb refactor PR.  One thing worth noting is that s3 claims that you have to start a new session each time in order to be thread-safe, so if we're parallelizing our downloads, we should probably account for that:

https://boto3.amazonaws.com/v1/documentation/api/1.14.31/guide/session.html#multithreading-or-multiprocessing-with-sessions

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -362,6 +450,130 @@ def launchCluster(self,
...
+        }, {
+            'Name': 'state',
+            'Values': ['available']
+        }]
+
+        # Fill in this collection
+        by_az = {}
+
+        # Go get all the subnets. There's no way to page manually here so it
+        # must page automatically.
```
DailyDreaming, line 512: It would be nice if AWS did that for all of its API calls.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -362,6 +450,130 @@ def launchCluster(self,
...
+            subnet_acls = set(self._get_subnet_acls(subnet.subnet_id))
+            if subnet_acls != acls:
+                # Reject this subnet because it has different ACLs
+                logger.debug('Subnet %s is a lot like subnet %s but has ACLs of %s instead of %s; skipping',
+                             subnet.subnet_id, base_subnet_id, subnet_acls, acls)
+                continue
+
+            if subnet.availability_zone not in by_az:
+                # Make sure we have a bucket of subnets for this AZ
+                by_az[subnet.availability_zone] = []
```
DailyDreaming, line 526: Not necessary at all, but DefaultDict is nice for this.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -362,6 +450,130 @@ def launchCluster(self,
...
+        # Compose a filter that selects the default subnet in the AZ
+        filters = [{
+            'Name': 'association.subnet-id',
+            'Values': [subnet]
+        }]
+
+        # TODO: Can't we use the resource's network_acls.filter(Filters=)?
+        return [item['NetworkAclId'] for item in self._pager(ec2.describe_network_acls,
+                                                             'NetworkAcls',
+                                                             Filters=filters)]
```
DailyDreaming, line 550: Is it possible for this to grow too large?  It's a habit now for me to make returning unbounded lists into generators that hand back one element (or a specified batch size) at a time for processing.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -362,6 +450,130 @@ def launchCluster(self,
...
+        }, {
+            'Name': 'availability-zone',
+            'Values': [zone]
+        }]
+
+        for subnet in self.aws.resource(zone, 'ec2').subnets.filter(Filters=filters):
+            # There should only be one result, so when we see it, return it
+            return subnet.subnet_id
+        # If we don't find a subnet, something is wrong. Maybe this zone was
+        # added after your account?
```
DailyDreaming, line 572: Can this comment/suggestion be added to the error message?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -702,11 +934,31 @@ def _toNameSpace(self) -> str:
...
 
+    def _namespace_name(self, name: str) -> str:
+        """
+        Given a name for a thing, add our cluster name to it in a way that
+        results in an acceptable name for something on AWS.
+        """
+
+        # This logic is a bit weird, but it's what Boto2Context used to use.
+        # Drop the leading / from the absolute-path-style "namespace" name and
+        # then encode underscores and slashes.
```
DailyDreaming, line 945: If it works, it works.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1236,23 +1552,12 @@ def _getRoleInstanceProfileNames(self, role_name: str) -> List[str]:
...
-                                                                           MaxItems=200,
-                                                                           Marker=response['Marker'])
-            else:
-                # No more pages
-                break
+        # Grab the connection we need to use for this operation.
+        iam = self.aws.client(self._region, 'iam')
 
-        return allProfiles
+        return [item['InstanceProfileName'] for item in self._pager(iam.list_instance_profiles_for_role,
```
DailyDreaming, line 1558: Same concern as above, though I guess we've never run into problems before, so maybe passing the whole list is always fine.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1376,29 +1659,82 @@ def kubernetes_policy(self) -> dict:
...
+            logger.debug('Created new IAM role')
+        except BotoServerError as e:
+            if e.status == 409 and e.error_code == 'EntityAlreadyExists':
+                logger.debug('IAM role already exists. Reusing.')
+                pass
+            else:
+                raise
+
+        # Delete superfluous policies
+        policy_names = set(iam.list_role_policies(aws_role_name).policy_names)
```
DailyDreaming, line 1692: Hmmm... I remember Asha added a way to supply policies from the user side.  Is it possible for this to unexpectedly prune a policy that, if supplied manually, might break on something else?  I believe it was "--awsEc2ProfileArn": https://github.com/DataBiosphere/toil/pull/2703/files

src/toil/test/provisioners/aws/awsProvisionerTest.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, methodName):
         self.numWorkers = ['2']
         self.numSamples = 2
         self.spotBid = 0.15
-        self.zone = get_current_aws_zone()
+        self.zone = get_best_aws_zone()
```
DailyDreaming, line 87: Toil zone is best zone.

DailyDreaming (commented):
Only major concern is the policy pruning.  The rest are nits and this looks great to me.

DailyDreaming (approved):

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1376,29 +1659,82 @@ def kubernetes_policy(self) -> dict:
...
+        for policy_name in policy_names.difference(set(list(policies.keys()))):
+            iam.delete_role_policy(aws_role_name, policy_name)
```
DailyDreaming, line 1694: Does this delete only duplicate policies or does it delete potentially policies it needs?  I'm only concerned that this might interfere with a user supplied policy a la `--awsEc2ProfileArn` if it is the second: https://github.com/DataBiosphere/toil/pull/2703

src/toil/provisioners/aws/awsProvisioner.py
adamnovak: I don't think people will put thousands of ACLs on a single subnet; there are only so many services in the world you could run, and each ACL is meant to hold multiple rules.

We could make this a generator, but how would that interact with retries? I don't think our default retry decorators would let us e.g. resume iteration from the right place.

src/toil/provisioners/aws/awsProvisioner.py
adamnovak: There's probably exactly one item in this list all the time; we make the roles we use, and when the role is made it gets one instance profile added to it.

adamnovak: You can specify your own `--awsEc2ProfileArn`, and maybe have it attached to a role that has a lot of instance profiles on it? But that seems like a pathological configuration we can support when a user actually has it.

src/toil/provisioners/aws/awsProvisioner.py
adamnovak: If someone specifies a --awsEc2ProfileArn, `_createProfileArn()` is never called, and it won't call `_setup_iam_ec2_role()`. This is supposed to run just on properly-cluster-namespaced, toil-managed roles, where Toil is allowed to fiddle with all the policies.

Now, I copied most of this over from the old Boto2Context, which used to be CGCloud and had ambitions of idempotency, so it has all this code to do things like fix up an existing role despite how you shouldn't ever `launch-cluster` with the same name as an existing cluster. We might want to simplify it away and just bail out if the role already exists.

PR #3767, opened 2021-09-01
Attempt to use absolute local paths when exportFile/importFile do not detect a schema.

DailyDreaming:
When importing relative file paths without a schema, it can be unclear how Toil is attempting to resolve those paths.  This forces the base importFile method to add a local "file://" schema when no schema is found, to adjust it to an absolute path, and include an appropriate error message (early) helpful to users if the file does not exist.

A short blurb in the docs has also been updated on use.

## Changelog Entry

 * Attempt to use absolute local paths when ImportFile does not detect a schema.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -20,12 +20,15 @@ import sys
 import tempfile
 import time
 import uuid
+
+from urllib.parse import urlparse
 from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser, _ArgumentGroup
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import requests
 
 from toil import logProcessContext, lookupEnvVar
+from toil.fileStores import FileID
 from toil.batchSystems.options import (add_all_batchsystem_options,
                                        set_batchsystem_config_defaults,
                                        set_batchsystem_options)
@@ -1014,7 +1017,10 @@ class Toil:
             logger.debug('Injecting user script %s into batch system.', userScriptResource)
             self._batchSystem.setUserScript(userScriptResource)
 
-    def importFile(self, srcUrl, sharedFileName=None, symlink=False):
+    def importFile(self,
+                   srcUrl: str,
+                   sharedFileName: Optional[str] = None,
+                   symlink: bool = False) -> Optional[Union[FileID, str]]:
         """
         Imports the file at the given URL into job store.
 
@@ -1022,9 +1028,10 @@ class Toil:
         full description
         """
         self._assertContextManagerUsed()
+        srcUrl = self.sanity_check_uri_and_format(srcUrl)
         return self._jobStore.importFile(srcUrl, sharedFileName=sharedFileName, symlink=symlink)
 
-    def exportFile(self, jobStoreFileID, dstUrl):
+    def exportFile(self, jobStoreFileID: Union[FileID, str], dstUrl: str) -> None:
         """
         Exports file to destination pointed at by the destination URL.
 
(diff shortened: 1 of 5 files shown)


adamnovak (changes requested):
We should test that `importFile` actually does work with paths now, if we want to follow our checklist.

DailyDreaming:
@adamnovak Added.  I refactored a couple of things too.

src/toil/common.py
```diff
@@ -1014,26 +1017,55 @@ def _setupAutoDeployment(self, userScript=None):
...
 
         See :func:`toil.jobStores.abstractJobStore.AbstractJobStore.exportFile` for a
         full description
         """
         self._assertContextManagerUsed()
+        dstUrl = self.sanity_check_uri_and_format(dstUrl)
         self._jobStore.exportFile(jobStoreFileID, dstUrl)
 
+    @staticmethod
+    def sanity_check_uri_and_format(uri: str, check_existence: bool = False) -> str:
```
adamnovak, line 1046: I'm going to come out against "sanity" checks; I don't think the term is appropriate and we don't want to scare people who aren't, strictly speaking, sane away from our project.

That's also not what this function does anyway; it returns a string. maybe it should be called `normalize_url` or `ensure_url` or `urlify_paths` or something.

src/toil/common.py
```diff
@@ -1014,26 +1017,55 @@ def _setupAutoDeployment(self, userScript=None):
...
+        if urlparse(uri).scheme == 'file':
+            uri = urlparse(uri).path  # this should strip off the local file schema; it will be added back
+
+        # account for the schema-less case, which should be coerced to a local absolute path
+        if urlparse(uri).scheme == '':
+            abs_path = os.path.abspath(uri)
+            if not os.path.exists(abs_path) and check_existence:
+                raise FileNotFoundError(
+                    f'Import failed for: "{uri}"\n'
+                    f'No schema (http://, s3://, file://, etc.) for path was included.\n'
```
adamnovak, line 1061: Isn't it "scheme" and not "schema"?

src/toil/common.py
```diff
@@ -1014,26 +1017,55 @@ def _setupAutoDeployment(self, userScript=None):
...
+                    f'Attempting to import as an absolute local file path: "{abs_path}"\n'
+                    f'Relative to directory: "{os.getcwd()}".\n'
+                    f'Please try to include the schema, and adjust the path as necessary, '
+                    f'using absolute paths where possible.')
```
adamnovak, line 1065: Is making the path absolute likely to help, or really any better?

src/toil/test/src/importExportFileTest.py
```diff
@@ -27,113 +27,122 @@
...
+            with self.subTest('Testing permissions are preserved for local importFile/exportFile'):
+                for executable in True, False:
+                    file_path = self.create_file(content='Hello', executable=executable)
+                    initial_permissions = os.stat(file_path).st_mode & stat.S_IXUSR
+                    file_id = toil.importFile(f'file://{file_path}')
+                    toil.exportFile(file_id, f'file://{self.output_file_path}')
+                    current_permissions = os.stat(self.output_file_path).st_mode & stat.S_IXUSR
+                    assert initial_permissions == current_permissions
+
+            with self.subTest('Testing relative paths without the file:// schema.'):
```
adamnovak, line 112: I didn't know `subTest()` was a thing!

src/toil/common.py
adamnovak: Also, didn't we just drop the `file:` scheme that may have been included?

adamnovak (changes requested):
This looks mostly good, but now I don't like the new function name.

I'm also having some second thoughts about the error message, though I suppose this one works. But maybe it should be more along the lines of "Could not find local file {original_uri}", or maybe "Could not find local file {absolute_path} to import {original_uri}", if the URI wasn't an absolute path. There's no real reason the user *should* use absolute paths, or even include the `file:` scheme, if we don't really need it. We just need to make Toil's interpretation of the input clear so they can tell if they aren't in the directory they think they are.

src/toil/common.py
DailyDreaming: Well, I suppose `normalize_url` has a certain ring to it, even though "normalization" has always sounded like what would immediately follow a oppressively political fellow aiming a very powerful laser at someone they disagree with.

src/toil/common.py
DailyDreaming: Hmmm... I've been using them interchangeably.  I'll switch over to scheme to be consistent with urlparse.  There's an interesting suggestion here that schema's modern usage is largely Kant's fault: https://english.stackexchange.com/questions/40702/difference-between-scheme-and-schema

And yes, the `file:` scheme was dropped so that the same scheme-less conditional would modify it to be an absolute path and then add `file://` back.

src/toil/common.py
DailyDreaming: It is often expedient and often works.  It is also, as I recall, the first suggestion you made to Benedict when he brought up the issue.

src/toil/test/src/importExportFileTest.py
DailyDreaming: Yes.  I am a fan.

adamnovak (approved):
I tweaked one of the docstrings, but I think this is good now.

PR #3778, opened 2021-09-10
Prevent exception from being raised when modifying dir permissions for clean up.

DailyDreaming:
I don't think the workflow should break if everything was successful but we can't clean up.

## Changelog Entry

 * Prevent exception from being raised when modifying dir permissions for clean up.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/worker.py
+++ b/src/toil/worker.py
@@ -611,7 +611,8 @@ def workerScript(jobStore: AbstractJobStore, config: Config, jobName: str, jobSt
             # Just chmod it for rwx for user. This can't work anyway if it isn't ours.
             try:
                 os.chmod(os.path.dirname(path),  stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
-            except PermissionError:
+            except PermissionError as e:
+                logger.error('Could not set permissions on %s to allow cleanup of %s: %s', os.path.dirname(path), path, e)
                 pass
         shutil.rmtree(localWorkerTempDir, onerror=make_parent_writable)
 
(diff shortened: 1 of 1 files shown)


src/toil/worker.py
```diff
@@ -609,10 +609,13 @@ def make_parent_writable(func: Callable[[str], Any], path: str, _: Any) -> None:
             up after itself.
             """
             # Just chmod it for rwx for user. This can't work anyway if it isn't ours.
-            os.chmod(os.path.dirname(path),  stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+            try:
+                os.chmod(os.path.dirname(path),  stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+            except PermissionError:
+                pass
```
mr-c, line 615: Can there at least be a log entry?

adamnovak, line 615: Yeah, @mr-c is right; this should `logger.error` a complaint.

adamnovak (changes requested):
We still want to know when cleanup doesn't work, but you're right it probably shouldn't fail the workflow.

Now someone will complain that a bunch of ostensibly successful workflows filled their disk. But the root cause will be that they're running in a weird environment with a recalcitrant filesystem and they need to do manual cleanup anyway, and they wouldn't have been able to swap the filesystem if the workflows had failed to complain about it. So it is probably better this way.

adamnovak (approved):
I just went and added the log entry I wanted.

mr-c (approved):

PR #3932, opened 2021-11-18
Issue/3917 toil kill find pid log (#2)

kannon92:
## Changelog Entry
1) Fix toil kill to find shared pid.log file.  
2) Add a toil unit test for toil kill.  
## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/test/utils/toilKillTest.py
+++ b/src/toil/test/utils/toilKillTest.py
@@ -1,0 +1,59 @@
+# Copyright (C) 2018 Regents of the University of California
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+import unittest
+import os
+import sys
+import time
+import shutil
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from toil import subprocess
+from toil.test import ToilTest, needs_cwl
+
+class ToilKillTest(ToilTest):
+    """A set of test cases for "toil kill"."""
+
+    def setUp(self):
+        """Shared test variables."""
+        self.cwl = os.path.abspath('src/toil/test/utils/ABCWorkflowDebug/sleep.cwl')
+        self.yaml = os.path.abspath('src/toil/test/utils/ABCWorkflowDebug/sleep.yaml')
+        self.jobstore = os.path.join(os.getcwd(), 'testkill')
+
+    def tearDown(self):
+        """Default tearDown for unittest."""
+        if os.path.exists(self.jobstore):
+            shutil.rmtree(self.jobstore)
+        if os.path.exists('tmp'):
+            shutil.rmtree('tmp')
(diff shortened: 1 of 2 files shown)


src/toil/test/utils/toilKillTest.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (C) 2018 Regents of the University of California
```
mr-c, line 1: @kannon92 Please change this to you or your copyright holder

kannon92: Oh.  That was silly of me.  I copied this from an older example and didn't update the header.  I think the update should suffice.

src/toil/test/utils/toilKillTest.py
```diff
@@ -0,0 +1,59 @@
...
+    def tearDown(self):
+        """Default tearDown for unittest."""
+        if os.path.exists(self.jobstore):
+            shutil.rmtree(self.jobstore)
+        if os.path.exists('tmp'):
+            shutil.rmtree('tmp')
+        unittest.TestCase.tearDown(self)
+
+    @needs_cwl
+    def testCWLToilKill(self):
```
adamnovak, line 45: I think this function ought to be named in `snake_case` according to our current rules.

src/toil/utils/toilKill.py
```diff
@@ -41,11 +41,12 @@ def main() -> None:
...
     # otherwise, kill the pid recorded in the jobstore
     else:
-        pid_log = os.path.join(os.path.abspath(config.jobStore), 'pid.log')
+        jobStore = Toil.resumeJobStore(config.jobStore)
+        pid_log = os.path.join(jobStore.sharedFilesDir, 'pid.log')
         with open(pid_log) as f:
             pid2kill = f.read().strip()
         try:
-            os.kill(int(pid2kill), signal.SIGKILL)
+            os.kill(int(pid2kill), signal.SIGTERM)
```
adamnovak, line 49: It would be nice if we could pick the signal from the command line, or send a kill if it's still running after a while, or if we could hold off on declaring the process "successfully terminated" until it actually dies. But those might all be separate feature requests; it's probably good enough to just make `toil kill` work at all again.

adamnovak (changes requested):
This is mostly good, but I think one of the new methods is supposed to be in `snake_case`, and I don't have access to tweak it myself.

src/toil/test/utils/toilKillTest.py
kannon92: Good call.  Too much time spent in javascript lately.  

I will push that up.

kannon92: I'm new to github but I believe I addressed this in a new commit.

src/toil/utils/toilKill.py
adamnovak: The simplest thing to start would be a CLI option to set the signal type, with SIGTERM being the default. Kind of like normal `kill` has.

Being super smart about checking if it stopped in a reasonable amount of time would be a harder design problem.

But neither needs to be part of *this* PR.

src/toil/utils/toilKill.py
```diff
@@ -41,11 +41,12 @@ def main() -> None:
         logger.info("All jobs SHOULD have been killed")
     # otherwise, kill the pid recorded in the jobstore
     else:
-        pid_log = os.path.join(os.path.abspath(config.jobStore), 'pid.log')
+        jobStore = Toil.resumeJobStore(config.jobStore)
+        pid_log = os.path.join(jobStore.sharedFilesDir, 'pid.log')
```
mr-c, line 45: https://ucsc-ci.com/databiosphere/toil/-/jobs/3277#L1049

> `src/toil/utils/toilKill.py:45:32: error: "AbstractJobStore" has no attribute
"sharedFilesDir"  [attr-defined]`

adamnovak (approved):
I'm going to approve and also close this; I glommed it into #3941 to make the minor edits it needs to actually merge.

PR #3956, opened 2021-12-02
AWS Batch Batch System

### LINKED ISSUE #3942: Implement AWS Batch batch system
If AGC isn't going to provide a TES endpoint, but instead an AWS Batch endpoint to run jobs against, then we will need an AWSBatchBatchSystem in Toil that can run jobs directly against the proprietary AWS Batch API, instead of just the standard TES API.

My guess is that it **should** be quite similar to the TES batch system, but we'll need to copy the code and swap out py-tes calls with whatever part of boto3 wraps AWS Batch. We might also need some custom-ish IAM stuff to really make sure that we're running jobs with the right rights to use an AWS job store if we have one.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1100)
┆epic: MVP - Implement Toil engine prototype
┆friendlyId: TOIL-1100

adamnovak:
This will fix #3942 (when complete), by implementing an `aws_batch` batch system.

We assume that you already have an AWS Batch queue to use, and probably that you have an IAM role set up that grants access to S3 so you can use a job store there. That's going to be a bit of a pain to use manually, but when integrating into AGC we probably will be able to get those from AGC. Hopefully. I haven't checked yet.

This doesn't support the weirdly restrictive Fargate-based AWS Batch, only the EC2-based AWS Batch. And I still need to figure out where EC2-based AWS Batch is likely to hide its big local disks in the filesystem.

AWS Batch doesn't actually let you request local scratch disk space. I'm not sure how AGC thinks it is going to reliably run CWL workflows that need nontrivial disk space without this feature. We will have to check with the AGC folks.

## Changelog Entry

 * Add a new `aws_batch` batch system to run jobs on AWS Batch.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak:
After talking to the AGC folks, we think that:

* Preemptible/non-preemptible, if we ever expose it, would be a second queue. For now we can go with one queue and let the whole workflow be preemptible or non-preemptible regardless of what the Toil task flags say. Wehn we do proper dupport for preemptability we would do something in AGC and its CDK template to stamp out 2 queues, and rake in a separate queue for jobs that can be preempted that has access to a second spot-based CE as a second priority after the main CE.
* Disk space is not really Toil's problem. The AGC nodes have their root disks dynamically resized at runtime as they appear to fill, by a daemon that watches them. So we might need to do some work on e.g. the CachingFileStore if we find that it gets confused by this in testing, but we should mostly be protected from running out of space without having to do anything.

So all that's left here is filling in the docs for all the new options and environment variables.

adamnovak:
Also satisfying MyPy.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,519 @@
...
+            if len(jobs) == 0:
+                # Job no longer exists at all
+                return
+            job = jobs[0]
+            if job.get('status') and job['status'] in ['SUCCEEDED', 'FAILED']:
+                # The job has stopped
+                return
+            # Otherwise the job is still going. Wait for it to stop.
+            logger.info('Waiting for killed job %s to stop', self.aws_id_to_bs_id.get(aws_id, aws_id))
+            time.sleep(2)
```
DailyDreaming, line 391: Should this have a timeout?  Also... in the case of long running processes (i.e. week+), we may want to include additional retries beyond the default, since one may get three errors one day and three more the next, etc. if it's intended to run that long.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,519 @@
...
+                'jobDefinitionName': 'toil-' + str(uuid.uuid4()),
+                'type': 'container',
+                'containerProperties': {
+                    'image': self.docker_image,
+                    # Unlike the Kubernetes batch system we always mount the Toil
+                    # workDir onto the host. Hopefully it has its ephemeral disks mounted there.
+                    # TODO: Where do the default batch AMIs mount their ephemeral disks, if anywhere?
+                    'volumes': [{'name': 'workdir', 'host': {'sourcePath': '/var/lib/toil'}}],
+                    'mountPoints': [{'containerPath': self.worker_work_dir, 'sourceVolume': 'workdir'}],
+                    # Requirements will always be overridden but mist be present anyway
```
DailyDreaming, line 410: `must`

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,519 @@
...
+                    'resourceRequirements': [
+                        {'type': 'MEMORY', 'value': str(math.ceil(max(4, to_mib(self.config.defaultMemory))))},
```
DailyDreaming, line 412: 4mb feels low, even though this reassuringly says this default's day will never come to pass.  Maybe 64mb?

src/toil/lib/aws/__init__.py
```diff
@@ -11,22 +11,258 @@
...
-    import boto3
-    return boto3.client(*args, **kwargs)
+logger = logging.getLogger(__name__)
 
+try:
+    # This file contains some stuff that needs boto/boto3 to be installed in
+    # order to be defined. But this __init__ file and files under it also
+    # contain some stuff that we need to be able to import when boto isn't
+    # installed. So everything that does need those modules gets defined in
+    # this try.
```
DailyDreaming, line 34: I guess I'm voting that this be put into a non-init file to avoid giant try/excepts like this.

src/toil/lib/aws/__init__.py
```diff
@@ -11,22 +11,258 @@
...
+    running in.
+
+    Finally, if we have boto2, and a default region is configured in Boto 2,
+    chooses a zone in that region.
+
+    Returns None if no method can produce a zone to use.
+    """
+    return try_providers([get_aws_zone_from_environment,
+                          get_aws_zone_from_metadata,
+                          get_aws_zone_from_boto])
```
DailyDreaming, line 265: Hmmm... why this over something like:

```
return get_aws_zone_from_environment() or get_aws_zone_from_metadata() or get_aws_zone_from_boto() or None
```

DailyDreaming (commented):
Looks good.  Mostly minor comments.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,417 @@
...
+
+from toil import applianceSelf
+from toil.batchSystems.abstractBatchSystem import (EXIT_STATUS_UNAVAILABLE_VALUE,
+                                                   BatchJobExitReason,
+                                                   UpdatedBatchJobInfo)
+from toil.batchSystems.cleanup_support import BatchSystemCleanupSupport
+from toil.common import Config, Toil
+from toil.job import JobDescription
+from toil.lib.aws import client
+from toil.lib.conversions import to_mib, from_mib
```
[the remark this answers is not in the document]
adamnovak: These aren't new functions, and they're already called a bunch in the Mesos code, vs. `convert_units` which appears once outside the conversions module, in `lsfHelper.py`.

There's very little call to e.g. convert from MB to KiB in the general codebase; I will try renaming the specialized conversion functions to be harder to misuse, and make them use the same backend.

src/toil/test/__init__.py
```diff
@@ -322,6 +322,19 @@ def get_temp_file(suffix: str = "", rootDir: Optional[str] = None) -> str:
...
+    Use as a decorator before test classes or methods to run only if the given
+    environment variable is set.
+
+    Can include a comment saying what the variable should be set to.
+    """
+    def decorator(test_item: MT) -> MT:
+        if not os.getenv(var_name):
+            return unittest.skip(f"Set {var_name}{' to ' + comment if comment else ''} to include this test.")(test_item)
+        return test_item
+    return decorator
```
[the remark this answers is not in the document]
adamnovak: I liked it better this way because then it matches all the other functions here, and we *could* decorate with it. But I will change it.

docs/appendices/environment_vars.rst
```diff
@@ -107,6 +107,20 @@ There are several environment variables that affect the way Toil runs.
 |                                  | deleted until all associated nodes have been       |
 |                                  | terminated.                                        |
 +----------------------------------+----------------------------------------------------+
+| TOIL_AWS_BATCH_REGION            | Region to use when using the AWS Batch batch       |
```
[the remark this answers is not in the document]
adamnovak: I think so? We don't have a `TOIL_AWS_REGION`, and Batch here doesn't need a *particular* zone, only a region. It wouldn't make sense to me to make people say `TOIL_AWS_BATCH_QUEUE=myqueue TOIL_AWS_ZONE=us-east-2c` when the queue really is in all of `us-east-2`.

We *can* pick up the region from `TOIL_AWS_ZONE` if it is set, but I didn't want to make people pick a zone when it doesn't matter.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,519 @@
...
+        kept_chars = [c for c in input_name if c.isalnum() or c == '-' or c == '_']
+        if len(kept_chars) == 0 or not kept_chars[0].isalnum():
+            # Make sure we start with something alphanumeric
+            kept_chars = ['j'] + kept_chars
+        # Keep no more than the limit of them
+        kept_chars = kept_chars[:128]
+        # And re-compose them into a string
+        return ''.join(kept_chars)
+
+    def __get_runtime(self, job_detail: Dict[str, Any]) -> Optional[float]:
```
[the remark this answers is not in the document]
adamnovak: I don't think we're going to be extending this class, so name mangling as a tool to avoid collisions doesn't really help us here.

But I do mean this method to be at class scope. I don't *want* this method to be re-usable elsewhere unless it is officially promoted to part of the API (and given a test, etc.). At that point we can rename it with find/replace because... it won't be mentioned anywhere else in the codebase. I could use some kind of custom decorator that audits the call stack to make sure nobody is calling it who shouldn't be, but the two underscores is standard, easier, and less obnoxious if someone *does* have to hack around it for a good reason.

I think when I was refactoring the leader I got mad about how the particular behavior of single-underscore-prefixed stuff had grown to matter in other parts of the codebase. I wanted to be able to change the semantics of or drop a thing that started with an underscore, but it turned out it was being used elsewhere or under test in the unit tests. The `AbstractJobStore` even has e.g. `_read_from_url()` as part of its defined API, which concrete job store classes are required to implement.

Maybe if I was using a proper IDE that knew about all the project files at once, I wouldn't trip over these usages because the IDE would be surfacing them for me?

Anyway, now when I add something that I don't want to grow secret connections to other parts of the codebase over time, I put two underscores on it.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,519 @@
...
+
+                    if job_detail['status'] == 'FAILED' and 'statusReason' in job_detail:
+                        # AWS knows why the job failed, so log the error
+                        logger.error('Job %s failed because: %s', bs_id, job_detail['statusReason'])
+
+                    # Compose a result
+                    result = UpdatedBatchJobInfo(jobID=bs_id, exitStatus=exit_code, wallTime=runtime, exitReason=exit_reason)
+
+                    # No more iteration needed, we found a result.
+                    break
```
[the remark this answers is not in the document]
adamnovak: It looks like I missed breaking one of the for loops here. I'll try and refactor this to make it clearer.

src/toil/batchSystems/awsBatch.py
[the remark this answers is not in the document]
adamnovak: PEP-8 recommends using __ only in base classes, to keep subclasses out of your hair, and that _ ought to be enough to discourage people in other parts of your codebase.

Maybe I will change to _ and put in the docstrings that nobody should call these? And maybe we need to make it a review check that nobody is calling _-prefixed things they shouldn't be?

src/toil/batchSystems/awsBatch.py
adamnovak: This is only used on killed jobs, so long-running jobs aren't a concern here.

I also don't think it can have a timeout, unless we add a system to remember jobs that are supposed to be dead, and filter them from our view of the jobs that exist/are running. `killBatchJobs` isn't allowed to return until the jobs it was sent to kill appear dead, and if we don't wait until they are well and truly dead on the AWS side we'll have to fake it.

If we want to protect the user from a stuck job on the AWS side, we could determine a time that it's reasonable to wait on AWS to kill the job, and crash Toil if it is exceeded.

src/toil/batchSystems/awsBatch.py
adamnovak: This isn't a default; it's a protection against sending the API out-of-spec requests. They specify 4 MiB as the minimum memory request.

If we want to pad the job memory requests for overhead, we need another mechanism for that.

I'll pull this out into a constant to be more clear what it is for.

src/toil/lib/aws/__init__.py
adamnovak: I'll move these to another file.

src/toil/lib/aws/__init__.py
adamnovak: Yeah I think I can actually just use `or` in both places I am using this.

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,537 @@
...
+                for job_detail in self._describe_jobs_in_batches():
+                    if job_detail.get('status') in ['SUCCEEDED', 'FAILED']:
+                        # This job is done!
+                        aws_id = job_detail['jobId']
+                        bs_id = self.aws_id_to_bs_id[aws_id]
+
+                        # Acknowledge it
+                        acknowledged.append((aws_id, bs_id))
+
+                        if job_detail['jobId'] in self.killed_job_aws_ids:
```
w-gao, line 310: Nit:
```suggestion
                        if aws_id in self.killed_job_aws_ids:
```

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,537 @@
+# Copyright (C) 2015-2021 Regents of the University of California
```
w-gao, line 1: Nit: `2022`

src/toil/batchSystems/awsBatch.py
```diff
@@ -0,0 +1,537 @@
...
+import math
+import os
+import pickle
+import tempfile
+import time
+import uuid
+from argparse import ArgumentParser, _ArgumentGroup
+from typing import Any, Callable, Dict, Iterator, List, Optional, Set, Union
+
+from requests.exceptions import HTTPError
```
w-gao, line 41: I don't think we're using `HTTPError` in this file anymore.

w-gao (approved):
Just a few comments. I wasn't able to test this, but this looks pretty good to me! It looks like Lon's comments have also been addressed.

I see that there are a few TODO's (places that may need retries, etc), but they should be fine for now.

DailyDreaming (approved):
LGTM.  Thank you for addressing my comments.

bollig:
Can you update the documentation to reflect this new option?

PR #3971, opened 2021-12-21
Add launch cluster support for ARM

### LINKED ISSUE #3966: Launch cluster support for ARM
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1111)
┆friendlyId: TOIL-1111

jonathanxu18:
Should fix #3966.

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/provisioners/abstractProvisioner.py
```diff
@@ -778,11 +779,19 @@ def addToilService(self, config: InstanceConfiguration, role: str, keyPath: str
...
+        if platform.machine() in ['arm', 'aarch64_be', 'aarch64', 'armv8b', 'armv8l']:
+            return 'arm64'
+        return 'amd64'
+
     def getKubernetesValues(self):
         """
         Returns a dict of Kubernetes component versions and paths for formatting into Kubernetes-related templates.
         """
         return dict(
+            ARCHITECTURE=self.getArchitecture(),
```
adamnovak, line 794: This not always going to be the same result as `self._architecture` on the AWSProvisioner. So if you're on an and64 machine running `launch-cluster` to launch an ARM cluster, you'll get an ARM instance and an ARM AMI and then try to install amd64 Kubernetes on it.

Maybe `AWSProvisioner` should override `getArchitecture()` to return that? Or maybe we should make `AbstractProvisioner` have an `_architecture` and in the `__init__` set `self._architecture = self.getArchitecture()` as a default?

adamnovak (changes requested):
It looks like the code to fill in the architecture fields for the instance types from the big downloaded JSON is missing.

You probably also want to run the downloader and commit its results, so we have architectures in our list of instance types.

Also, the problem where there are two different and potentially mismatched notions of the cluster's architecture both in use (`self._architecture` and `self.getArchitecture()` needs to be fixed).

src/toil/lib/ec2nodes.py
```diff
@@ -192,11 +197,14 @@ def fetchEC2InstanceDict(awsBillingJson: Dict[str, Any], region: str) -> Dict[st
             normal_use = i.get('usagetype').endswith('BoxUsage:' + i['instanceType'])  # not reserved or unused
             if normal_use:
                 disks, disk_capacity = parseStorage(v["attributes"]["storage"])
+                architecture = 'arm64' if re.search(".*\dg.*\..*", i["instanceType"]) else 'amd64'
```
adamnovak, line 200: Oops, I was wrong, you do have the code to determine the architecture, right here.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -407,7 +407,7 @@ def launchCluster(self,
         createdSGs = self._createSecurityGroups()
         bdms = self._getBoto3BlockDeviceMappings(leader_type, rootVolSize=leaderStorage)
 
-        userData = self._getIgnitionUserData('leader')
+        userData = self._getIgnitionUserData('leader', self._architecture)
```
adamnovak, line 410: This is going to pass the architecture as the second argument (`keyPath`). I think you need to say `architecture=self._architecture` here.

adamnovak (changes requested):
I think you're missing a keyword on what you meant to pass as a keyword argument, but other than that I don't see any problems.

adamnovak:
It looks like tests are failing in `toil.test.provisioners.aws.awsProvisionerTest.PreemptableDeficitCompensationTest.test` because `_architecture` isn't actually getting set on the provisioner, because `launchCluster` is never called in that Toil execution.

I think this is because we're operating the provisioner inside the workflow, in this case (we're using the Toil-run autoscaling of a Mesos cluster). When that happens, we never `launchCluster` in that particular Toil execution, but we do add nodes.

I think in that case (i.e. when reading `self._architecture` throws an AttributeError) we want to be using the current architecture of the node we are running on, or else maybe the architecture of the node we are trying to add, if we can get it.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -842,7 +842,7 @@ def addNodes(self, nodeTypes: Set[str], numNodes, preemptable, spotBid=None) ->
             subnet_id = next(iter(self._worker_subnets_by_zone[zone]))
 
         keyPath = self._sseKey if self._sseKey else None
-        userData = self._getIgnitionUserData('worker', keyPath, preemptable)
+        userData = self._getIgnitionUserData('worker', keyPath, preemptable, self._architecture)
```
adamnovak, line 845: This could use `type_info.architecture` which will always be defined. Or you could default this when undefined, like you are doing in `_discoverAMI`.

But maybe what you really should do is bring back `getArchitecture()`, and have it return `self._architecture` if it is set, or the current machine's architecture otherwise, and then call it everywhere you need an architecture to use? Then you won't need to repeat the same try/catch everywhere.

Or, you could set `self._architecture` to the current machine's architecture in `__init__`, so we default to assuming the same architecture as we're running on unless someone specifies differently in a `launchCluster()` call.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -940,7 +940,10 @@ def _discoverAMI(self) -> str:
         :return: The AMI ID (a string like 'ami-0a9a5d2b65cce04eb') for Flatcar.
         :rtype: str
         """
-        return get_flatcar_ami(self.aws.client(self._region, 'ec2'))
+        try:
+            return get_flatcar_ami(self.aws.client(self._region, 'ec2'), self._architecture)
+        except:
```
adamnovak, line 945: This probably wants to be `except AttributeError` and have a comment about how sometimes we are running in the in-workflow Toil scaler and self._architecture isn't set.

Also, if we don't pass an architecture, the default is amd64, which only happens to work because the in-workflow scaler is only ever used with amd64 Mesos clusters. We might need to note that or this will seem wrong.

adamnovak (changes requested):
My previous complaints have been satisfied, but it looks like when we did pair programming on this we forgot about the in-workflow Toil autoscaler, where the provisioner is used to add and remove nodes but `launchCluster()` never gets called.

In that case, we never set `self._architecture`, and so when we go to read it it is undefined and throws errors.

You need to do something to fix that, such as setting it in `__init__` or making sure all accesses allow a default value. As long as we only use the Toil-integrated autoscaler on Mesos clusters, it's safe to assume that whenever it's not set we ought to be amd64 (`getattr(self, '_architecture', 'amd64')`).

src/toil/provisioners/aws/awsProvisioner.py
adamnovak: This might be better as a getattr call with a default value.

adamnovak (approved):
I think this will work!

DailyDreaming (approved):
LGTM.  Thank you @jonathanxu18 !

adamnovak:
Sadly a dependency made a release that breaks our typechecking, so this isn't going to pass until it has #4011 I think.

PR #3974, opened 2021-12-21
Use an `init` process that reaps zombies on toil clusters

w-gao:
The `--init` flag to `docker run` enables [`tini`](https://github.com/krallin/tini), which is an init process that properly reaps zombies. This should be able to clean up zombie processes on the Toil appliance.

## Changelog Entry

 * Use an `init` process that reaps zombies on toil clusters

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/batchSystems/singleMachine.py
+++ b/src/toil/batchSystems/singleMachine.py
@@ -334,7 +334,6 @@ class SingleMachineBatchSystem(BatchSystemSupport):
                 # The group is not only gone but reused
                 pass
 
-
     def _pollForDoneChildrenIn(self, pid_to_popen):
         """
         See if any children represented in the given dict from PID to Popen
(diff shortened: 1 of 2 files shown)


src/toil/batchSystems/singleMachine.py
```diff
@@ -570,31 +569,6 @@ def _handleChild(self, pid: int) -> None:
             # If we killed it then it shouldn't show up in the queue.
             self.outputQueue.put(UpdatedBatchJobInfo(jobID=jobID, exitStatus=statusCode, wallTime=time.time() - info.time, exitReason=None))
 
-        # Make absolutely sure all processes in the group have received their
```
w-gao, line 573: We might still want this loop so users know that there are child processes that can't be reaped, if they are not running on a toil cluster?

DailyDreaming (commented):
Adding `tini` / `--init` looks like a great idea.  I'm on the side of keeping the loop, see comment.  Thanks for adding this.

src/toil/batchSystems/singleMachine.py
w-gao: I agree! I will bring the loop back and update the comment accordingly.

adamnovak:
@DailyDreaming If we keep the loop, and we run inside someone else's container, running their user code that leaves orphan zombie child processes, we will get stuck.

I think that's what's happening to Cactus in https://github.com/ComparativeGenomicsToolkit/cactus/issues/610; I don't think our Docker is being used there, so removing the loop is I think the only lever we have to solve that problem.

adamnovak:
@w-gao Is this still a draft? Or can we un-draft it?

adamnovak (approved):
I think this is good and should fix the issues with Cactus runs getting stuck.

We could make our warnings a little more clearly about zombie processes possibly being left behind and less about "surviving process groups" that are "known to batch systems", but that's not a new message so it can't really count against the PR.

src/toil/batchSystems/singleMachine.py
w-gao: > @DailyDreaming If we keep the loop, and we run inside someone else's container, running their user code that leaves orphan zombie child processes, we will get stuck.

I updated so that we'll just leave a warning message if there are zombie processes left behind instead, since the loop wouldn't actually do anything except blocking the workflow from completion. Please let me know what you think though.

w-gao:
> @w-gao Is this still a draft? Or can we un-draft it?

Should be ready. I also updated the logging message so please let me know if that's more informative now.

adamnovak (approved):
This looks great now!

PR #4038, opened 2022-02-10
Add CWL tests under ARM

### LINKED ISSUE #3890: Run CWL tests under Arm
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1078)
┆epic: Graviton2 Summer 2021 Project
┆friendlyId: TOIL-1078

jonathanxu18:
Should fix #3890.

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -215,6 +215,9 @@ cwl_on_arm:
     - pwd
     - virtualenv -p ${MAIN_PYTHON_PKG} venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[cwl,aws]
     - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
+    # This reads GITLAB_SECRET_FILE_SSH_KEYS
+    - python setup_gitlab_ssh.py
+    - chmod 400 /root/.ssh/id_rsa
     - make test tests=src/toil/test/cwl/cwlTest.py::CWLOnARMTest
 
 cwl_v1.0_kubernetes:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -97,9 +97,7 @@ py37_main:
     - virtualenv -p python3.7 venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[all] packages='htcondor awscli'
     - make test tests=src/toil/test/src
     - make test tests=src/toil/test/utils
-    - make test tests=src/toil/test/lib/test_ec2.py
-    - make test tests=src/toil/test/lib/aws
-    - make test tests=src/toil/test/lib/test_conversions.py
+    - TOIL_SKIP_DOCKER=true make test tests=src/toil/test/lib
 
 py37_appliance_build:
   stage: basic_tests
@@ -118,9 +116,7 @@ py38_main:
     - virtualenv -p python3.8 venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[all] packages='htcondor awscli'
     - make test tests=src/toil/test/src
     - make test tests=src/toil/test/utils
-    - make test tests=src/toil/test/lib/test_ec2.py
-    - make test tests=src/toil/test/lib/aws
-    - make test tests=src/toil/test/lib/test_conversions.py
+    - TOIL_SKIP_DOCKER=true make test tests=src/toil/test/lib
 
 py38_appliance_build:
   stage: basic_tests
@@ -139,6 +135,7 @@ py39_main:
     - virtualenv -p python3.9 venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[all] packages=htcondor
     - make test tests=src/toil/test/src
     - make test tests=src/toil/test/utils
+    - TOIL_SKIP_DOCKER=true make test tests=src/toil/test/lib
 
 py39_appliance_build:
   stage: basic_tests
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,8 +28,9 @@ before_script:
   # where buildx can't use a bearer token from a kube config and falls back to anonymous instead
   # of using the system's service account.
   - KUBECONFIG=/dev/null docker buildx create --use --name toilbuilder --platform=linux/amd64,linux/arm64 --node=buildkit-amd64 --driver=kubernetes --driver-opt="nodeselector=kubernetes.io/arch=amd64"
-  # Dump the builder info
-  - docker buildx inspect --bootstrap || true
+  # Dump the builder info, and make sure it exists.
+  - docker buildx inspect --bootstrap || (echo "Docker builder deployment can't be found in our Kubernetes namespace! Are we on the right Gitlab runner?" && exit 1)
+  # This will hang if we can't talk to the builder
   - (echo "y" | docker buildx prune --keep-storage 80G) || true
   - sudo apt-get update
   - sudo apt-get install -y software-properties-common build-essential virtualenv
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -209,6 +209,14 @@ cwl_v1.2:
     - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
     - make test tests=src/toil/test/cwl/cwlTest.py::CWLv12Test
 
+cwl_on_arm:
+  stage: main_tests
+  script:
+    - pwd
+    - virtualenv -p ${MAIN_PYTHON_PKG} venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[cwl,aws]
+    - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
+    - make test tests=src/toil/test/cwl/cwlTest.py::CWLOnARMTest
+
 cwl_v1.0_kubernetes:
   stage: main_tests
   only: []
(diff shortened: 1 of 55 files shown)


src/toil/test/cwl/cwlTest.py
```diff
@@ -631,6 +639,162 @@ def test_kubernetes_cwl_conformance(self, **kwargs):
...
+        assert self.zone is not None, "Could not determine AWS availability zone to test in; is TOIL_AWS_ZONE set?"
+        # We need a boto2 connection to EC2 to check on the cluster
+        self.boto2_ec2 = boto.ec2.connect_to_region(zone_to_region(self.zone))
+
+        # Where should we put our virtualenv?
+        self.venvDir = '/tmp/venv'
+        # Where should we put our data to work on?
+        # Must exist in the Toil container; the leader will try to rsync to it
+        # (for the SSE key) and not create it.
+        self.dataDir = '/tmp'
```
adamnovak, line 664: Some of these fields (`dataDir`, for example) aren't used in this new copy and should be removed.

src/toil/test/cwl/cwlTest.py
```diff
@@ -631,6 +639,162 @@ def test_kubernetes_cwl_conformance(self, **kwargs):
...
+        # If this fails, no tests will run.
+        self.destroyCluster()
+
+    def tearDown(self):
+        # Note that teardown will run even if the test crashes.
+        super().tearDown()
+        self.destroyCluster()
+        subprocess.check_call(['toil', 'clean', self.jobStore])
+
+    def sshUtil(self, command):
```
adamnovak, line 690: Would it be possible to create a base class for this class and the class we copied this from, so we can have just one `sshUtil` definition? Or otherwise refactor it out so we only have to have one copy?

src/toil/test/cwl/cwlTest.py
```diff
@@ -631,6 +639,162 @@ def test_kubernetes_cwl_conformance(self, **kwargs):
...
+        self.leader = self.cluster.getLeader()
+
+        # Clone Toil repo
+        # We retrieve the Toil version needed to run the CWL tests
+        # Use either a git command, CI_COMMIT_SHA, or with version.py to find the version
+        # After finding the correct commit and checkout
+        # Activate the virtualenv
+        # In the commit we run the tests using pytest
+
+        commit = os.environ['CI_COMMIT_SHA']
```
adamnovak, line 785: If the test needs this variable set, we should put a marker on it to require that.

I think we should bring back the `needs_env_var` decorator from https://github.com/DataBiosphere/toil/pull/3956/files/f1b1b854d9f8bd288ed51252884043b6cf176173#diff-3425fd4766c61da9e32e2b8f70b5d5dece213fd8ceed813bf8370034913f7d28R325-R336 and put `@needs_env_var('CI_COMMIT_SHA')` on this test. Otherwise it will fail instead of being skipped if someone tries to run it locally and doesn't fake the variable.

src/toil/test/cwl/cwlTest.py
```diff
@@ -631,6 +639,162 @@ def test_kubernetes_cwl_conformance(self, **kwargs):
...
+        self.sshUtil(['git', 'clone', 'https://github.com/DataBiosphere/toil.git'])
+        self.sshUtil(['bash', '-c', f'cd toil && git checkout {commit}'])
+        # --never-download prevents silent upgrades to pip, wheel and setuptools
+        # Activate the virtualenv
+        venv_command = ['virtualenv', '--system-site-packages', '--never-download', self.venvDir]
+        self.sshUtil(venv_command)
+        self.sshUtil(['bash', '-c', 'pwd && ls && cd toil && ls && pwd'])
+        self.sshUtil(['bash', '-c', f'. .{self.venvDir}/bin/activate && cd toil && make prepare && make develop extras=[all]'])
+        # Runs the CWLv12Test on an ARM instance
+        self.sshUtil(['bash', '-c', f'. .{self.venvDir}/bin/activate && ls ./sys/fs/cgroup && cd toil && pytest --log-cli-level DEBUG -r s src/toil/test/cwl/cwlTest.py::CWLv12Test::test_run_conformance'])
```
adamnovak, line 796: Is there a good reason to be looking at the cgroups here?

src/toil/test/cwl/cwlTest.py
```diff
@@ -631,6 +639,162 @@ def test_kubernetes_cwl_conformance(self, **kwargs):
...
+        self.boto2_ec2 = boto.ec2.connect_to_region(zone_to_region(self.zone))
+
+        # Where should we put our virtualenv?
+        self.venvDir = '/tmp/venv'
+        # Where should we put our data to work on?
+        # Must exist in the Toil container; the leader will try to rsync to it
+        # (for the SSE key) and not create it.
+        self.dataDir = '/tmp'
+
+    def destroyCluster(self) -> None:
```
adamnovak, line 666: Technically, we're only supposed to be taking PRs where everything is `snake_case` now, but a bunch of these copy-pasted methods are `camelCase` and thus ought to be renamed.

adamnovak (changes requested):
This looks like it should work, but it has some unused variables we ought to get rid of. Also, it would be good if all the new methods and variables were translated to `snake_case` names.

Also, is it possible to share `sshUtil`'s implementation with the place it was copy-pasted from? Maybe factor it out from both places into a new base class or a utility function you can import and call? It's a lot of code and I don't think any changes have been made to it.

adamnovak (changes requested):
This looks like a good design, and I think we can leave the copy-pasted camelCase method names for now, but I'm concerned that the machine we're creating to run tests on might not really be an ARM instance type.

src/toil/test/cwl/cwlTest.py
```diff
@@ -791,7 +795,7 @@ def test_cwl_on_arm(self):
 
         # Runs the CWLv12Test on an ARM instance
         self.sshUtil(['bash', '-c', f'. .{self.venvDir}/bin/activate && cd toil && pytest --log-cli-level DEBUG -r s src/toil/test/cwl/cwlTest.py::CWLv12Test::test_run_conformance'])
-        self.cluster.destroyCluster()
+        self.tearDown()
```
adamnovak, line 798: Doesn't `tearDown` automatically run after all the tests?

adamnovak (approved):
This looks like it ought to work.

.gitlab-ci.yml
```diff
@@ -209,6 +209,14 @@ cwl_v1.2:
     - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
     - make test tests=src/toil/test/cwl/cwlTest.py::CWLv12Test
 
+cwl_on_arm:
+  stage: main_tests
+  script:
+    - pwd
+    - virtualenv -p ${MAIN_PYTHON_PKG} venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[cwl,aws]
+    - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
+    - make test tests=src/toil/test/cwl/cwlTest.py::CWLOnARMTest
```
adamnovak, line 218: Actually, the tests are failing because we don't actually go get the private keys we need in order to log into the AWS instances we make, [like we do for the integration tests](https://github.com/DataBiosphere/toil/blob/c758dd2986281b2c9a95c171ecd162263e1cb358/.gitlab-ci.yml#L295-L297). We need to run the magic key-getting script:

```suggestion
    - pwd
    - virtualenv -p ${MAIN_PYTHON_PKG} venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[cwl,aws]
    - python setup_gitlab_docker.py  # login to increase the docker.io rate limit
    # This reads GITLAB_SECRET_FILE_SSH_KEYS
    - python setup_gitlab_ssh.py
    - chmod 400 /root/.ssh/id_rsa
    - make test tests=src/toil/test/cwl/cwlTest.py::CWLOnARMTest
```

adamnovak (changes requested):
I found the reason the tests fail; we need to give the test runner the secret private key so it can log in.

adamnovak:
@jonathanxu18 Now it's able to ssh to the cluster, but it still doesn't appear to be working (because it can't start the Toil appliance), and I don't know why. [6fd7a08](https://github.com/DataBiosphere/toil/pull/4038/commits/6fd7a08c1306e675e565a967dfc5fe994a4dbfb3) worked as far as we tested it, but we didn't run the Gitlab tests through on it so we didn't try the test we're adding here.

You are still able to start up an ARM cluster with e.g. `TOIL_APPLIANCE_SELF=quay.io/ucsc_cgl/toil:5.7.0a1-8c55d50a15cb1ee2dbe7f853a7fb4d35e02f6280-py3.9`, right?

I did see that the node exporter service is stuck constantly restarting. I think that's because it [uses a Docker container](https://github.com/DataBiosphere/toil/blob/c758dd2986281b2c9a95c171ecd162263e1cb358/src/toil/provisioners/abstractProvisioner.py#L707) that probably isn't available for ARM. We could stop including that service in ARM clusters, but I don't think it's causing the problem here.

adamnovak:
OK, the problem is that we're running this:
```
'toil', 'launch-cluster', '-p=aws', '-z', 'us-west-2a', '--keyPairName=id_rsa', '--leaderNodeType=t4g.medium', '--logDebug', 'cwl-test-b77bffa7-440c-42e7-93eb-110a2de8e9e9'
```

And this defaults to making a Mesos cluster. Mesos has no ARM binaries, so the Mesos leader that is the main process in the Toil appliance container can't start, so the Toil appliance container can't start.

When launching a cluster for ARM testing, we need to make sure we have `--clusterType=kubernetes` in the command, in addition to `--leaderNodeType=t4g.medium`. @jonathanxu18 I think the best weay to implement that would be another variable like `self.leaderNodeType`, maybe `self.cluster_type`. You would set it to `mesos` by default, and set it to `kubernetes` in the class for this test, and pass along `f"--clusterType={self.cluster_type}"` as another argument when launching the cluster.

adamnovak (approved):
This looks good to me!

PR #4052, opened 2022-02-22
Run CWL conformance tests via WES

### LINKED ISSUE #3876: WES: Run CWL conformance tests via WES
Run through the CWL conformance tests via WES manually and/or through the CI.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1067)
┆friendlyId: TOIL-1067

w-gao:
Closes #3876. We are skipping a few tests because `cwltool --print-deps` doesn't include secondary files from the default. See [example](https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/mixed-versions/wf-v10.cwl#L4-L10). Also see https://github.com/DataBiosphere/toil/pull/4052#discussion_r886037170.

These tests are also not run under CI at the moment, but I've ran them against a local and remote Toil WES server.

## Changelog Entry

 * Run CWL conformance tests via WES

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/server/wes/abstract_backend.py
```diff
@@ -255,7 +255,11 @@ def collect_attachments(self, run_id: Optional[str], temp_dir: Optional[str]) ->
             self.log_for_run(run_id, "Using workflow_url '%s'" % body.get("workflow_url"))
         else:
             raise MalformedRequestException("Missing 'workflow_url' in submission")
+
         if "workflow_params" not in body:
             raise MalformedRequestException("Missing 'workflow_params' in submission")
+        # TODO: make sure workflow_url and body["workflow_params"] does not access files outside of its own
+        #  workflow directory.
+        # this is hard for WDL files especially, unless we interpret every string to be a path
```
adamnovak, line 263: I don't think this is too important; we aren't meant to provide multi-tenant security here.

src/toil/test/__init__.py
```diff
@@ -612,6 +612,22 @@ def needs_server(test_item: MT) -> MT:
         return test_item
 
 
+def needs_wes(test_item: MT) -> MT:
+    """Use as a decorator before test classes or methods to run only if WES is available."""
```
adamnovak, line 616:
```suggestion
def needs_wes_server(test_item: MT) -> MT:
    """Use as a decorator before test classes or methods to run only if a WES server is available to run against."""
```

src/toil/test/cwl/cwlTest.py
```diff
@@ -707,6 +721,53 @@ def test_kubernetes_cwl_conformance_with_caching(self):
...
+        password = os.environ.get("TOIL_WES_PASSWORD")
+
+        extra_args = [f"--wes_endpoint={endpoint}"]
+        if user and password:
+            # TODO: if we run this on the CI, make sure we don't log this...
+            #  we output the entire command before we run, and cwltest logs the command if a test fails.
+            extra_args.extend([
+                f"--wes_user={user}",
+                f"--wes_password={password}"
+            ])
```
adamnovak, line 750: The WES runner script should probably take these from the environment itself instead of having them set here.

Or, if we're running against local a local WES server managed by the tests, we can use no authentication and make sure we only listen on the loopback interface.

src/toil/test/cwl/cwlTest.py
```diff
@@ -707,6 +721,53 @@ def test_kubernetes_cwl_conformance_with_caching(self):
...
+        #   - 332: iwd-fileobjs2
+
+        # Main issues:
+        # 1. `cwltool --print-deps` doesn't seem to include secondary files from the default
+        #     e.g.: https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/mixed-versions/wf-v10.cwl#L4-L10
+        # 2. We need a way to reinterpret relative paths inside the workflow files
+        #    e.g.: https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/iwd/iwd-fileobjs1.cwl#L7-L11
+
+        return self.test_run_conformance(
+            runner=["python", os.path.abspath("src/toil/test/server/wes_cwl_runner.py")],
```
adamnovak, line 766: Maybe add another entry point for this script to expose it as a command, like we have for `toil-cwl-runner`?

src/toil/test/server/wes_client.py
```diff
@@ -0,0 +1,212 @@
+# Modified from wes-service under Apache License, Version 2.0.
+# https://github.com/common-workflow-language/workflow-service/blob/main/LICENSE
```
adamnovak, line 2: Instead of copy-pasting all this, we could just depend on the `wes-service` package, and e.g. `from wes_client.util import WESClient`, right? What kind of changes do we need to make that would justify reimplementing the bindings here? And can we make them by inheriting most of the code from the other package and just overriding some methods?

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+                    get_deps(file.get("secondaryFiles"))
+
+    get_deps(json_result.get("secondaryFiles", []))
+    return deps
+
+
+def submit_run(client: WESClient,
+               cwl_file: str,
+               input_file: Optional[str] = None,
+               engine_options: Optional[List[str]] = None) -> str:
```
adamnovak, line 90: This function is missing a docstring.

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+    --logDebug
+```
+"""
+
+logger = logging.getLogger(__name__)
+
+
+def get_deps_from_cwltool(cwl_file, input_file, input_deps: bool = False):
+    """
+    Return a list of dependencies of the given workflow from cwltool.
```
adamnovak, line 41: What does `input_deps` do here? Can we get only either the deps for the input file or the CWL file? Why do we need the input file if we aren't getting deps for it?

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+    --disableCaching \
+    --clean=always \
+    --logDebug
+```
+"""
+
+logger = logging.getLogger(__name__)
+
+
+def get_deps_from_cwltool(cwl_file, input_file, input_deps: bool = False):
```
adamnovak, line 39:
```suggestion
def get_deps_from_cwltool(cwl_file: str, input_file: Optional[str], input_deps: bool = False):
```

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+
+def poll_run(client: WESClient, run_id: str) -> bool:
+    """ Return True if the given workflow run is in a finished state."""
+    status_result = client.get_run_status(run_id)
+    state = status_result.get("state")
+
+    return state in ("COMPLETE", "CANCELING", "CANCELED", "EXECUTOR_ERROR", "SYSTEM_ERROR")
+
+
+def print_logs_and_exit(client: WESClient, run_id: str) -> None:
```
adamnovak, line 113: This should have a docstring too.

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+    # cwltest checks for stdout, stderr, and return code
+    data = client.get_run_log(run_id)
```
adamnovak, line 115: It doesn't look like this makes more requests to actually download the stdout and stderr data form the URLs that the WES server is supposed to give.

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+    # first two positional arguments are passed in by cwltest
+    parser.add_argument("cwl_file", type=str)
+    parser.add_argument("input_file", type=str, nargs="?", default=None)
+    # a few arguments useful for the WES runner
+    parser.add_argument("--wes_endpoint", default="http://localhost:8080",
+                        help="The http(s) URL of the WES server.  (default: %(default)s)")
+    parser.add_argument("--wes_user", default=None,
+                        help="Username to use for basic authentication to the WES server.")
+    parser.add_argument("--wes_password", default=None,
+                        help="Password to use for basic authentication to the WES server.")
```
adamnovak, line 135: It would be great if this one in particular could come from an environment variable.

src/toil/test/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,165 @@
...
+
+    wes_user = options.wes_user
+    wes_password = options.wes_password
+    client = WESClient(base_url=endpoint,
+                       auth=(wes_user, wes_password) if wes_user and wes_password else None)
+
+    run_id = submit_run(client, cwl_file, input_file, engine_options=rest)
+    assert run_id
+
+    # logger.warning(f"Submitted run. Waiting for {run_id} to finish...")
```
adamnovak, line 154: You should drop these or uncomment them and put them at a lower log level.

adamnovak (changes requested):
This looks pretty good. I kind of want the WES-based CWL runner to be brought out as a command-line tool, and to pull in somebody else's WES client bindings if we can get away with it.

src/toil/test/server/wes_cwl_runner.py
w-gao: > Can we get only either the deps for the input file or the CWL file?

I think so. We have to call `cwltool --print-deps cwl_file` to get the dependencies inside the CWL file, and call `cwltool --print-input-deps cwl_file input_file` to get the dependencies inside the input file.

I intended this to be two function calls based on `input_deps`, but I now realize that we don't need `input_deps` as we can just infer it if `input_file` is provided. Or, perhaps we could also execute them both here so we only need to call this function once.

src/toil/test/server/wes_client.py
adamnovak: I talked to @w-gao and we need to duplicate code here because trying to install `wes-service` gets stuck in an infinite loop on a `NameError: name 'Str' is not defined` in the install script for `ruamel.yaml==0.15.77`.

The changes made include support for workflow engine parameters, in a way that's tough to nicely do with inheritance; we'd duplicate a bunch of request building code and end up with a `WESClientWithWorkflowEngineParameters` or something.

So the plan is to keep this in Toil, but contribute the additional features back upstream, and get @mr-c or someone else who has touched `wes-service` to release a version that has these features and also installs. Then we can replace our copy with theirs.

src/toil/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,334 @@
...
+    options, rest = parser.parse_known_args()
+
+    cwl_file = options.cwl_file
+    input_file = options.input_file
+
+    # Initialize client and run the workflow
+    endpoint = options.wes_endpoint
+
+    # For security reasons, username and password can only come from environment variables
+    wes_user = os.environ.get("TOIL_WES_USER", None)
```
mr-c, line 315:
```suggestion
    wes_user = os.environ.get("TOIL_WES_PASSWORD", None)
```

src/toil/server/wes_cwl_runner.py
```diff
@@ -0,0 +1,334 @@
...
+    wes_password = os.environ.get("TOIL_WES_USER", None)
```
mr-c, line 316:
```suggestion
    wes_password = os.environ.get("TOIL_WES_PASSWORD", None)
```

src/toil/test/cwl/cwlTest.py
```diff
@@ -807,14 +814,94 @@ def test_kubernetes_cwl_conformance_with_caching(self):
...
+        #   - 311: mixed_version_v11_wf
+        #   - 312: mixed_version_v12_wf
+        #   - 331: iwd-fileobjs1
+        #   - 332: iwd-fileobjs2
+
+        # Main issues:
+        # 1. `cwltool --print-deps` doesn't seem to include secondary files from the default
+        #     e.g.: https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/mixed-versions/wf-v10.cwl#L4-L10
+        # 2. We need a way to reinterpret relative paths inside the workflow files
+        #    e.g.: https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/iwd/iwd-fileobjs1.cwl#L7-L11
```
mr-c, line 846: Let me know if you'd like assistance to either of these. I could fix `cwltool --print-deps` to include defaults, for example.

src/toil/server/wes_cwl_runner.py
w-gao: Good catch! Thank you @mr-c

src/toil/test/cwl/cwlTest.py
w-gao: I would really appreciate if you could fix `cwltool --print-deps` to include the defaults! I don't think I am familiar with `cwltool` enough to be able to fix that :(. I should be able to fix the second issue because it has something to do with the new WES runner.

mr-c: Okay, I'll work on that today!

This runner is very helpful, thank you! I had a hacky bash script that I discarded in favor of your code. With your WES runner I have successfully run all the conformance tests for CWL versions 1.0 and 1.1 against local and remote installations of toil 5.6.0.

Would it be okay with you if later I split this out as a separate `cwl-wes-runner` project inside one of the CWL GitHub organizations? I'd then remove the code from `workflow-service` to use it instead.

mr-c: fixing `cwltool --print-deps` to add the secondary files when there is a default `clas: File` literal is going to be a lot of work; it is a very obscure situation that can be worked around by adding the `secondaryFiles` manually to the `class: File` literal, so feel free to skip those tests for now. The standalone `cwl-wes-runner` is more important, so I'm going to focus on that first.

src/toil/test/server/wes_cwl_runner.py
w-gao: Ah, I'm using the `outputs` data from the GetRunLog response here, since it's the same as the stdout for `toil-cwl-runner`.

adamnovak:
@w-gao It looks like this is coming together. Do you think it's ready to not be a draft?

adamnovak (changes requested):
I complained about a bunch of stuff, but I think the only thing that really *needs* to change is the handling of path names in the `File` and `Directory` objects vs. in attachment naming. I think the attachment name and the `location` always need to fully match, but right now I'm not convinced that they will. Either they need to be generated by the same code, or the places where they are generated need to both say what restrictions/transformations exactly are meant to be applied.

src/toil/server/cli/wes_cwl_runner.py
```diff
@@ -0,0 +1,342 @@
...
+            "auth": auth,
+            "proto": proto,
+            "host": host
+        })
+
+    @staticmethod
+    def get_version(extension: str, workflow_file: str) -> str:
+        """Determines the version of a .py, .wdl, or .cwl file."""
+        if extension == "py":
+            return "3.8"
```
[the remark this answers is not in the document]
w-gao: This is mostly a placeholder for now. If we ever want it to be accurate, we could ask the user to provide it at least for Toil workflows.

docs/appendices/environment_vars.rst
```diff
@@ -204,6 +204,15 @@ There are several environment variables that affect the way Toil runs.
...
 |                                  | or ``google``.                                     |
 +----------------------------------+----------------------------------------------------+
+| TOIL_WES_ENDPOINT                | URL to the WES server to use for the WES-based CWL |
+|                                  | runner ("toil-wes-cwl-runner") and tests.          |
++----------------------------------+----------------------------------------------------+
+| TOIL_WES_USER                    | Username to use with HTTP Basic Authentication to  |
+|                                  | log into the WES server.                           |
++----------------------------------+----------------------------------------------------+
+| TOIL_WES_PASSWORD                | Password to use with HTTP Basic Authentication to  |
+|                                  | log into the WES server.                           |
```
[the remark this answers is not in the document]
w-gao: I moved them to the docstring inside the runner to avoid potential confusion.

src/toil/server/cli/wes_cwl_runner.py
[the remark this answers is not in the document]
adamnovak: I think that should be recorded in a comment here. We want whoever looks at the code next not to waste a lot of time when they see this and realize it can't possibly actually work.
```suggestion
            # For Toil testing, we only ever run CWL workflows.
            # Just pretend all Python workflows are version 3.8 for now.
            return "3.8"
```

src/toil/server/cli/wes_cwl_runner.py
```diff
@@ -0,0 +1,342 @@
...
+    Return a list of dependencies of the given workflow from cwltool.
+
+    :param cwl_file: The CWL file.
+    :param input_file: Omit to get the dependencies from the CWL file. If set,
+                       this returns the dependencies from the input file.
+    """
+
+    option = '--print-input-deps' if input_file else '--print-deps'
+
+    args = ['cwltool', option, '--relative-deps', 'cwd', cwl_file]
```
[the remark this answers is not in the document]
adamnovak: It looks like `printdeps` in the cwltool code eventually calls [`find_deps`](https://github.com/common-workflow-language/cwltool/blob/898fddf9167620645177eb94c8ea68ede0667f96/cwltool/main.py#L598) to actually get the deps. I would recommend importing and calling that, rather than running the command and parsing the output, unless replicating the setup for it would be hard.

src/toil/server/cli/wes_cwl_runner.py
```diff
@@ -0,0 +1,342 @@
...
+    TODO: Propose a PR in wes-service to include workflow_engine_params.
+    """
+    def __init__(self, endpoint: str, auth: Optional[Tuple[str, str]] = None) -> None:
+        """
+        :param endpoint: The http(s) URL of the WES server. Must include the
+                         protocol.
+        :param auth: Authentication information that will be attached to every
+                     request to the WES server.
+        """
+        proto, host = endpoint.split("://")
```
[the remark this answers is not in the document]
adamnovak: I would usually prefer to use the library URL parsing functions because it's harder to get things wrong with them.

Really I'd want to stop using `proto` and `host` and just keep a `urllib.parse` type around as a base URL, but it looks like we need them for the base class.

Maybe this way is fine?

mr-c:
I'm getting `[CRITICAL] WORKER TIMEOUT ` errors from `toil server` with a big workflow; does anyone know where to adjust the gunicorn timeout value? https://docs.gunicorn.org/en/stable/settings.html#timeout

src/toil/server/cli/wes_cwl_runner.py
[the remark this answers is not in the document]
w-gao: Here is a script that I'm using to find the dependencies for the CWL file, and it seems to work:

```python
from cwltool.context import LoadingContext
from cwltool.load_tool import resolve_tool_uri
from cwltool.main import find_deps
from schema_salad.exceptions import ValidationException

TOOL = "file:///Users/wlgao/code/GI/toil/test_files/hello.cwl"


try:
    loading_context = LoadingContext()
    uri, tool_file_uri = resolve_tool_uri(
        TOOL,
        resolver=loading_context.resolver,
        fetcher_constructor=loading_context.fetcher_constructor,
    )

    deps = find_deps(None, loading_context.loader, uri)
    print(deps)

except ValidationException:
    # handle error?
    raise
```

But there will be a lot more things to set up and load for `print_input_deps` to work, and we would need to duplicate a lot of code from `cwltool`: https://github.com/common-workflow-language/cwltool/blob/898fddf9167620645177eb94c8ea68ede0667f96/cwltool/main.py#L1287-L1322

w-gao:
@mr-c Unfortunately I don't think there is a way to specify it currently, but after https://github.com/DataBiosphere/toil/pull/4052/commits/880af6cf66defceb4354a8ababf7235c00f46645 we would be able to configure Gunicorn using the [`GUNICORN_CMD_ARGS`](https://docs.gunicorn.org/en/stable/settings.html#settings) environment variable. I could also expose this dict so users can pass Gunicorn configs [here](https://github.com/DataBiosphere/toil/blob/releases/5.6.x/src/toil/server/app.py#L106-L109) through the "toil server" command.

Do you happen to know why it is timing out with big workflows? (Is it because it takes more than 30 seconds to upload the large files?)

mr-c:
I replaced the big data files with empty ones for testing, but now I see that `toil-wes-cwl-runner` struggles with relative paths in `run`

from the workflow
``` cwl
    bwa_mem_paired:                                                             
        run: ../wrappers/bwa-mem.cwl 
```
```
schema_salad.exceptions.ValidationException: Whole_Exome-Seq.cwl:202:1: checking field `steps`
Whole_Exome-Seq.cwl:234:5:   checking object `Whole_Exome-Seq.cwl#bwa_mem_paired`
Whole_Exome-Seq.cwl:235:9:     Field `run` contains undefined reference to
                               `file:///home/ubuntu/workflows/run-32c8bd49b4d74d5aa17f880e1a4aa03a/wrappers/bwa-mem.cwl`
```
``` sh
~$ ls /home/ubuntu/workflows/run-32c8bd49b4d74d5aa17f880e1a4aa03a/
total 20
drwxrwxr-x 4 ubuntu ubuntu 4096 Jun  8 09:46 execution
drwxrwxr-x 2 ubuntu ubuntu 4096 Jun  8 09:46 outputs
-rw-rw-r-- 1 ubuntu ubuntu 1607 Jun  8 09:46 request.json
-rw-rw-r-- 1 ubuntu ubuntu 7812 Jun  8 09:46 stderr
-rw-rw-r-- 1 ubuntu ubuntu    0 Jun  8 09:46 stdout
~$ ls /home/ubuntu/workflows/run-32c8bd49b4d74d5aa17f880e1a4aa03a/execution/
total 28
-rw-rw-r-- 1 ubuntu ubuntu 15829 Jun  8 09:46 Whole_Exome-Seq.cwl
drwxrwxr-x 5 ubuntu ubuntu  4096 Jun  8 09:46 data
-rw-rw-r-- 1 ubuntu ubuntu  1276 Jun  8 09:46 wes_inputs.json
drwxrwxr-x 2 ubuntu ubuntu  4096 Jun  8 09:46 wrappers

```

src/toil/server/cli/wes_cwl_runner.py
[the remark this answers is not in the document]
adamnovak: Maybe calling it on the command line is the best way then.

adamnovak (approved):
This looks mostly good to me; it's probably done enough to merge. Then we can handle any remaining problems with timeouts and relative paths as separate issues, once we can characterize them. Any relative path problems might also want to become CWL conformance tests.

src/toil/server/cli/wes_cwl_runner.py
```diff
@@ -0,0 +1,448 @@
...
+    Where "message.txt" is resolved to "../input_files/message.txt".
+
+    For the workflow attachment "../input_files/message.txt", we must replace
+    this relative path to something else that is not above the CWL workflow
+    file, otherwise the Toil WES server will drop all the traversals up to
+    contain attachments within its "execution" folder.
+
+    TODO: we need to find a way to call this function and replace the relative
+     paths referenced inside the CWL workflow directly.
+     e.g.: https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/iwd/iwd-fileobjs1.cwl#L7-L11
```
adamnovak, line 88: Rather then replacing the relative paths in the workflow files, an approach might be to scan everything, find the highest-level directory traversed to, and then send everything as relative paths down from there.

So in this example, we'd see that `~/toil` is the common ancestor of the hierarchy we are working in, and we'd send everything as paths down from there, using that as the `base_dir` instead of wherever the CWL file lives. We'd send the workflow file as `workflows/hello.cwl`, and send the inputs as `input_files/hello.yaml` and `input_files/message.txt`. Then if the CWL file wants to look at `../input_files/message.txt` relative to itself, it can.

adamnovak (changes requested):
I think we might have lost non-local-file workflow support on accident, but otherwise this looks right to me.

src/toil/server/cli/wes_cwl_runner.py
```diff
@@ -245,20 +214,22 @@ def build_wes_request(
...
-            with open(file, "rb") as f:
-                filename = generate_attachment_path_name(base_dir, file)
-                workflow_attachments.append((filename, BytesIO(f.read())))
+        common_path, attachment_paths = generate_attachment_path_names(attachments)
+        for src, dest in zip(attachments, attachment_paths):
+            with open(src, "rb") as f:
+                workflow_attachments.append((dest, BytesIO(f.read())))
+
+        # Make sure we let the server know where the main CWL file is located
+        data["workflow_url"] = os.path.relpath(workflow_file, common_path)
```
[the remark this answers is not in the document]
w-gao: That's right, thanks for catching this!


I also made more changes that need additional review. It turns out that we do need to modify the input file (at least when the CWL file is in a different directory).

w-gao:
> I replaced the big data files with empty ones for testing, but now I see that `toil-wes-cwl-runner` struggles with relative paths in `run`
> 
> from the workflow
> 
> ```cwl
>     bwa_mem_paired:                                                             
>         run: ../wrappers/bwa-mem.cwl 
> ```
> 
> ```
> schema_salad.exceptions.ValidationException: Whole_Exome-Seq.cwl:202:1: checking field `steps`
> Whole_Exome-Seq.cwl:234:5:   checking object `Whole_Exome-Seq.cwl#bwa_mem_paired`
> Whole_Exome-Seq.cwl:235:9:     Field `run` contains undefined reference to
>                                `file:///home/ubuntu/workflows/run-32c8bd49b4d74d5aa17f880e1a4aa03a/wrappers/bwa-mem.cwl`
> ```
> 
> ...

@mr-c I haven't tested this workflow directly, but I think it was failing because it's importing from a parent directory inside the CWL file (similar to [this conformance test](https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/iwd/iwd-fileobjs1.cwl#L7-L11)), which should be fixed now that we've implemented this: https://github.com/DataBiosphere/toil/pull/4052#discussion_r892631339.

mr-c:
> @mr-c I haven't tested this workflow directly, but I think it was failing because it's importing from a parent directory inside the CWL file (similar to [this conformance test](https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/tests/iwd/iwd-fileobjs1.cwl#L7-L11)), which should be fixed now that we've implemented this: [#4052 (comment)](https://github.com/DataBiosphere/toil/pull/4052#discussion_r892631339).

I did a basic test and I can confirm the path fix. I'll do a full test with real data to confirm all the fixes later today.

Thank you very much @w-gao for all your help! The ELIXIR Compute Platform sends their appreciation.

mr-c:
Hmm.. some paths issues still remain:

`FileNotFoundError: [Errno 2] No such file or directory: '/home/ubuntu/workflows/run-54af27c0e1e24b00b6e403bcff545d94/data/temp_data'`

wrong path, instead of `run-54af27c0e1e24b00b6e403bcff545d94/execution/data/temp_data` which does exist:
```
:~/workflows/run-54af27c0e1e24b00b6e403bcff545d94$ ls -l /home/ubuntu/workflows/run-54af27c0e1e24b00b6e403bcff545d94/execution/data/
total 12
drwxrwxr-x 2 ubuntu ubuntu 4096 Jun  9 09:28 humandb
drwxrwxr-x 2 ubuntu ubuntu 4096 Jun  9 09:28 temp_data
drwxrwxr-x 2 ubuntu ubuntu 4096 Jun  9 09:28 ucsc_hg19_chr1_subset
```
caused by `execution/wes_inputs.json` having a relative path still
```
~/workflows/run-54af27c0e1e24b00b6e403bcff545d94$ jq . < execution/wes_inputs.json 
{
  "raw_files_directory": {
    "class": "Directory",
    "path": "../data/temp_data"
  },
```

w-gao:
> Hmm.. some paths issues still remain:
> 
> `FileNotFoundError: [Errno 2] No such file or directory: '/home/ubuntu/workflows/run-54af27c0e1e24b00b6e403bcff545d94/data/temp_data'`

My bad, I accidentally reverted my uncommitted changes to handle `path`s here when I was refactoring: https://github.com/DataBiosphere/toil/blob/63dbfe35727119ea1972656916663690bfd075ef/src/toil/server/cli/wes_cwl_runner.py#L151

I added it back.


I did realize that the `cwltool --print-input-deps` output is missing the last two secondary files here: https://github.com/athenarc/whole-exome-seq/blob/f5afcdd8894e87b8299d21c42fc7b163a4e2e126/workflows/Whole_Exome-Seq.cwl#L111-L112

```cwl
    HaplotypeCaller_reference:
        type: File
        secondaryFiles:
            - .amb
            - .ann
            - .bwt
            - .pac
            - .sa
            - .fai
            - ^.dict
```

perhaps because it only includes the secondary files for the first occurrence of `../data/ucsc_hg19_chr1_subset/hg19_chr1_subset.fa` here: https://github.com/athenarc/whole-exome-seq/blob/f5afcdd8894e87b8299d21c42fc7b163a4e2e126/workflows/Whole_Exome-Seq.cwl#L21-L28?

```cwl
    bwa_mem_ref_genome:
        type: File
        secondaryFiles:
            - .amb
            - .ann
            - .bwt
            - .pac
            - .sa
```

I'm not sure what the best approach here would be, but, as a quick fix, I added them manually to `bwa_mem_ref_genome` for `cwltool --print-input-deps` to pick them up.  After that I was able to run this workflow successfully on my local machine.

```yaml
bwa_mem_ref_genome:
  class: File
  path: ../data/ucsc_hg19_chr1_subset/hg19_chr1_subset.fa
  secondaryFiles:
    - class: File
      path: ../data/ucsc_hg19_chr1_subset/hg19_chr1_subset.fa.fai
    - class: File
      path: ../data/ucsc_hg19_chr1_subset/hg19_chr1_subset.dict
```

adamnovak (approved):

PR #4084, opened 2022-04-18
Make `toil kill` work when the leader is on another machine

### LINKED ISSUE #3917: `toil kill` doesn't cancel batch system jobs
As noted in https://gitter.im/bd2k-genomics-toil/Lobby?at=618eb0f3fb3dcd4e88818830, `toil kill` doesn't seem to be using the normal shared file machinery to find `pid.log`, and when we reorganized where shared files live, it lost the ability to find it.

We should change `toil kill` to use the normal shared file read code, so it will always be able to find the files it is looking for, no matter where we store them.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1087)
┆friendlyId: TOIL-1087

w-gao:
Closes #3917. We decided that it is not worth it to have every worker constantly poll the job store and self-terminate, since that could create too many requests to S3 or google storage (if using a cloud job store) and most of the time the leader is able to clean up all its jobs. 

## Changelog Entry

 * Make `toil kill` work when the leader is on another machine

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import os
 import pickle
 import re
 import shutil
@@ -1451,6 +1452,26 @@ class AbstractJobStore(ABC):
 
     # A few shared files useful to Toil, but probably less useful to the users.
 
+    def write_leader_pid_file(self) -> None:
+        """
+        Write the pid of this process to a file in the job store.
+
+        Overwriting the current contents of pid.log is a feature, not a bug of
+        this method. Other methods will rely on always having the most current
+        pid available. So far there is no reason to store any old pids.
+        """
+        with self.write_shared_file_stream('pid.log') as f:
+            f.write(str(os.getpid()).encode('utf-8'))
+
+    def read_leader_pid_file(self) -> int:
+        """
+        Read the pid of the leader process to a file in the job store.
+
+        :raise NoSuchFileException: If the PID file doesn't exist.
+        """
+        with self.read_shared_file_stream('pid.log') as f:
+            return int(f.read().strip())
+
     def write_leader_node_id(self) -> None:
         """
         Write the leader node id to the job store. This should only be called
@@ -1462,6 +1483,8 @@ class AbstractJobStore(ABC):
     def read_leader_node_id(self) -> str:
         """
         Read the leader node id stored in the job store.
+
+        :raise NoSuchFileException: If the node ID file doesn't exist.
         """
         with self.read_shared_file_stream("leader_node_id.log") as f:
             return f.read().decode('utf-8').strip()
@@ -1473,9 +1496,8 @@ class AbstractJobStore(ABC):
         The initialized file contains the characters "NO". This should only be
         changed when the user runs the "toil kill" command.
 
-        Removing this file or changing this file to a "YES" triggers a kill of
-        the leader process. The workers are expected to be cleaned up by the
-        leader.
+        Changing this file to a "YES" triggers a kill of the leader process. The
+        workers are expected to be cleaned up by the leader.
         """
         with self.write_shared_file_stream("_toil_kill_flag") as f:
             f.write(("YES" if kill else "NO").encode('utf-8'))
@@ -1489,8 +1511,8 @@ class AbstractJobStore(ABC):
             with self.read_shared_file_stream("_toil_kill_flag") as f:
                 killed = f.read().decode() == "YES"
         except NoSuchFileException:
-            # Deletion of the flag can also signal kill.
-            killed = True
+            # The kill flag file doesn't exist yet.
+            killed = False
         return killed
 
     # Helper methods for subclasses
--- a/src/toil/jobStores/abstractJobStore.py
+++ b/src/toil/jobStores/abstractJobStore.py
@@ -44,7 +44,7 @@ from uuid import uuid4
 from requests.exceptions import HTTPError
 from typing_extensions import Literal, overload
 
-from toil.common import Config, safeUnpickleFromStream
+from toil.common import Config, safeUnpickleFromStream, getNodeID
 from toil.fileStores import FileID
 from toil.job import (
     CheckpointJobDescription,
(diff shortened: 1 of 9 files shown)


src/toil/leader.py
```diff
@@ -959,16 +970,39 @@ def _getStatusHint(self):
...
 
-    def _reportWorkflowStatus(self):
+    def _reportWorkflowStatus(self) -> None:
         """
         Report the current status of the workflow to the user.
         """
 
         # For now just log our scheduling status message to the log.
-        # TODO: make this update fast enought to put it in the progress
+        # TODO: make this update fast enough to put it in the progress
```
DailyDreaming, line 979: :100:

src/toil/utils/toilKill.py
```diff
@@ -25,28 +26,47 @@
...
+    if local_leader:
+        try:
+            os.kill(pid_to_kill, signal.SIGKILL if options.force else signal.SIGTERM)
+            logger.info("Toil process %i successfully terminated.", pid_to_kill)
+        except OSError:
+            logger.error("Could not signal process %i. Is it still running?", pid_to_kill)
+            sys.exit(1)
+    else:
+        # Flip the flag inside the job store to signal kill
+        with job_store.write_shared_file_stream("_toil_kill_flag") as f:
```
DailyDreaming, line 70: Suggestion: Centralize the `"_toil_kill_flag"` file name with an environment variable like `LEADER_KILL_FLAG_FILENAME` and import it.

Also consider writing the function `_write_kill_flag_file(self)` to have a `kill=True`/`kill=False` input var, and reuse here.

src/toil/utils/toilKill.py
```diff
@@ -25,28 +26,47 @@
...
+    # Get the leader PID
     with job_store.read_shared_file_stream("pid.log") as f:
         pid_to_kill = int(f.read().strip())
 
-    # TODO: We assume this is a local PID so we can't kill the leader if the
-    #  process is on a different machine. We also rely on the leader to be
-    #  responsive to clean up all its jobs.
+    # Check if the leader is on the same machine
+    with job_store.read_shared_file_stream("leader_node_id.log") as f:
+        leader_node_id = f.read().decode('utf-8').strip()
```
DailyDreaming, line 49: Could this be a reciprocal `read_node_id_file()` function alongside `write_node_id_file()` above?

src/toil/leader.py
```diff
@@ -959,16 +970,39 @@ def _getStatusHint(self):
...
         # bar/status line.
         logger.info(self._getStatusHint())
 
+    def _write_kill_flag_file(self) -> None:
+        """
+        Write a file inside the job store that serves as a kill flag.
```
DailyDreaming, line 985: Consider including a sentence to explain that writing `NO` is the initialized file, should only be changed when the user runs the `kill` command, and that this triggers a kill of the leader process (as opposed to a worker).

adamnovak:
@w-gao Are you working on any changes for this or should we just merge it as is, since it's approved?

src/toil/utils/toilKill.py
w-gao: Correct me if I'm wrong but I feel like using an environment variable for this might not be appropriate? If users are allowed to change this, `toil kill` and the leader might not be working with the same file and the kill won't have any effect, especially if they are running `toil kill` from a different machine.

I like the idea of reusing these functions, although I think we'd have to move them to the job store since `toil kill` won't have a Leader instance.

w-gao:
> @w-gao Are you working on any changes for this or should we just merge it as is, since it's approved?

Please don't merge it yet, I'm still experimenting with changes.

Lon suggests that we reuse the `_write_kill_flag_file()` and `_read_kill_flag_file()` functions (and for reading/writing the leader node ID), which makes sense, but since `toil kill` doesn't have access to the leader, I'm thinking about moving them to the job store?

Let me know if you think these functions would belong to the job store.

src/toil/utils/toilKill.py
adamnovak: I also suspect an environment variable might not really be useful here, except maybe as a way to run a workflow that itself pathologically wants to use the filename `_toil_kill_flag`.

But it would make sense to use a constant for this instead of typing the same string in two places, and also to encapsulate this in a `write_kill_flag(True/False)` and `read_kill_flag()`. I guess those would go on the `AbstractJobStore`?

src/toil/utils/toilKill.py
adamnovak: Maybe. It would be weird for it to be a member of `toil.common.Toil` though. Maybe just in this file?

adamnovak:
Yeah, I think if we make functions to work with the kill flag and the leader node ID, we'd put them on the AbstractJobStore. Or maybe make them free-floating and pass a JobStore. And they wouldn't have the leading underscores if it were up to me, since they'd be intended for use outside the class.

src/toil/test/utils/utilsTest.py
```diff
@@ -300,8 +300,8 @@ def testGetPIDStatus(self):
         wf.wait()
         self.check_status('COMPLETED', status_fn=ToilStatus.getPIDStatus)
 
-        # TODO: we need to reach into the FileJobStore's files and
-        # delete this shared file. We assume we know its internal layout.
+        # TODO: we need to reach into the FileJobStore's files and delete this
+        #  shared file. We assume we know its internal layout.
         os.remove(os.path.join(self.toilDir, 'files/shared/pid.log'))
```
w-gao, line 305: We need a `delete_shared_file()` method in the job store to refactor this.

src/toil/utils/toilKill.py
DailyDreaming: My mistake, I did not mean an environment variable, but a constant variable.  File names can change and it's just good to have it in one place

src/toil/utils/toilKill.py
DailyDreaming: That makes sense to me.

src/toil/test/utils/utilsTest.py
adamnovak: This makes sense to leave as is; we can fix the test hack if we ever do add that method.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1449,6 +1450,71 @@ def read_logs(self, callback: Callable[..., Any], read_all: bool = False) -> int
...
+    def read_leader_pid_file(self) -> int:
+        """
+        Read the pid of the leader process to a file in the job store.
+
+        :raise NoSuchFileException: If the PID file doesn't exist.
+        """
+        with self.read_shared_file_stream('pid.log') as f:
+            return int(f.read().strip())
+
+    def write_leader_node_id(self) -> None:
```
adamnovak, line 1475: The names are a bit weird; the PID file is a `_file`, but this is just an `_id`. Maybe it should also have a `_file` suffix? Or maybe we should drop it from the PID functions so we get `write_leader_pid` and `write_leader_node_id`?

src/toil/utils/toilKill.py
```diff
@@ -25,28 +26,48 @@
...
-        logger.info("Toil process %i successfully terminated.", pid_to_kill)
-    except OSError:
-        logger.error("Toil process %i could not be terminated.", pid_to_kill)
-        raise
+    # Check if the leader is on the same machine
+    leader_node_id = job_store.read_leader_node_id()
+    local_leader = leader_node_id == getNodeID()
+
+    if local_leader:
+        # Check if we can send signal to the leader. If not, process might be in
```
adamnovak, line 55:
```suggestion
        # Check if we can send signals to the leader. If not, process might be in
```

adamnovak (changes requested):
Now I want the function names to be more consistent, and I think I found a typo. But other than that this is good!

PR #4085, opened 2022-04-27
Issues/3514 missing permission warning aws

Hexotical:
## Changelog Entry

Add code for warning check to be used when launching cluster with AWS.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,140 @@
...
+        ))
+
+        if role_policy["PolicyDocument"]["Statement"][0]["Effect"] == "Allow":
+            if role_policy["PolicyDocument"]["Statement"][0]["Resource"] not in allowed_actions.keys():
+                allowed_actions[role_policy["PolicyDocument"]["Statement"][0]["Resource"]] = []
+
+            allowed_actions[role_policy["PolicyDocument"]["Statement"][0]["Resource"]].append(
+                role_policy["PolicyDocument"]["Statement"][0]["Action"])
+
+    check_policy_warnings(allowed_actions)
```
adamnovak, line 132: I don't think this call should be here; we just want to return the permission grants and we will check them against something later, right?

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,140 @@
...
+    aws = AWSConnectionManager()
+
+    region = zone_to_region(get_best_aws_zone() or "us-west-2a" )
+
+
+    iam: IAMClient = cast(IAMClient, aws.client(region, 'iam'))
+
+    response = iam.get_instance_profile(InstanceProfileName="fakename_toil")
+
+    role_name = response['InstanceProfile']['Roles'][0]['RoleName']
```
adamnovak, line 101: Is this somehow going to be the caller's role? Why?

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,140 @@
...
+
+        if role_policy["PolicyDocument"]["Statement"][0]["Effect"] == "Allow":
+            if role_policy["PolicyDocument"]["Statement"][0]["Resource"] not in allowed_actions.keys():
+                allowed_actions[role_policy["PolicyDocument"]["Statement"][0]["Resource"]] = []
+
+            allowed_actions[role_policy["PolicyDocument"]["Statement"][0]["Resource"]].append(
+                role_policy["PolicyDocument"]["Statement"][0]["Action"])
+
+    check_policy_warnings(allowed_actions)
+    return allowed_actions
```
adamnovak, line 133: Once we handle all the actions allowed by policies on the role, we might want to handle actions allowed by policies on the user directly, if there are any. I think I can have an AWS account with a user in it, and have the user not have any role, but attach policies to the user to give them rights. Or can I not do that?

Do we also maybe have to deal with free-floating policies in the account? Or do policies have to be attached to something?

contrib/admin/mypy-with-ignore.py
```diff
@@ -91,6 +91,7 @@ def main():
         'src/toil/lib/encryption/conftest.py',
         'src/toil/lib/encryption/__init__.py',
         'src/toil/lib/aws/__init__.py',
+        #'src/toil/lib/aws/iam.py',
```
adamnovak, line 94: We shouldn't add this commented-out line.
```suggestion
```

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Check whether necessary permissions are permitted for AWS
+
+    :param allowed_actions: Dictionary containing actions allowed by resource
+    :param launching_perms: Set of required actions to launch a cluster on AWS
+    """
+    permissions = [x for x in launching_perms if helper_permission_check(x, allowed_actions["*"])]
+
+    if not launching_perms.issubset(set(permissions)):
+        raise RuntimeError("Missing permissions", permissions)
```
adamnovak, line 40: This is raising an error and not raising a warning or logging at warning level. So if it appears that we don't have e.g. `iam:DeleteRole` granted on `*`, Toil will bail out and refuse to create the cluster, with no way to override it.

I don't think that's the behavior we want. For one thing, we aren't 100% confident that we're actually testing all the ways a user could have permission to do something, so there will be cases where we do have the permissions we need but we don't know it.

We're also checking for permission grants on `*` (i.e. all resources); there might be ways to grant a permission on less than `*` that will let Toil work, but right now Toil can't understand them and will bail out if it sees them.

We probably want the user to still be able to use Toil in those cases. That could look like some kind of `--force-permissions` flag or something to override the checks, or it could look like just making the checks produce warning-level log messages when they fail, and going on to attempt the things we think we can't do anyway. Or we could pause and make the user hit `y` or something.

src/toil/utils/toilLaunchCluster.py
```diff
@@ -106,6 +108,10 @@ def main() -> None:
                              "the extra security groups do not have the same name as the cluster name.")
     options = parser.parse_args()
     set_logging_from_options(options)
+
+    #TODO implement permission check aws
+    check_policy_warnings(get_allowed_actions())
```
adamnovak, line 113: I think that `options.zone` needs to feed in here.

Also, if the user passed `--ec2InstanceProfileArn` already, we need different permissions: we don't need to be able to create arbitrary roles and attach policies, but we do need to be able to assign that particular instance profile to instances. So information about that value probably should come through here as well (at least eventually). As a hack we could just skip the check if that option is used.

adamnovak (changes requested):
OK, I looked through this, and it seems like we have a core engine here to check the permissions and see if the actions we want to do are going to be allowed.

But I'm not convinced we're actually managing to fetch the permission grants that apply to the caller, and it looks like we're just checking some of the IAM permissions right now and we don't actually have anything hooked up for the EC2 permissions yet (which is fine but means we'd still need to do more work to fully close the issue).

And I found some other things that look like they need to be threaded through still, like the zone we are supposed to be deploying the cluster in.

src/toil/test/lib/aws/test_iam.py
```diff
@@ -0,0 +1,62 @@
...
+        try:
+            iam.check_policy_warnings(launch_tester)
+        except RuntimeError:
+            pass
+
+    def test_allowed_actions(self):
+        allowed = iam.get_allowed_actions()
+        for action in allowed:
+            logger.info("allowed %s", action)
+        #logger.info(allowed)
```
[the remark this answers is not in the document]
Hexotical: yeah basically

src/toil/lib/aws/iam.py
Hexotical: During one of my pair programming sessions with Lon we were testing ways to get associated permissions, and determined we couldn't do it through the user without asking for more parameters because to get the permissions directly attached to a user we need a username and a policy name so instead we got the role associated with the instance profile and check the permissions on the role

src/toil/lib/aws/iam.py
Hexotical: policies haee to be attatched to a policy document, but I couldn't ifnd a way to find an associated policy name with the document associated with the user account

src/toil/test/provisioners/aws/utilProvisionerTest.py
```diff
@@ -0,0 +1,86 @@
...
+    provisioner = AWSProvisioner('fakename', 'mesos', "us-west-2a", 10000, None, None)
+
+
+
+    botocore_session = get_session()
+    botocore_session.get_component('credential_provider').get_provider(
+        'assume-role').cache = JSONFileCache()
+    session = Session(botocore_session=botocore_session)
+
+    client = session.client('iam')
```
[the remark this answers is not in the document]
Hexotical: yeah essentially

src/toil/lib/aws/iam.py
adamnovak: This produces a role name even if `fakename_toil` doesn't exist as an instance profile, right? Some kind of default role? And then you are checking the permissions on that role.

But it doesn't seem to me that the permissions on that default role are particularly likely to also be the permissions that `toil launch-cluster` has when it is running. Do you have a justification for thinking that that I don't know about? (Maybe when you run this from an EC2 machine that default role is always the same role as the current machine?) Or are we just checking the permissions of some random role instead of what we are supposed to accomplish, which is checking the permissions that are currently granted to us?

adamnovak:
I am still due to re-review this.

I think the CI failures are #4168 so those aren't really the fault of this PR.

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,182 @@
...
+                                  "iam:ListAttatchedRolePolicies",
+                                  "iam:ListPolicies",
+                                  "iam:ListRoleTags",
+                                  "iam:PutRolePolicy",
+                                  "iam:RemoveRoleFromInstanceProfile",
+                                  "iam:TagRole"
+                                  }
+
+
+def check_policy_warnings(allowed_actions: Dict[str, List[str]] = {'*': []}, launching_perms: Set[str] = _CLUSTER_LAUNCHING_PERMISSIONS) -> bool:
```
adamnovak, line 30: This function should have a better name.

If I see `y = check_policy_warnings(x)`, then `y` is probably a `bool`, but I'm going to assume that `x` somehow represents "policy warnings" that need to be checked. And that's not what's happening at all.

If you want to usually call it with only the one argument, you might want to call it something like `permission_grants_allow_launching_cluster`. Personally I think it might be even better to have the user import both the function and the permissions list, so the calls would look something like:

```
if not permission_grants_allow(grants, CLUSTER_LAUNCHING_PERMISSIONS):
    logger.warning('You may not have sufficient permissions!')
```

It's very clear what that code is doing, versus:

```
if not check_policy_warnings(allowed_actions):
    # Does the reader have any idea what the code now knows? If the sense of the if statement was backward, how quickly would you be able to tell?
```

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,182 @@
...
+    Check whether necessary permissions are permitted for AWS
+
+    :param allowed_actions: Dictionary containing actions allowed by resource
+    :param launching_perms: Set of required actions to launch a cluster on AWS
+    """
+    permissions = [x for x in launching_perms if check_permission_allowed(x, allowed_actions["*"])]
+
+    if not launching_perms.issubset(set(permissions)):
+        for perm in permissions:
+            logger.warning('Permission %s is missing', perm)
```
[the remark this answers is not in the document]
adamnovak: If you made the function `get_missing_permissions(grants, required_permissions)`, then you could test that you were getting the right missing permissions to warn about.

src/toil/lib/aws/iam.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def check_policy_warnings(allowed_actions: Dict[str, List[str]] = {'*': []}, launching_perms: Set[str] = _CLUSTER_LAUNCHING_PERMISSIONS) -> bool:
+    """
+    Check whether necessary permissions are permitted for AWS
+
+    :param allowed_actions: Dictionary containing actions allowed by resource
+    :param launching_perms: Set of required actions to launch a cluster on AWS
+    """
+    permissions = [x for x in launching_perms if check_permission_allowed(x, allowed_actions["*"])]
```
adamnovak, line 37: Here's where we're getting the requirement that all the grants be on the `*` resource. We should probably note that in a comment, so that when we come back to change it later we know where to start.

src/toil/test/lib/aws/test_iam.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_negative_permissions_iam(self):
+        launch_tester = {'*': ['ec2:*',  's3:*', 'sdb:*']}
+        assert iam.check_policy_warnings(launch_tester) is False
+    '''
+    def test_allowed_actions(self):
+        allowed = iam.get_allowed_actions()
+        for action in allowed:
+            logger.info("allowed %s", action)
+    '''
```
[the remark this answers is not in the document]
adamnovak: Also, I think we probably do need to try calling `get_allowed_actions()` in a test that `@needs_aws`, so we can make sure that we can in fact fetch the grants.

adamnovak (changes requested):
This PR has improved!

We aren't going to bail out on all of Toil when we think we might not have the permissions, and it looks like there's fallback from user permissions to role permissions in a way that ought to work.

But I don't think the tests are sufficient; they should exercise the talking-to-IAM logic and more exhaustively test the edge cases of the permission wildcard matching, in a way that wouldn't be passable by trivial code that didn't actually work.

I misunderstood what the tests thought they were doing calling `check_policy_warnings` for about 20 minutes, which I think means that things have either the wrong names or not enough comments.

Also, I think the core permission-vs-wildcard-pattern check could delegate to Python's `fnmatch` and save a lot of code.

contrib/admin/mypy-with-ignore.py
```diff
@@ -91,6 +91,7 @@ def main():
         'src/toil/lib/encryption/conftest.py',
         'src/toil/lib/encryption/__init__.py',
         'src/toil/lib/aws/__init__.py',
+        'src/toil/lib/aws/iam.py',
```
adamnovak, line 94: I still don't want to merge this with this in here. When I take it out, here's what I get from `make mypy`:

```
/home/anovak/workspace/toil/contrib/admin/mypy-with-ignore.py
src/toil/lib/aws/iam.py: note: In function "policy_permissions_allow":
src/toil/lib/aws/iam.py:54:116: error: Incompatible default for argument
"given_permissions" (default has type "Dict[<nothing>, <nothing>]", argument has
type "Set[str]")  [assignment]
    ...tr, List[str]] = {'*': []}, given_permissions: Set[str] = {}) -> bool:
                                                                 ^
src/toil/lib/aws/iam.py: note: In function "allowed_actions_roles":
src/toil/lib/aws/iam.py:103:50: error: Argument 1 to "dict" has incompatible
type "GetRolePolicyResponseTypeDef"; expected
"SupportsKeysAndGetItem[str, List[str]]"  [arg-type]
            role_policy: Dict[str, List[str]] = dict(iam.get_role_policy(
                                                     ^
src/toil/lib/aws/iam.py:103:50: note: Following member(s) of "GetRolePolicyResponseTypeDef" have conflicts:
src/toil/lib/aws/iam.py:103:50: note:     Expected:
src/toil/lib/aws/iam.py:103:50: note:         def __getitem__(self, str) -> List[str]
src/toil/lib/aws/iam.py:103:50: note:     Got:
src/toil/lib/aws/iam.py:103:50: note:         def __getitem__(self, str) -> object
src/toil/lib/aws/iam.py:108:26: error: No overload variant of "__getitem__" of
"list" matches argument type "str"  [call-overload]
            for statement in role_policy["PolicyDocument"]["Statement"]:
                             ^
src/toil/lib/aws/iam.py:108:26: note: Possible overload variants:
src/toil/lib/aws/iam.py:108:26: note:     def __getitem__(self, SupportsIndex) -> str
src/toil/lib/aws/iam.py:108:26: note:     def __getitem__(self, slice) -> List[str]
src/toil/lib/aws/iam.py:113:54: error: Argument 1 to "append" of "list" has
incompatible type "List[str]"; expected "str"  [arg-type]
    ...                allowed_actions[resource].append(role_policy[statement...
                                                        ^
src/toil/lib/aws/iam.py: note: In function "get_aws_account_num":
src/toil/lib/aws/iam.py:178:12: error: Incompatible return value type (got
"Optional[str]", expected "str")  [return-value]
        return boto3.client('sts').get_caller_identity().get('Account')
               ^
Found 5 errors in 1 file (checked 62 source files)
make: *** [Makefile:250: mypy] Error 1
```

I think we can probably sort these out with around half an hour of work.

adamnovak (changes requested):
This is good now overall, but I don't want to merge it until we can get the typing right, because we want the number of ignored files to only decrease.

adamnovak (approved):
I think this looks good.

It could have some tests for fetching the policies, but those would be hard to set up for in a portable way.

adamnovak (approved):

adamnovak:
@Hexotical I think you have the argument order wrong when you call `client` AKA `get_client` and the AWS service name and zone are being swapped. THe CI tests that launch clusters seem to be failing to do so.

Can you fix that and launch a cluster manually to confirm it actually works?

adamnovak:
It looks like some CWL conformance tests timed out when running against Kuberentes. This PR shouldn't actually affect that; I think maybe our Kubernetes was too busy and our tests can't handle that happening.

https://ucsc-ci.com/databiosphere/toil/-/jobs/20491#L2558

I am going to rerun the offending tests.

PR #4105, opened 2022-05-19
Contrib hook scripts

### LINKED ISSUE #4104: I keep pushing commits that fail CI due to MyPy
I keep pushing commits that fail CI. I would have caught it if I had run MyPy, but I do not run MyPy all the time because it is slow (especially on Courtyard) and I am lazy.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1171)
┆friendlyId: TOIL-1171

adamnovak:
I got tired of pushing commits that don't type-check, so I wrote some scripts I can use as git hooks to fix #4104.

This should do it without me generally needing to wait more than 10 seconds for MyPy to happen, and without me needing to actually bother to run MyPy all the time. 

## Changelog Entry

 * There are some potentially useful git hook scripts in `contrib/hooks`

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/requirements-cwl.txt
+++ b/requirements-cwl.txt
@@ -1,4 +1,4 @@
-cwltool==3.1.20220406080846
+cwltool==3.1.20220502060230
 galaxy-tool-util
 ruamel.yaml>=0.15,<=0.17.21
 ruamel.yaml.clib>=0.2.6
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,7 +6,7 @@ stubserver>=1.1,<2
 setuptools>=51.3.3,<63
 sphinx>=4,<5
 cwltest>=2.2.20211116163652
-mypy==0.941
+mypy==0.960
 types-requests
 types-setuptools
 types-boto
--- a/requirements-google.txt
+++ b/requirements-google.txt
@@ -1,2 +1,2 @@
 apache-libcloud>=2.2.1,<3
-google-cloud-storage==1.6.0
+google-cloud-storage>=1.6.0,<2


contrib/hooks/lib.py
```diff
@@ -0,0 +1,139 @@
...
+
+def announce(message):
+    sys.stderr.flush()
+    sys.stderr.write(message)
+    sys.stderr.write('\n')
+    sys.stderr.flush()
+
+cached_hostname = None
+def get_hostname() -> str:
+    global cached_hostname
```
DailyDreaming, line 27: Suggestion: use `functools.lru_cache()` here instead of a global, and a warning log message if it fails.  i.e.

```
@functools.lru_cache
def get_hostname() -> str:
    try:
        return socket.getfqdn()
    except Exception as e:
        warning.warn(f'An error occurred while attempting to describe the host: {e}')
    return 'localhost'

File: contrib/hooks/lib.py [hunk 2]
```diff
@@ -0,0 +1,139 @@
...
+    if 'SSH_TTY' in os.environ:
+        protocol = 'sftp'
+    else:
+        protocol = 'file'
+    
+    realpath = os.path.realpath(path)
+    if not text:
+        text = path    
+        
+    return f'\033]8;;{protocol}://{get_hostname()}{realpath}\033\\{text}\033]8;;\033\\'
```
DailyDreaming, line 47: Could there be a comment here about the purpose of the `\033]8;;` / `\033]8;;\033\\` prefix/suffixes?

I found this, so assume it is affecting/minimizing terminal size for some reason?  https://stackoverflow.com/questions/68931529/printf-0338widthheightt-works-for-a-specific-height-and-width-but-is-t

DailyDreaming (approved):
Oh, I like this PR.  Thanks for doing this!  Only minor comments, otherwise this looks great.

Suggestion: include some information about these scripts in the contributing doc.

contrib/hooks/lib.py
adamnovak: I didn't want to use this because it's fiddly between versions; what you wrote won't work on 3.7, for example, because the max size (or maybe just some parens) are required.

But I can swap over.

contrib/hooks/lib.py
adamnovak: This is making [hyperlinks in the terminal with OSC 8](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).

I've rearranged this function to make it less magic and actually explain what it is up to and how the sequences fit together.

adamnovak:
Lon thinks we could mention these in CONTRIBUTING.md.

Issue #4111, opened 2022-05-20
Celery tasks end with an exception when canceled instead of finishing properly


adamnovak:
I keep seeing WES server logs like this:
```
Fri, 20 May 2022 16:12:13 -0700	[toil] INFO:werkzeug:10.0.112.123 - - [20/May/2022 23:12:13] "POST /ga4gh/wes/v1/runs/run-e294bea6d7364c6da1b1d68b37574d3b/cancel HTTP/1.1" 200 -
Fri, 20 May 2022 16:12:13 -0700	[celery] [2022-05-20 23:12:13,933: INFO/MainProcess] Terminating run-e294bea6d7364c6da1b1d68b37574d3b (Signals.SIGUSR1)
Fri, 20 May 2022 16:12:13 -0700	[celery] [2022-05-20 23:12:13,934: WARNING/ForkPoolWorker-3] SoftTimeLimitExceeded()
Fri, 20 May 2022 16:12:17 -0700	[toil] INFO:werkzeug:10.0.155.56 - - [20/May/2022 23:12:17] "GET /engine/v1/status HTTP/1.1" 200 -
Fri, 20 May 2022 16:12:18 -0700	[toil] INFO:werkzeug:10.0.174.246 - - [20/May/2022 23:12:18] "GET /engine/v1/status HTTP/1.1" 200 -
Fri, 20 May 2022 16:12:18 -0700	[toil] INFO:werkzeug:10.0.112.123 - - [20/May/2022 23:12:18] "GET /engine/v1/status HTTP/1.1" 200 -
[...]

Issue #4109, opened 2022-05-20
S3 State Store for WES server sticks a leading slash into key names instead of storing state where it is supposed to go


adamnovak:
Under AGC I am getting my state stored at e.g. `s3://agc-318423852362-us-west-2//project/CWLDemo/userid/anovak4rBHiA/context/myContext/toil-execution/state/run-00190a13ef2d41ba8e0ce2e15499b186`. Note the two slashes after the bucket name.

We should fix this before AGC release or else we'll need code to migrate from there.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1174)
┆friendlyId: TOIL-1174

adamnovak:
This should be fixed now.

PR #4107, opened 2022-05-19
Update Cactus and exclude broken networkx

adamnovak:
This should fix #4106 by pinning what probably was a working networkx version.

## Changelog Entry
To be copied to the [draft changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog) by merger:

 * Toil now excludes the broken NetworkX 2.8.1 release, which breaks pickling and the Cactus integration tests

## Reviewer Checklist

 * [ ] Make sure it is coming from `issues/XXXX-fix-the-thing` in the Toil repo, or from an external repo.
    * [ ] If it is coming from an external repo, make sure to pull it in for CI with:
        ```
        contrib/admin/test-pr otheruser theirbranchname issues/XXXX-fix-the-thing
        ```
    * [ ] If there is no associated issue, [create one](https://github.com/DataBiosphere/toil/issues/new).
* [ ] Read through the code changes. Make sure that it doesn't have:
    * [ ] Addition of trailing whitespace.
    * [ ] New variable or member names in `camelCase` that want to be in `snake_case`.
    * [ ] New functions without [type hints](https://docs.python.org/3/library/typing.html).
    * [ ] New functions or classes without informative docstrings.
    * [ ] Changes to semantics not reflected in the relevant docstrings.
    * [ ] New or changed command line options for Toil workflows that are not reflected in `docs/running/{cliOptions,cwl,wdl}.rst` 
    * [ ] New features without tests.
* [ ] Comment on the lines of code where problems exist with a review comment. You can shift-click the line numbers in the diff to select multiple lines.
* [ ] Finish the review with an overall description of your opinion.

## Merger Checklist

* [ ] Make sure the PR passes tests.
* [ ] Make sure the PR has been reviewed **since its last modification**. If not, review it.
* [ ] Merge with the Github "Squash and merge" feature.
    * [ ] If there are multiple authors' commits, add [Co-authored-by](https://github.blog/2018-01-29-commit-together-with-co-authors/) to give credit to all contributing authors.
* [ ] Copy its recommended changelog entry to the [Draft Changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog).
* [ ] Append the issue number in parentheses to the changelog entry.

Issue #4106, opened 2022-05-19
Toil Cactus integration test failing due to NetworkX bug


adamnovak:
https://ucsc-ci.com/databiosphere/toil/-/jobs/15482#L2236

	Traceback (most recent call last):
	  File "/usr/local/lib/python3.9/dist-packages/toil/worker.py", line 381, in workerScript
	    job = Job.loadJob(jobStore, jobDesc)
	  File "/usr/local/lib/python3.9/dist-packages/toil/job.py", line 2292, in loadJob
	    job = cls._unpickle(userModule, fileHandle, requireInstanceOf=Job)
	  File "/usr/local/lib/python3.9/dist-packages/toil/job.py", line 1918, in _unpickle
	    runnable = unpickler.load()
	  File "/usr/local/lib/python3.9/dist-packages/networkx/classes/reportviews.py", line 1290, in __setstate__
	    self._adjdict = G._pred if hasattr(G, "pred") else G._adj
	AttributeError: 'DiGraph' object has no attribute '_adj'

I think maybe the Cactus we use is hitting https://github.com/networkx/networkx/issues/5655 which was maybe introduced in a new NetworkX release.

We will probably have to ignore the Cactus integration test until Cactus or NetworkX fixes this, and then update to a fixed Cactus.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1172)
┆friendlyId: TOIL-1172

adamnovak:
NetworkX 2.8.1 released yesterday: https://pypi.org/project/networkx/#history

adamnovak:
This should actually be fixed now.

PR #4113, opened 2022-05-23
AGC megabranch

adamnovak:
This PR combines all the stuff we wanted for AGC integration, and brings back a line I deleted in #4107 when fixing the Cactus integration test, which broke it in a different way.

Since we're supposed to merge PRs one at a time, and they take so long to test, and CI has been failing at the last step anyway, we haven't really been able to keep up with fixes for the bugs Amazon is sending us. So hopefully we can merge this in one CI cycle and clear the backlog.

## Changelog Entry

* Cactus integration tests should work again (#4106).

From #4110 which I merged prematurely and had to walk back:
 * Toil no longer stores state in S3 at keys with an extra leading `/`. (#4109)
 * Added more tests for Toil's server mode and its integration with AGC components.
 * Reduced the time to time out CANCELING WES workflows to CANCELED.
 * Actually handle the case where the Toil workflow takes longer than our timeout to clean up after the Celery task is canceled. (#4111).
 
From #4098:
 * Toil `JobStore` implementations now support listing files, and polling files to see if they are directories, for URLs they implement.
 * CWL `Directory` objects can now come from Amazon S3 (#4094)
 * S3 access can better work around not being allowed to `GetBucketLocation` (#4095)
 * S3 file import works even if Toil is running in a Virtual Private Cloud with a special local-network S3 Endpoint set up, and the source bucket is in a different region from Toil and the job store bucket.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/cwl/cwltoil.py
```diff
@@ -1116,6 +1224,8 @@ def size(self, path: str) -> int:
         # prevent multiple downloads as in
         # https://github.com/DataBiosphere/toil/issues/3665
         if path.startswith("toilfile:"):
+            if self.file_store is None:
+                raise RuntimeError("URL requires a file store:" + path)
```
w-gao, line 1228:
```suggestion
                raise RuntimeError("URL requires a file store: " + path)
```

src/toil/jobStores/aws/jobStore.py
```diff
@@ -430,28 +432,38 @@ def get_empty_file_store_id(self, jobStoreID=None, cleanup=False, basename=None)
...
+                    info = self.FileInfo.loadOrCreate(jobStoreFileID=jobStoreFileID,
+                                                      ownerID=str(self.sharedFileOwnerID),
+                                                      encrypted=None)
+                info.copyFrom(srcObj)
+                info.save()
+                return FileID(info.fileID, size) if shared_file_name is None else None
+        except ServerSideCopyProhibitedError:
+            # AWS refuses to do this copy for us
+            logger.warning("Falling back to copying via the local machine. This could get expensive!")
+            pass
```
w-gao, line 453: Also line 466, does `pass` do anything here?

src/toil/lib/aws/utils.py
```diff
@@ -205,18 +205,87 @@ def create_s3_bucket(
...
+        """
+        # If that also doesn't work, we can try HEAD-ing the bucket and looking
+        # for an 'x-amz-bucket-region' header on the response, which can tell
+        # us where the bucket is. See
+        # <https://github.com/aws/aws-sdk-cpp/issues/844#issuecomment-383747871>
+        # This won't typecheck until https://github.com/youtype/mypy_boto3_builder/issues/147 is fixed.
+        info = s3_client.head_bucket(Bucket=bucket_name)  # type: ignore
+        return cast(str, info['ResponseMetadata']['HTTPHeaders']['x-amz-bucket-region'])
+
+    # Compose a list of strategise we want to try in order, which may work.
```
w-gao, line 255: `strategies`

src/toil/lib/aws/utils.py
```diff
@@ -232,9 +301,9 @@ def get_object_for_url(url: ParseResult, existing: Optional[bool] = None) -> Obj
         """
         s3_resource = cast(S3ServiceResource, session.resource('s3'))
```
w-gao, line 302: This looks like it's unused since there is `s3` below?

w-gao (approved):
This looks pretty good to me and it seems like Lon's comments in #4098 have been addressed as well! I also reviewed the changes in #4110 and they look great to me. Just a few minor comments.

The Cactus integration test has passed as well!

src/toil/jobStores/aws/jobStore.py
adamnovak: Not really; it's not needed to hold the block in existence because there's a logging line there.

I think I put it in to make it clear that I mean to discard this error, and so that if I decided not to have that logging line after all the code would still work.

src/toil/lib/aws/utils.py
adamnovak:
```suggestion
    # Compose a list of strategies we want to try in order, which may work.
```

src/toil/lib/aws/utils.py
adamnovak: Hm. We probably don't need this.
```suggestion
```

src/toil/jobStores/aws/jobStore.py
w-gao: That makes sense.

PR #4172, opened 2022-08-03
Do not submit empty jobs

jsmedmar:
Thanks again for such a great tool.

Here I'm proposing a change, which I don't think its production ready, for a `leader.py` `TODO` item. When Toil finishes jobs, it walks up the DAG submitting empty jobs (jobs that actually do no work) requesting the original resources. Provisioning cloud machines for these jobs is not great.

Here I'm proposing to delete these jobs directly in `leader.py`. Full disclaimer, my knowledge of Toil's inner workings is limited, so this should be considered as a starting point. 

All the best

-----------
I had created an issue about this before: https://github.com/DataBiosphere/toil/issues/3131

### DIFF
--- a/src/toil/leader.py
+++ b/src/toil/leader.py
@@ -571,18 +571,22 @@ class Leader:
                          readyJob)
             self.serviceManager.kill_services(self.toilState.servicesIssued[readyJob.jobStoreID], error=False)
         else:
-            #There are no remaining tasks to schedule within the job, but
-            #we schedule it anyway to allow it to be deleted. Remove the job
+            # There are no remaining tasks to schedule within the job
 
+            # previous suggestions -- resizing works, or delete here... this branch is deleting the job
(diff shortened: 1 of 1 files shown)


mr-c:
Thank you for your PR, @jsmedmar ! I've copied it into our repo for testing at https://github.com/DataBiosphere/toil/tree/issues/3131-do-not-submit-empty-jobs , so check https://ucsc-ci.com/databiosphere/toil/-/pipelines/2001 for results from the CI

adamnovak:
It looks like this is a patch submitted from a long-running fork that branched off before the JobNode/JobGraph -> Job/JobDescription refactor, and so the patch doesn't work anymore. We ought to be able to port it over though.

adamnovak:
I think #4259 integrated this but Github didn't pick it up for some reason.

PR #4188, opened 2022-08-24
Slurm testing

Hexotical:
## Changelog Entry

Adds basic testing for slurm using a slurm docker cluster by running sample workflows.
 

## Reviewer Checklist

        ```
        ```

## Merger Checklist


.gitlab-ci.yml
```diff
@@ -179,6 +179,18 @@ batch_systems:
     - make test tests=src/toil/test/mesos/MesosDataStructuresTest.py
     - kill $(jobs -p) || true
 
+slurm_test:
+  image:
+    name: "quay.io/vgteam/vg_ci_prebake:latest"
+    entrypoint: ["/usr/bin/env"]
```
adamnovak, line 185: Do we still need to override the image here if we aren't doing anything fancy in the script section anymore and we're instead just calling a full Bash script?

Because Lon changed the default image for the tests to be a Toil-specific prebake one with dependencies installed, and dropped the dependency install code from the Gitlab setup steps, so the image specified here might not work anymore.

```suggestion
```

contrib/slurm-test/slurm_test.sh
```diff
@@ -0,0 +1,21 @@
...
+docker exec slurm-test_slurmmaster_1 sudo apt install python3-pip -y
+docker exec slurm-test_slurmmaster_1 pip3 install "git+https://github.com/DataBiosphere/toil.git@${GIT_COMMIT}"
+docker exec slurm-test_slurmmaster_1 sinfo -N -l
+docker exec slurm-test_slurmmaster_1 python3 /home/admin/toil_workflow.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
+docker cp slurm-test_slurmmaster_1:/home/admin/output.txt output_Docker.txt
+docker exec slurm-test_slurmmaster_1 python3 /home/admin/sort.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
+docker cp slurm-test_slurmmaster_1:/home/admin/sortedFile.txt sortedFile.txt
+docker-compose stop
+./check_out.sh
+rm sort.py
```
adamnovak, line 20: Having the Gitlab script copy sort.py into here, and then deleting it, makes the setup very tightly coupled: you can *only* run this script as part of the Gitlab script. If you run it on its own, manually, it will complain that sort.py isn't there. If you provide it, it will run once, but delete it, so you can't run it again to test something else without providing a new copy of sort.py.

If the test fails, sort.py sticks around, untracked and un-cleaned-up.

I would recommend symlinking sort.py into here from its normal location, so there can be just one copy, but so that it will also be available here. I am pretty sure `docker cp` will follow the symlink to get the real sort.py. Then you won't have to deal with copying the file around, or with wanting to clean up afterward.

contrib/slurm-test/slurm_test.sh
```diff
@@ -0,0 +1,21 @@
...
+docker cp toil_workflow.py slurm-test_slurmmaster_1:/home/admin
+#Assumes sort.py copied to this directory during gitlab run for ease of access
+docker cp sort.py slurm-test_slurmmaster_1:/home/admin
+docker cp fileToSort.txt slurm-test_slurmmaster_1:/home/admin
+docker cp toil_workflow.py slurm-test_slurmmaster_1:/home/admin
+GIT_COMMIT=$(git rev-parse HEAD)
+docker exec slurm-test_slurmmaster_1 sudo apt install python3-pip -y
+docker exec slurm-test_slurmmaster_1 pip3 install "git+https://github.com/DataBiosphere/toil.git@${GIT_COMMIT}"
+docker exec slurm-test_slurmmaster_1 sinfo -N -l
+docker exec slurm-test_slurmmaster_1 python3 /home/admin/toil_workflow.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
```
adamnovak, line 14:
```suggestion
# Test 1: make sure a very simple workflow runs under Slurm
docker exec slurm-test_slurmmaster_1 python3 /home/admin/toil_workflow.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
```

contrib/slurm-test/slurm_test.sh
```diff
@@ -0,0 +1,21 @@
...
+docker cp slurm-test_slurmmaster_1:/home/admin/output.txt output_Docker.txt
+docker exec slurm-test_slurmmaster_1 python3 /home/admin/sort.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
```
adamnovak, line 16:
```suggestion
# Test 2: make sure the "sort" workflow runs under Slurm
docker exec slurm-test_slurmmaster_1 python3 /home/admin/sort.py file:my-job-store --batchSystem slurm --disableCaching --retryCount 0
```

adamnovak (changes requested):
I think we might need to change the image stuff before merging, since Lon already merged the change to use `quay.io/ucsc_cgl/toil_ci_prebake:latest`.

I also suggested some other potential improvements.

At some point we should maybe think about whether all the stuff in `contrib` really belongs there vs. elsewhere.

contrib/slurm-test/slurm_test.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -e
+docker-compose up -d
+docker ps
+docker cp toil_workflow.py slurm-test_slurmmaster_1:/home/admin
+ln -s /toil/src/test/sort/sort.py sort.py
```
adamnovak, line 6: Assuming we run this script here in `<repo root>/contrib/slurm-test`, then this file is 2 levels up to the main Toil directory, and from there in `src/test/sort/sort.py`. It's not going to be at an absolute path under `/toil`, generally.

```suggestion
ln -s ../../src/test/sort/sort.py sort.py
```

Really, though, I would commit the symlink to Git. Run that command from here and then `git add sort.py` and you won't have to worry about creating and destroying symlinks at test runtime.

adamnovak (changes requested):
Looks like the symlink stuff isn't quite going to work yet.

adamnovak (approved):
I would still recommend *committing* a relative symlink, rather than making a new one every time, which isn't much better than copying every time. But it looks like it will work properly this way.

contrib/slurm-test/sort.py
```diff
@@ -0,0 +1 @@
+/home/lukas/Desktop/toil/src/toil/test/sort/sort.py
```
mr-c, line 1: I don't think this symlink will work. It should be relative, not an absolute path

PR #4179, opened 2022-08-17
Move coordination directory to places we can use

adamnovak:
Fixes #4170 by only using /var/run/user if XDG tells us we have it in our
session. Otherwise we will try other places, including /run/lock/toil. We now
make sure to mount /run/lock across form the host on Mesos and Kubernetes,
instead of /var/run/user, so we should end up using that there.

## Changelog Entry
To be copied to the [draft changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog) by merger:

 * PR submitter writes their recommendation for a changelog entry here

## Reviewer Checklist

 * [ ] Make sure it is coming from `issues/XXXX-fix-the-thing` in the Toil repo, or from an external repo.
    * [ ] If it is coming from an external repo, make sure to pull it in for CI with:
        ```
        contrib/admin/test-pr otheruser theirbranchname issues/XXXX-fix-the-thing
        ```
    * [ ] If there is no associated issue, [create one](https://github.com/DataBiosphere/toil/issues/new).
* [ ] Read through the code changes. Make sure that it doesn't have:
    * [ ] Addition of trailing whitespace.
    * [ ] New variable or member names in `camelCase` that want to be in `snake_case`.
    * [ ] New functions without [type hints](https://docs.python.org/3/library/typing.html).
    * [ ] New functions or classes without informative docstrings.
    * [ ] Changes to semantics not reflected in the relevant docstrings.
    * [ ] New or changed command line options for Toil workflows that are not reflected in `docs/running/{cliOptions,cwl,wdl}.rst` 
    * [ ] New features without tests.
* [ ] Comment on the lines of code where problems exist with a review comment. You can shift-click the line numbers in the diff to select multiple lines.
* [ ] Finish the review with an overall description of your opinion.

## Merger Checklist

* [ ] Make sure the PR passes tests.
* [ ] Make sure the PR has been reviewed **since its last modification**. If not, review it.
* [ ] Merge with the Github "Squash and merge" feature.
    * [ ] If there are multiple authors' commits, add [Co-authored-by](https://github.blog/2018-01-29-commit-together-with-co-authors/) to give credit to all contributing authors.
* [ ] Copy its recommended changelog entry to the [Draft Changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog).
* [ ] Append the issue number in parentheses to the changelog entry.

PR #4191, opened 2022-08-25
Make Toil clusters put temporary files in `/var/tmp` by default.

### LINKED ISSUE #4148: Toil clusters have `/tmp` in memory when it should be on disk
As noted in https://github.com/DataBiosphere/toil/issues/4147#issuecomment-1159085124, when you launch a Toil cluster and connect to it, your `/tmp` is a `tmpfs` in-memory filesystem.

It should be an on-disk filesystem instead. We might need to do something with the Flatcar base image to make this work, because we might also need to share it between the Toil appliance container and the host.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1190)
┆friendlyId: TOIL-1190

adamnovak:
This should fix #4148.

It is [the opinion of the `systemd` devs](https://systemd.io/TEMPORARY_DIRECTORIES/) and therefore the incontrovertible truth, that if a file is "big" it has no business being in `/tmp` and really needs to be in `/var/tmp` instead. A `tmpfs`-backed `/tmp` is [a thing Arch shows you how to do](https://wiki.archlinux.org/title/Tmpfs), for example, and also looks like a default `systemd` behavior in some setups.

Neither Python's `tempfile` nor the `TMPDIR`/`TEMP`/`TMP` environment variables understand a concept of tinty-temp-file vs. I-need-to-download-this-big-archive-somewhere temp file. Also, we make large temporary files in dependency code, I think.

Also, if we didn't have a `tmpfs` `/tmp`, we'd want it really to be on our big ephemeral disks anyway, which we keep over in `/var` already.

So this changes Toil to use a `TMPDIR` environment variable of `/var/tmp` on its clusters, and on its Kubernetes pods, by default.

## Changelog Entry

 * Toil clusters now have `/var/tmp` as the default temporary directory, since they often make large temporary files

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -494,10 +494,10 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
             # Provision Toil WorkDir from a HostPath volume, to share with other pods.
             # Create the directory if it doesn't exist already.
             mount_host_path('workdir', self.host_path, self.worker_work_dir, create=True)
-            # We also need to mount across /var/run/user, where we will put
+            # We also need to mount across /run/lock, where we will put
             # per-node coordiantion info.
             # Don't create this; it really should always exist.
-            mount_host_path('coordination', '/var/run/user', '/var/run/user')
+            mount_host_path('coordination', '/run/lock', '/run/lock')
         else:
             # Provision Toil WorkDir as an ephemeral volume
             ephemeral_volume_name = 'workdir'
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -141,6 +141,12 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
             # TODO: Keep this in sync with the Dockerfile.
             self.worker_work_dir = '/var/lib/toil'
 
+        # A Toil-managed Kubernetes cluster will have most of its temp space at
+        # /var/tmp, which is where really large temp files really belong
+        # according to https://systemd.io/TEMPORARY_DIRECTORIES/. So we will
+        # set the default temporary directory to there for all our jobs.
+        self.environment['TMPDIR'] = '/var/tmp'
+
         # Get the name of the AWS secret, if any, to mount in containers.
         self.aws_secret_name = os.environ.get("TOIL_AWS_SECRET_NAME", None)
 
@@ -477,7 +483,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         def mount_host_path(volume_name: str, host_path: str, mount_path: str, create: bool = False) -> None:
             """
             Add a host path volume with the given name to mount the given path.
-            
+
             :param create: If True, create the directory on the host if it does
                    not exist. Otherwise, when the directory does not exist, the
                    pod will wait for it to come into existence.
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -59,7 +59,7 @@ from toil import logProcessContext, lookupEnvVar
 from toil.batchSystems.options import (add_all_batchsystem_options,
                                        set_batchsystem_config_defaults,
                                        set_batchsystem_options)
-from toil.bus import (MessageBus, 
+from toil.bus import (MessageBus,
                       JobIssuedMessage,
                       JobCompletedMessage,
                       JobFailedMessage,
@@ -124,6 +124,7 @@ class Config:
         self.jobStore: Optional[str] = None  # type: ignore
         self.logLevel: str = logging.getLevelName(root_logger.getEffectiveLevel())
         self.workDir: Optional[str] = None
+        self.coordination_dir: Optional[str] = None
         self.noStdOutErr: bool = False
         self.stats: bool = False
 
@@ -299,6 +300,11 @@ class Config:
                 logger.warning(f'Length of workDir path "{self.workDir}" is {len(self.workDir)} characters.  '
                                f'Consider setting a shorter path with --workPath or setting TMPDIR to something '
                                f'like "/tmp" to avoid overly long paths.')
+        set_option("coordination_dir")
+        if self.coordination_dir is not None:
+            self.coordination_dir = os.path.abspath(self.coordination_dir)
+            if not os.path.exists(self.coordination_dir):
+                raise RuntimeError(f"The path provided to --coordinationDir ({self.coordination_dir}) does not exist.")
 
         set_option("noStdOutErr")
         set_option("stats")
@@ -377,14 +383,14 @@ class Config:
         set_option("doubleMem")
         set_option("maxJobDuration", int, iC(1))
         set_option("rescueJobsFrequency", int, iC(1))
-        
+
         # Log management
         set_option("maxLogFileSize", h2b, iC(1))
         set_option("writeLogs")
         set_option("writeLogsGzip")
         set_option("writeLogsFromAllJobs")
         set_option("write_messages")
-        
+
         assert not (self.writeLogs and self.writeLogsGzip), \
             "Cannot use both --writeLogs and --writeLogsGzip at the same time."
         assert not self.writeLogsFromAllJobs or self.writeLogs or self.writeLogsGzip, \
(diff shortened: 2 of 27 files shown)


adamnovak:
I made this on top of #4179 in hopes that maybe we can merge them both at once?

src/toil/batchSystems/kubernetes.py
```diff
@@ -141,6 +141,12 @@ def __init__(self, config, maxCores, maxMemory, maxDisk):
             # TODO: Keep this in sync with the Dockerfile.
             self.worker_work_dir = '/var/lib/toil'
 
+        # A Toil-managed Kubernetes cluster will have most of its temp space at
+        # /var/tmp, which is where really large temp files really belong
+        # according to https://systemd.io/TEMPORARY_DIRECTORIES/. So we will
```
DailyDreaming, line 146: Thanks for including the link explaining the paradigm.

src/toil/common.py
```diff
@@ -1294,32 +1304,78 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
...
+            """
+            return f(o) if o is not None else None
 
+        def ofilter(o: Optional[T], p: Callable[[T], bool]) -> Optional[T]:
+            """
+            Filter an optional with a predicate function.
+            """
+            return None if (o is None or not p(o)) else o
+
+        def try_path(path: str) -> Optional[str]:
```
DailyDreaming, line 1338: This function in particular, but also possibly omap and ofilter could probably live in lib and eventually be reused.  It's definitely something I could see reusing.

src/toil/common.py
```diff
@@ -1294,32 +1304,78 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
...
+            lambda: cls.getToilWorkDir(config_work_dir)
+        ]
+
+        for source in sources:
+            # Go through all the sources and use the first one that works.
+            result = source()
+            if result is not None:
+                return result
+
+        raise RuntimeError("Could not determine a coordination directory by any method!")
```
DailyDreaming, line 1378: Debated whether to leave this comment or not... as simpler code is more of a style I like rather than anything functional.  I'll leave the comment here and you can change it or not at your discretion.

This might be simpler as an `or` statement.  `ofilter` and `omap` also might be simplified out using a statement like: `try_path(os.getenv('XDG_RUNTIME_DIR') or try_path(os.path.join(p, 'toil'))`.

```
source: Optional[str] = (
    # First try an override env var
    os.getenv('TOIL_COORDINATION_DIR_OVERRIDE') or
    # Then the value from the config
    config_coordination_dir or
    # Then a normal env var
    # TODO: why/how would this propagate when not using single machine?
    os.getenv('TOIL_COORDINATION_DIR') or
    # Then try a `toil` subdirectory of the XDG runtime directory
    # (often /var/run/users/<UID>). But only if we are actually in a
    # session that has the env var set. Otherwise it might belong to a
    # different set of sessions and get cleaned up out from under us
    # when that session ends.
    try_path(os.getenv('XDG_RUNTIME_DIR') or try_path(os.path.join(p, 'toil')) or
    # Try under /run/lock. It might be a temp dir style sticky directory.
    try_path('/run/lock') or
    # Finally, fall back on the work dir and hope it's a legit filesystem.
    cls.getToilWorkDir(config_work_dir)
)

if source:
    return source

raise RuntimeError("Could not determine a coordination directory by any method!")
```

src/toil/common.py
```diff
@@ -1294,32 +1304,78 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
...
-            return in_memory_base
-        else:
-            try:
-                # Try to make it opportunistically.
-                os.makedirs(in_memory_base, exist_ok=True)
-                return in_memory_base
-            except:
-                pass
+        # Define higher-order functions over Optional because I learned Haskell
+        # once.
```
DailyDreaming, line 1323: I've heard it's not voluntary.

src/toil/lib/aws/ami.py
```diff
@@ -27,11 +28,20 @@ def get_flatcar_ami(ec2_client: BaseClient, architecture: str = 'amd64') -> str:
     :param ec2_client: Boto3 EC2 Client
     :param architecture: The architecture type for the new AWS machine. Can be either amd64 or arm64
     """
+
+    # How many times should we bang on the Flatcar release feed before moving
+    # on if it can't get us an AMI?
+    MAX_TRIES = 3
```
DailyDreaming, line 34: Third times the charm.  Looks about right to me.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -606,12 +606,22 @@ def addVolumesService(self, config: InstanceConfiguration):
...
         Add a service to prepare and mount local scratch volumes.
         """
+
+        # TODO: when
+        # https://www.flatcar.org/docs/latest/setup/storage/mounting-storage/
+        # describes how to collect all the ephemeral disks declaratively and
+        # make Ignition RAID them, stop doing it manually. Might depend on real
+        # solution for https://github.com/coreos/ignition/issues/1126
+        #
+        # TODO: check what kind of isntance this is, and what ephemeral volumes
```
DailyDreaming, line 616: `isntance`

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -766,6 +776,8 @@ def addToilService(self, config: InstanceConfiguration, role: str, keyPath: str
             entryPointArgs = " ".join(["'" + customDockerInitCommand + "'", entryPoint, entryPointArgs])
             entryPoint = "customDockerInit.sh"
 
+        # Set up the service. Make sure to make it default to using the
+        # actually-big temp directory.
```
DailyDreaming, line 780: Suggestion: Might be good to include `/var/tmp` and possibly the link used earlier here (https://systemd.io/TEMPORARY_DIRECTORIES/).

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -301,6 +301,11 @@ def launchCluster(self,
         :return: None
         """
 
+        # First, pre-flight-check pur permissions before makign anything.
```
DailyDreaming, line 304: `pur`
`makign`

src/toil/test/cwl/cwlTest.py
```diff
@@ -405,7 +405,10 @@ def test_bioconda(self):
             out_name="output1",
         )
 
+    # TODO: work out why we can't fulfil the SoftwareRequirement and it crashes
```
DailyDreaming, line 408: `fulfil`

Though this is fixed now, correct?

DailyDreaming (approved):
Overall, this looks good to me, though I think the galaxy `xfail` needs removal before merging.  Take a look at the other comments, but they are minor.

Thank you for the detailed PR description and the links outlining the `systemd` paradigm.

src/toil/test/cwl/cwlTest.py
```diff
@@ -405,7 +405,10 @@ def test_bioconda(self):
...
+    # the galaxy module 
     @needs_docker
+    @pytest.mark.xfail
     def test_biocontainers(self):
```
mr-c, line 412:
```suggestion
    @needs_docker
    def test_biocontainers(self):
```

src/toil/test/cwl/cwlTest.py
adamnovak: I guess it must be if it is working in mainline.

src/toil/common.py
adamnovak: `or` and short-circuiting get us most of the way there. But I don't see how we'd do the `os.path.join` onto `os.getenv('XDG_RUNTIME_DIR')` only when that's an allowed argument for `os.path.join`. It would have to be something exploiting how `and` happens to return its second argument:

```
...
or ('XDG_RUNTIME_DIR' in os.environ and try_path(os.path.join(os.environ['XDG_RUNTIME_DIR'], 'toil')))
...
```

adamnovak: This way also takes `''` as equivalent to `None`, but that would be fine here.

adamnovak: I think this way is probably better than bringing in monads actually. And I think I was just doing the wrong thing before and using `$XDG_RUNTIME_DIR/toil` only if *that* already existed.

DailyDreaming (approved):
LGTM.  Thanks for addressing the comments!

PR #4195, opened 2022-08-30
Add new version of Toil logo from Slack [ci skip]

adamnovak:
Here's the logo update we were kicking around in Slack.

I think the ends of the antennae might need to be hidden behind the main outline still, but we can run with this.

I am going to see if Gitlab is willing to skip CI-ing this.


DailyDreaming (approved):
Awesome.  Thanks for adding it.

PR #4196, opened 2022-08-30
Google fixes

boukn:
Several bugfixes needed to get this almost working on google cloud. Some like --network are just to make it more configurable. Others, like the convert_bytes before the vm launch or the source image update fix bugs,


src/toil/provisioners/gceProvisioner.py
```diff
@@ -34,13 +34,24 @@
 logging.getLogger("urllib3.connectionpool").setLevel(logging.WARNING)
 
 
+# pulled from stack overflow
+# fix for objects getting passed to a json serializer which dies on bytes
+def convert_bytes(data):
+    if isinstance(data, bytes):  return data.decode('ascii')
```
DailyDreaming, line 40: This should be `data.decode('utf-8')` instead of `ascii`, unless there's a compelling reason not to.

src/toil/provisioners/gceProvisioner.py
```diff
@@ -34,13 +34,24 @@
...
+    if isinstance(data, dict):   return dict(map(convert_bytes, data.items()))
+    if isinstance(data, tuple):  return tuple(map(convert_bytes, data))
+    if isinstance(data, list):   return list(map(convert_bytes, data))
+    if isinstance(data, set):    return set(map(convert, data))
```
DailyDreaming, line 44: Should `convert` here be `convert_bytes`?

src/toil/provisioners/gceProvisioner.py
```diff
@@ -34,13 +34,24 @@
 logging.getLogger("urllib3.connectionpool").setLevel(logging.WARNING)
 
 
+# pulled from stack overflow
+# fix for objects getting passed to a json serializer which dies on bytes
+def convert_bytes(data):
```
DailyDreaming, line 39: Suggestion, move this function to: https://github.com/DataBiosphere/toil/blob/master/src/toil/lib/compatibility.py

DailyDreaming (commented):
Thanks for doing this!  Our priorities aren't currently focused on our Google support, so any fixes and patches made are greatly appreciated!

Just a couple of comments, but otherwise these look like reasonable changes.

PR #4220, opened 2022-09-20
Issues/3653 arbitrary aws tags

Hexotical:
## Changelog Entry

Address issue 3653 via adding a new environment variable which takes a json dictionary of tags to apply when creating buckets, adds testing for util functions to gitlab

## Reviewer Checklist

        ```
        ```

## Merger Checklist


.gitlab-ci.yml
```diff
@@ -286,6 +286,7 @@ jobstore_and_provisioning:
   script:
     - pwd
     - ${MAIN_PYTHON_PKG} -m virtualenv venv && . venv/bin/activate && pip install -U pip wheel && make prepare && make develop extras=[all] packages=htcondor
+    - make test tests=src/toil/test/lib/aws/
```
DailyDreaming, line 289: `TOIL_SKIP_DOCKER=true make test tests=src/toil/test/lib` is run `basic_tests`.  We could probably separate these more cleanly.

docs/appendices/environment_vars.rst
```diff
@@ -224,6 +224,11 @@ There are several environment variables that affect the way Toil runs.
 |                                  | at UCSC to stop a bot we have that terminates      |
 |                                  | untagged resources.                                |
 +----------------------------------+----------------------------------------------------+
+| TOIL_AWS_TAG                     | This will tag cloud resources with any arbitrary   |
+|                                  | tags given in a JSON format. For information on    |
+|                                  | valid AWS tags, see 'AWS Tags'_.                   |
++----------------------------------+----------------------------------------------------+
++----------------------------------+----------------------------------------------------+
```
DailyDreaming, line 231: Redundant line.

docs/appendices/environment_vars.rst
```diff
@@ -224,6 +224,11 @@ There are several environment variables that affect the way Toil runs.
 |                                  | at UCSC to stop a bot we have that terminates      |
 |                                  | untagged resources.                                |
 +----------------------------------+----------------------------------------------------+
+| TOIL_AWS_TAG                     | This will tag cloud resources with any arbitrary   |
```
DailyDreaming, line 227: s/TOIL_AWS_TAG/TOIL_AWS_TAGS/g

src/toil/test/lib/aws/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.DEBUG)
+
+class TagGenerationTest(ToilTest):
+    """
+    Test for tag generation from environment variables
+    """
+    def test_build_tag(self):
+        environment = dict()
+        environment["TOIL_OWNER_TAG"] = "😀"
```
DailyDreaming, line 33: Excellent use of unicode.

src/toil/lib/aws/utils.py
```diff
@@ -387,4 +388,27 @@ def list_objects_for_url(url: ParseResult) -> List[str]:
...
+            if isinstance(json_user_tags, dict):
+                tags.update(json.loads(user_tags))
+            else:
+                logger.error('TOIL_AWS_TAGS must be in JSON format: {"key" : "value", ...}')
+                exit(1)
+        except json.decoder.JSONDecodeError:
+            logger.error('TOIL_AWS_TAGS must be in JSON format: {"key" : "value", ...}')
+            exit(1)
+    return tags
+def flatten_tags(tags: Dict[str, str]) -> List[Dict[str, str]]:
```
DailyDreaming, line 410: Should be newline or two even above this.

DailyDreaming (changes requested):
We should probably allow different services to be tagged in different ways.  Something like:

```
{
  's3': {'OWNER': 'TheBucketBaron',
           'EXPIRES': 'AtTheEndOfAllBuckets'},
   'ec2': {'OWNER': 'IAmAFreeInstance'}
}
```

We only need `s3` and `ec2` for now, so let's just start with those.  We can add more later if we need/want them.

We can also remove and stop using the old `TOIL_OWNER_TAG` env var.  Assuming this works, `TOIL_AWS_TAGS` should be able to replace it entirely.

It would also be nice to have a `default_aws_tags.json`, whose contents could be shown in the docs as an example of how to set this.

adamnovak:
@DailyDreaming It looks like right now [`toil launch-cluster` has a `--tag` option](https://github.com/DataBiosphere/toil/blob/da56ac10b4f1f8744be1dcea163527bcb2839473/src/toil/utils/toilLaunchCluster.py#L52-L62) to let you add tags to the things it makes, across all services, and this new variable is applied to AWS Job Store buckets (and not yet merged in with `--tags` by `launch-cluster`).

I'm not sure we need support for different tags for different services. We might want to let the environment variable work as a fallback for `--tag` before we merge, but I'd be happy with just one set of tags across services, at least for now.

adamnovak:
@Hexotical What do you think makes the most sense to do from your perspective?

Hexotical:
@adamnovak #4226 seems to want individual tagging for each service, will add this

src/toil/utils/toilLaunchCluster.py
```diff
@@ -108,7 +109,7 @@ def main() -> None:
     options = parser.parse_args()
     set_logging_from_options(options)
 
-    tags = create_tags_dict(options.tags) if options.tags else dict()
+    tags = create_tags_dict(options.tags) if options.tags else build_tag_dict()
```
adamnovak, line 112: These functions could have better names to explain why we need two of them that kind of do the same thing. Or we could unify them somehow.

src/toil/utils/toilLaunchCluster.py
```diff
@@ -19,6 +19,7 @@
 
 from toil import applianceSelf
 from toil.common import parser_with_common_options
+from toil.lib.aws.utils import build_tag_dict
```
adamnovak, line 22: It looks like `toil.lib.aws.utils` is designed to be able to import even if the Boto modules (which only come with the `aws` extra) aren't installed. So even if we import this we should still be able to `launch-cluster` if the user is only using Google Cloud and the Google modules.

docs/running/cloud/amazon.rst
```diff
@@ -118,6 +118,10 @@ To further break down each of these commands:
 
     **--keyPairName id_rsa** --- The name of your key pair, which should be "id_rsa" if you've followed this tutorial.
 
+.. note::
+
+   You can set the TOIL_AWS_TAGS with a JSON dictionary to specify arbitrary tags for AWS resources
```
adamnovak, line 123: "the TOIL_AWS_TAGS" sounds pretty weird, as does "set ... with". And JSON doesn't call them "dictionaries" but instead "objects". And it's not explicitly stated here that the format is going to be tag name to tag value. And an example would be nice. Also the variable name should maybe be formatted as code.

```suggestion
   You can set the ``TOIL_AWS_TAGS`` environment variable to a JSON object to specify arbitrary tags for AWS resources. For example, if you ``export TOIL_AWS_TAGS='{"project-name": "variant-calling"}'`` in your shell before using Toil, AWS resources created by Toil will be tagged with a ``project-name`` tag with the value ``variant-calling``.
```

adamnovak (approved):
I think this is pretty good. All of @DailyDreaming's concerns look to have been addressed.

I added some suggestions, but we could also just merge this.

DailyDreaming (changes requested):
Thanks!  All of my concerns appear to be addressed.  There's one bad import for python3.7 though that needs to be corrected before this can be merged though (`ImportError: cannot import name 'MutableMapping' from 'typing_extensions' in toil/src/toil/lib/aws/utils.py`).

src/toil/lib/aws/utils.py
```diff
@@ -40,7 +41,7 @@
                             retry)
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal, MutableMapping
 else:
     from typing_extensions import Literal
```
adamnovak, line 46:
```suggestion
    from typing import Literal
else:
    from typing_extensions import Literal
from typing import MutableMapping
```

[The `typing` module has MutableMapping all the way back to 3.7](https://docs.python.org/3.7/library/typing.html#typing.MutableMapping), so we can always get it from there, not just when we're on Python 3.8 or newer. And we can't just not import it when we're on 3.7, since we actually need it.

adamnovak (changes requested):
I think you still need to actually import `MutableMapping` though.

adamnovak (approved):

DailyDreaming (approved):
LGTM.  Thank you @Hexotical !

PR #4267, opened 2022-11-02
Account for overhead in Cluster Scaler

### LINKED ISSUE #2103: The cluster scalar uses total memory instead of available memory.
The cluster scaler calculation to determine which machine types to start uses the total memory value. The Mesos slaves will return available memory, though, which might not be sufficient to run the jobs. This leads to the situation where the clusterScalar will not spin up additional machines because it thinks that there are enough, but some jobs will hang forever since the memory offered by Mesos is insufficient. This occurred in issue #2078, which was resolved by lowering memory requirements, but did not fix the underlying issue.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-233)
┆friendlyId: TOIL-233

adamnovak:
This will fix #2103. This is also relevant to https://github.com/DataBiosphere/toil/issues/4147#issuecomment-1179587561 where a node with memory size very close to the job's requirements was provisioned but not able to be used.

This changes the cluster scaler so that it preprocesses its node sizes to remove some resources for overhead (i.e. the kernel, other processes on the system, and Mesos/Kubernetes). Currently we reserve memory and disk.

To guess at how much memory might be used, in the absence of much real data, we use [a formula from Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu) which describes how much memory GKE clusters keep for themselves, as a function of node memory size. We assume that Toil-managed Mesos and Kubernetes clusters use that much memory or less.

To guess at how much disk space might be used, we take 5% of the disk space (commonly a fraction reserved for superuser) and add 5 GiB. If this is more than the size of the disk on the node, we complain that the disk is weirdly small and assume all of it will be reserved (which may come back to haunt the autoscaler later if we ever try to use ~6 GB nodes).

None of this is user-configurable. **Unless you pass `--assumeZeroOverhead` in which case we turn it off.**

I've added some tests for made-up and real-world cases to constrain the overhead behavior to something sensible. Those pass, but I'm not sure if any live Toil autoscaler tests will now not scale up because they think they can't fit jobs where they will actually fit.

## Changelog Entry

 * Toil's built-in autoscaler now guesses that some memory and disk space on nodes will not actually be available for jobs; pass `--assumeZeroOverhead` to revert to the old behavior (#2103) 

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -447,8 +447,10 @@ class ClusterScaler:
         outputs.
         """
 
-        # See https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu for how EKS works.
-        # We are going to just implement their scheme and hope it is a good enough overestimate of actual overhead.
+        # See
+        # https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#memory_cpu
+        # for how EKS works. We are going to just implement their scheme and
+        # hope it is a good enough overestimate of actual overhead.
         smaller = copy.copy(full_node)
 
         # Take away some memory to account for how much the scheduler is going to use.
@@ -456,7 +458,7 @@ class ClusterScaler:
         # And some disk to account for the OS and possibly superuser reservation
         # TODO: Figure out if the disk is an OS disk of a scratch disk
         smaller.disk -= self._disk_overhead(smaller.disk)
-        
+
         logger.debug('Node shape %s can hold jobs of shape %s', full_node, smaller)
 
         return smaller
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -129,7 +129,7 @@ class BinPackedFit:
                     # the user to raise the memory on the smallest machine?
                     fewest_constraints = failures
             
-            return jobShape, fewest_constraints
+            return jobShape, fewest_constraints if fewest_constraints is not None else []
 
         # grab current list of job objects appended to this instance type
         nodeReservations = self.nodeReservations[chosenNodeShape]
@@ -1047,7 +1047,7 @@ class JobTooBigError(Exception):
         self.msg = ''.join(parts)
         super().__init__()
 
-    def __str__(self):
+    def __str__(self) -> str:
         """
         Stringify the exception, including the message.
         """
--- a/src/toil/provisioners/clusterScaler.py
+++ b/src/toil/provisioners/clusterScaler.py
@@ -11,12 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import json
 import logging
+import math
 import os
 import time
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Union
 
 from toil.batchSystems.abstractBatchSystem import (AbstractBatchSystem,
                                                    AbstractScalableBatchSystem,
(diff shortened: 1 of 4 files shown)


src/toil/provisioners/clusterScaler.py
```diff
@@ -416,6 +433,91 @@ def _round(self, number: float) -> int:
             # If we get here, something has gone wrong.
             raise RuntimeError(f"Could not round {number}")
 
+    def _reserve_overhead(self, full_node: Shape) -> Shape:
```
w-gao, line 436:
```
None of this is user-configurable.
```

Would we ever want to disable / overwrite the overhead? Some tests are not passing because the limit is set to a very small value, but I don't know if that will happen realistically.

src/toil/test/provisioners/clusterScalerTest.py
```diff
@@ -438,6 +450,121 @@ def testBetaInertia(self):
             scaler.smoothEstimate(c4_8xlarge_preemptable, 100)
         self.assertEqual(scaler.smoothEstimate(c4_8xlarge_preemptable, 100), 100)
 
+    def test_overhead_accounting_large(self):
```
w-gao, line 453: great tests!

w-gao (commented):
This looks great to me! Some cluster scaler tests are not passing though:

```
INFO     toil.test:__init__.py:113 Setting up toil.test.provisioners.clusterScalerTest.ScalerThreadTest.testClusterScaling ...
WARNING  toil.provisioners.clusterScaler:clusterScaler.py:512 All 10.0 bB of disk on a node type are likely to be needed by the OS! The node probably cannot do any useful work!
WARNING  toil.provisioners.clusterScaler:clusterScaler.py:105 Couldn't fit job with requirements Shape(wallTime=9.671000000000003, memory=10, cores=10, disk=10, preemptable=False) into any nodes in the nodeTypes list.
```
https://github.com/DataBiosphere/toil/blob/0b754e37572a45fce6feb5b6ad46099a16f7034f/src/toil/test/provisioners/clusterScalerTest.py#L631

src/toil/provisioners/clusterScaler.py
```diff
@@ -416,6 +433,91 @@ def _round(self, number: float) -> int:
...
+            # Below each breakpoint, reserve the matching portion of the memory
+            # since the previous breakpoint, like a progressive income tax.
+            limit = min(breakpoint, memory_bytes)
+            reservation = fraction * (limit - accounted)
+            logger.debug('Reserve %sB of memory between %sB and %sB', bytes2human(reservation), bytes2human(accounted), bytes2human(limit))
+            reserved += reservation
+            accounted = limit
+            if accounted >= memory_bytes:
+                break
+        logger.debug('Reserved %sB/%sB memory for overhead', bytes2human(reserved), bytes2human(memory_bytes))
```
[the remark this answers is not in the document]
adamnovak: I kept ending up with like `123 Gi` when I really wanted `GiB`. But I suppose `Gi` is fine.

src/toil/provisioners/clusterScaler.py
adamnovak: I think users might run into situations where we thing a job won't fit when it really will fit. It might make sense to give people a way to turn this back off, since it's possible that they already know a workflow works and we'd be guessing it wouldn't work. But I'm having trouble thinking of a good way to let people actually tune this.

adamnovak:
OK @w-gao I think I have addressed your comments. I *think( I've fixed the failing tests, by adapting the test cases to allow for the estimated overhead. I've also added an `--assumeZeroOverhead` flag that you can pass to skip shrinking the node sizes, and use the old autoscaling behavior.

src/toil/provisioners/clusterScaler.py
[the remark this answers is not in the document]
w-gao: Ohh, that makes sense. I think the only problem was that something like `200 b` would end up being `200 bB`

w-gao (approved):
Looks great!

PR #4292, opened 2022-12-02
Allow determining of which jobs are running via toil status --replayBus

Hexotical:
## Changelog Entry

Allow determining of which jobs are running via toil status <job store> --printStatus

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/batchSystems/singleMachine.py
```diff
@@ -221,9 +221,16 @@ def daddy(self):
...
                             break
+                        elif result is not False:
+                            #Result is a PID
+
+                            if self._outbox is not None:
+                                # Annotate the job with the PID generated.
+                                self._outbox.publish(
+                                   BatchSystemMessage(jobID, result, self.__class__.__name__))
 
                         # Otherwise it's a PID if it succeeded, or False if it couldn't
```
adamnovak, line 232: I think now "otherwise" (to all the if cases above), the result can only be `False`.

src/toil/bus.py
```diff
@@ -145,6 +155,18 @@ class JobAnnotationMessage(NamedTuple):
     # The annotation data
     annotation_value: str
 
+class BatchSystemMessage(NamedTuple):
```
adamnovak, line 158: I'm not sure about the name here. All the other messages are named after events they represent (like `JobFailedMessage` or after data they are there to carry (like `QueueSizeMessage`). This one is named after the batch system, but it doesn't exist to tell you what the batch system is (though that is in there). It exists to tell you what the external batch system/backing scheduler's ID is for a particular job. 

Maybe this should be an `ExternalBatchIdMessage`? Or `ExternalIdMessage`?

src/toil/bus.py
```diff
@@ -599,5 +623,84 @@ def __init__(self) -> None:
...
+
+        We must not have connected to anything yet.
+        """
+
+        # We need to call the two inherited connection methods
+        super()._set_bus_and_message_types(bus, wanted_types)
+        self._set_bus(bus)
 
 
+def replay_message_bus(path: str) -> Dict[str, Any]:
```
adamnovak, line 638: This should have a better type signature, or else a *very* clear description of what exactly we return and what the caller is going to be allowed to do with it.

Maybe to do that we want to promote `JobStatus` out of the function?

src/toil/bus.py
```diff
@@ -599,5 +623,84 @@ def __init__(self) -> None:
...
+                # Remember the last value of any annotation that is set
+                job_statuses[event.job_id].annotations[event.annotation_name] = event.annotation_value
+            elif isinstance(event, BatchSystemMessage):
+                if event.toil_batch_id in batch_to_job_id:
+                    job_statuses[batch_to_job_id[event.toil_batch_id]].external_batch_id = event.external_batch_id
+                    job_statuses[batch_to_job_id[event.toil_batch_id]].batch_system = event.batch_system
+
+    return job_statuses
+
+def gen_messBus_path() -> str:
```
adamnovak, line 699: This name isn't `snake_case`. I also don't think it's a good idea to abbreviate the message bus as `messBus`; we don't call this enough for that to be worth it.

src/toil/bus.py
```diff
@@ -599,5 +623,84 @@ def __init__(self) -> None:
...
+    """
+    If given a path, return absolute path otherwise generate path to store message bus at
```
adamnovak, line 701: This docstring is wrong; the function takes no arguments and so can't be given a path.

adamnovak: This should also explain that the generated path has been created by the function, and that the caller is responsible for ensuring that it eventually gets deleted (or taking the blame if it does not).

src/toil/job.py
```diff
@@ -1158,6 +1158,17 @@ def pre_update_hook(self) -> None:
         self._job_version += 1
         logger.debug("New job version: %s", self)
 
+    def get_job_type(self) -> str:
```
adamnovak, line 1161: What does this function do? It needs a docstring. It definitely *doesn't* return the "type" of the job in the sense of a Python class type.

src/toil/leader.py
```diff
@@ -159,6 +160,8 @@ def __init__(self,
             # Message bus messages need to go to the given file.
             # Keep a reference to the return value so the listener stays alive.
             self._message_subscription = self.toilState.bus.connect_output_file(self.config.write_messages)
+        else:
+            raise RuntimeError("Write messages not set")
```
adamnovak, line 164: Since `write_messages` is now type `str`, it isn't allowed by the type system to be None, so we probably don't actually need to check if it is None. We can just use it as a string.

The case we *might* have to handle is if it is a file path that can't be opened (like `""`). But `connect_output_file` will probably take care of throwing an error in that case.

src/toil/server/wes/toil_backend.py
```diff
@@ -63,15 +63,17 @@
...
         launching workflow runs via Celery and retrieving data generated from
         them.
 
-        :param base_scratch_dir: The directory where workflows keep their
+        :param base_work_dir: The directory where workflows keep their
                                  output/scratch directories under their run
-                                 IDs.
+                                 IDs. Includes the MessageBus which requires
+                                 a reliable location, thus should always be
+                                 the work_dir specified by the user.
```
adamnovak, line 76: I don't think you did anything to the places where `ToilWorkflow()` is constructed to ensure this new constraint, and I don't really understand what you mean. Right now `base_scratch_dir` comes from the Toil *Server*'s `work_dir`, whicyh comes from https://github.com/DataBiosphere/toil/blob/88d4589592eba433d10b3c835ef2a0be5653336a/src/toil/server/app.py#L57-L59. But that's not really "specified by the user"; it's a fixed parameter of the WES server that we set at server startup and which doesn't. as fart as I know, get overridden by the user-supplied workflows. We probably want to rename the server parameter *away* from `work_dir`, because of the potential for confusion with an individual workflow's `work_dir` parameter, which is different.

And when it is passe din here as a `base_scratch_dir`, it is still one fixed directiry for all the workflows, so it doesn't really include the message bus log; it contains all the workflows' message bus logs, in subdirectories by `run_id`.

src/toil/batchSystems/singleMachine.py
```diff
@@ -221,9 +221,16 @@ def daddy(self):
                             # cores/memory/disk individually)?
                             self.inputQueue.put(args)
                             break
+                        elif result is not False:
+                            #Result is a PID
+
+                            if self._outbox is not None:
+                                # Annotate the job with the PID generated.
+                                self._outbox.publish(
+                                   BatchSystemMessage(jobID, result, self.__class__.__name__))
```
adamnovak, line 230: At this point `result` is a numeric PID, right? But `BatchSystemMessage` needs to hold a string. So don't we need a `str()` call to convert?

src/toil/bus.py
```diff
@@ -599,5 +623,84 @@ def __init__(self) -> None:
...
+                job_statuses[event.job_id].name = event.job_type
+                job_statuses[event.job_id].exit_code = event.exit_code
+            elif isinstance(event, JobFailedMessage):
+                job_statuses[event.job_id].name = event.job_type
+                if job_statuses[event.job_id].exit_code == 0:
+                    # Record the failure if we never got a failed exit code.
+                    job_statuses[event.job_id].exit_code = 1
+            elif isinstance(event, JobAnnotationMessage):
+                # Remember the last value of any annotation that is set
+                job_statuses[event.job_id].annotations[event.annotation_name] = event.annotation_value
```
adamnovak, line 691: Since nobody ever sends a `JobAnnotationMessage` anymore, and instead we always send `BatchSystemMessage` to communicate e.g. the AWS Batch job ID, then the annotations here will always be empty.

1. If they are always empty, it might make sense to remove the field, and the `JobAnnotationMessage` altogether.
2. Whether you do that or not, properly supporting AGC's dialect of WES requires always including the AWS Batch job ID in the task names that the server reports. Right now we do that by [selecting a filter function](https://github.com/DataBiosphere/toil/blob/88d4589592eba433d10b3c835ef2a0be5653336a/src/toil/server/wes/toil_backend.py#L602-L611) that gets to see the annotations and modify the WES tasks, and then [applying the filter function](https://github.com/DataBiosphere/toil/pull/4292/files#diff-e05a9b10abd96d5732ff2a3577a21843f0c6efa1074f2e44d752670449124edbR238-R249) when we go to generate the tasks. But if we no longer keep the AWS Batch job ID in the annotations, the current filter function for AWS Batch won't work, and we will break AWS Batch support.

We really ought to have the AWS WES dialect under test if we want to not break it. But, failing that, we definitely need to adapt the filter function setup to be able to put the new `external_batch_id` field from the `JobStatus` into the task name, instead of looking at the annotations. Maybe we need to give the filter_function access to the whole `JobStatus` and not just the annotations field?

src/toil/utils/toilStatus.py
```diff
@@ -219,6 +223,23 @@ def getStatus(jobStoreName: str) -> str:
                 pass
         return 'RUNNING'
 
+
+    def readBusMessages(self) -> None:
```
adamnovak, line 227: This should be `snake_case`.

src/toil/utils/toilStatus.py
```diff
@@ -219,6 +223,23 @@ def getStatus(jobStoreName: str) -> str:
...
+        """
+        Goes through bus messages, returns a list of tuples which have correspondence between
+        PID on assigned batch system and
+
+        Prints a list of the currently running jobs
+        """
+
+        print("\nMessage bus path: ", self.message_bus_path)
+        replayed_messages = replay_message_bus(self.message_bus_path)
```
adamnovak, line 236: How are we going to handle the case where the user doesn't actually have access to the message bus log file (because e.g. it is on another machine and not on a shared filesystem)? Will we just crash, or do we want to catch `FileNotFoundError` and print some kind of suggestion?

src/toil/utils/toilStatus.py
```diff
@@ -329,6 +350,8 @@ def main() -> None:
                         help="Print children of each job. default=%(default)s",
                         default=False)
 
+    parser.add_argument("--replayBus", action="store_true",
+                        help="Determine which jobs are currently running and the associated batch system ID")
```
adamnovak, line 354: Is this the right name for this option? The fact that there's a bus that gets replayed seems like an implementation detail that the user shouldn't have to care about; if the user never hears about the existence of a "message bus", so much the better.

To match the other options I guess this could be `--printBatch` or `--printSchedulerInfo` or `--printJobs` or something. If it wasn't already `toil status` it could be `--printStatus`.

Is there a good reason to make this an option at all, and not just always replay the message bus and print information from it if we can find it?

src/toil/utils/toilStatus.py
adamnovak: Also all the other function names start with "print" and this one starts with "read" but actually it prints anyway.

.gitlab-ci.yml
```diff
@@ -62,15 +62,15 @@ cwl_dependency_is_stand_alone:
   stage: linting_and_dependencies
   script:
     - pwd
-    - ${MAIN_PYTHON_PKG} -m virtualenv venv && . venv/bin/activate && make prepare && make develop extras=[cwl]
+    - ${MAIN_PYTHON_PKG} -m virtualenv venv && . venv/bin/activate && make prepare && make develop extras=[cwl, server]
```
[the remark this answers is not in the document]
adamnovak: It looks like this also doesn't work because it contains a space; the stuff after the space [doesn't make it into the `pip install` command](https://ucsc-ci.com/databiosphere/toil/-/jobs/28601#L472) and so the install fails in the test.

src/toil/bus.py
Hexotical: Maybe later but I think job status overlaps with with how toil thinks of job status internally already and I don't think there's a need to promote it outside.

src/toil/bus.py
Hexotical: This should probably be its own issue

src/toil/bus.py
```diff
@@ -599,5 +621,95 @@ def __init__(self) -> None:
...
+
+    job_store_id: str
+    name: str
+    exit_code: int
+    annotations: Dict[str, str]
+    toil_batch_id: int
+    external_batch_id: str
+    batch_system: str
+
+    def __repr__(self):
```
adamnovak, line 650:
```suggestion
    def __repr__(self) -> str:
```

src/toil/bus.py
```diff
@@ -599,5 +621,95 @@ def __init__(self) -> None:
...
+
+        # We need to call the two inherited connection methods
+        super()._set_bus_and_message_types(bus, wanted_types)
+        self._set_bus(bus)
 
 
+@dataclass
+class JobStatus:
+    """
+    Records the status of a job.
```
adamnovak, line 639: Since this is a public thing now it could use some more documentation for what the fields mean.

src/toil/server/wes/toil_backend.py
```diff
@@ -63,15 +63,16 @@
...
         """
         Class to represent a Toil workflow. This class is responsible for
         launching workflow runs via Celery and retrieving data generated from
         them.
 
-        :param base_scratch_dir: The directory where workflows keep their
+        :param base_work_dir: The directory where workflows keep their
                                  output/scratch directories under their run
-                                 IDs.
+                                 IDs. Includes the MessageBus which requires
```
adamnovak, line 74: I think the file is really a log of the message bus's messages, rather than being the bus *itself*, like a socket that lets you talk top it or something.

adamnovak (approved):
I think this is good now! There's a missing type that trips up mypy but I think I can just fix it.

PR #4299, opened 2022-12-13
Issues/4218 non cache default

Hexotical:
## Changelog Entry

Add a --caching option which explicitly states whether to use caching with a workflow, sets a default value depending on whether or not we are using the file job store if not specified.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/common.py
```diff
@@ -563,10 +563,12 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None) -> None:
...
     move_exports.set_defaults(moveExports=False)
-    file_store_options.add_argument('--disableCaching', dest='disableCaching', action='store_true',
-                                    default=False,
-                                    help='Disables caching in the file store. This flag must be set to use '
-                                         'a batch system that does not support cleanup, such as Parasol.')
+
+    caching = file_store_options.add_mutually_exclusive_group()
+    caching_help = ("Enable or disable caching for your workflow, specifying this overrides default from job store")
+    caching.add_argument('--disableCaching', dest='caching', action='store_false', help=caching_help)
+    caching.add_argument('--caching', dest='caching', type=bool, help=caching_help)
```
adamnovak, line 570: I saw where you put a `--caching=FALSE` example, and I was worried that `"FALSE"` might not parse as `False` because of the case, so I tested it:

```
Python 3.8.10 (default, Nov 14 2022, 12:59:47) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> bool('FALSE')
True
```

It turns out that `type=bool` just doesn't work properly, [because `bool()` of any non-empty string is `True`](https://stackoverflow.com/q/15008758).

I think we actually need to do something more like https://stackoverflow.com/a/59579733

```suggestion
    from distutils.util import strtobool
    caching.add_argument('--caching', dest='caching', type=lambda val: bool(strtobool(val)), help=caching_help)
```

adamnovak: This also suggests that maybe we could use a test for the parsing? I'm not sure the best way to do that though.

docs/running/cliOptions.rst
```diff
@@ -433,6 +433,9 @@ systems have issues!).
 
 **Miscellaneous Options**
 
+  --caching BOOL        Set caching options. This must be set to FALSE
+                        to use a batch system that does not support
+                        cleanup, such as Parasol.
```
adamnovak, line 438: This ought to be under *Data Storage Options*, next to `--disableCaching`.

adamnovak (changes requested):
This looks good, except that I apparently lied about `type=bool`, so you can only use `--caching` to turn it on and not to turn it off.

adamnovak (approved):
Today I noticed that we already have something set up that ought to let us say `type="bool"` and get the right behavior:

https://github.com/DataBiosphere/toil/blob/8a0d05c27c353914d95c672ddd025c3d4a95819d/src/toil/common.py#L479

This way looks fine, but it would be better if we could deduplicate these.

PR #4300, opened 2022-12-14
CWL: fix false warning when `outputSource` contains only one `None` value

### LINKED ISSUE #3991: Conditional workflows that evaluate to false log a warning about pickValue if not specified.
In issue 3990, I also see a warning posted from the toil-cwl-runner that says you need to have pickValue set to handle null outputs.  

This is not warning for cwltool.  

See issue 3990 for a use case.
```
You had a conditional step that did not run, but you did not use pickValue to handle the skipped input.
{
    "out1": null
}
```

In general, the behavior for outputs is expected.

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-1122)
┆friendlyId: TOIL-1122

w-gao:
Fix #3991 by adding a condition to check if the input is `SkipNull` in `ResolveSource::pick_value()`.

```cwl
# example from the issue
...
outputs:
  out1:
    type: File
    outputSource:
      - echo/echoOut
    pickValue: first_non_null
...
```

When the `outputSource` only has one element, we internally store it as a tuple instead of a list of tuples. This source could evaluate to null but our [`pick_value()` function](https://github.com/DataBiosphere/toil/blob/releases/5.7.x/src/toil/cwl/cwltoil.py#L370) would return early and not filter out the SkipNull. 

## Changelog Entry

- CWL: fix false warning when `outputSource` contains only one `None` value

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -299,12 +299,6 @@ class ResolveSource:
         self.name, self.input, self.source_key = name, input, source_key
 
         source_names = aslist(self.input[self.source_key])
-        # with open("/Users/wlgao/Desktop/log.txt", "a") as f:
-        #     f.write(f"{name=}\n")
-        #     f.write(f"{input=}\n")
-        #     f.write(f"{source_key=}\n")
-        #     f.write(f"{source_names=}\n\n")
-
         # Rule is that source: [foo] is just foo
         #                      unless it also has linkMerge: merge_nested
         if input.get("linkMerge") or len(source_names) > 1:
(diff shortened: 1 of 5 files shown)


mr-c (changes requested):
Thanks! Can you add a test case to confirm the fix?

adamnovak (approved):
Yeah a test would make this better. But are we set up to test for warnings? If that's not hard we should do it, but if testing for warnings turns out to be hard and there aren't already tests we can add on to for ResolveSource I'm not sure it makes sense to try and start with this.

w-gao:
I don't think we have existing tests for ResolveSource.

pytest does have a [caplog fixture](https://docs.pytest.org/en/7.1.x/reference/reference.html#std-fixture-caplog) and it seems to be straightforward to capture the logs and check if a log message is present in the output.

I added a small test that works well locally, so hopefully it passes on the CI as well.

mr-c (approved):

PR #4348, opened 2023-01-30
black + isort everything

mr-c:
## Changelog Entry

 * [Misc] All source code is now formatted by `black`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


mr-c:
@adamnovak I'd like to merge this PR, and then open a separate PR to add the final commit ID to `.git-blame-ignore-revs` which GitHub will then ignore automatically when doing a git blame view. The follow-on PR will also update `docs/contributing/contributing.rst` with the developer instructions to get the same benefit (`git config blame.ignoreRevsFile .git-blame-ignore-revs`).

DailyDreaming:
@mr-c Oof, we can discuss this, but I'm not sure we're ready to convert.

Adding 18,000 fresh lines is quite verbose, and it doesn't look good to me.  Black would be better if users could configure it, since I don't like a lot of their decisions (including shorter line lengths and some arguably awkward wrapping).

adamnovak (changes requested):
It looks like this ended up reverting some other semantically-important changes to the repo?

Speciffically e.g. https://github.com/DataBiosphere/toil/pull/4348/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R44

adamnovak:
Setting up a bypass for git blame for this is a good idea.

@DailyDreaming I think the *point* of Black is that users can't configure it; there shall be one true source code form for any Python abstract syntax tree, and the users shall never again need or be allowed to argue over it.

I'm still not sure that Black-ifying all the source is actually worth the trouble, though; we don't really waste a lot of time arguing over or fiddling with source code formatting, as far as I can tell. Most of our formatting-related costs go into thinking about whether we ought to add/adopt/enforce Black or not. (I've also spent a fairly substantial amount of time resolving merge conflicts when somebody rearranged imports and I added a new import.) Which to me says no, we shouldn't bother adding Black, it's not going to pay off.

But as long as we *haven't* adopted Black, the possibility of adopting it will remain, while if we *do* adopt it we'll almost certainly not decide to abandon it. So maybe the simplest possible solution is to do it, because if we don't there will be at least one more issue proposing that we do?

adamnovak:
@mr-c You just ran the formatters once, right? Won't we need a process that somehow applies `black` to incoming changesets automatically, to prevent needing another giant Blackification PR in the future? How is that going to work?

I really want to avoid PRs failing *because* they didn't follow Black, because that makes e.g. the Github web editor kind of useless, and PRs from external contributors without our preferred dev setup much more of a pain to land. Maybe there's some kind of BlackBot we can turn lose on incoming PRs?

mr-c:
> Black would be better if users could configure it, since I don't like a lot of their decisions (including shorter line lengths and some arguably awkward wrapping).

Line length is configurable with `--line-length 100` or via the following in `pyproject.toml`:
``` toml
[tool.black]
line-length = 100
```

Honestly, I don't care what tool is used, as long as it is comprehensive. `black` is used by almost all of the other projects I work on, so that would be a nice choice :-)

Makefile
```diff
@@ -267,13 +266,10 @@ remove_unused_imports: $(PYSOURCES)
 remove_trailing_whitespace:
 	$(CURDIR)/contrib/admin/remove_trailing_whitespace.py
 
-format: $(wildcard src/toil/cwl/*.py)
+format: $(PYSOURCES)
```
mr-c, line 269: @adamnovak Yes, here is the relevant line

adamnovak:
I'm going to close this in favor of #4777 which changes the formatter only.

PR #4356, opened 2023-01-31
Kubernetes batch system: support `--maxCores`, `--maxDisk`, and `--maxMemory`

### LINKED ISSUE #2864: Kubernetes batch system ignores --maxCores, --maxDisk, and --maxMemory and just throws all the jobs at the cluster at once
In theory, `--maxCores`, `--maxDisk`, and `--maxMemory` should limit how much of each resource Toil is willing to request/use for its jobs at a time.

In practice, this is implemented in the single machine batch system, but not in some of the others, including the Kubernetes batch system.

I would like to be able to run my Toil run against our lab cluster without swamping it and making other people's jobs wait for my jobs to finish. So I would like to leave some cluster resources free for others, by restricting the maximum resources Toil will try to use at once. `--maxCores`, `--maxDisk`, and `--maxMemory` seem to be the right way to do this, but they are implemented per batch system, and they are not implemented in the Kubernetes batch system.

Someone should implement these features in the Kubernetes batch system. Perhaps the `ResourcePool` from the single machine batch system could be re-used?

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-457)
┆Issue Number: TOIL-457

w-gao:
Fixes #2864 by introducing a jobs queue system to restrict the maximum resources given to kubernetes at once. Proposed changes to the Kubernetes batch system:

- Maintain a queue of jobs in Toil to prevent swamping the cluster when resources are over a set limit (`--maxCores`, `--maxMemory`, and `--maxDisk`)
    - All jobs would go through the queue to be submitted to the kubernetes cluster
- Use a separate scheduler thread to process the jobs queue
    - This thread wakes up when a new job is put into the queue or when any resource is freed
    - When it wakes up, it loops over the entire queue and tries to launch as many jobs as possible
      - When any resource is not available for the job, it is put back into the queue

## Changelog Entry

- Kubernetes batch system: support `--maxCores`, `--maxDisk`, and `--maxMemory`

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/batchSystems/kubernetes.py
+++ b/src/toil/batchSystems/kubernetes.py
@@ -104,7 +104,7 @@ from toil.batchSystems.abstractBatchSystem import (EXIT_STATUS_UNAVAILABLE_VALUE
 from toil.batchSystems.cleanup_support import BatchSystemCleanupSupport
 from toil.batchSystems.contained_executor import pack_job
 from toil.batchSystems.options import OptionSetter
-from toil.common import Config, Toil
+from toil.common import Config, Toil, SYS_MAX_SIZE
 from toil.job import JobDescription, Requirer
 from toil.lib.conversions import human2bytes
 from toil.lib.misc import get_user_name, slow_down, utc_now
@@ -235,10 +235,11 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         self.run_id = f'toil-{self.unique_id}'
 
         # Keep track of available resources.
+        maxMillicores = int(SYS_MAX_SIZE if self.maxCores == SYS_MAX_SIZE else self.maxCores * 1000)
         self.resource_sources = [
             # A pool representing available CPU in units of millicores (1 CPU
             # unit = 1000 millicores)
-            ResourcePool(int(self.maxCores / 1000), 'cores'),
+            ResourcePool(maxMillicores, 'cores'),
             # A pool representing available memory in bytes
             ResourcePool(self.maxMemory, 'memory'),
             # A pool representing the available space in bytes
@@ -268,6 +269,7 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         self.schedulingThread.start()
 
         logger.warning("On branch issues/2864-k8s-queue-jobs")
+        logger.warning(f"available resources: {maxMillicores} cores, {self.maxMemory} memory, {self.maxDisk} disk")
 
     def _pretty_print(self, kubernetes_object: Any) -> str:
         """
@@ -554,7 +556,9 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
         while True:
             with self._work_available:
                 # Wait until we get notified to do work and release lock.
+                logger.debug("!! [scheduler] waiting for work to do")
                 self._work_available.wait()
+                logger.debug("!! [scheduler] more work")
 
                 if self.shutting_down.is_set():
                     # We're shutting down.
@@ -582,7 +586,6 @@ class KubernetesBatchSystem(BatchSystemCleanupSupport):
                         result = self._launch_job(job_name, job_desc, spec)
                         if result is False:
                             # Not enough resources to launch this job.
-                            logger.debug(f"!! not enough resources to run job id: {job_id}")
                             jobs.put(job)
                         else:
                             self._queued_job_ids.remove(job_id)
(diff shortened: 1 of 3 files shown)


src/toil/batchSystems/kubernetes.py
```diff
@@ -797,54 +890,150 @@ def mount_host_path(volume_name: str, host_path: str, mount_path: str, create: b
...
+        Release all resources acquired for a job.
+
+        :param resources: The resources in the order: core, memory, disk.
+        :param notify: If True, notify the threads that are waiting on the
+                       self._work_available condition.
+        """
+        # TODO: accelerators?
+
+        # What pools and sets do we want resources from
+
```
adamnovak, line 904: This doesn't really make sense to me.
```suggestion
```

src/toil/batchSystems/kubernetes.py
```diff
@@ -1603,40 +1788,47 @@ def killBatchJobs(self, jobIDs: List[int]) -> None:
...
+            if job_id not in issued_on_kubernetes:
                 # It never went to Kubernetes (or wasn't there when we just
                 # looked), so we can't kill it on Kubernetes.
+
+                # It is likely that the job is still waiting inside the jobs
+                # queue and hasn't been submitted to Kubernetes. In this case,
+                # it is not reliable to loop through the queue and try to
+                # delete it. Instead, let's keep track of it and don't submit
+                # when we encounter it.
+                self._killed_queue_jobs.add(job_id)
```
adamnovak, line 1805: This (and the `_getIssuedNonLocalBatchJobIDs()` call) needs to be protected with a lock that stops the scheduler from submitting the job to Kubernetes at the same time we are checking to see if it is there and if not adding it to the list of jobs not to submit.

I think you can probably just do it inside a `with self._work_available:`.

adamnovak: Also `_killed_queue_jobs` will get arbitrarily big if we kill a lot of jobs that aren't actually running or in the queue (maybe because they finished?). Not sure if there's anything reasonable to do about that though.

adamnovak (changes requested):
This looks pretty good!

I found a little debugging code left in that probably shouldn't be there, and I think a race between deleting a job that's in the queue and sending it to Kubernetes.

It looks like there's now a lot of shared or nearly-shared code between KubernetesBatchSystem and SingleMachineBatchSystem, around queuing up the jobs and managing the resource pools. Did you consider introducing a new base class or widget to hold it? Or is it all too coupled to what a job-to-be-submitted-later actually is to do that without making more giant terrible tuples like we use for this sort of thing in AbstractGridEngineBatchSystem?

src/toil/batchSystems/kubernetes.py
```diff
@@ -230,6 +234,43 @@ def __init__(self, config: Config, maxCores: int, maxMemory: int, maxDisk: int)
...
+        self._acquired_resources: Dict[str, List[int]] = {}
+
+        # Queue for jobs to be submitted to the Kubernetes cluster
+        self._jobs_queue: Queue[Tuple[int, JobDescription, V1PodSpec]] = Queue()
+
+        # A set of job IDs that should be killed
+        self._killed_queue_jobs: Set[int] = set()
+
+        # We use this event to signal shutdown
+        self.shutting_down = Event()
```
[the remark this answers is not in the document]
w-gao: Good catch - this was from the single machine batch system and I forgot to change it.

src/toil/batchSystems/kubernetes.py
w-gao: Yeah, this was also from the single machine batch system where we have ResourceSets for the accelerators. (We might want to consider the accelerators here as well, but I wanted to make sure the resource pool works first).

src/toil/batchSystems/kubernetes.py
w-gao: That is a good catch! I was worried that there may be race conditions that I didn't account for. I think it would make sense to use the same lock as the scheduler for this.

> Also _killed_queue_jobs will get arbitrarily big if we kill a lot of jobs that aren't actually running or in the queue (maybe because they finished?). Not sure if there's anything reasonable to do about that though.

This is also true, but yeah I don't think we can do anything about it? We could check if the job is actually in the queue, but it might just be possible that a job hasn't been added to our queue yet and we end up not considering it.

w-gao:
@adamnovak Thanks for the review! I did consider introducing a new base class to queue up the jobs and manage the resource pools, and I think it would make sense to do that when we want to support this for the other batch systems like TES, AWS Batch, and even the Grid Engine batch systems. But the single machine batch system already has its own queue for the subprocesses with additional logic to handle finished child processes, and I didn't want to mess with that in this PR.

adamnovak (approved):
This looks good to me now!

PR #4463, opened 2023-05-04
Added type hinting to resource.py #938

stxue1:
- Type hinting to functions in resource.py
- Added return type to inVirtualEnv() in __init__.py
- Added None checks to some function bodies

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -523,9 +523,9 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
             # toil is being run interactively, in which case
             # we can reasonably assume that we are not running
             # on a worker node.
-            if mainModule.__file__ is None:
-                return False
             try:
+                if mainModule.__file__ is None:
+                    return False
                 mainModuleFile = os.path.basename(mainModule.__file__)
             except AttributeError:
                 return False
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -90,9 +90,8 @@ class Resource(namedtuple('Resource', ('name', 'pathHash', 'url', 'contentHash')
         with cls._load(leaderPath) as src:
             with jobStore.write_shared_file_stream(shared_file_name=pathHash, encrypted=False) as dst:
                 userScript = src.read()
-                if isinstance(userScript, bytes):
-                    contentHash.update(userScript)
-                    dst.write(userScript)
+                contentHash.update(userScript)
+                dst.write(userScript)
         return cls(name=os.path.basename(leaderPath),
                    pathHash=pathHash,
                    url=jobStore.getSharedPublicUrl(sharedFileName=pathHash),
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -90,9 +90,8 @@ class Resource(namedtuple('Resource', ('name', 'pathHash', 'url', 'contentHash')
         with cls._load(leaderPath) as src:
             with jobStore.write_shared_file_stream(shared_file_name=pathHash, encrypted=False) as dst:
                 userScript = src.read()
-                if isinstance(userScript, bytes):
-                    contentHash.update(userScript)
-                    dst.write(userScript)
+                contentHash.update(userScript)
+                dst.write(userScript)
         return cls(name=os.path.basename(leaderPath),
                    pathHash=pathHash,
                    url=jobStore.getSharedPublicUrl(sharedFileName=pathHash),
@@ -524,8 +523,6 @@ class ModuleDescriptor(namedtuple('ModuleDescriptor', ('dirPath', 'name', 'fromV
             # toil is being run interactively, in which case
             # we can reasonably assume that we are not running
             # on a worker node.
-            if mainModule.__file__ is None:
-                return False
             try:
                 if mainModule.__file__ is None:
                     return False
--- a/src/toil/resource.py
+++ b/src/toil/resource.py
@@ -24,10 +24,17 @@ from contextlib import closing
 from io import BytesIO
 from pydoc import locate
 from tempfile import mkdtemp
-from typing import Any
 from urllib.error import HTTPError
 from urllib.request import urlopen
 from zipfile import ZipFile
+
+from typing import (TYPE_CHECKING,
+                    Optional,
+                    Callable,
+                    IO,
+                    Type,
+                    Sequence,
+                    BinaryIO)
 
 from toil import inVirtualEnv
 from toil.lib.iterables import concat
@@ -35,8 +42,12 @@ from toil.lib.memoize import strict_bool
 from toil.lib.retry import ErrorCondition, retry
 from toil.version import exactPython
 
-logger = logging.getLogger(__name__)
+from types import ModuleType
 
+if TYPE_CHECKING:
+    from toil.jobStores.abstractJobStore import AbstractJobStore
+
+logger = logging.getLogger(__name__)
 
 class Resource(namedtuple('Resource', ('name', 'pathHash', 'url', 'contentHash'))):
     """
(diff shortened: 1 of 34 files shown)


src/toil/resource.py
```diff
@@ -24,19 +24,33 @@
...
 from toil.lib.iterables import concat
 from toil.lib.memoize import strict_bool
 from toil.lib.retry import ErrorCondition, retry
 from toil.version import exactPython
 
-logger = logging.getLogger(__name__)
+if TYPE_CHECKING:
+    from toil.jobStores.abstractJobStore import AbstractJobStore
+    from types import ModuleType
+    from io import FileIO
```
adamnovak, line 51: We might need to import these unconditionally; I think we only want to use `TYPE_CHECKING` to break circular imports, and then the things need to be quoted strings when actually used in types, since they won't be defined at runtime.

src/toil/resource.py
```diff
@@ -81,14 +95,15 @@ def create(cls, jobStore, leaderPath):
         with cls._load(leaderPath) as src:
             with jobStore.write_shared_file_stream(shared_file_name=pathHash, encrypted=False) as dst:
                 userScript = src.read()
-                contentHash.update(userScript)
-                dst.write(userScript)
+                if isinstance(userScript, bytes):
```
adamnovak, line 98: I might do an `is not None` here to be a little clearer as to why it might not be a `bytes`.

src/toil/resource.py
```diff
@@ -181,27 +196,27 @@ def localPath(self):
...
+        return locate(className)(*json.loads(_json)) # type: ignore
 
     @classmethod
-    def _pathHash(cls, path):
+    def _pathHash(cls, path: str) -> str:
         return hashlib.md5(path.encode('utf-8')).hexdigest()
 
     @classmethod
-    def _load(cls, path):
+    def _load(cls, path: str) -> Union[IO[bytes], IO[str]]:
```
adamnovak, line 219: This doesn't look like a useful type for this to be. This is the same `_load` I was complaining about above where you only use the result if it is a `bytes`, right?

I don't think discarding `str` versions is the right approach; we should make this always `IO[bytes]` and then anything that isn't currently returning an `IO[bytes]` probably should be changed to do so.

adamnovak: Also you can drop all the `:rtype:` comments now that we have real type hints.

src/toil/resource.py
```diff
@@ -243,15 +258,15 @@ class FileResource(Resource):
     """A resource read from a file on the leader."""
 
     @classmethod
-    def _load(cls, path):
+    def _load(cls, path: str) -> TextIO:
         return open(path)
```
adamnovak, line 262: If we harmonize on binary-mode streams, this would change to a binary mode open and a binary-mode return value.

src/toil/resource.py
```diff
@@ -378,8 +393,10 @@ def forModule(cls, name: str) -> Any:
         method assumes that the module with the specified name has already been loaded.
         """
         module = sys.modules[name]
-        filePath = os.path.abspath(module.__file__)
-        filePath = filePath.split(os.path.sep)
+        fileAbsPath = ""
+        if module.__file__ is not None:
```
adamnovak, line 397: We might want to throw a specific error if `__file__` is None. Right now it will complain that the (nonexistent) file doesn't end in `.py`.

src/toil/resource.py
```diff
@@ -442,13 +459,13 @@ def _check_conflict(cls, dirPath, name):
             sys.path = old_sys_path
 
     @property
-    def belongsToToil(self):
+    def belongsToToil(self) -> Any:
```
adamnovak, line 462: This should return `bool`, right?

src/toil/resource.py
```diff
@@ -499,7 +517,7 @@ def stash(tmpDirPath):
                                   name=self.name,
                                   fromVirtualEnv=self.fromVirtualEnv)
 
-    def _runningOnWorker(self):
+    def _runningOnWorker(self) -> Union[bool, str]:
```
adamnovak, line 520: I think this is also meant to always return `True` or `False`.

src/toil/resource.py
```diff
@@ -511,7 +529,10 @@ def _runningOnWorker(self):
             # we can reasonably assume that we are not running
             # on a worker node.
             try:
-                mainModuleFile = os.path.basename(mainModule.__file__)
+                if mainModule.__file__ is not None:
```
adamnovak, line 532: I think we are trying to detect if there is no `__file__` set, and in that case return `False`. With this check we no longer get an exception so I think the function won't work anymore.

src/toil/resource.py
```diff
@@ -537,7 +558,7 @@ def globalize(self) -> "ModuleDescriptor":
                                   fromVirtualEnv=fromVirtualEnv)
 
     @property
-    def _resourcePath(self):
+    def _resourcePath(self) -> Any:
```
adamnovak, line 561: Why can't this be `str`?

src/toil/resource.py
```diff
@@ -557,35 +578,35 @@ def _resourcePath(self):
...
     @classmethod
-    def _initModuleName(cls, dirPath):
+    def _initModuleName(cls, dirPath: str) -> Optional[str]:
         for name in ('__init__.py', '__init__.pyc', '__init__.pyo'):
             if os.path.exists(os.path.join(dirPath, name)):
                 return name
         return None
 
-    def _rootPackage(self):
+    def _rootPackage(self) -> Any:
```
adamnovak, line 587: Doesn't this always return `str`? Since `self.name` would be a `str` if it is being split?

src/toil/resource.py
```diff
@@ -557,35 +578,35 @@ def _resourcePath(self):
...
         try:
             head, tail = self.name.split('.', 1)
         except ValueError:
-            raise ValueError('%r is stand-alone module.' % self)
+            raise ValueError('%r is stand-alone module.' % self.__repr__())
         else:
             return head
 
-    def toCommand(self):
+    def toCommand(self) -> Tuple[str, ...]:
```
adamnovak, line 595: Maybe use `Sequence[str]` as the return type so you can say all the elements are `str` and not just the first one?

adamnovak: Then it would be allowable to pass the result of `toCommand()` to `fromCommand()`, which would make sense.

adamnovak (changes requested):
This is pretty great! But there are some places where `Any` is being used for reasons that aren't made clear, and a couple places where I am concerned about functionality possibly breaking.

adamnovak (approved):
I'm still worried that check against bytes type is going to come back to bite us, but the tests probably will catch it? So I think we can go ahead with this.

stxue1:
I think I forgot to take the check against byte type out. It is no longer needed as the `_load()` function should return `IO[bytes]` that makes `read` open it in bytes automatically. I added an extra commit to the issues branch that removes it.

DailyDreaming (approved):
This looks good!  One last thing before merging... please remove this line so that mypy checks (and maintains) the correct type checking of this file: https://github.com/DataBiosphere/toil/blob/master/contrib/admin/mypy-with-ignore.py#L32

adamnovak:
OK, I synced up the branches so `issues/938-type-resource` is now the same commit in @stxue1's repo being PR'd and in the DataBiosphere repo being tested, and both are up to date. I think this will probably pass again and then we can merge it!

DailyDreaming (dismissed):

src/toil/resource.py
```diff
@@ -511,6 +524,8 @@ def _runningOnWorker(self):
             # we can reasonably assume that we are not running
             # on a worker node.
             try:
+                if mainModule.__file__ is None:
```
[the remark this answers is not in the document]
stxue1: I think this should be fixed in [this commit I pushed today](https://github.com/DataBiosphere/toil/pull/4463/commits/d6f93fa516ff2a5253a878de6e7d24209143472e), there were two checks for `__file__`, so I removed the redundant one and the other one should be caught within the try block.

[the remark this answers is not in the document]
adamnovak: Yeah this looks fixed.

adamnovak:
@stxue1 it looks like you've made different commits to issues/938-type-resource in your repo and issues/938-type-resource in the DataBiosphere/toil repo, when to merge this we need to have both those branches pointing at the same commit hash. I'm going to try and merge them again.

If you want to make another commit, make sure to make the commit once and then push the same local issues/938-type-resource branch to both your repo and to DataBiosphere/toil.

PR #4545, opened 2023-07-21
Fall back to 1 core when # CPUs unavailable

tjni:
## Changelog Entry

 * Fall back to assuming machine has 1 core when CPU count is unavailable.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak (changes requested):
At the end of this function we do:

```
    # Return the smaller of the actual thread count and the cgroup's limit, minimum 1.
    result = cast(int, max(1, min(min(affinity_size, cgroup_size), total_machine_size)))
```

So I think if we let `total_machine_size` fall back to 1, we won't ever be able to get an answer above 1, even of e.g. `os.sched_getaffinity()` is available and working properly.

If we really want to properly support cases where `psutil.cpu_count(logical=True)` is unavailable, we need to also change the way we combine the values so that we don't use the CPU count as an upper limit when it's not available.

adamnovak:
@tjni Do you want to change they way the values get combined, or should we do it?

tjni:
You're correct, I'm sorry I missed that case. I can work on it given that it's my odd environment triggering this corner case, and I'll add a test for this too.

My environment is building this package bundled in [nixpkgs](https://github.com/NixOS/nixpkgs) using the nix package manager. This package manager has a sandbox mode, which prevents access to a lot of system functionality (in my case, it uses [the Apple sandbox](https://reverse.put.as/wp-content/uploads/2011/09/Apple-Sandbox-Guide-v1.0.pdf)). This includes access to [sysctlbyname](https://github.com/giampaolo/psutil/blob/962cb9ad58ca3bc21b20130f435f8d507d931051/psutil/arch/osx/cpu.c#L39) unless explicitly allowed. Aside from that, one way to fix this in nixpkgs is just to disable the tests require going outside of this sandbox, but I noticed from psutil docs that this function could return None so thought it would be nice to fill that gap.

adamnovak:
@tjni Are you still planning to revise this, or should we take over?

adamnovak (approved):

adamnovak:
I'm going to close this in favor of the same commits in #4780 from within the repo.

PR #4569, opened 2023-08-17
Add config file support

### LINKED ISSUE #4538: Add config file for Toil
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1374)
┆Issue Number: TOIL-1374

stxue1:
## Changelog Entry
This will resolve #4538 
 * Added config file support
    * Toil will now use configargparse in the place of argparse
    * Centralized CLI argument's defaults and environmental variables
    * Arguments that take a value have a camel case name
 * Changes to tests, cwltoil, and wdltoil to use configpargparse
 * Added a `toil config` tool to generate a default config file
    * Added a test to CI to test `toil config`'s functionality

 * Breaking changes: Existing Toil workflows using argparse will no longer properly support environmental variables. This can be fixed by using configargparse's `ArgumentParser()` or `Job.Runner.getDefaultArgumentParser()`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/common.py
```diff
@@ -463,6 +477,63 @@ def __hash__(self) -> int:
...
+    """
+    Write a Toil config file to the given path.
+
+    Safe to run simultaneously in multiple processes. No process will see an
+    empty or partially-written file at the given path.
+    """
+    # this is placed in common.py rather than toilConfig.py to prevent circular imports
+
+    # configargparse's write_config does not write options with a None value, so do this instead
+    omit = ("help", "config", "defaultAccelerators", "nodeTypes", "nodeStorageOverrides", "setEnv", "minNodes", "maxNodes", "logCritical", "logDebug", "logError", "logInfo", "logOff", "logWarning", "linkImports", "noLinkImports", "moveExports", "noMoveExports", "disableCaching")
```
adamnovak, line 502: I don't really understand what "this" is. Why do these have to be excluded from the default config? Since we're just dumping YAML ourselves, why do we care what their `write_config` does? Are we trying to mimic it for some reason?

Instead of just listing facts about some options here (which will change when the options are added/removed/adjusted), can't we look at the stored default value instead to detect the options we want to skip?

src/toil/common.py
```diff
@@ -498,20 +568,116 @@ def parser_with_common_options(
...
+    class CheckCoordinationDir(_StoreAction):
+        def __call__(self, parser: Any, namespace: Any, values: Any, option_string: Any=None) -> None:
+            coordination_dir = values
+            if coordination_dir is not None:
+                coordination_dir = os.path.abspath(coordination_dir)
+                if not os.path.exists(coordination_dir):
+                    raise RuntimeError(
+                        f"The path provided to --coordinationDir ({coordination_dir}) does not exist.")
+            setattr(namespace, self.dest, values)
+    return CheckCoordinationDir
```
adamnovak, line 581: Can't this just be a normal class? Why do we need the factory function if it takes no arguments?

src/toil/common.py
```diff
@@ -498,20 +568,116 @@ def parser_with_common_options(
...
+
+
+def make_workdir_check_action() -> Type[_StoreAction]:
```
adamnovak, line 584: Same for this one. Also we might want docstrings for all these actions.

src/toil/common.py
```diff
@@ -498,20 +568,116 @@ def parser_with_common_options(
...
+                if not is_within(values):
+                    raise parser.error(f"{option_string} ({values}) must be within the range: [{min}, {'infinity' if max is None else max}]")
+            except AssertionError:
+                raise RuntimeError(f"The {option_string} option has an invalid value: {values}")
+            setattr(namespace, self.dest, values)
+    return RangeValidation
+
+
+def make_check_ssekey_action() -> Type[_StoreAction]:
+    class SSEKeyValidation(_StoreAction):
```
adamnovak, line 644: I feel like the action names could be better. Some of these are imperatives ("Check whatever"), some are nouns for processes ("Whatever validation"). And in default argparse you say things like `action="store_true"` where the name is a noun for the action that is going to happen.

We should settle on one naming scheme, preferably one that argparse already has some evidence of using for its actions, and we should use it consistently for all our custom actions. Maybe this would be `StoreSSEKey`? Or `ParseSSEKey`? Or, since Argparse ships a `BooleanOptionalAction`, maybe this is an `SSEKeyAction`? Or `ValidatedSSEKeyAction`?

src/toil/common.py
```diff
@@ -498,20 +568,116 @@ def parser_with_common_options(
...
     """
 
-    if config is None:
-        config = Config()
     if not (isinstance(parser, ArgumentParser) or isinstance(parser, _ArgumentGroup)):
         raise ValueError(f"Unanticipated class: {parser.__class__}.  Must be: argparse.ArgumentParser or ArgumentGroup.")
 
+    if isinstance(parser, ArgParser):
+        # in case the user passes in their own configargparse instance instead of calling getDefaultArgumentParser()
+        parser._config_file_parser = YAMLConfigFileParser()  # type: ignore[misc]
```
adamnovak, line 670: Why does this need to happen?

src/toil/common.py
```diff
@@ -498,20 +568,116 @@ def parser_with_common_options(
...
+    else:
+        # configargparse advertises itself as a drag and drop replacement, and running the normal argparse ArgumentParser
+        # through this code still seems to work (with the exception of config file and environmental variables)
+        warnings.warn(f'Using deprecated library argparse for options parsing.'
+                      f'This will not parse config files or use environmental variables.'
```
adamnovak, line 675:
```suggestion
                      f'This will not parse config files or use environment variables.'
```

src/toil/common.py
```diff
@@ -591,24 +761,27 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
 
     caching = file_store_options.add_mutually_exclusive_group()
     caching_help = ("Enable or disable caching for your workflow, specifying this overrides default from job store")
-    caching.add_argument('--disableCaching', dest='caching', action='store_false', help=caching_help)
-    caching.add_argument('--caching', dest='caching', type=lambda val: bool(strtobool(val)), help=caching_help)
-    caching.set_defaults(caching=None)
+    caching.add_argument('--caching', dest='caching', type=opt_strtobool, default=None, help=caching_help)
+    caching.add_argument('--enableCaching', dest='caching', action='store_true', help=argparse.SUPPRESS)
+    caching.add_argument('--disableCaching', dest='caching', action='store_false', help=argparse.SUPPRESS)
+    # caching.set_defaults(caching=None) # default is None according to PR 4299, seems to be generated at runtime
```
adamnovak, line 784:
```suggestion
    # default is None according to PR 4299, seems to be generated at runtime
```

src/toil/common.py
```diff
@@ -617,20 +790,21 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
+                                          f"choices are {provisioner_choices}.  The default is {None}.")
+    autoscaling_options.add_argument('--node_types', default=[], dest="nodeTypes", type=parse_node_type, nargs="*",
                                      help="Specifies a list of comma-separated node types, each of which is "
                                           "composed of slash-separated instance types, and an optional spot "
                                           "bid set off by a colon, making the node type preemptible. Instance "
                                           "types may appear in multiple node types, and the same node type "
                                           "may appear as both preemptible and non-preemptible.\n"
                                           "Valid argument specifying two node types:\n"
-                                          "\tc5.4xlarge/c5a.4xlarge:0.42,t2.large\n"
+                                          "\t[c5.4xlarge/c5a.4xlarge:0.42, t2.large]\n"
```
adamnovak, line 807:
```suggestion
                                          "\tc5.4xlarge/c5a.4xlarge:0.42,t2.large\n"
```

I don't think we're actually able to parse brace-enclosed, space-separated lists form the command line are we?

src/toil/common.py
```diff
@@ -639,79 +813,90 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
                                           f"average of the estimated number of nodes. A value of 0.0 disables any "
                                           f"smoothing, and a value of 0.9 will smooth so much that few changes will "
-                                          f"ever be made.  Must be between 0.0 and 0.9.  default={config.betaInertia}")
-    autoscaling_options.add_argument("--scaleInterval", dest="scaleInterval", default=None,
+                                          f"ever be made.  Must be between 0.0 and 0.9.  default={0.1}")
+    autoscaling_options.add_argument("--scale_interval", "--scaleInterval", dest="scaleInterval", default=60, type=int,
                                      help=f"The interval (seconds) between assessing if the scale of "
-                                          f"the cluster needs to change. default={config.scaleInterval}")
-    autoscaling_options.add_argument("--preemptibleCompensation", "--preemptableCompensation", dest="preemptibleCompensation", default=None,
+                                          f"the cluster needs to change. default={60}")
```
adamnovak, line 844: I think we want to use `%(default)s` in these messages instead of putting the default value again.

src/toil/common.py
```diff
@@ -639,79 +813,90 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
                                      help="Enable the prometheus/grafana dashboard for monitoring CPU/RAM usage, "
                                           "queue size, and issued jobs.")
-    autoscaling_options.add_argument("--assumeZeroOverhead", dest="assume_zero_overhead", default=False, action="store_true",
+    autoscaling_options.add_argument("--assume_zero_overhead", "--assumeZeroOverhead", dest="assume_zero_overhead", default=False, type=convert_bool,
                                      help="Ignore scheduler and OS overhead and assume jobs can use every last byte "
                                           "of memory and disk on a node when autoscaling.")
 
     # Parameters to limit service jobs / detect service deadlocks
-    if not config.cwl:
+    # if not config.cwl:
```
adamnovak, line 874:
```suggestion
```

src/toil/common.py
```diff
@@ -729,60 +914,65 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
-    resource_options.add_argument('--defaultDisk', dest='defaultDisk', default=None, metavar='INT',
-                                  help=resource_help_msg.format('default', 'disk', disk_mem_note, bytes2human(config.defaultDisk)))
-    resource_options.add_argument('--defaultAccelerators', dest='defaultAccelerators', default=None, metavar='ACCELERATOR[,ACCELERATOR...]',
-                                  help=resource_help_msg.format('default', 'accelerators', accelerators_note, config.defaultAccelerators))
-    resource_options.add_argument('--defaultPreemptible', '--defaultPreemptable', dest='defaultPreemptible', metavar='BOOL',
-                                  type=bool, nargs='?', const=True, default=False,
+
+    h2b = lambda x: human2bytes(str(x))
+
+    resource_options.add_argument("--default_memory", '--defaultMemory', dest='defaultMemory', default=2147483648, type=h2b, action=make_int_range_validation_action(1),
```
adamnovak, line 920: Shouldn't these magic numbers also be `SYS_MAX_SIZE`?

src/toil/common.py
```diff
@@ -791,70 +981,87 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
-    misc_options.add_argument('--disableProgress', dest='disableProgress', action='store_true', default=False,
+    misc_options.add_argument("--disable_progress", '--disableProgress', dest='disableProgress', type=convert_bool, default=False,
                               help="Disables the progress bar shown when standard error is a terminal.")
+    # If using argparse instead of configargparse, this should just not parse when calling parse_args()
+    # defaults are already populated at config init
+    # env_config = os.getenv("TOIL_CONFIG")
+    # home_config_path = os.path.join(os.path.expanduser("~"), "/.config/toil.cfg")
+    # home_config = home_config_path if os.path.exists(home_config_path) else None
+    # default_config_list = [env_config, home_config]
+    # default_config = None if len(default_config_list) == 0 else default_config_list[0]
```
adamnovak, line 1043:
```suggestion
```

src/toil/common.py
```diff
@@ -791,70 +981,87 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
+    default_config = None
+    misc_options.add_argument('--config', dest='config', is_config_file_arg=True, default=default_config,
+                              help="Get options from a config file.")
```
adamnovak, line 1046:
```suggestion
    misc_options.add_argument('--config', dest='config', is_config_file_arg=True, default=None,
                              help="Get options from a config file.")
```

src/toil/leader.py
```diff
@@ -35,7 +35,7 @@
                       JobIssuedMessage,
                       JobMissingMessage,
                       JobUpdatedMessage,
-                      QueueSizeMessage)
+                      QueueSizeMessage, gen_message_bus_path)
```
adamnovak, line 38:
```suggestion
                      QueueSizeMessage,
                      gen_message_bus_path)
```

src/toil/leader.py
```diff
@@ -115,6 +115,11 @@ def __init__(self,
         # state change information about jobs.
         self.toilState = ToilState(self.jobStore)
 
+        if self.config.write_messages is None:
+            # The user hasn't specified a place for the message bus so we
+            # should make one.
+            self.config.write_messages = gen_message_bus_path()
+
         if self.config.write_messages is not None:
```
adamnovak, line 123: But now this is always true and so not needed, right?

src/toil/provisioners/__init__.py
```diff
@@ -81,6 +81,62 @@ def add_provisioner_options(parser: argparse.ArgumentParser) -> None:
                                            "Must be lowercase and may not contain the '_' character.")
 
 
+def parse_node_type(node_type_spec: Optional[str]) -> Optional[Tuple[Set[str], Optional[float]]]: # move later
```
adamnovak, line 84: Where is this supposed to move?

src/toil/server/app.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+from configargparse import ArgumentParser
```
adamnovak, line 15: Why do we sometimes import `ArgParser` and sometimes import `ArgumentParser`? Can we do one consistently?

src/toil/test/src/fileStoreTest.py
```diff
@@ -472,7 +472,15 @@ def _testValidityOfCacheEvictTest(self):
             # the cache hence this test is redundant (caching will be free).
             if not self.options.jobStore.startswith(('aws', 'google')):
                 workDirDev = os.stat(self.options.workDir).st_dev
-                jobStoreDev = os.stat(os.path.dirname(self.options.jobStore)).st_dev
+                if self.options.jobStore.startswith("file:"):
+                    # Before #4538, options.jobStore would have the raw path while the Config object would prepend the
+                    # filesystem to the path (/path/to/file vs file:/path/to/file)
+                    # The options namespace and the Config object now have the exact same behavior
+                    # which means parse_jobstore will be called with argparse rather than with the config object
+                    # so remove the prepended filesystem
```
adamnovak, line 480:
```suggestion
                    # so remove the prepended file: scheme
```

src/toil/test/utils/ABCWorkflowDebug/debugWorkflow.py
```diff
@@ -139,7 +139,7 @@ def broken_job(job, num):
 if __name__=="__main__":
     jobStorePath = sys.argv[1] if len(sys.argv) > 1 else tempfile.mkdtemp("debugWorkflow")
     options = Job.Runner.getDefaultOptions(jobStorePath)
-    # options.clean = "always"
+    options.clean = "never"
```
adamnovak, line 142:
```suggestion
```

src/toil/test/utils/utilsTest.py
```diff
@@ -115,6 +115,25 @@ def statusCommand(self, failIfNotComplete=False):
             commandTokens.append('--failIfNotComplete')
         return commandTokens
 
+    def test_config_functionality(self):
+        """Ensure that creating and reading back the config file works"""
+        config_file = os.path.abspath("config.yaml")
+        config_command = [self.toilMain, 'config', config_file]
+        try:
+            system(config_command)
```
adamnovak, line 123: Shouldn't we be using `subprocess.check_call` here?

src/toil/utils/toilStatus.py
```diff
@@ -360,6 +360,7 @@ def main() -> None:
         sys.exit(0)
 
     config = Config()
+    config.clean = "never"
```
adamnovak, line 363: Is there a reason we need this here?

adamnovak (changes requested):
This looks pretty good! But I am worried about the duplicated default values in the help messages, and how it seems to move us over to snake_case options as the main names without really explaining that or updating the docs.

src/toil/common.py
```diff
@@ -115,113 +140,166 @@ class Config:
...
+                logger.debug("Loaded default configuration: %s", json.dumps(s, indent=4))
+        except:
+            # Something went wrong reading the default config, so dump its
+            # contents to the log.
+            logger.info("Configuration file contents: %s", open(DEFAULT_CONFIG_FILE, 'r').read())
+            raise
+        parser = ArgParser()
+        addOptions(parser, self, jobstore_as_flag=True)
+        ns = parser.parse_args(f"--config={DEFAULT_CONFIG_FILE}")
+        self.setOptions(ns)
```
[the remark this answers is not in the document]
stxue1: If it's caught during `parse_args()`, it should just say which argument has the issue, such as `example.py: error: --retryCount (-1) must be within the range: [0, infinity)`. The behavior should be the same as passing in an invalid argument through the CLI.

src/toil/common.py
stxue1: It seems like `defaultMemory` and `defaultDisk` controls the amount of resources that the system will request, and setting it to `SYS_MAX_SIZE` will return InsufficientResources. The docs say defaultMemory should be 2GB and the value here is 2GiB, so maybe the default should be changed to match the docs?

src/toil/provisioners/__init__.py
```diff
@@ -81,6 +81,62 @@ def add_provisioner_options(parser: argparse.ArgumentParser) -> None:
...
+    for AWS), and an optional bid in dollars after a colon.
+
+    Raises ValueError if the node type cannot be parsed.
+
+    Inputs should look something like this:
+
+    >>> parse_node_types('c5.4xlarge/c5a.4xlarge:0.42,t2.large')
+    ({'c5.4xlarge', 'c5a.4xlarge'}, 0.42)
+    >>> parse_node_types('t2.large')
+    ({'t2.large'}, None)
```
adamnovak, line 100: This probably has to change to talk about this function and not the function it was copied from.

src/toil/provisioners/__init__.py
```diff
@@ -81,6 +81,62 @@ def add_provisioner_options(parser: argparse.ArgumentParser) -> None:
...
+            parsed = (instance_types, float(parts[1]) if len(parts) > 1 else None)
+        except Exception as e:
+            if isinstance(e, ValueError):
+                raise
+            else:
+                raise ValueError(f'Could not parse node type "{node_type_spec}"')
+
+    return parsed
+
 def parse_node_types(node_type_specs: Optional[str]) -> List[Tuple[Set[str], Optional[float]]]:
```
adamnovak, line 140: Shouldn't this now call into `parse_one_node_type`?

src/toil/provisioners/__init__.py
```diff
@@ -81,6 +81,62 @@ def add_provisioner_options(parser: argparse.ArgumentParser) -> None:
...
+    ({'c5.4xlarge', 'c5a.4xlarge'}, 0.42)
+    >>> parse_node_types('t2.large')
+    ({'t2.large'}, None)
+
+    :param node_type_specs: A string defining a node type
+
+    :returns: the parsed node type, where the type is the set of
+              instance types, and the float bid, or None.
+    """
+    # Collect together all the node types
```
adamnovak, line 107: I think a bunch of comments in here are also inaccurate now, if this is working just on one node type.

src/toil/common.py
```diff
@@ -617,101 +839,111 @@ def addOptions(parser: ArgumentParser, config: Optional[Config] = None, jobstore
...
     # TODO: Better consolidate this provisioner arg and the one in provisioners/__init__.py?
     autoscaling_options.add_argument('--provisioner', '-p', dest="provisioner", choices=provisioner_choices,
+                                     default=None,
                                      help=f"The provisioner for cluster auto-scaling.  This is the main Toil "
                                           f"'--provisioner' option, and defaults to None for running on single "
                                           f"machine and non-auto-scaling batch systems.  The currently supported "
-                                          f"choices are {provisioner_choices}.  The default is {config.provisioner}.")
-    autoscaling_options.add_argument('--nodeTypes', default=None,
+                                          f"choices are {provisioner_choices}.  The default is %(default)s.")
+    autoscaling_options.add_argument('--node_types', default=[], dest="node_types", type=parse_one_node_type, nargs="*",
```
adamnovak, line 850: It looks like this is meant to now be used like `--node_types t2.large c5.4xlarge`. But the documentation now shows you passing a bracket-enclosed list that I don't think `parse_one_node_type` can actually parse. Also, using `nargs="*"` can cause trouble because it doesn't really know when to stop consuming arguments. If I have `toil-cwl-runner --logDebug script.cwl inputs.json` and I want to set a node type and I do `toil-cwl-runner --logDebug --node_types c5.4xlarge script.cwl inputs.json`, then suddenly my positional arguments have been eaten.

We might want to consider `action="append"` here instead, and a singular option that can repeat. Like `--node_type t2.large:1.5 --node_type c5.4xlarge/c5a.4xlarge:0.42`.

Either way we have to make sure to fix the docs/examples to show it the way that actually works on the command line, and not with square brackets.

src/toil/test/utils/utilsTest.py
```diff
@@ -115,6 +115,25 @@ def statusCommand(self, failIfNotComplete=False):
...
         return commandTokens
 
+    def test_config_functionality(self):
+        """Ensure that creating and reading back the config file works"""
+        config_file = os.path.abspath("config.yaml")
+        config_command = [self.toilMain, 'config', config_file]
+        try:
+            subprocess.check_call(config_command)
+        except subprocess.CalledProcessError:  # This happens when the script fails due to having unfinished jobs
+            self.assertRaises(subprocess.CalledProcessError, system)
```
adamnovak, line 125: This looks like it invokes the `system` function with no arguments and makes sure it raises an error, but I can't figure out what it is doing here.

src/toil/common.py
```diff
@@ -23,10 +23,15 @@
...
+
+from ruamel.yaml import YAML
+from ruamel.yaml.comments import CommentedMap
+from configargparse import ArgParser, YAMLConfigFileParser
+from argparse import (SUPPRESS,
+                      ArgumentDefaultsHelpFormatter,
                       ArgumentParser,
                       Namespace,
-                      _ArgumentGroup)
+                      _ArgumentGroup, _StoreAction, _AppendAction, _StoreFalseAction, _StoreTrueAction)
```
adamnovak, line 34: Do these get used?

src/toil/common.py
```diff
@@ -463,6 +488,99 @@ def __hash__(self) -> int:
...
+    #   linkImports, moveExports, disableCaching, are deprecated in favor of --link_imports, --move_exports,
+    #   and --caching respectively
+    # Skip these two to ensure that values from the config file and argparse defaults are one to one:
+    # (ex: an argparse option that parsed a string list of "1,2,3" -> [1,2,3] has been changed to just have its list
+    # representation)
+    #   setEnv is deprecated in favor of --environment, as this lets the config file have a native dictionary
+    #   minNodes, maxNodes, and nodeStorageOverrides, and nodeTypes are deprecated in favor of min_nodes, max_nodes,
+    #   node_storage_overrides, and node_types in order for the argument to be represented as a native list
+    # Also skip StoreTrue and StoreFalse options that have opposite defaults as including it in the config would
+    # override those defaults
```
adamnovak, line 533: This makes a lot more sense!

adamnovak (changes requested):
This is better, but there are still some pieces that seem unfinished/bits of what looks like commented-out code.

Also, I'm still not quite happy with the new options.
* I don't think the command line help/option documentation in the RST files is showing how they actually end up needing to be used.
* I feel like we ought to be able to do some trickery with the actions or the `type` functions to be able to have the typed/structured representation we want in the config file while still having a natural CLI representation. We shouldn't be asking people to e.g. send in JSON on the command line to specify environment variables, but we should be able to have them use a YAML dictionary in the config file. What's the obstacle that is preventing that?
* I don't want the non-deprecated options for Toil to be a mix of camelCase and snake_case options on the same command line. @DailyDreaming do you have an opinion about what that ought to look like? Do you think this would really be fine and we should start gradually moving over to snake_case options?

src/toil/batchSystems/options.py
```diff
@@ -17,6 +17,10 @@
 from argparse import ArgumentParser, _ArgumentGroup
 from typing import Any, Callable, List, Optional, TypeVar, Union, cast
 
+from configargparse import ArgParser
+
+SYS_MAX_SIZE = 9223372036854775807
```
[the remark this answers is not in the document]
stxue1: I think the original defaults were sys.maxsize, but similar defaults in common.py use SYS_MAX_SIZE. This option was duplicated from common.py since python complained about circular imports. I can probably import SYS_MAX_SIZE within the add_all_batchsystem_options function, or switch to using sys.maxsize, but I would like the default values to be static.

src/toil/common.py
stxue1: Some of the custom action classes and the config file generator use these classes. Ideally I'd like for them to only use argparse.Action, but I haven't found a workaround for that.

docs/running/cliOptions.rst
```diff
@@ -463,10 +477,10 @@ systems have issues!).
                         for server-side encryption on awsJobStore or
                         googleJobStore. SSE will not be used if this flag is
                         not passed.
-  --setEnv NAME, -e NAME
-                        NAME=VALUE or NAME, -e NAME=VALUE or NAME are also valid.
+  --environment NAME_DICT, -e NAME_DICT
+                        ``-e="{NAME: VALUE,...}"`` or ``-e="{NAME: null,...}"``.
```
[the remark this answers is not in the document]
stxue1: I think I found a way to preserve the previous implementations

adamnovak (changes requested):
I think some of the docs formatting is wrong, but it looks like the compatibility problems are all fixed!

adamnovak (approved):
The only remaining problem I see here is the proliferation of magic numbers for the argument defaults. But if you really want it like that for some reason @stxue1 I guess we can leave it?

stxue1:
I can change it to its string representation pretty easily. It also looks like the docs used GB instead of GiB, so I updated the docs to match as well.

PR #4589, opened 2023-09-06
Add String to File functionality into toil-wdl-runner

### LINKED ISSUE #4554: Support conversion from String to File in WDL
Some WDL workflows rely on being able to convert between a String and a File:

https://github.com/gatk-workflows/seq-format-conversion/blob/6b69dde1877de997effedd3388aeaca929db04da/paired-fastq-to-unmapped-bam.wdl#L65-L66

This is apparently [obvious and unambiguous](https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#type-coercion), but the spec doesn't really say in what filesystem this is supposed to happen, and MiniWDL doesn't seem to magically call the Toil file import.

We should make sure that string to file conversions actually get around to putting the file into the file store so that this workflow can work.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1381)
┆Issue Number: TOIL-1381

stxue1:
This will resolve #4554.
This uses monkey patching as MiniWDL just stores the filepath for these types of conversions. If there is a new extension point in MiniWDL, this should probably be replaced with a newer implementation. It also looks like workflow nodes by default already [run on the leader node](https://github.com/DataBiosphere/toil/blob/10a25897064b2637701ce43060b5ce6928f02c14/src/toil/wdl/wdltoil.py#L957-L959), so I didn't explicitly add those arguments for workflow nodes.

## Changelog Entry

 * Support string to file conversion in toil-wdl-runner

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/wdl/wdltoil.py
+++ b/src/toil/wdl/wdltoil.py
@@ -408,7 +408,7 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
     Standard library implementation for WDL as run on Toil.
     """
 
-    def __init__(self, file_store: AbstractFileStore):
+    def __init__(self, file_store: AbstractFileStore, execution_dir: Optional[str] = None):
         """
         Set up the standard library.
         """
@@ -425,6 +425,8 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
 
         # Keep the file store around so we can access files.
         self._file_store = file_store
+
+        self._execution_dir = execution_dir
 
     def _is_url(self, filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:', TOIL_URI_SCHEME]) -> bool:
         """
@@ -465,7 +467,12 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
             result = self._file_store.readGlobalFile(imported)
         else:
             # This is a local file
-            result = filename
+            # To support relative paths, join the execution dir and filename
+            # if filename is already an abs path, join() will do nothing
+            if self._execution_dir is not None:
+                result = os.path.join(self._execution_dir, filename)
+            else:
+                result = filename
 
         logger.debug('Devirtualized %s as openable file %s', filename, result)
         assert os.path.exists(result), f"Virtualized file {filename} looks like a local file but isn't!"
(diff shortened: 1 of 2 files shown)


src/toil/wdl/wdltoil.py
```diff
@@ -2297,12 +2331,44 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
         self.defer_postprocessing(workflow_job)
         return workflow_job.rv()
 
+# Monkey patch miniwdl's WDL.Value.Base.coerce() function
+# miniwdl recognizes when a string needs to be converted into a file
+# However miniwdl's string to file conversions is to just store the filepath
+# Toil needs to virtualize the file into the jobstore
+# So monkey patch coerce to always virtualize whenever a file is expected
+#   _virtualize_filename should detect if the value is already a file and return immediately if so
+# Sometimes string coerce is called instead, so monkeypatch string coerce
```
adamnovak, line 2346: I think some of this should be a docstring, and some looks like it ought to be comments inside the actual function.

adamnovak (changes requested):
This looks like it ought to work, I think. But the documentation for `monkeypatch_coerce` should be better.

src/toil/wdl/wdltoil.py
```diff
@@ -2314,12 +2347,44 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+@contextmanager
+def monkeypatch_coerce(standard_library: ToilWDLStdLibBase) -> Any:
```
adamnovak, line 2364: This function needs a docstring.

src/toil/wdl/wdltoil.py
```diff
@@ -735,15 +748,20 @@ def evaluate_decl(node: WDL.Tree.Decl, environment: WDLBindings, stdlib: WDL.Std
 
     return evaluate_named_expression(node, node.name, node.type, node.expr, environment, stdlib)
 
-def evaluate_call_inputs(context: Union[WDL.Error.SourceNode, WDL.Error.SourcePosition], expressions: Dict[str, WDL.Expr.Base], environment: WDLBindings, stdlib: WDL.StdLib.Base) -> WDLBindings:
+def evaluate_call_inputs(context: Union[WDL.Error.SourceNode, WDL.Error.SourcePosition], expressions: Dict[str, WDL.Expr.Base], environment: WDLBindings, stdlib: WDL.StdLib.Base, inputs_list: Optional[List[WDL.Tree.Decl]] = None) -> WDLBindings:
```
adamnovak, line 751: I think instead of taking `inputs_list`, we should take what's being stored in `inputs_dict` (a mapping from string to type). That makes more sense alongside `expressions`, which is a mapping from string to expression, and keeps `WDL.Tree` out of the function. We should push the code to convert from the list of `Decl`s to a collection of expected types out into the caller.

The docstring should explain what the new parameter is and how it needs to relate to `expressions`, and that is a lot easier if we can say it is "an optional dict from name to expected type, for the names in `expressions`", versus if we have to say it is "An optional list of WDL Decl workflow nodes, where there is a Decl for each entry in `expressions`, and the name of the Decl is the same as the name used for the expression, and the type of the Decl is the expected type for the expression".

It also would make it a lot easier to write a test for, since it's harder to construct `WDL.Tree.Decl` objects by hand.

adamnovak (changes requested):
This looks pretty good, but I think the API on `evaluate_call_inputs` needs to be changed to use more consistent argument representations, and it looks like there's a missing docstring on one of the new functions.

PR #4610, opened 2023-10-02
Ensure sibling files in toil-wdl-runner

### LINKED ISSUE #4595: Fix sibling file support in Toil WDL runner
We want to pass the sibling file conformance tests

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1396)
┆Issue Number: TOIL-1396

stxue1:
This will resolve #4595
## Changelog Entry

 * Ensure that sibling files are preserved between node uploads in toil-wdl-runner

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/wdl/wdltoil.py
+++ b/src/toil/wdl/wdltoil.py
@@ -306,7 +306,7 @@ def recursive_dependencies(root: WDL.Tree.WorkflowNode) -> Set[str]:
 
 TOIL_URI_SCHEME = 'toilfile:'
 
-def pack_toil_uri(file_id: FileID, file_dir: str, file_basename: str) -> str:
+def pack_toil_uri(file_id: FileID, dir_id: uuid.UUID, file_basename: str) -> str:
     """
     Encode a Toil file ID and its source path in a URI that starts with the scheme in TOIL_URI_SCHEME.
     """
@@ -314,7 +314,7 @@ def pack_toil_uri(file_id: FileID, file_dir: str, file_basename: str) -> str:
     # We urlencode everything, including any slashes. We need to use a slash to
     # set off the actual filename, so the WDL standard library basename
     # function works correctly.
-    return f"{TOIL_URI_SCHEME}{quote(file_id.pack(), safe='')}/{quote(file_dir)}/{quote(file_basename, safe='')}"
+    return f"{TOIL_URI_SCHEME}{quote(file_id.pack(), safe='')}/{quote(str(dir_id))}/{quote(file_basename, safe='')}"
 
 def unpack_toil_uri(toil_uri: str) -> Tuple[FileID, str, str]:
     """
@@ -431,7 +431,7 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
     """
     Standard library implementation for WDL as run on Toil.
     """
-    def __init__(self, file_store: AbstractFileStore, stdlib_id: Optional[Any] = None, execution_dir: Optional[str] = None):
+    def __init__(self, file_store: AbstractFileStore, execution_dir: Optional[str] = None):
         """
         Set up the standard library.
         """
@@ -450,7 +450,7 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
         self._file_store = file_store
 
         # UUID to differentiate which node files are virtualized from
-        self.stdlib_id = stdlib_id or uuid.uuid4()
+        self._parent_dir_to_ids: Dict[str, uuid.UUID] = dict()
 
         self._execution_dir = execution_dir
 
@@ -462,16 +462,6 @@ class ToilWDLStdLibBase(WDL.StdLib.Base):
             if filename.startswith(scheme):
                 return True
         return False
-
-    def _is_uuid(self, id: str) -> bool:
-        """
-        Test if a string is a valid UUID
-        """
-        try:
-            uuid.UUID(id)
-            return True
-        except ValueError:
-            return False
 
     @memoize
     def _devirtualize_filename(self, filename: str) -> str:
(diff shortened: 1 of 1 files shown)


adamnovak:
Looks like the MiniWDL self test does not pass:
```
	  File "/builds/databiosphere/toil/src/toil/worker.py", line 406, in workerScript
	    job._runner(jobGraph=None, jobStore=jobStore, fileStore=fileStore, defer=defer)
	  File "/builds/databiosphere/toil/src/toil/job.py", line 2811, in _runner
	  File "/builds/databiosphere/toil/src/toil/job.py", line 2728, in _run
	    return self.run(fileStore)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 1428, in run
	    bindings = devirtualize_files(bindings, standard_library)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 829, in devirtualize_files
	    return map_over_files_in_bindings(environment, stdlib._devirtualize_filename)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 984, in map_over_files_in_bindings
	    return map_over_typed_files_in_bindings(bindings, lambda _, x: transform(x))
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 974, in map_over_typed_files_in_bindings
	    return environment.map(lambda b: map_over_typed_files_in_binding(b, transform))
	  File "/builds/databiosphere/toil/venv/lib/python3.9/site-packages/WDL/Env.py", line 151, in map
	    fb = f(b)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 974, in <lambda>
	    return environment.map(lambda b: map_over_typed_files_in_binding(b, transform))
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 993, in map_over_typed_files_in_binding
	    return WDL.Env.Binding(binding.name, map_over_typed_files_in_value(binding.value, transform), binding.info)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 1019, in map_over_typed_files_in_value
	    new_path = transform(value.type, value.value)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 984, in <lambda>
	    return map_over_typed_files_in_bindings(bindings, lambda _, x: transform(x))
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 488, in _devirtualize_filename
	    file_id, parent_id, file_basename = unpack_toil_uri(filename)
	  File "/builds/databiosphere/toil/src/toil/wdl/wdltoil.py", line 334, in unpack_toil_uri
	    raise ValueError(f"Wrong number of path segments in URI: {toil_uri}")
	ValueError: Wrong number of path segments in URI: toilfile:77%3A0%3Afiles%2Ffor-job%2Fkind-WDLWorkflowNodeJob%2Finstance-5d763d7k%2Ffile-9c4f99f31c1e412382d40794ba45b968%2Ftmpatj37zsh//builds/databiosphere/toil/wdl-conformance-tests/tmpatj37zsh
```

src/toil/wdl/wdltoil.py
```diff
@@ -470,11 +485,21 @@ def _devirtualize_filename(self, filename: str) -> str:
...
+            file_id, parent_id, file_basename = unpack_toil_uri(filename)
 
             # Decide where it should be put
-            file_dir = self._file_store.getLocalTempDir()
-            dest_path = os.path.join(file_dir, file_basename)
+            if self._is_uuid(parent_id):
+                # This is a URI with the "parent" UUID attached to the filename
+                # Use UUID as folder name rather than a new temp folder to reduce internal clutter
+                if not os.path.exists(parent_id):
+                    os.mkdir(parent_id)
```
adamnovak, line 495: It looks like we implicitly rely on `self._file_store.localTempDir` being the current directory here?

Probably we should construct the path to the directory we want to put the file in, make it, construct the file path from that, and return that.

src/toil/wdl/wdltoil.py
```diff
@@ -470,11 +485,21 @@ def _devirtualize_filename(self, filename: str) -> str:
...
+                # Put the UUID in the destination path in order for tasks to see where to put files depending on their parents
+                dest_path = os.path.join(self._file_store.localTempDir, parent_id, file_basename)
+            else:
+                # In case a file was not virtualized with a UUID
+                # But this shouldn't happen
```
adamnovak, line 500: Why not make the UUID mandatory in `pack_toil_uri`? If it should always be there, we probably want to actually enforce that. We could also move all the type conversion into the pack/unpack functions, so only real UUID objects go in and out.

src/toil/wdl/wdltoil.py
```diff
@@ -470,11 +485,21 @@ def _devirtualize_filename(self, filename: str) -> str:
...
             # Deserialize the FileID
-            file_id, file_basename = unpack_toil_uri(filename)
+            file_id, parent_id, file_basename = unpack_toil_uri(filename)
 
             # Decide where it should be put
-            file_dir = self._file_store.getLocalTempDir()
-            dest_path = os.path.join(file_dir, file_basename)
+            if self._is_uuid(parent_id):
+                # This is a URI with the "parent" UUID attached to the filename
+                # Use UUID as folder name rather than a new temp folder to reduce internal clutter
```
adamnovak, line 493: So the algorithm is to set up all the files from a particular previous job in a single host directory, under their basenames, and then mount them into the container so that everything in a given host directory stays together.

That definitely will mostly work, but it won't handle the case where multiple files with the same basename, but different paths, are uploaded by a job.

I can have a job make `dir1/index.txt` and `dir2/index.txt` as outputs, and use them both as inputs to another job. It looks like with the current implementation one of those is going to overwrite the other, since they have the same creator UUID and the same basename.

adamnovak (changes requested):
It looks like the implementation here doesn't account for anything about a file except for its basename and its creating job. In particular, the other parts of the file path other than just the file's own name are ignored. So e.g. `dir1/index.txt` and `dir2/index.txt` from the same job will not both be able to be used by a subsequent job.

We might need to add a test for that and then fix the implementation to pass it.

src/toil/wdl/wdltoil.py
```diff
@@ -1548,7 +1532,7 @@ def patched_run_invocation(*args: Any, **kwargs: Any) -> List[str]:
         # container-determined strings that are absolute paths to WDL File
         # objects, and like MiniWDL we can say we only support
         # working-directory-based relative paths for globs.
-        outputs_library = ToilWDLStdLibTaskOutputs(file_store, host_stdout_txt, host_stderr_txt, current_directory_override=workdir_in_container, stdlib_id=stdlib_id)
+        outputs_library = ToilWDLStdLibTaskOutputs(file_store, host_stdout_txt, host_stderr_txt, current_directory_override=workdir_in_container)
```
adamnovak, line 1535: @stxue1 If we don't pass the directory to ID mapping between the different standard library instantiations, is it possible to get two different standard library instances to upload two sibling files, but with different source directory UUIDs? Maybe by abusing the string-to-File logic?

adamnovak (approved):
I feel like it might still be possible to defeat this, but it's definitely better and should work correctly for most non-adversarial workflows.

src/toil/wdl/wdltoil.py
stxue1: I think string-to-file logic will probably not support sibling files since the standard library instances might be different. The directory to ID mappings would probably have to be passed around in order to support this.

PR #4611, opened 2023-10-03
take any nvidia-smi exception as not having gpu

glennhickey:
Toil runs `nvidia-smi` to count the available GPUs, and if the command fails, it catches it and returns 0.  This is fine, but only when it fails with an exception that Toil expects.  This came up once before in https://github.com/ComparativeGenomicsToolkit/cactus/issues/937 where an unhandled permission error wasn't caught, which completely blocked the user from running Cactus. This got fixed in #4391 by adding a check for `PermissionError`.  

But now somebody's run into what looks like the exact same issue again: https://github.com/ComparativeGenomicsToolkit/cactus/issues/1185.  This time it looks like they're getting a `NotADirectoryError` when running `nvidia-smi`.  They don't want to use GPUs but the failing exception completely breaks Cactus...

Anyway, this PR just makes the check return gpus=0 (and not crash) for *any* exception which, as a user, I think is more desirable. 

## Changelog Entry

 * More robust `nvidia-smi` check

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/lib/accelerators.py
+++ b/src/toil/lib/accelerators.py
@@ -100,13 +100,7 @@ def count_nvidia_gpus() -> int:
                 .firstChild,
             ).data
         )
-    except (
-        FileNotFoundError,
-        subprocess.CalledProcessError,
-        IndexError,
-        ValueError,
-        PermissionError,
-    ):
+    except: 
         return 0
 
     # TODO: Parse each gpu > product_name > text content and convert to some


adamnovak (approved):
We'll have to see what not catching misspelled variables does for us in terms of surprise problems. Hopefully it causes fewer than `nvidia-smi` finding new and innovative ways to not work.

adamnovak:
I've pulled this into `issues/4611-tolerate-broken-nvidia-smi` for testing.

adamnovak:
I'm going to merge this without waiting for a re-test because I don't see how it could fight with the config file work.

PR #4618, opened 2023-10-16
Fix configargparse in CWL

### LINKED ISSUE #4615: Fix configargparse parsing wrong in toil-cwl-runner
In toil-cwl-runner, configargparse sometimes doesn’t work due to a bug it has. We need a workaround for it.

If nargs=argparse.remainder on an argument, configargparse will not collect the right arguments, because the arguments from the config get tacked on the end of the argument list and can get eaten.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1407)
┆Issue Number: TOIL-1407

stxue1:
This resolves #4615.
## Changelog Entry

 * Fix config file support for CWL

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -3572,6 +3572,7 @@ def get_options(args: List[str]) -> argparse.Namespace:
     addOptions(parser, jobstore_as_flag=True, cwl=True)
     add_cwl_options(parser)
 
+    options: argparse.Namespace
     options, cwl_options = parser.parse_known_args(args)
     options.cwljob.extend(cwl_options)
 
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -32,8 +32,9 @@ import stat
 import sys
 import tempfile
 import textwrap
-import urllib
 import uuid
+from collections import OrderedDict
+from io import StringIO
 from threading import Thread
 from typing import (
     IO,
@@ -66,7 +67,8 @@ import cwltool.load_tool
 import cwltool.main
 import cwltool.resolver
 import schema_salad.ref_resolver
-from configargparse import ArgParser
+from configargparse import ArgParser, already_on_command_line, _COMMAND_LINE_SOURCE_KEY, _ENV_VAR_SOURCE_KEY, \
+    ConfigFileParserException, ACTION_TYPES_THAT_DONT_NEED_A_VALUE, _DEFAULTS_SOURCE_KEY, _CONFIG_FILE_SOURCE_KEY
 from cwltool.loghandler import _logger as cwllogger
 from cwltool.loghandler import defaultStreamHandler
 from cwltool.mpi import MpiConfig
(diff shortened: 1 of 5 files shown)


adamnovak:
The right way to do this is probably actually one config file with all the CWL, WDL, and base Toil options in it. We need to do some gymnastics to make the parser accept all of those options, but *only* those options, in the config file, and still reject options from the wrong runner on the command line.

We also need to handle the CWL or WDL extras not being installed, when parsing options. We want the config file to have the same contents and be parseable whether `cwltool` or `miniwdl` are installed or not. This might mean moving the option definitions to separate files.

src/toil/cwl/cwltoil.py
```diff
@@ -3574,8 +3557,73 @@ def main(args: Optional[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
...
+
+    cmd_line_args_actions = {y for x in cmd_line_args for y in x.split("=")}
+    # remove parser actions that are not specified on the command line
+    i = 0
+    while i < len(parser._actions):
+        action = parser._actions[i]
+        if set(action.option_strings).isdisjoint(cmd_line_args_actions) and len(action.option_strings) > 0:
+            del parser._actions[i]
+        else:
+            i += 1
```
adamnovak, line 3609: Is this really going to work right in all cases? Can't it be broken by specifying something like `--tesPassword "I=hate=the=--defaultDisk=option"`, which would then make the default disk from the config not apply?

src/toil/cwl/cwltoil.py
```diff
@@ -3574,8 +3557,73 @@ def main(args: Optional[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
...
+    # Parse only cmd line args
+    cmd_line_options = parser.parse_args(cmd_line_args)
+    # Merge namespaces, with command line taking precedence over config
```
adamnovak, line 3612:
```suggestion
    # Merge namespaces, with command line taking precedence over config.
    # Because we removed all the options not actually specified on the command line,
    # nothing will get clobbered with a default.
```

adamnovak (changes requested):
This looks like it mostly works, but I think the help output needs to be cleaned up a bit, and I am concerned that the pass to remove the options from the parser that aren't actually used is not perfectly correct.

Maybe there's a way to run the option parsing and find out what options get parsed by instrumenting something, and then you could use that list instead of a list made by manually parsing with a different implementation from the real one?

src/toil/cwl/cwltoil.py
```diff
@@ -3324,7 +3307,7 @@ def main(args: Optional[List[str]] = None, stdout: TextIO = sys.stdout) -> int:
         help="Do not delete Docker container used by jobs after they exit",
         dest="rm_container",
     )
-    extra_dockergroup = parser.add_argument_group()
+    extra_dockergroup = parser.add_argument_group("extra_dockergroup")
```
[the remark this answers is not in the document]
stxue1: I probably added this on accident, so I'll leave it as before

adamnovak (approved):

stxue1:
`cwlTest.py::test_usage_message` is failing because of this. It tests that improperly ordered arguments will return a certain error message, ex: `toil-cwl-runner job --logInfo`, but since argparse.REMAINDER was replaced, improperly ordered arguments will parse normally.

adamnovak:
@stxue1 We probably want to remove that test then, if we no longer intend that behavior.

DailyDreaming (approved):
Thanks for the fix @stxue1 !

PR #4682, opened 2023-11-14
Support CWL 1.2.1

### LINKED ISSUE #4570: Support CWL v1.2.1
Running the new tests shows 4 failures and 2 unsupported features: https://github.com/DataBiosphere/toil/pull/4565

We're also being solicited for feedback on the changes: https://github.com/DataBiosphere/toil/issues/4564

Will update this ticket once we've looked over the changes and discussed how to support the latest changes and features.

stxue1:
This resolves #4570 and should let the new CWL tests to pass on CI. It doesn't deal with 2 of the tests though as they rely on InplaceUpdateRequirement, and cwltoil [doesn't seem to support](https://github.com/DataBiosphere/toil/blob/dbba7bad6d05a35e83ab28a34279578dcd6e9c04/src/toil/cwl/cwltoil.py#L3134-L3139) that as of now.
This also has some changes to the file jobstore code. Before it wasn't quoting/unquoting URLs, which cwltoil seemed to have some issues with. 

Fixes https://github.com/DataBiosphere/toil/issues/3870 as well

## Changelog Entry

 * Support CWL 1.2.1
   * Encode/decode URLs in internal Toil functions

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/cwl/cwltoil.py
```diff
@@ -1033,7 +1043,9 @@ def __str__(self) -> str:
 
 class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
     """Subclass the cwltool command line tool to provide the custom ToilPathMapper."""
-
+    def __init__(self, toolpath_object: 'CommentedMap', loadingContext: 'cwltool.context.LoadingContext'):
```
mr-c, line 1046:
```suggestion
    def __init__(self, toolpath_object: CommentedMap, loadingContext: cwltool.context.LoadingContext):
```
These types aren't hidden behind a `if TYPE_CHECKING:` block, so you can use them directly.

src/toil/cwl/cwltoil.py
```diff
@@ -1033,7 +1043,9 @@ def __str__(self) -> str:
...
+        loadingContext.relax_path_checks = True
```
mr-c, line 1047: This is an interesting policy decision you've made.

In cwltool this defaults to `False` unless the `--relax-path-checks` option is passed.

This is to encourage the use of unsurprising path components as other tools might have issues with spaces, emojis, colons, etc.. 

For running the CWL Conformance tests with `cwltool` we do of course [set `--relax-path-checks`](https://github.com/common-workflow-language/cwl-v1.2/blob/0d538a0dbc5518f3c6083ce4571926f65cb84f76/.github/workflows/cwltool.yml#L18)

mr-c (changes requested):

src/toil/cwl/cwltoil.py
```diff
@@ -3285,12 +3297,46 @@ def generate_default_job_store(
...
+                if output.get("nameroot") is not None:
+                    del output["nameroot"]
+                if output.get("nameext") is not None:
+                    del output["nameext"]
+            # Recurse through nested objects
+            if output.get("listing") is not None:
+                for nested_object in output["listing"]:
+                    change_output(nested_object)
+    # Both the 'result' and 'log' output objects need to be changed
+    if output.get("result") is not None:
```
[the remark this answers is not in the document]
stxue1: I originally thought the issue the tests weren't passing was because something about the format of the output was incorrect. Looking at the compare function in cwltest though, I think the real reason why the tests was failing was because [the path it was grabbing](https://github.com/common-workflow-language/cwltest/blob/1987ca25d76975e41974bea3a34591051ece6379/cwltest/compare.py#L165C32-L165C32) was URL encoded and thus did not exist. I think the proper fix is to add the `path` field too when `location` is added during the file staging step, and I'll remove this function as it's excessive/unnecessary.

src/toil/cwl/cwltoil.py
stxue1: Yeah, I wasn't aware that `--relax-path-checks` is set for cwltool's conformance tests, I think that explains why toil-cwl-runner didn't pass the conformance tests, as I don't believe toil-cwl-runner currently applies that on CI. This can be removed then, and I can pass `--relax-path-checks` in Toil's CWL tests instead.

src/toil/cwl/cwltoil.py
stxue1: I removed this addition since it's no longer necessary when passing `--relax-path-checks` to the runner

mr-c:
> It doesn't deal with 2 of the tests though as they rely on InplaceUpdateRequirement, and cwltoil [doesn't seem to support](https://github.com/DataBiosphere/toil/blob/dbba7bad6d05a35e83ab28a34279578dcd6e9c04/src/toil/cwl/cwltoil.py#L3134-L3139) that as of now.

Well, it does with `--bypass-file-store ` when running in single machine mode; where are the CWL conformance tests being run?

mr-c (approved):

mr-c:
Thank you @stxue1 !

stxue1:
> > It doesn't deal with 2 of the tests though as they rely on InplaceUpdateRequirement, and cwltoil [doesn't seem to support](https://github.com/DataBiosphere/toil/blob/dbba7bad6d05a35e83ab28a34279578dcd6e9c04/src/toil/cwl/cwltoil.py#L3134-L3139) that as of now.
> 
> Well, it does with `--bypass-file-store ` when running in single machine mode; where are the CWL conformance tests being run?

It might be possible to add `--bypass-file-store` to the [runner arguments](https://github.com/DataBiosphere/toil/blob/ff5bacc79109731da8f874657c5bef84a746c715/src/toil/test/cwl/cwlTest.py#L133-L138). The only CWL test I'd be concerned about would be CWLonARM, which I know runs on kubernetes, but it looks like it only specifies the [leader node type](https://github.com/DataBiosphere/toil/blob/ff5bacc79109731da8f874657c5bef84a746c715/src/toil/test/cwl/cwlTest.py#L1009-L1015), so maybe it's effectively being ran as single machine there as well.

stxue1:
Seems like CI isn't too happy about `--bypass-file-store` being on by default, so I'll revert it. Although it looks like one of the conformance tests already [runs with `--bypass-file-store`](https://github.com/DataBiosphere/toil/blob/d7102135477650f3e9c0d55675ae59b3daecf4c7/src/toil/test/cwl/cwlTest.py#L912), so maybe just having that is enough?

src/toil/cwl/cwltoil.py
mr-c, line 1: please run `make format`

mr-c (changes requested):

mr-c:
> it looks like one of the conformance tests already [runs with `--bypass-file-store`](https://github.com/DataBiosphere/toil/blob/d7102135477650f3e9c0d55675ae59b3daecf4c7/src/toil/test/cwl/cwlTest.py#L912), so maybe just having that is enough?

Okay, looks like that gets run by the CI: https://ucsc-ci.com/databiosphere/toil/-/jobs/56461#L1019

mr-c (approved):
Huzzah!

PR #4758, opened 2024-01-19
Note --disableProgress's argument

adamnovak:
@juklucas noted that we don't say we need an argument here, but we do need one.

## Changelog Entry

 * `--disableProgress` option argument is now documented.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -505,7 +505,8 @@ systems have issues!).
                         to provision mesos for autoscaling.
   --statusWait INT      Seconds to wait between reports of running jobs.
                         (default=3600)
-  --disableProgress     Disables the progress bar shown when standard error is
+  --disableProgress BOOL
+			Disables the progress bar shown when standard error is
                         a terminal.
 
 **Debug Options**


docs/running/cliOptions.rst
```diff
@@ -505,7 +505,8 @@ systems have issues!).
                         to provision mesos for autoscaling.
   --statusWait INT      Seconds to wait between reports of running jobs.
                         (default=3600)
-  --disableProgress     Disables the progress bar shown when standard error is
+  --disableProgress BOOL
+			Disables the progress bar shown when standard error is
                         a terminal.
```
mr-c, line 510: Wouldn't it be better to not require the bool at all?

Change `type=convert_bool` to `action="store_true"` and drop `metavar="BOOL"` at

https://github.com/DataBiosphere/toil/blob/67a8b471dd77db33c25bf9db071def18c8d53f1f/src/toil/options/common.py#L695-L697

(matching [what is done for `--stats`](https://github.com/DataBiosphere/toil/blob/67a8b471dd77db33c25bf9db071def18c8d53f1f/src/toil/options/common.py#L308C71-L309))

mr-c (changes requested):

docs/running/cliOptions.rst
adamnovak: That's basically what we used to do? @stxue1 you had a reason for switching this to take an argument, right?

If we're already making incompatible changes we might be better off with `--progress=True`/`--progress=False` and keeping the old `--disableProgress` without argument as a thing that doesn't show up in the config.

stxue1: When I was adding metavars, the previous implementation looked like it took a true/false argument. Though it doesn't make much sense to take a bool with its argument name, so it should probably be `action="store_true"` for this one.
https://github.com/DataBiosphere/toil/blob/e80db2cd9d8f800befd94bb454581fd781130ba3/src/toil/common.py#L1537-L1543

adamnovak:
I'm closing this in favor of #4778.

PR #4811, opened 2024-02-29
Make leader wait for expected updates to be visible in the job store, or fail the job

### LINKED ISSUE #3841: The scheduler and file system can race
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1054)
┆Issue Number: TOIL-1054

adamnovak:
This fixes #3841 by adding an equivalent of Snakemake's `--latency-wait` option, here `--jobStoreTimeout`, which is the number of seconds we allow the job store to lag behind the scheduler. If this time elapses and we don't see an ostensibly successful job either remove itself or save an updated version of itself (with an incremented version clock), we will fail the job with a new `PARTITION` reason (since as far as we know there's a netsplit in the job store distributed system).

We should see this kind of failure in cases where the shared filesystem or other job store storage is lagging badly behind the scheduler (in which case you can increase `--jobStoreTimeout`), or in cases where the the job's writes are genuinely lost. We will also get warning messages for cases where a job already failed but didn't commit any updates back, which can happen when the scheduler can't even start it.

The messages look like:

```
Job foobarbaz v1 has no new version available after waiting 30 seconds. Either worker updates to the job store are delayed longer than your --jobStoreTimeout, or the worker trying to run the job vanished.
Marking ostensibly successful job foobarbaz v1 that did not report in to the job store before --jobStoreTimeout as having been partitioned from us.
```

Right now I've only plugged this in for finished jobs; there are two other points where we refresh jobs from the job store, but I'm not confident we actually expect the jobs to have always been modified, so I am leaving them alone.

## Changelog Entry

 * Toil will now wait `--jobStoreTimeout` seconds (default: 30) to see an update to/removal of a job that was run, and will not let the job succeed unless it is seen to make progress.
 * Toil job descriptions no longer have a `command` field, and track the link to the job body and the command to invoke the Toil worker separately.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


DailyDreaming:
@adamnovak Is this ready for review?

adamnovak:
@DailyDreaming Yes, it should be.

Do you want me to always request review from you specifically when something is ready? Usually what I do is mark it draft if it's definitely *not* ready, and only send a request if I get bored of waiting for you to come by and review it.

DailyDreaming:
> @DailyDreaming Yes, it should be.
> 
> Do you want me to always request review from you specifically when something is ready? Usually what I do is mark it draft if it's definitely _not_ ready, and only send a request if I get bored of waiting for you to come by and review it.

@adamnovak It would help me to be flagged for review.  I generally skip PRs without passing tests unless I'm flagged specifically, so that I know that despite some flaky test, you think that the PR is in a good state, and I'll look for that test(s), rerun or comment as necessary.

adamnovak:
@DailyDreaming This should be ready!

src/toil/job.py
```diff
@@ -779,14 +778,19 @@ def makeString(x: Union[str, bytes, None]) -> str:
...
+        # None if no body needs to run.
         #
-        # Gets replaced with/rewritten into the real, executable command when
-        # the leader passes the description off to the batch system to be
-        # executed.
-        self.command: Optional[str] = command
+        # Used to be the "command" until it was extended beyond making sense as
+        # an actual command, so now it's just a string with several
+        # space-separated fields. See attach_body() and get_body().
+        self._body_spec: Optional[str] = None
```
DailyDreaming, line 793: Suggestion: Change to `self._body` instead of `self._body_spec` (so as to be in line with `attach_body()` and `get_body()`).

Also, not necessary, but I feel like it might be clearer (and self documenting) if this could either be separate `self` variables like `self.filestore_id` and `self.command` OR a json like `{"filestore_id": 1234, "command": ["echo", "echo"]}`.  That way I wouldn't have to scroll to `attach_body()` and `get_body()` to intuit what the contents of `self._body_spec` might be.

src/toil/job.py
```diff
@@ -779,14 +778,19 @@ def makeString(x: Union[str, bytes, None]) -> str:
...
+        # JobDescription describes, and the module(s) needed to unpickle it.
+        # None if no body needs to run.
         #
-        # Gets replaced with/rewritten into the real, executable command when
-        # the leader passes the description off to the batch system to be
-        # executed.
-        self.command: Optional[str] = command
+        # Used to be the "command" until it was extended beyond making sense as
+        # an actual command, so now it's just a string with several
+        # space-separated fields. See attach_body() and get_body().
```
DailyDreaming, line 792: I'm not sure this comment is as useful as just describing the contents of `self._body_spec` as a space-separated string of filestore_id and command.  Would suggest changing these three lines to ` # a space-separated string of filestore_id and command` .

src/toil/job.py
```diff
@@ -714,7 +714,7 @@ class JobDescription(Requirer):
     """
     Stores all the information that the Toil Leader ever needs to know about a Job.
 
-    (requirements information, dependency information, commands to issue,
+    (requirements information, dependency information, pickled job body to run,
```
DailyDreaming, line 717: I know that this wasn't your comment, but "requirements" and "dependencies" are kind of synonyms.

Could the contents of "pickled job body" be clearly described here?

src/toil/leader.py
```diff
@@ -1092,7 +1096,7 @@ def reissueOverLongJobs(self) -> None:
                                 str(runningJobs[jobBatchSystemID]),
                                 str(maxJobDuration))
                     jobsToKill.append(jobBatchSystemID)
-            reissued = self.killJobs(jobsToKill)
+            reissued = self.killJobs(jobsToKill, exit_reson=BatchJobExitReason.MAXJOBDURATION)
```
DailyDreaming, line 1099: `exit_reson`

DailyDreaming (commented):
Thanks for doing this.

I think that there's only one misspelled variable, otherwise my only comments are minor style choices and this looks good to me.

src/toil/job.py
```diff
@@ -954,7 +958,71 @@ def services(self):
...
+            raise RuntimeError(f"An invalid body spec {self._body_spec} was found in {self}.")
+        return parts[1], ModuleDescriptor.fromCommand(parts[2:])
+
+    def set_worker_command(self, worker_command: str) -> None:
+        """
+        Used by the leader to store the command to run this job.
+        """
+        # TODO: This probably should become a new issueBatchJob argument
+        # instead.
+        self._worker_command = worker_command
```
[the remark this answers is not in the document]
adamnovak: I wanted the check on the getter, so I didn't want to expose the value itself without an `_`, so then I needed a setter.

I could make this a `@property`, but we'd still have the setter.

Or I could do the refactor to stop shipping the worker command through the job altogether and change the `issueBatchJob` interface, and all its implementations.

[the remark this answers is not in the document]
DailyDreaming: @adamnovak It's not a big deal then and I don't think it's worth a big refactor.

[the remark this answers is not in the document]
adamnovak: I tried just moving the parameter to `issueBatchJob`, since I already had a TODO that said it ought to be there.

adamnovak:
@DailyDreaming It's not letting me dismiss your review to say I did the things you wanted. Maybe because you disn't mark it as needing changes? So to merge this I have to override the branch protection rules, which doesn't seem right.

Or am I really supposed to be getting *you* to say I did the things you wanted?

PR #4819, opened 2024-03-07
Measure CPU and memory usage in WDL Docker containers

adamnovak:
This uses code like [the MiniWDL plugin](https://github.com/chanzuckerberg/miniwdl/blob/6dfe83781f74a8e248727eb61c31e1c7562bd26e/examples/plugin_log_task_usage/miniwdl_log_task_usage.py) to measure CPU and memory usage from within the MiniWDL-managed Docker container by injecting a prefix to the user Bash command script.

It sends the output via a hidden directory in the working directory, which hopefully is allowed by the spec. Stats are recorded once per second, and the log grows forever as long as the command runs.

You can try it with:
```
cat >workflow.wdl <<'EOF'
version 1.1
import "https://raw.githubusercontent.com/chanzuckerberg/miniwdl/6dfe83781f74a8e248727eb61c31e1c7562bd26e/examples/plugin_log_task_usage/StressTest.wdl"
workflow hello {
  input {
  }

  call StressTest.StressTest {
  }

  output {
  }
}
EOF

toil-wdl-runner workflow.wdl --logDebug --jobStore ./store --stats

toil stats ./store
```

## Changelog Entry

 * Toil can now measure CPU and memory in WDL containers run with Docker.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/lib/resources.py
```diff
@@ -18,6 +18,28 @@
...
+_extra_memory_ki: int = 0
+
+def record_extra_memory(peak_ki: int) -> None:
+    """
+    Become responsible for the given peak memory usage, in kibibytes.
+
+    The memory will be treated as if it was used by a child process at the time
+    our real child processes were also using their peak memory.
+    """
+    global _extra_memory_ki
```
DailyDreaming, line 32: Is there a way to do this without globals?  We've managed to avoid them so far in Toil.

adamnovak: I don't think there's a way to avoid module-level state and still have these functions affect the operation of the other functions, with the functions where they are.

I could hide the state as an attribute of the function objects, or in a list that I can write to elements of, and avoid the `global` keyword.

Or I could introduce a class with some class-level attributes and put the state or the state and all the functions in there. Maybe making everybody static methods on a `ResourceMonitor` class would be better, even though I'd have to fix all the existing usages.

DailyDreaming (approved):
LGTM.

Issue #4807, opened 2024-02-23
Implement FUSE on unprivileged clusters that have FUSE installed already


stxue1:
Ex: If there is FUSE implemented systemwide on Phoenix. FUSE in unprivileged mode may lack permissions to talk to the kernel directly, and have behavioral differences compared to running in privileged mode.

stxue1:
Latest versions of Singularity always attempt to mount with FUSE before sandboxing. Maybe updating Singularity in the Dockerfile is all that is needed. One downside is that old Docker images won't be supported anymore, but that's likely considered more of a user error.

stxue1:
For future reference:
As of now, for Kubernetes, [I don't think this is possible.](https://github.com/kubernetes/kubernetes/issues/7890)
The only two options I have seen are to use a third party plugin (which runs a privileged pod adjacently) or to simply run the pod as privileged.

Plugins like [smarter-device-manager](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-766088805), [fuse-device-plugin](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-1441213446), or [meta-fuse-csi-plugin](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-1822140474) look promising, but some of these appear to be [storage drivers](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-1324909377). 

For toil, this is not what we want; I believe we essentially want an unprivileged pod to have freeform [access to fuse to mount inside it's own pod space](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-1943097424).

The [Docker workaround](https://github.com/docker/for-linux/issues/321#issuecomment-677744121) cannot apply in Kubernetes as Kubernetes does not have a `--device` argument, though there could be more reasons why it won't work as well.

It does look like the Docker workaround works well enough, so as long as Toil detects it is inside an unprivileged docker container and has access to `/dev/fuse`, it will be able to mount with FUSE.

stxue1:
There's currently some code in [issues/4807-unprivileged-fuse](https://github.com/DataBiosphere/toil/tree/issues/4807-unprivileged-fuse) that gives toil-wdl-runner the ability to use FUSE in an unprivileged context if it's available. The way it detects that FUSE can be used and is unprivileged is probably not ideal though (and the branch doesn't set the envvar TOIL_INSTALLED_INSIDE_DOCKER in dockerfile.py), but it should work in theory.
[...]

PR #4824, opened 2024-03-09
Enable FUSE for privileged Toil clusters

### LINKED ISSUE #4806: Implement FUSE on Toil for privileged environments
Ex: on AWS where we have root permissions, we can install FUSE on the cluster.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1506)
┆Issue Number: TOIL-1506

stxue1:
This should close #4806
This is also related to #4807

The Dockerfile pulls from Github releases for now, but I will change it to a Debian mirror once the package managers get around to updating Singularity.

I think the namespace rules might not be needed now from my basic testing. My guess is the new version of Singularity now is able to handle it, but I'll test it more thoroughly later.

## Changelog Entry

 * Set Toil-managed clusters to run in privileged mode by default
   * Allow Toil-managed Kubernetes clusters support FUSE
   * Added option in the Kubernetes batchsystem to ask for privileged pods
   * Added option in `launch-cluster` to enable FUSE support
 * Reduced configuration debug spam

## Reviewer Checklist

        ```
        ```

## Merger Checklist


stxue1:
I think this is just blocked by the debian mirrors being outdated. I can get the .deb file from github, but the debian file is for amd64, and CI wants to build arm
```commandline
$ dpkg -i ./singularity-ce_4.1.2-jammy_amd64.deb 
dpkg: error processing archive ./singularity-ce_4.1.2-jammy_amd64.deb (--install):
 package architecture (amd64) does not match system (arm64)
Errors were encountered while processing:
 ./singularity-ce_4.1.2-jammy_amd64.deb
```

Docker was truncating the log output. With apt install, the full log is:
```commandline
apt install ./singularity-ce_4.1.2-jammy_amd64.deb 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'singularity-ce:amd64' instead of './singularity-ce_4.1.2-jammy_amd64.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 singularity-ce:amd64 : Depends: libc6:amd64 (>= 2.34) but it is not installable
                        Depends: libfuse2:amd64 (>= 2.6) but it is not installable
                        Depends: libglib2.0-0:amd64 (>= 2.35.8) but it is not installable
                        Depends: libseccomp2:amd64 (>= 2.5.0) but it is not installable
                        Depends: zlib1g:amd64 (>= 1:1.1.4) but it is not installable
                        Depends: crun:amd64 (>= 1.5.0) but it is not installable or
                                 runc:amd64 but it is not installable
                        Depends: fuse:amd64 but it is not installable
                        Depends: fuse2fs:amd64 but it is not installable
                        Depends: squashfs-tools:amd64 but it is not installable
                        Recommends: squashfs-tools-ng:amd64 but it is not installable
```

Makefile
```diff
@@ -169,7 +169,7 @@ define tag_docker
 	@printf "$(green)Tagged appliance image $1 as $2.$(normal)\n"
 endef
 
-docker: toil_docker prometheus_docker grafana_docker mtail_docker
+docker: toil_docker
```
adamnovak, line 172: I think we still need these.

```suggestion
docker: toil_docker prometheus_docker grafana_docker mtail_docker
```

docker/Dockerfile.py
```diff
@@ -24,15 +24,17 @@
...
+                   'python3.12': ['python3.12-distutils', 'python3.12-venv']}
+
+dependencies = ' '.join(python_packages[python] +
+                        ['libffi-dev',  # For client side encryption for extras with PyNACL
                          python,
                          f'{python}-dev',
-                         'python3.8-distutils' if python == 'python3.8' else '',
-                         'python3.9-distutils' if python == 'python3.9' else '',
-                         'python3.10-distutils' if python == 'python3.10' else '',
-                         'python3.11-distutils' if python == 'python3.11' else '',
```
adamnovak, line 34: Should this just be `f'{python}-dev', f'{python}-distutils', f'{python}-venv'`? I think we used to have a python version that we needed that *didn't* have with the numbered packages available, but it seems like that's no longer the case.

docker/Dockerfile.py
```diff
@@ -141,6 +145,15 @@ def heredoc(s):
...
     # So we need to make sure to install a downgraded squashfs first.
+    #
+    # TODO: Singularity has since resolved this on their end
+    # https://github.com/sylabs/singularity/pull/267
+    # This works by checking that the UID of the caller is not root
+    # In a Kubernetes pod, the default setup will have UID 0 even if the pod is unprivileged
+    # https://github.com/sylabs/singularity/issues/2727
+    # It is possible to avoid this by changing the user of the pod (such as runAsUser: 1000)
+    # but this may cause permission issues, ex: changed read/write permissions
+    # so the downgraded squashfs stays, but options for updated squashfs are possible
```
adamnovak, line 156: What options do you mean?

docs/contributing/contributing.rst
```diff
@@ -310,6 +311,31 @@ with ``worker`` to get shell access in your worker.
     if Docker can't find the file/directory on the host it will silently fail and mount
     in an empty directory.
 
+
+.. admonition:: Enabling FUSE
+
+    When running toil-wdl-runner with Singularity, Singularity will sandbox images
```
adamnovak, line 317: This sounds like we mean it will restrict what the images are allowed to do, but we really mean that it will decompress the image files that store the container image filesystems and turn them into built-out directories full of files, which it calls "sandbox directories".

```suggestion
    When running toil-wdl-runner with Singularity, Singularity will decompress images to sandbox directories
```

docs/contributing/contributing.rst
```diff
@@ -310,6 +311,31 @@ with ``worker`` to get shell access in your worker.
...
+    by default. This can take time if a workflow has lots of images. To avoid this, FUSE can be given
+    to the Docker container at startup. There are 2 main ways to do this. Either run all the Docker
```
adamnovak, line 319: FUSE is a whole project/kernel feature; I think it sounds odd to say that we give FUSE itself to the container.
```suggestion
    by default. This can take time if a workflow has lots of images. To avoid this, access to FUSE can be given
    to the Docker container at startup. There are 2 main ways to do this. Either run all the Docker
```

docs/running/cloud/clusterUtils.rst
```diff
@@ -125,6 +125,11 @@ exist yet, Toil will create it for you.
                         to override the default value from ``--nodeStorage`` for the
                         specified nodeType(s). This is useful for heterogeneous jobs
                         where some tasks require much more disk than others.
+  --allowFuse BOOL
+                        Whether to allow FUSE mounts for faster runtimes with Singularity.
+                        Note: This will result in the Toil container running as privileged.
+                        For Kubernetes, pods will be asked to run as privileged. If this is not
+                        allowed, images will be sandboxed instead.
```
adamnovak, line 132:
```suggestion
                        allowed, Singularity containers will use sandbox directories instead.
```

docs/running/cloud/kubernetes.rst
```diff
@@ -188,6 +188,8 @@ To configure your workflow to run on Kubernetes, you will have to configure seve
 
 #. ``TOIL_KUBERNETES_OWNER`` **should** be set to the username of the user running the Toil workflow. The jobs that Toil creates will include this username, so they can be more easily recognized, and cleaned up by the user if anything happens to the Toil leader. In this example we are using ``demo-user``.
 
+#. ``TOIL_KUBERNETES_PRIVILEGED`` **can** be set to True or False. When true, this allows pods to run as privileged, enabling FUSE mounts for Singularity for faster runtimes. If this is not set to true, Singularity will sandbox images. This is unset/False by default except in Toil-managed clusters.
```
adamnovak, line 191:
```suggestion
#. ``TOIL_KUBERNETES_PRIVILEGED`` **can** be set to True or False. When true, this allows pods to run as privileged, enabling FUSE mounts for Singularity for faster runtimes. If this is not set to true, Singularity will extract images to sandbox directories. This is unset/False by default except in Toil-managed clusters.
```

src/toil/batchSystems/kubernetes.py
```diff
@@ -881,6 +883,11 @@ def mount_host_path(volume_name: str, host_path: str, mount_path: str, create: b
                                                   name="runner-container",
                                                   resources=resources,
                                                   volume_mounts=mounts)
+
+        # In case security context rules are not allowed to be set
+        if self.config.kubernetes_privileged:
```
adamnovak, line 888: This comment is kind of backwards from the if. We're going to do the thing the if controls if the user *does* want to do this thing, not if the thing isn't allowed to be done.
```suggestion
        # In case security context rules are not allowed to be set, we only apply
        # a security context at all if we need to turn on privileged mode.
        if self.config.kubernetes_privileged:
```

src/toil/common.py
```diff
@@ -276,7 +277,7 @@ def prepare_restart(self) -> None:
         # exist and that can't safely be re-made.
         self.write_messages = None
 
-    def setOptions(self, options: Namespace) -> None:
+    def setOptions(self, options: Namespace, debug_mute: bool = False) -> None:
```
adamnovak, line 280: I might call this `quiet`. Also, we should explain what it does if you set it, since it's kind of non-obvious why you would ever need this.

adamnovak: Alternately, instead of having logging as an externally-exposed part of this function's contract, we could just not log inside the function, and find where we're calling it where we actually want logging, and add logging *outside* the function in those places.

Otherwise the docstring needs to be something weird like `Creates a config object from the options object and also optionally logs all the options.` and our function no longer has one clearly defined job.

src/toil/options/__init__.py
```diff
@@ -0,0 +1,13 @@
+from typing import Optional
+
+from toil.lib.conversions import strtobool
+
+
+def convert_bool(b: str) -> bool:
+    """Convert a string representation of bool to bool"""
+    return bool(strtobool(b))
```
adamnovak, line 8: This function just does exactly what `strtobool` does. We probably should get rid of it and just use `strtobool`.

src/toil/options/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def opt_strtobool(b: Optional[str]) -> Optional[bool]:
+    """Convert an optional string representation of bool to None or bool"""
+    return b if b is None else convert_bool(b)
```
adamnovak, line 13: Maybe this should just live in `toil.lib.conversions` next to the non-`opt_` version `strtobool`? Than we could avoid adding this file.

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -167,6 +168,9 @@ def __init__(
         # Kubernetes joining information as a dict for Kubernetes clusters.
         self._leaderWorkerAuthentication = None
 
+        # Whether or not to use FUSE on the cluster. If true, the cluster will be launched in privileged mode
```
adamnovak, line 171:
```suggestion
        # Whether or not to use FUSE on the cluster. If true, the cluster's Toil containers will be launched in privileged mode
```

src/toil/provisioners/abstractProvisioner.py
```diff
@@ -813,13 +817,11 @@ def add_toil_service(self, config: InstanceConfiguration, role: str, keyPath: st
                 -v /etc/kubernetes:/etc/kubernetes \\
                 -v /etc/kubernetes/admin.conf:/root/.kube/config \\
                 # Pass in a path to use for singularity image caching into the container
+                {"-e TOIL_KUBERNETES_PRIVILEGED=True --privileged" if self.enable_fuse else 
+                "--security-opt seccomp=unconfined --security-opt systempaths=unconfined"} \\
                 -e TOIL_KUBERNETES_HOST_PATH=/var/lib/toil \\
                 -e SINGULARITY_CACHEDIR=/var/lib/toil/singularity \\
```
adamnovak, line 823: The comment for `SINGULARITY_CACHEDIR` has gotten separated from `SINGULARITY_CACHEDIR` here and should be reattached, instead of adding more lines between them.

adamnovak (changes requested):
This looks pretty good; I found a few pieces that could be cleaned up.

docker/Dockerfile.py
stxue1: I separated the python packages into 2 as `python3.13-distutils` doesn't seem to exist in apt, and I'm not sure if it will be added as it was supposed to be [removed in Python 3.12](https://peps.python.org/pep-0632/).

docker/Dockerfile.py
stxue1: I did try to see if Singularity would work with an updated squashfs-tools package by running under a user that isn't UID 0. Singularity did run, but the user won't have sufficient permissions for things such as writing files. I haven't followed through yet as I'm not too sure about the best way to give the user permissions, since existing normal Toil workflows could be writing to root owned directories currently. I could probably simply chmod certain directories that I think are necessary (such as `/` and other directories that aren't kernel/OS related).

adamnovak (approved):

stxue1:
I left code in https://github.com/DataBiosphere/toil/tree/issues/4807-unprivileged-fuse for a possible implementation on how to get unprivileged FUSE with singularity to work

PR #4828, opened 2024-03-09
Add colored logging

### LINKED ISSUE #4827: Add colored logs to Toil
Toil's logging can be very verbose and large sometimes. Colored logs might help with readability. I took a short glance at how MiniWDL does colored logs and it looks fairly easy to add colors. MiniWDL does make its text bold though, and I'm not sure if Toil should as well.

On a related note, I also believe configargparse is fully compatible with argcomplete, so it's possible to add that too to make using Toil on CLI easier.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1516)
┆Issue Number: TOIL-1516

stxue1:
This will close #4827 

## Changelog Entry

 * Add colored logging to Toil
   * Added a new option `--no-color` to disable colored logging

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/statsAndLogging.py
+++ b/src/toil/statsAndLogging.py
@@ -18,7 +18,6 @@ import logging
 import os
 import time
 from argparse import ArgumentParser, Namespace
-from datetime import datetime
 from logging.handlers import RotatingFileHandler
 from threading import Event, Thread
 from typing import IO, TYPE_CHECKING, Any, Callable, List, Optional, Union
--- a/src/toil/statsAndLogging.py
+++ b/src/toil/statsAndLogging.py
@@ -249,7 +249,7 @@ def install_log_color(set_logger: Optional[logging.Logger] = None) -> None:
     """Make logs colored."""
     # Most of this code is taken from miniwdl
     # delayed import
-    import coloredlogs  # type ignore[import-untyped]
+    import coloredlogs  # type: ignore[import-untyped]
 
     level_styles = dict(coloredlogs.DEFAULT_LEVEL_STYLES)
     level_styles["debug"]["color"] = 242
--- a/src/toil/statsAndLogging.py
+++ b/src/toil/statsAndLogging.py
@@ -18,10 +18,12 @@ import logging
 import os
 import time
 from argparse import ArgumentParser, Namespace
+from datetime import datetime
 from logging.handlers import RotatingFileHandler
 from threading import Event, Thread
 from typing import IO, TYPE_CHECKING, Any, Callable, List, Optional, Union
 
+from toil.lib.conversions import strtobool
 from toil.lib.expando import Expando
 from toil.lib.resources import get_total_cpu_time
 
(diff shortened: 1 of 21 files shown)


src/toil/statsAndLogging.py
```diff
@@ -231,17 +232,45 @@ def shutdown(self) -> None:
...
+    level_styles = dict(coloredlogs.DEFAULT_LEVEL_STYLES)
+    level_styles["debug"]["color"] = 242
+    level_styles["notice"] = {"color": "green", "bold": True}
+    level_styles["error"]["bold"] = True
+    level_styles["warning"]["bold"] = True
+    level_styles["info"] = {}
+    field_styles = dict(coloredlogs.DEFAULT_FIELD_STYLES)
+    field_styles["asctime"] = {"color": "blue"}
+    field_styles["name"] = {"color": "magenta"}
+    fmt = "%(asctime)s.%(msecs)03d %(name)s %(message)s"  # colors obviate levelname
```
adamnovak, line 263: I don't think we should rely on the colors to convey level information:

1. Not everyone can actually see the difference between any given pair of colors.
2. You won't always necessarily be able to see the colors that were originally used. Maybe someone has copy-pasted the logs, or maybe you are looking at the logs on CI in the raw log view.
3. You might want to search the logs for error or warning messages, but you can't really search by color.

src/toil/statsAndLogging.py
```diff
@@ -262,6 +291,8 @@ def add_logging_options(parser: ArgumentParser) -> None:
     group.add_argument("--logFile", dest="logFile", help="File to log in.")
     group.add_argument("--rotatingLogging", dest="logRotating", action="store_true", default=False,
                        help="Turn on rotating logging, which prevents log files from getting too big.")
+    group.add_argument("--no-color", action="store_false", dest="colored_logs", default=True,
+                       help="Enable colored logging")
```
adamnovak, line 295: The help here seems wrong; `--no-color` would *disable* colored logging.

I would maybe do another `=True/=False` option like all the other new flags we've been adding?

Also, `--no-color` is kebab case while the other Toil option names are usually camelCase. We should probably use camelCase if it is multiple words.

src/toil/statsAndLogging.py
```diff
@@ -231,17 +232,45 @@ def shutdown(self) -> None:
...
+        install_log_color(set_logger)
 
     # Suppress any random loggers introduced by libraries we use.
     # Especially boto/boto3.  They print too much.  -__-
     suppress_exotic_logging(__name__)
 
 
+def install_log_color(set_logger: Optional[logging.Logger] = None) -> None:
+    """Make logs colored."""
+    # Most of this code is taken from miniwdl
```
adamnovak, line 250: Is it taken from MiniWDL in compliance with the terms of MiniWDL's license? Do we need to add anything to Toil's `LICENSE` file to go with it?

src/toil/statsAndLogging.py
```diff
@@ -231,17 +232,45 @@ def shutdown(self) -> None:
         # in addition to cleaning on exceptions, onError should clean if there are any failed jobs
 
 
-def set_log_level(level: str, set_logger: Optional[logging.Logger] = None) -> None:
+def set_log_level(level: str, set_logger: Optional[logging.Logger] = None, colored_logs: bool = False) -> None:
```
adamnovak, line 235: Maybe turning the colors on and off should have its own method? It's not really part of setting the log level to also set the colors. Or maybe this method's name could change?

adamnovak (changes requested):
It would be cool to have colored logs!

I'm worried about *only* conveying some information by color when color is on, though; just because color was on when the message was logged doesn't mean whoever is looking at the logs now can actually *see* those colors.

Also, I'm worried about what this might do to the raw logs view on Gitlab, or other times when you do something like log from Toil to a file and then open and look at the file. Are we going to get a bunch of escape characters cluttering up the text? Or is the backing module smart enough to know when it is looking at a real terminal on stderr versus CI's output capture or a file?

Other comments are inline.

stxue1:
I managed to mimic the old logging format exactly; the actual text should remain unaffected except for having colors.
image

MiniWDL is under the MIT license which I think is mostly compatible with Apache, but I included it in the LICENSE file just in case.

PR #4859, opened 2024-04-03
Replace use of boto with boto3 for `awsProvisioner.py`

### LINKED ISSUE #4718: Usage of boto (not boto3) prevents toil from working with Python3.12
Hi,
Debian is currently transitioning to Python3.12.  Inside `setup.py` of toil  I've found the usage of `imp` and the comment that it is known that this modules is deprecated but usage of its successor ` importlib` would not be possible.  To work around this I wrote a [small patch](https://salsa.debian.org/med-team/toil/-/blob/master/debian/patches/python3.12.patch) where I'm simply using a shell command to obtain the version.  I'd recommend to replace `imp` somehow (more?) sensibly to follow Python3 progress sooner or later.
The more concerning issue is the usage of boto.  In [issue #3951 of boto](https://github.com/boto/boto/issues/3951) authors wrote:

```
This repo, `boto` was deprecated in 2020 and the last supported Python 3 version was 3.4. It's unlikely this package will ever work with Python 3.12.
 
The current AWS SDK for Python, boto3, can be found [here](https://github.com/boto/boto3) and does support Python 3.12.
```

Thus I strongly recommend to switch to `boto3` before your next release.

Kind regards, Andreas.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1468)
┆Issue Number: TOIL-1468

### LINKED ISSUE #4860: Replace use of boto with boto3 for `awsProvisioner.py`
Related to #4718.

This issue is to only replace `boto` in `awsProvisioner.py`, and so Github will allow me to unlink the [other issue](https://github.com/DataBiosphere/toil/issues/4718).

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1536)
┆Issue Number: TOIL-1536

stxue1:
This will close #4860 
This is related but does not close issue 4718. This only deals with the use of boto for `awsProvisioner.py` and certain test files.
I'm splitting the boto to boto3 conversion into separate PRs to avoid one giant PR.
## Changelog Entry

 * Replace use of boto with boto3 for `awsProvisioner.py`
 * Add type hinting to `awsProvisioner.py`

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/lib/aws/__init__.py
+++ b/src/toil/lib/aws/__init__.py
@@ -23,6 +23,8 @@ from urllib.request import urlopen
 
 from mypy_boto3_s3.literals import BucketLocationConstraintType
 
+AWSRegionName = Union[BucketLocationConstraintType, Literal["us-east-1"]]
+
 logger = logging.getLogger(__name__)
 
 # This file isn't allowed to import anything that depends on Boto or Boto3,
@@ -129,7 +131,7 @@ def get_current_aws_zone() -> Optional[str]:
         get_aws_zone_from_environment_region() or \
         get_aws_zone_from_boto()
 
-def zone_to_region(zone: str) -> Union["BucketLocationConstraintType", Literal["us-east-1"]]:
+def zone_to_region(zone: str) -> AWSRegionName:
     """Get a region (e.g. us-west-2) from a zone (e.g. us-west-1c)."""
     # re.compile() caches the regex internally so we don't have to
     availability_zone = re.compile(r'^([a-z]{2}-[a-z]+-[1-9][0-9]*)([a-z])$')
--- a/src/toil/lib/aws/__init__.py
+++ b/src/toil/lib/aws/__init__.py
@@ -16,19 +16,19 @@ import logging
 import os
 import re
 import socket
+import toil.lib.retry
 from http.client import HTTPException
 from typing import Dict, MutableMapping, Optional, Union, Literal
 from urllib.error import URLError
 from urllib.request import urlopen
 
-from toil.lib.retry import ErrorCondition
 from botocore.exceptions import ClientError
 
 from mypy_boto3_s3.literals import BucketLocationConstraintType
 
 AWSRegionName = Union[BucketLocationConstraintType, Literal["us-east-1"]]
 
-AWSServerErrors = ErrorCondition(
+AWSServerErrors = toil.lib.retry.ErrorCondition(
     error=ClientError,
     error_codes=[404, 500, 502, 503, 504]
 )
--- a/src/toil/lib/aws/__init__.py
+++ b/src/toil/lib/aws/__init__.py
@@ -21,9 +21,17 @@ from typing import Dict, MutableMapping, Optional, Union, Literal
 from urllib.error import URLError
 from urllib.request import urlopen
 
+from toil.lib.retry import ErrorCondition
+from botocore.exceptions import ClientError
+
 from mypy_boto3_s3.literals import BucketLocationConstraintType
 
 AWSRegionName = Union[BucketLocationConstraintType, Literal["us-east-1"]]
+
+AWSServerErrors = ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
+)
 
 logger = logging.getLogger(__name__)
 
--- a/src/toil/lib/aws/__init__.py
+++ b/src/toil/lib/aws/__init__.py
@@ -28,6 +28,8 @@ from mypy_boto3_s3.literals import BucketLocationConstraintType
 
 AWSRegionName = Union[BucketLocationConstraintType, Literal["us-east-1"]]
 
+# These are errors where we think something randomly
+# went wrong on the AWS side and we ought to retry.
 AWSServerErrors = toil.lib.retry.ErrorCondition(
     error=ClientError,
     error_codes=[404, 500, 502, 503, 504]
--- a/src/toil/lib/aws/__init__.py
+++ b/src/toil/lib/aws/__init__.py
@@ -17,9 +17,11 @@ import os
 import re
 import socket
 from http.client import HTTPException
-from typing import Dict, MutableMapping, Optional
+from typing import Dict, MutableMapping, Optional, Union, Literal
 from urllib.error import URLError
 from urllib.request import urlopen
+
+from mypy_boto3_s3.literals import BucketLocationConstraintType
 
 logger = logging.getLogger(__name__)
 
@@ -67,11 +69,10 @@ def get_aws_zone_from_metadata() -> Optional[str]:
         # metadata.
         try:
             # Use the EC2 metadata service
-            import boto
-            str(boto)  # to prevent removal of the import
-            from boto.utils import get_instance_metadata
+            from ec2_metadata import ec2_metadata
+
             logger.debug("Fetch AZ from EC2 metadata")
-            return get_instance_metadata()['placement']['availability-zone']
+            return ec2_metadata.availability_zone
         except ImportError:
             # This is expected to happen a lot
             logger.debug("No boto to fetch ECS metadata")
@@ -128,7 +129,7 @@ def get_current_aws_zone() -> Optional[str]:
         get_aws_zone_from_environment_region() or \
         get_aws_zone_from_boto()
 
-def zone_to_region(zone: str) -> str:
+def zone_to_region(zone: str) -> Union["BucketLocationConstraintType", Literal["us-east-1"]]:
     """Get a region (e.g. us-west-2) from a zone (e.g. us-west-1c)."""
     # re.compile() caches the regex internally so we don't have to
     availability_zone = re.compile(r'^([a-z]{2}-[a-z]+-[1-9][0-9]*)([a-z])$')
--- a/src/toil/lib/aws/utils.py
+++ b/src/toil/lib/aws/utils.py
@@ -36,7 +36,7 @@ from toil.lib.retry import (DEFAULT_DELAYS,
                             get_error_code,
                             get_error_status,
                             old_retry,
-                            retry)
+                            retry, ErrorCondition)
 
 if sys.version_info >= (3, 8):
     from typing import Literal
@@ -77,11 +77,14 @@ THROTTLED_ERROR_CODES = [
         'EC2ThrottledException',
 ]
 
-@retry(errors=[BotoServerError])
+@retry(errors=[ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
+)])
 def delete_iam_role(
     role_name: str, region: Optional[str] = None, quiet: bool = True
 ) -> None:
-    from boto.iam.connection import IAMConnection
+    # from boto.iam.connection import IAMConnection
 
     # TODO: the Boto3 type hints are a bit oversealous here; they want hundreds
     # of overloads of the client-getting methods to exist based on the literal
@@ -92,9 +95,8 @@ def delete_iam_role(
     # we wanted MyPy to be able to understand us. So at some point we should
     # consider revising our API here to be less annoying to explain to the type
     # checker.
-    iam_client = cast(IAMClient, session.client('iam', region_name=region))
-    iam_resource = cast(IAMServiceResource, session.resource('iam', region_name=region))
-    boto_iam_connection = IAMConnection()
+    iam_client = session.client('iam', region_name=region)
+    iam_resource = session.resource('iam', region_name=region)
     role = iam_resource.Role(role_name)
     # normal policies
     for attached_policy in role.attached_policies.all():
@@ -104,16 +106,19 @@ def delete_iam_role(
     for inline_policy in role.policies.all():
         printq(f'Deleting inline policy: {inline_policy.policy_name} from role {role.name}', quiet)
         # couldn't find an easy way to remove inline policies with boto3; use boto
-        boto_iam_connection.delete_role_policy(role.name, inline_policy.policy_name)
+        iam_client.delete_role_policy(RoleName=role.name, PolicyName=inline_policy.policy_name)
     iam_client.delete_role(RoleName=role_name)
     printq(f'Role {role_name} successfully deleted.', quiet)
 
 
-@retry(errors=[BotoServerError])
+@retry(errors=[ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
+)])
 def delete_iam_instance_profile(
     instance_profile_name: str, region: Optional[str] = None, quiet: bool = True
 ) -> None:
-    iam_resource = cast(IAMServiceResource, session.resource("iam", region_name=region))
+    iam_resource = session.resource("iam", region_name=region)
     instance_profile = iam_resource.InstanceProfile(instance_profile_name)
     if instance_profile.roles is not None:
         for role in instance_profile.roles:
@@ -123,11 +128,14 @@ def delete_iam_instance_profile(
     printq(f'Instance profile "{instance_profile_name}" successfully deleted.', quiet)
 
 
-@retry(errors=[BotoServerError])
+@retry(errors=[ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
+)])
 def delete_sdb_domain(
     sdb_domain_name: str, region: Optional[str] = None, quiet: bool = True
 ) -> None:
-    sdb_client = cast(SimpleDBClient, session.client("sdb", region_name=region))
+    sdb_client = session.client("sdb", region_name=region)
     sdb_client.delete_domain(DomainName=sdb_domain_name)
     printq(f'SBD Domain: "{sdb_domain_name}" successfully deleted.', quiet)
 
@@ -162,7 +170,10 @@ def retry_s3(delays: Iterable[float] = DEFAULT_DELAYS, timeout: float = DEFAULT_
     """
     return old_retry(delays=delays, timeout=timeout, predicate=predicate)
 
-@retry(errors=[BotoServerError])
+@retry(errors=[ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
+)])
 def delete_s3_bucket(
     s3_resource: "S3ServiceResource",
     bucket: str,
@@ -238,7 +249,7 @@ def enable_public_objects(bucket_name: str) -> None:
     would be a very awkward way to do it. So we restore the old behavior.
     """
 
-    s3_client = cast(S3Client, session.client('s3'))
+    s3_client = session.client('s3')
 
     # Even though the new default is for public access to be prohibited, this
     # is implemented by adding new things attached to the bucket. If we remove
@@ -261,7 +272,7 @@ def get_bucket_region(bucket_name: str, endpoint_url: Optional[str] = None, only
     :param only_strategies: For testing, use only strategies with 1-based numbers in this set.
     """
 
-    s3_client = cast(S3Client, session.client('s3', endpoint_url=endpoint_url))
+    s3_client = session.client('s3', endpoint_url=endpoint_url)
 
     def attempt_get_bucket_location() -> Optional[str]:
         """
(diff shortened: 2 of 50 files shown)


src/toil/lib/ec2.py
```diff
@@ -244,7 +258,12 @@ def spotRequestNotFound(e):
         if instance_ids:
             # This next line is the reason we batch. It's so we can get multiple instances in
             # a single request.
-            yield ec2.get_only_instances(instance_ids)
+            for instance_id in instance_ids:
+                for attempt in retry_ec2():
+                    with attempt:
+                        # Increase hop limit from 1 to use Instance Metadata V2
+                        boto3_ec2.modify_instance_metadata_options(InstanceId=instance_id, HttpPutResponseHopLimit=3)
```
adamnovak, line 265: This should just be a call to `increase_instance_hop_limit()`, right?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -199,14 +223,20 @@ def __init__(self, clusterName, clusterType, zone, nodeStorage, nodeStorageOverr
         # or readClusterSettings()
         super().__init__(clusterName, clusterType, zone, nodeStorage, nodeStorageOverrides)
 
+        if self._zone is None:
+            logger.warning("Leader zone was never initialized before creating AWS provisioner. Defaulting to cluster zone.")
```
adamnovak, line 227: Maybe this should be about how a zone was not provided (by the user) instead of never initialized (by us)?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -216,32 +246,44 @@ def createClusterSettings(self):
...
         self._worker_subnets_by_zone = self._get_good_subnets_like(self._leader_subnet)
 
-        self._leaderPrivateIP = instanceMetaData['local-ipv4']  # this is PRIVATE IP
-        self._keyName = list(instanceMetaData['public-keys'].keys())[0]
-        self._tags = {k: v for k, v in self.getLeader().tags.items() if k != _TAG_KEY_TOIL_NODE_TYPE}
+        from boto.utils import get_instance_metadata
+
+        self._keyName = instance["KeyName"]
+        logger.error(f"{self._keyName}")
+        # self._keyName = list(instanceMetaData['public-keys'].keys())[0]
```
adamnovak, line 271:
```suggestion
```

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -216,32 +246,44 @@ def createClusterSettings(self):
...
         # Determine where to deploy workers.
         self._worker_subnets_by_zone = self._get_good_subnets_like(self._leader_subnet)
 
-        self._leaderPrivateIP = instanceMetaData['local-ipv4']  # this is PRIVATE IP
-        self._keyName = list(instanceMetaData['public-keys'].keys())[0]
-        self._tags = {k: v for k, v in self.getLeader().tags.items() if k != _TAG_KEY_TOIL_NODE_TYPE}
+        from boto.utils import get_instance_metadata
+
+        self._keyName = instance["KeyName"]
+        logger.error(f"{self._keyName}")
```
adamnovak, line 270: This looks left in accidentally?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -216,32 +246,44 @@ def createClusterSettings(self):
...
+        # self._keyName = list(instanceMetaData['public-keys'].keys())[0]
+        self._leaderPrivateIP = ec2_metadata.private_ipv4  # this is PRIVATE IP
+        self._tags = {k: v for k, v in (self.getLeader().tags or {}).items() if k != _TAG_KEY_TOIL_NODE_TYPE}
         # Grab the ARN name of the instance profile (a str) to apply to workers
-        self._leaderProfileArn = instanceMetaData['iam']['info']['InstanceProfileArn']
+        leader_info = ec2_metadata.iam_info
+        if leader_info is None:
+            logger.error("Failed to query leader IAM metadata. This may cause issues with the cluster in the future.")
```
adamnovak, line 277: Can we be more specific about what kind of issues and why they might or might not happen? It's something like the leader might not have permissions to access AWS services, right?

adamnovak: Also won't this always happen if the leader is not itself in AWS?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -777,10 +819,10 @@ def _recover_node_type_bid(self, node_type: Set[str], spot_bid: Optional[float])
 
         return spot_bid
 
-    def addNodes(self, nodeTypes: Set[str], numNodes, preemptible, spotBid=None) -> int:
+    def addNodes(self, nodeTypes: Set[str], numNodes: int, preemptible: bool, spotBid: Optional[float]=None) -> int:
         # Grab the AWS connection we need
-        ec2 = self.aws.boto2(self._region, 'ec2')
-
+        # ec2 = self.aws.boto2(self._region, 'ec2')
```
adamnovak, line 824:
```suggestion
```

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -822,21 +867,41 @@ def addNodes(self, nodeTypes: Set[str], numNodes, preemptible, spotBid=None) ->
...
+                           'BlockDeviceMappings': bdm,
+                           'IamInstanceProfile': {
+                               'Arn': self._leaderProfileArn
+                           },
+                           'Placement': {
+                               'AvailabilityZone': zone
+                           },
+                           'SubnetId': subnet_id}
+                       }
+        # _TAG_KEY_TOIL_NODE_TYPE
```
adamnovak, line 890: What is this comment telling me?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -845,41 +910,45 @@ def addNodes(self, nodeTypes: Set[str], numNodes, preemptible, spotBid=None) ->
                 # every request in this method
                 if not preemptible:
                     logger.debug('Launching %s non-preemptible nodes', numNodes)
-                    instancesLaunched = create_ondemand_instances(ec2,
-                                                                  image_id=self._discoverAMI(),
-                                                                  spec=kwargs, num_instances=numNodes)
+                    instancesLaunched = create_ondemand_instances(boto3_ec2=boto3_ec2,
+                                                                                         image_id=self._discoverAMI(),
+                                                                                         spec=on_demand_kwargs, num_instances=numNodes)
```
adamnovak, line 915: This looks indented oddly.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1121,96 +1225,90 @@ def _getBoto3BlockDeviceMappings(cls, type_info: InstanceType, rootVolSize: int
...
+        ec2: EC2Client = self.aws.client(self._region, 'ec2')
 
-        requests = ec2.get_all_spot_instance_requests()
-        tags = ec2.get_all_tags({'tag:': {_TAG_KEY_TOIL_CLUSTER_NAME: self.clusterName}})
-        idsToCancel = [tag.id for tag in tags]
-        return [request for request in requests if request.id in idsToCancel]
+        requests: List[SpotInstanceRequestTypeDef] = ec2.describe_spot_instance_requests()["SpotInstanceRequests"]
+        tag_filter: FilterTypeDef = {"Name": "tag:" + _TAG_KEY_TOIL_CLUSTER_NAME, "Values": [self.clusterName]}
+        tags: List[TagDescriptionTypeDef] = ec2.describe_tags(Filters=[tag_filter])["Tags"]
+        # tags = ec2.get_all_tags({'tag:': {_TAG_KEY_TOIL_CLUSTER_NAME: self.clusterName}})
```
adamnovak, line 1263:
```suggestion
```

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1121,96 +1225,90 @@ def _getBoto3BlockDeviceMappings(cls, type_info: InstanceType, rootVolSize: int
...
         # The VPC connection can do anything the EC2 one can do, but also look at subnets.
-        vpc = self.aws.boto2(self._region, 'vpc')
+        boto3_ec2: EC2Client = self.aws.client(region=self._region, service_name="ec2")
 
-        def groupNotFound(e):
-            retry = (e.status == 400 and 'does not exist in default VPC' in e.body)
-            return retry
-        # Security groups need to belong to the same VPC as the leader. If we
-        # put the leader in a particular non-default subnet, it may be in a
-        # particular non-default VPC, which we need to know about.
```
adamnovak, line 1180: Isn't this still true or important?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1121,96 +1225,90 @@ def _getBoto3BlockDeviceMappings(cls, type_info: InstanceType, rootVolSize: int
...
         try:
-            web = vpc.create_security_group(self.clusterName,
-                                            'Toil appliance security group', vpc_id=vpcId)
-        except EC2ResponseError as e:
-            if e.status == 400 and 'already exists' in e.body:
-                pass  # group exists- nothing to do
+            if vpc_id is None:
+                web_response: CreateSecurityGroupResultTypeDef = boto3_ec2.create_security_group(GroupName=self.clusterName, Description='Toil appliance security group')
+            else:
+                web_response = boto3_ec2.create_security_group(GroupName=self.clusterName, Description='Toil appliance security group', VpcId=vpc_id)
```
adamnovak, line 1288: This might be better done as building a dict, conditionally adding a key, and `**`ing it into the boto3 call's kwargs.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1481,22 +1579,26 @@ def _createWorkerAutoScalingGroup(self,
...
             yield from getattr(result, result_attribute_name)
             if result.is_truncated == 'true':
                 marker = result.marker
             else:
                 break
 
-    def _pager(self, requestor_callable: Callable, result_attribute_name: str, **kwargs) -> Iterable[Dict[str, Any]]:
+    # A = TypeVar("A", bound=LiteralString)
+    # B = TypeVar("B")
+    # func: Gettable[Literal["PolicyNames"], str] = ListRolePoliciesResponseTypeDef(PolicyNames= ["a"], IsTruncated= True, Marker="a", ResponseMetadata=ResponseMetadataTypeDef(RequestId="",HostId="",HTTPStatusCode=1,HTTPHeaders={},RetryAttempts=1))
```
adamnovak, line 1599:
```suggestion
```

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -1539,11 +1643,13 @@ def _getInstanceProfileNames(self) -> List[str]:
...
         results = []
-        for result in self._boto2_pager(self.aws.boto2(self._region, 'iam').list_instance_profiles,
-                                        'instance_profiles'):
+        boto3_iam = self.aws.client(self._region, 'iam')
+        for result in self._pager(boto3_iam.list_instance_profiles,
+                                  'InstanceProfiles'):
             # For each Boto2 role object
             # Grab out the name
-            name = result['instance_profile_name']
+            cast(InstanceProfileTypeDef, result)
```
adamnovak, line 1651: Doesn't the result of the cast need to be kept?

src/toil/test/provisioners/aws/awsProvisionerTest.py
```diff
@@ -191,28 +200,27 @@ def _test(self, preemptibleJobs=False):
...
+            time.sleep(10)
+        if volumes is None or len(volumes) > 0:
             self.fail('Volume with ID %s was not cleaned up properly' % volumeID)
 
         assert len(self.cluster._getRoleNames()) == 0
 
 
 @integrative
-@needs_mesos
+# @needs_mesos
```
adamnovak, line 223: Doesn't this need the Mesos extra? It can't work on Python too new to have PyMesos, right?

src/toil/test/provisioners/clusterTest.py
```diff
@@ -40,7 +41,12 @@ def __init__(self, methodName: str) -> None:
         # We need a boto2 connection to EC2 to check on the cluster.
         # Since we are protected by needs_aws_ec2 we can import from boto.
         import boto.ec2
-        self.boto2_ec2 = boto.ec2.connect_to_region(zone_to_region(self.zone))
+        self.region = zone_to_region(self.zone)
+        self.boto2_ec2 = boto.ec2.connect_to_region(self.region)
```
adamnovak, line 45: Shouldn't this come off the AWSConnectionManager()?

src/toil/test/provisioners/clusterTest.py
```diff
@@ -40,7 +41,12 @@ def __init__(self, methodName: str) -> None:
...
+
+        # Get connection to AWS with boto3/boto2
+        self.aws = AWSConnectionManager()
```
adamnovak, line 48: Does this need to know the default region we took in? Or do we always pass it the region when using it?

src/toil/lib/aws/__init__.py
```diff
@@ -128,7 +129,7 @@ def get_current_aws_zone() -> Optional[str]:
         get_aws_zone_from_environment_region() or \
         get_aws_zone_from_boto()
 
-def zone_to_region(zone: str) -> str:
+def zone_to_region(zone: str) -> Union["BucketLocationConstraintType", Literal["us-east-1"]]:
```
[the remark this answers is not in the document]
stxue1: I believe some Toil functions down the line require this specific type, such as this: https://github.com/DataBiosphere/toil/blob/6f00e5929ed64467c984afde31e2fb625e49475c/src/toil/lib/aws/utils.py#L198, so this is to satisfy mypy for Toil code.

src/toil/lib/ec2.py
stxue1: I don't think this is a boto3 call, and this seems like the main way to adjust HTTP related settings in boto3

src/toil/provisioners/aws/awsProvisioner.py
stxue1: I think the RuntimeError above probably catches whether the zone is None, and the leader_subnet seems to just be a mirror of zone, which would be caught if it was None. So this logger message probably would never be reached. I can probably get rid of this logger message which I believe I contained originally for mypy.

src/toil/provisioners/aws/awsProvisioner.py
stxue1: `iam_info` may return None, which has information that is later used to provision new workers/nodes such as spot instances. `ec2_metadata.iam_info` returns None by catching a 404 response, so it is probably due to the leader not being on AWS or AWS is unreachable. I think it's probably better to wrap this in a retry with a timeout. If the metadata isn't reachable, there probably is a greater error, and Toil's provisioning will probably break.

src/toil/provisioners/aws/awsProvisioner.py
stxue1: I think this still applies as the code is just a port, so I can add the comment back

src/toil/test/provisioners/clusterTest.py
stxue1: This should be fixed in my next PR, I mainly kept the boto2 part here to avoid spending too much time on porting over tests

src/toil/test/provisioners/clusterTest.py
stxue1: We should be always passing the region when using it, I don't think we can pass in a default region directly

src/toil/lib/aws/utils.py
```diff
@@ -104,16 +106,19 @@ def delete_iam_role(
...
-        boto_iam_connection.delete_role_policy(role.name, inline_policy.policy_name)
+        iam_client.delete_role_policy(RoleName=role.name, PolicyName=inline_policy.policy_name)
     iam_client.delete_role(RoleName=role_name)
     printq(f'Role {role_name} successfully deleted.', quiet)
 
 
-@retry(errors=[BotoServerError])
+@retry(errors=[ErrorCondition(
+    error=ClientError,
+    error_codes=[404, 500, 502, 503, 504]
```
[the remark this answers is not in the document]
stxue1: I think this exact signature is repeated elsewhere across files. There probably should be a constant ErrorCondition instance

src/toil/lib/aws/__init__.py
```diff
@@ -16,13 +16,23 @@
...
 from urllib.error import URLError
 from urllib.request import urlopen
 
+from botocore.exceptions import ClientError
+
 from mypy_boto3_s3.literals import BucketLocationConstraintType
 
+AWSRegionName = Union[BucketLocationConstraintType, Literal["us-east-1"]]
+
+AWSServerErrors = toil.lib.retry.ErrorCondition(
```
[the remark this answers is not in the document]
adamnovak: Also this is named like a type but really maybe it ought to be a constant like `AWS_SERVER_ERRORS`?

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -264,19 +259,20 @@ def readClusterSettings(self) -> None:
...
         self._leaderPrivateIP = ec2_metadata.private_ipv4  # this is PRIVATE IP
         self._tags = {k: v for k, v in (self.getLeader().tags or {}).items() if k != _TAG_KEY_TOIL_NODE_TYPE}
         # Grab the ARN name of the instance profile (a str) to apply to workers
-        leader_info = ec2_metadata.iam_info
+        leader_info = None
+        for attempt in old_retry(timeout=300, predicate=lambda e: isinstance(e, ValueError)):
+            with attempt:
+                leader_info = ec2_metadata.iam_info
+                if leader_info is None:
+                    raise ValueError(f"Failed to query leader IAM metadata. EC2 instance is not reachable.")
```
adamnovak, line 270: This seems a little like a misuse of `ValueError`. Will `ec2_metadata` raise ValueError when it fails?

If you're using `old_retry` I think you can just go around the loop again and generate a new `attempt` if you don't get an exception but also don't like the answer.

Also, don't you need a `break` to get out of this loop when you do have the metadata?

Also also, I don't think "EC2 instance is not reachable" is the right message here. At this point *we're* the EC2 instance. Were trying to talk to the IMDS to get metadata about ourselves. It's odd to say that the machine we're on is not reachable by the service we assume is supposed to be talking to us.

adamnovak (changes requested):
I think the metadata retry loop might be wrong, but other than that I think it's basically good now.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -263,11 +263,10 @@ def readClusterSettings(self) -> None:
         self._tags = {k: v for k, v in (self.getLeader().tags or {}).items() if k != _TAG_KEY_TOIL_NODE_TYPE}
         # Grab the ARN name of the instance profile (a str) to apply to workers
         leader_info = None
-        for attempt in old_retry(timeout=300, predicate=lambda e: isinstance(e, ValueError)):
+        for attempt in old_retry(timeout=300, predicate=lambda e: True):
```
adamnovak, line 266: We want to be careful with this too because this will retry on e.g. `ec2_metadata` not being defined, or other errors we would like to treat as genuine permanent problems. But the code here might be small enough that we can get away with it.

src/toil/provisioners/aws/awsProvisioner.py
```diff
@@ -263,11 +263,10 @@ def readClusterSettings(self) -> None:
...
             with attempt:
                 leader_info = ec2_metadata.iam_info
-                if leader_info is None:
-                    raise ValueError(f"Failed to query leader IAM metadata. EC2 instance is not reachable.")
+                assert leader_info is not None
```
adamnovak, line 269: It's *possible* for Python to run with asserts off, so if we actually expect this to sometimes happen we should manually throw something.

```suggestion
                if leader_info is None:
                    raise RuntimeError("Could not get EC2 metadata IAM info")
```

adamnovak (changes requested):
I don't think we want to rely on asserts being on for the logic, but otherwise this is OK.

src/toil/provisioners/aws/awsProvisioner.py
stxue1: It might be best to just define a new exception local to the scope of where ec2metadata is imported and being called.

PR #4868, opened 2024-04-10
Replace all usage of boto2 with boto3

### LINKED ISSUE #4718: Usage of boto (not boto3) prevents toil from working with Python3.12

stxue1:
Extension to #4859. Some of the fixes in that PR haven't been merged into this one yet, which I'll merge downstream once that PR is merged into master.

This will also close #4718.

I also moved the WDL tests on Kubernetes into it's own file as there was a circular dependency otherwise.

## Changelog Entry

 * Replace all instances of boto2 with boto3 for all Toil AWS code

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/jobStores/aws/jobStore.py
```diff
@@ -136,15 +138,15 @@ def __init__(self, locator: str, partSize: int = 50 << 20) -> None:
         self.region = region
         self.namePrefix = namePrefix
         self.partSize = partSize
-        self.jobsDomain = None
-        self.filesDomain = None
+        self.jobsDomainName: Optional[str] = None
+        self.filesDomainName: Optional[str] = None
```
adamnovak, line 142: Maybe we should refactor these to snake_case while we're touching them?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -377,49 +419,53 @@ def delete_job(self, job_id):
...
+                attributes = self.db.get_attributes(DomainName=self.jobsDomainName,
+                                                    ItemName=compat_bytes(job_id),
+                                                    ConsistentRead=True).get("Attributes", [])
         # If the overlargeID has fallen off, maybe we partially deleted the
         # attributes of the item? Or raced on it? Or hit SimpleDB being merely
         # eventually consistent? We should still be able to get rid of it.
-        self._checkItem(item, enforce = False)
-        if item.get("overlargeID", None):
+        self._checkAttributes(attributes, enforce=False)
+        overlarge_id_value = next((attribute["Value"] for attribute in attributes if attribute["Name"] == "overlargeID"), None)
```
adamnovak, line 431: I'm getting tired of reading basically the same comprehension repeatedly. This could be wrapped in a function to save code, I think.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -676,15 +723,16 @@ def get_shared_public_url(self, shared_file_name):
...
-        db = boto.sdb.connect_to_region(self.region)
-        if db is None:
+        sdb = boto3_session.client(service_name="sdb", region_name=self.region)
+        # db = boto.sdb.connect_to_region(self.region)
+        if sdb is None:
             raise ValueError("Could not connect to SimpleDB. Make sure '%s' is a valid SimpleDB region." % self.region)
-        monkeyPatchSdbConnection(db)
-        return db
+        # monkeyPatchSdbConnection(db)
+        return sdb
```
adamnovak, line 735: Probably we should drop this function and just use `boto3_session.client()`, or a `self.sdb` set from `boto3_session.client()`. All it does is call a method and check for a None that I think can't happen anymore.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -815,8 +863,10 @@ def bucket_retry_predicate(error):
 
                 return bucket
 
-    def _bindDomain(self, domain_name, create=False, block=True):
+    def _bindDomain(self, domain_name: str, create: bool = False, block: bool = True) -> Optional[str]:
```
adamnovak, line 866: Since there's no domain object anymore, this now returns a string that you passed it to indicate success, or None to indicate failure.

That's not a design that we would have arrived at if we had set out to build this on Boto 3 to start with.

Can we get it to not return anything and raise an exception if create=False and it doesn't exist? And change the usages to work with that?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -827,7 +877,7 @@ def _bindDomain(self, domain_name, create=False, block=True):
         :param bool block: If False, return None if the domain doesn't exist. If True, wait until
                domain appears. This parameter is ignored if create is True.
 
-        :rtype: Domain|None
+        :rtype: str domain_name | None
         :raises SDBResponseError: If `block` is True and the domain still doesn't exist after the
```
adamnovak, line 881: This sounds like it isn't true anymore; isn't `SDBResponseError` a Boto 2 thing?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -969,18 +1021,21 @@ def presenceIndicator(cls):
...
+                                                            ConsistentRead=True).get("Attributes", []))
 
         @classmethod
         def load(cls, jobStoreFileID):
             for attempt in retry_sdb():
                 with attempt:
                     self = cls.fromItem(
-                        cls.outer.filesDomain.get_attributes(item_name=compat_bytes(jobStoreFileID),
-                                                             consistent_read=True))
+                        {
```
adamnovak, line 1034: The closing bracket for this should be at the same indentation level.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1023,27 +1078,24 @@ def strOrNone(s):
...
                         raise AssertionError('Content is encrypted but no key was provided.')
                     if content is not None:
                         content = encryption.decrypt(content, sseKeyPath)
-                self = cls(fileID=item.name, ownerID=ownerID, encrypted=encrypted, version=version,
+                self = cls(fileID=item["Name"], ownerID=ownerID, encrypted=encrypted, version=version,
                            content=content, numContentChunks=numContentChunks, checksum=checksum)
                 return self
 
-        def toItem(self):
+        def toItem(self) -> Tuple[Dict[str, str], int]:
```
adamnovak, line 1098: We can cut the rtype fields from the docstrings if we now have real type annotations.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1023,27 +1078,24 @@ def strOrNone(s):
...
             """
             Convert this instance to an attribute dictionary suitable for SDB put_attributes().
```
adamnovak, line 1100: This might not be the right way to describe the result anymore if it now requires further processing with `SDBHelper` to actually send to AWS. Maybe we want to call it just a dictionary from attribute name to value?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1096,14 +1154,21 @@ def save(self):
                 self._previousVersion = self._version
                 if numNewContentChunks < self._numContentChunks:
                     residualChunks = range(numNewContentChunks, self._numContentChunks)
-                    attributes = [self._chunkName(i) for i in residualChunks]
+                    residual_chunk_names = [self._chunkName(i) for i in residualChunks]
+                    # boto3 requires providing the value as well as the name in the attribute, and we don't store it locally
```
adamnovak, line 1158: I don't think this is true on the backend. The PHP bindings were changed to not need it: https://github.com/aws/aws-sdk-php/issues/185

Is this a boto3 bug? What happens if you send a value different than what is stored, such as an empty string? Or if you don't actually send the value? Maybe the docs are wrong that value is required?

src/toil/jobStores/aws/utils.py
```diff
@@ -143,10 +143,41 @@ def binaryToAttributes(cls, binary):
...
         n = cls.maxValueSize
         chunks = (encoded[i:i + n] for i in range(0, len(encoded), n))
-        attributes = {cls._chunkName(i): chunk for i, chunk in enumerate(chunks)}
-        attributes.update({'numChunks': len(attributes)})
+        attributes = {cls._chunkName(i): chunk.decode("utf-8") for i, chunk in enumerate(chunks)}
+        attributes.update({'numChunks': str(len(attributes))})
         return attributes
 
+    @classmethod
+    def attributeDictToBoto3(cls, attributes: Dict[str, str]) -> List[AttributeTypeDef]:
```
adamnovak, line 151: Since we no longer have Boto 2 and Boto 3, maybe this should be called `attributeDictToList` or something?

src/toil/jobStores/aws/utils.py
```diff
@@ -143,10 +143,41 @@ def binaryToAttributes(cls, binary):
...
+        """
+        Convert the attribute dict (ex: from binaryToAttributes) into a list of attribute typed dicts
+        to be compatible with boto3 argument syntax
+        :param attributes: Dict[str, str], attribute in object form
+        :return: List[AttributeTypeDef], list of attributes in typed dict form
+        """
+        return [{"Name": name, "Value": value} for name, value in attributes.items()]
+
+    @classmethod
+    def attributeBoto3ToDict(cls, attributes: List[AttributeTypeDef]) -> Dict[str, str]:
```
adamnovak, line 161: And similarly this might not make sense to describe as converting from the Boto 3 format if there's no Boto 2.

src/toil/test/provisioners/aws/awsProvisionerTest.py
```diff
@@ -220,7 +220,7 @@ def _test(self, preemptibleJobs=False):
 
 
 @integrative
-@needs_mesos
+# @needs_mesos
```
adamnovak, line 223: Doesn't this still need Mesos?
```suggestion
@needs_mesos
```

src/toil/test/provisioners/clusterTest.py
```diff
@@ -40,9 +40,7 @@ def __init__(self, methodName: str) -> None:
         assert self.zone is not None, "Could not determine AWS availability zone to test in; is TOIL_AWS_ZONE set?"
         # We need a boto2 connection to EC2 to check on the cluster.
         # Since we are protected by needs_aws_ec2 we can import from boto.
```
adamnovak, line 42: This isn't true anymore, right?
```suggestion
```

src/toil/test/provisioners/clusterTest.py
```diff
@@ -40,9 +40,7 @@ def __init__(self, methodName: str) -> None:
...
-        import boto.ec2
         self.region = zone_to_region(self.zone)
-        self.boto2_ec2 = boto.ec2.connect_to_region(self.region)
 
         # Get connection to AWS with boto3/boto2
```
adamnovak, line 45:
```suggestion
        # Get connection to AWS
```

adamnovak (changes requested):
This looks pretty good. I found some commented-out code to drop.

I feel like it would be better with a wrapper function for fetching values from those lists of per-attribute dicts. And the domain binding function could have a different interface (and name?) since "binding" an SDB domain by making an object for it doesn't actually happen anymore.

And I think we should try harder to try and get out of having to send the current attribute value to delete an attribute. It makes no sense that we would need to do that, since it's not required by the server as far as I can tell, so either we don't actually have to provide it, or else it's a boto3 client bug.

src/toil/jobStores/aws/jobStore.py
stxue1: I think the docs are wrong in this case. I was using the stubs/boto3 docs when originally looking at this; Pycharm thinks the code is wrong as omitting the "Value" member of the dictionary means the argument type is no longer `List[AttributeTypeDef]`

stxue1: It might just be a quirk of boto3 as the `AttributeNames` argument must match a specific typed dict.
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Missing required parameter in Attributes[0]: "Value"
```

adamnovak: Does it matter what we pass there? Does it still work if we don't bother to get the real current values and always pass an empty string? Or does it get treated like an expect and not do the delete if the value isn't the one that's there?

adamnovak:
It looks like maybe we neglected to clean up some SimpleDB domains and ran out of them?
```
botocore.errorfactory.NumberDomainsExceeded: An error occurred (NumberDomainsExceeded) when calling the CreateDomain operation: Number of domains limit exceeded.
```

I'll try running the cleanup script.

adamnovak:
OK, we shoudl no longer be overwhelmed with SimpleDB domains. But @stxue1 you might want to go over the code here to be sure the delete is working properly.

src/toil/jobStores/aws/jobStore.py
stxue1: If we pass the empty string, I think it tries to delete an attribute with an empty string value.
```python
>>> response = sdb.put_attributes(DomainName=domain_name, ItemName="test-item", Attributes=[{"Name": "test-name", "Value": "test-value", "Replace": True}])
>>> response = sdb.get_attributes(DomainName=domain_name, ItemName="test-item", AttributeNames=["test-name"])
>>> response
{'Attributes': [{'Name': 'test-name', 'Value': 'test-value'}], 'ResponseMetadata': {'RequestId': '1a3a63b6-fc69-43a1-a25e-e00310ef7f39', 'BoxUsage': '0.0000093222', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Fri, 26 Apr 2024 18:39:23 GMT', 'content-type': 'text/xml', 'transfer-encoding': 'chunked', 'connection': 'keep-alive', 'vary': 'Accept-Encoding', 'server': 'Amazon SimpleDB'}, 'RetryAttempts': 0}}
>>> response = sdb.delete_attributes(DomainName=domain_name, ItemName="test-item", Attributes=[{"Name": "test-name", "Value": ""}])
>>> response
{'ResponseMetadata': {'RequestId': 'fe5d7a31-c52f-497a-df0a-4176aaf13d0b', 'BoxUsage': '0.0000219909', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Fri, 26 Apr 2024 18:40:40 GMT', 'content-type': 'text/xml', 'transfer-encoding': 'chunked', 'connection': 'keep-alive', 'vary': 'Accept-Encoding', 'server': 'Amazon SimpleDB'}, 'RetryAttempts': 0}}
>>> response = sdb.get_attributes(DomainName=domain_name, ItemName="test-item", AttributeNames=["test-name"])
>>> response
{'Attributes': [{'Name': 'test-name', 'Value': 'test-value'}], 'ResponseMetadata': {'RequestId': '13fa4c91-c855-3d88-c1d0-bf576bb568c0', 'BoxUsage': '0.0000093222', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Fri, 26 Apr 2024 18:40:51 GMT', 'content-type': 'text/xml', 'transfer-encoding': 'chunked', 'connection': 'keep-alive', 'vary': 'Accept-Encoding', 'server': 'Amazon SimpleDB'}, 'RetryAttempts': 0}}
>>> response = sdb.delete_attributes(DomainName=domain_name, ItemName="test-item", Attributes=[{"Name": "test-name", "Value": "test-value"}])
>>> response = sdb.get_attributes(DomainName=domain_name, ItemName="test-item", AttributeNames=["test-name"])
>>> response
{'ResponseMetadata': {'RequestId': 'cf383b93-220f-7a7a-b6de-561792180d1e', 'BoxUsage': '0.0000093202', 'HTTPStatusCode': 200, 'HTTPHeaders': {'date': 'Fri, 26 Apr 2024 18:41:04 GMT', 'content-type': 'text/xml', 'transfer-encoding': 'chunked', 'connection': 'keep-alive', 'vary': 'Accept-Encoding', 'server': 'Amazon SimpleDB'}, 'RetryAttempts': 0}}
```

stxue1:
I think the SDB domain deletion code here should work, I ran some of the AWS provisioner tests and no extra SDB domains were left over. I think those domains were from running tests locally and failing to delete them there, as while debugging, I'd sometimes run certain tests in debug mode but terminate them before they fully complete.

src/toil/jobStores/aws/jobStore.py
adamnovak: OK, then I guess we do need the extra round trip.

adamnovak (approved):
It looks like everything I complained about was fixed.

PR #4873, opened 2024-04-18
Detect if the GridEngine worker thread has crashed to prevent hanging the workflow

### LINKED ISSUE #4856: Failed `sbatch` may hang a workflow
Mira noticed that a workflow she was running reported a failed `sbatch` command, and then reported a long run of `squeue` commands succeeding while apparently doing nothing.

It may be that a failed `sbatch` can kill a job submission thread that we need to be alive for the workflow not to get stuck.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1533)
┆Issue Number: TOIL-1533

stxue1:
Resolves #4856 

A failed `sbatch` command would cause the background thread (used to be called worker) to crash and die, hanging the entire workflow as there was no logic to detect that the thread is now gone, and python's `Threads` doesn't carry through exceptions.

The `squeue` commands being reported is due to the command call function spamming the debug log: https://github.com/DataBiosphere/toil/blob/78fe93db1f10fb7ba77db90694549fecc0f6470d/src/toil/lib/misc.py#L141

This is useful from a debugging perspective, but can be vague and spammy at times, such as in this case, where the `squeue` commands were due to the leader checking for deadlocks (and overlong jobs): https://github.com/DataBiosphere/toil/blob/78fe93db1f10fb7ba77db90694549fecc0f6470d/src/toil/leader.py#L814

I'm hesitant to add more debug messages as the leader main loop will call these functions repeatedly. Ideally there is some way to have Toil log something even lower priority than debug which will then be used for maintenance. It is already at debug level though, so it shouldn't be too user facing.
## Changelog Entry

 * Add logic to detect if the worker thread of the GridEngineBatchSystem dies unexpectedly
 * Rename GridEngineBatchSystem's Worker to GridEngineThread

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -428,7 +434,12 @@ def killBatchJobs(self, jobIDs):
         for jobID in jobIDs:
             self.killQueue.put(jobID)
         while jobIDs:
-            killedJobId = self.killedJobsQueue.get()
+            try:
+                killedJobId = self.killedJobsQueue.get(timeout=10)
+            except queue.Empty:
```
adamnovak, line 439: It looks like we already had `from queue import Empty, Queue`, so we should probably use the shorter name:
```suggestion
            except Empty:
```

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import queue
```
adamnovak, line 15: If we already import `Queue` and `Empty` later, we don't need the whole module probably.
```suggestion
```

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -277,14 +280,17 @@ def run(self):
                 while self._runStep():
                     pass
             except Exception as ex:
-                logger.error("GridEngine like batch system failure", exc_info=ex)
-                raise
+                self.exception = ex
+                logger.error("GridEngine like batch system failure: %s", ex)
+                # dont raise exception as is_alive will still be set to false,
```
adamnovak, line 285:
```suggestion
                # don't raise exception as is_alive will still be set to false,
```

src/toil/lib/retry.py
```diff
@@ -572,6 +572,7 @@ def old_retry(
     >>> i
     1
     """
+    timeout = timeout if timeout else DEFAULT_TIMEOUT
```
adamnovak, line 575: Won't this break `timeout=0`? Maybe this should be:
```suggestion
    if timeout is None:
        timeout = DEFAULT_TIMEOUT
```

Or maybe we really want `timeout=None` to represent not using a timeout?

adamnovak (changes requested):
This looks pretty good, but I'm not sure what's going on with the change to the retry timeout logic.

I'm also not 100% convinced that `GridEngineThread` is the best name for the thread that talks to the scheduler. But I don't have a better idea for a name, so maybe that will have to stay.

src/toil/lib/retry.py
stxue1: I think setting it to a default might be better. With the current retry code the only way to simply set an infinite timeout is to just set the timeout to a really high number.

adamnovak (approved):

Issue #4910, opened 2024-05-02
Do not write to /tmp when --tmpdir-prefix is set


arunisaac:
Even when I set `--tmpdir-prefix` to a directory other than `/tmp`, toil-cwl-runner writes to `/tmp`. This can be quite bad on a HPC machine where `/tmp` is part of the root filesystem and writing to `/tmp` is frowned upon. For reference, here's my full toil command:

```
$ toil-cwl-runner --batchSystem single_machine --workDir work --tmpdir-prefix tmp-toil --tmp-outdir-prefix tmp-toil-out --outdir outputs/hornbeam-assemble/ --jobStore jobstore/hornbeam-assemble assemble.cwl hornbeam-assemble-inputs.yml
```

and here are the contents of the file toil-cwl-runner wrote to `/tmp`.

```
$ cat /tmp/tmp5gs4l2gp
toil.bus.JobUpdatedMessage	kind-CWLWorkflow/instance-gzk2lelm	0
toil.bus.JobIssuedMessage	assemble.cwl	kind-CWLWorkflow/instance-gzk2lelm	1
toil.bus.QueueSizeMessage	1
toil.bus.JobCompletedMessage	assemble.cwl	kind-CWLWorkflow/instance-gzk2lelm	0
toil.bus.ExternalBatchIdMessage	1	4925	SingleMachineBatchSystem
toil.bus.JobUpdatedMessage	kind-CWLWorkflow/instance-gzk2lelm	0
toil.bus.JobIssuedMessage	assemble.cwl.bwa-index._:249e87ff-7b2e-4209-8e3f-60f56bef1135	kind-CWLJob/instance-krz6hyqg	2
toil.bus.QueueSizeMessage	1
toil.bus.ExternalBatchIdMessage	2	4927	SingleMachineBatchSystem
```

Running `TMPDIR=/some/other/path toil-cwl-runner ...` works around this. But, this should not be necessary given the existence of the `--tmpdir-prefix` argument.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1559)
┆Issue Number: TOIL-1559

adamnovak:
That looks like the message bus log file that the leader writes. It's part of the Toil core, while `--tmpdir-prefix` is part of only `toil-cwl-runner`.

[...]

PR #4914, opened 2024-05-03
Try /tmp before the workdir for the Toil coordination directory

### LINKED ISSUE #4877: Coordination directory moves with the work directory when that's not advisable
For a few reasons, on our Slurm cluster, Toil is putting the coordination directory at the location of the work directory, since it can't use any of the available tmpfs locations.

But then when a user has a workflow that can't fit its scratch in node-local scratch, and moves the work directory to our Ceph share, the coordination directory goes with it. Then the aLastProcessStandingArena will hammer Ceph with O(n^2) file locking attempts in rapid succession as workers try and see if they are the last one alive on their node, and this can hang the Ceph MDS, since the Ceph MDS apparently isn't *quite* correct in its internal locking setup, and in any case isn't designed for this workload.

The workaround is to set `TOIL_COORDINATION_DIR=/data/tmp` whenever you set the work directory onto shared storage.

The fix is to not make the coordination directory follow the work directory, and to work harder to keep it local to the node. We should maybe just use `/tmp` if it's there, even if it's not technically the selected temp directory.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1543)
┆Issue Number: TOIL-1543

stxue1:
Closes #4877.

This will try the /tmp directory before the work directory in case the work directory is actually shared, as /tmp should be local to the node. I'm not sure of a proper way to test if a directory/filesystem is shared or not, the closest that I could find is using a nonconsistent linux command: https://unix.stackexchange.com/questions/72223/check-if-folder-is-a-mounted-remote-filesystem

## Changelog Entry

 * Try the /tmp/toil_coordination before the Toil workdir
## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1306,6 +1306,9 @@ class Toil(ContextManager["Toil"]):
                     os.path.join(os.environ['XDG_RUNTIME_DIR'], 'toil'))) or
                 # Try under /run/lock. It might be a temp dir style sticky directory.
                 try_path('/run/lock') or
+                # Before trying the workdir, try the /tmp directory as tmp directories
+                # should be local to the node and not shared like a work directoy might be
+                try_path('/tmp/toil_coordination') or
                 # Finally, fall back on the work dir and hope it's a legit filesystem.
                 cls.getToilWorkDir(config_work_dir)
         )


DailyDreaming (approved):
LGTM.

adamnovak:
@stxue1 What are the implications for multiple users of using `/tmp/toil_coordination` rather than just `/tmp`? If my Toil creates `/tmp/toil_coordination` on a machine, will your Toil be able to write to it? Or will it be forced to skip it because I already used the name?

Shouldn't we check `$TMPDIR` instead of always using `/tmp`? On our Phoenix Slurm cluster we would probably want this to pick `/data/tmp`, which is our TMPDIR, and not just plain `/tmp`.

See also #4910 where we're going to have to find and remove all places where `/tmp` is written to that aren't moved when `toil-cwl-runner` gets a `--tmpdir-prefix`. This looks to add a new one.

stxue1:
I think `/tmp/toil_coordination` is too unreliable then, as the directory will be associated to a certain user. I think I should make the tmpdir fallback rely on /tmp itself. I'll probably use `gettempdir()` to get a valid temp directory while omitting it's fallback to the current directory. This should catch all expected temp environmental variables. It might be possible for me to carry through the `--tmpdir-prefix` as well, and if the bus's temp file is where I think it is, I can try to adjust that too.

PR #4918, opened 2024-05-06
Add better logic in finding a temp directory for the Toil coordination directory

stxue1:
Addresses #4914 and #4910

This currently lets the Toil bus tmpdir respect the `--tmpdir-prefix` from cwltool, if it exists. However, a lot of internal code depends on directory paths instead of prefixed paths, so the tmp coordination directory selection does not currently respect the full prefix path.

This does add another option to Toil's config object, without an associated add_argument (at least for wdl + default toil). If that is not ideal, maybe Toil should get a command line argument to specify where the tmp directory should be?

## Changelog Entry

 * Toil coordination directory will use an available tmp directory if available (respects environmental variables)
 * Message bus's tmp directory will respect cwltool's `--tmpdir-prefix`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/common.py
```diff
@@ -1268,6 +1273,36 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
...
+        """
+        Get a working temp directory, testing if it is accessible
+        If tmpdir_prefix is given, assuming it is accessible, it will return its parent directory
+        if the path itself is not a directory path
+        :param tmpdir_prefix: tmp directory to override
+        :return: Path or none
+        """
+        # if tmpdir_prefix exists, return its parent or None depending on if it is accessible
+        if tmpdir_prefix is not None:
+            return tmpdir_prefix if os.path.isdir(tmpdir_prefix) else os.path.dirname(tmpdir_prefix)
```
adamnovak, line 1287: I don't think we want `tmpdir_prefix if os.path.isdir(tmpdir_prefix)`. If `tmpdir_prefix` is `/tmp/cwltmp-` but the directory `/tmp/cwltmp-` happens to exist, we still want to make temproary files and directories like `/tmp/cwltmp-XXXX` and not change to making them *in* that directory. The CWL-style tmpdir prefix will end in `/` if it is itself a directory we are meant to make stuff in.

src/toil/common.py
```diff
@@ -1268,6 +1273,36 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
...
+
+        # Priority will be: TMPDIR > TEMP > TMP > /tmp
+        # gettempdir returns the current working directory as a last resort
+        # but we don't want that, so return None if the cwd is reached
```
adamnovak, line 1291: What if `TMPDIR` is `/tmp` and we happen to be standing in `/tmp` when running the workflow? I don't think we want to disqualify a directory *just because* it's the current directory. If it has to fall all the way back to the current directory as our temp directory, maybe we should in fact use it as our temp directory?

src/toil/common.py
```diff
@@ -1268,6 +1273,36 @@ def getToilWorkDir(configWorkDir: Optional[str] = None) -> str:
             raise RuntimeError(f'The directory specified by --workDir or TOIL_WORKDIR ({workDir}) does not exist.')
         return workDir
 
+    @staticmethod
+    def get_working_tmpdir(tmpdir_prefix: Optional[str] = None) -> Optional[str]:
+        """
+        Get a working temp directory, testing if it is accessible
+        If tmpdir_prefix is given, assuming it is accessible, it will return its parent directory
+        if the path itself is not a directory path
```
adamnovak, line 1281: I might recommend creating a directory with that prefix. Then we could obey it.

src/toil/bus.py
```diff
@@ -741,12 +741,12 @@ def replay_message_bus(path: str) -> Dict[str, JobStatus]:
...
     return job_statuses
 
-def gen_message_bus_path() -> str:
+def gen_message_bus_path(tmpdir_prefix: Optional[str] = None) -> str:
     """
     Return a file path in tmp to store the message bus at.
     Calling function is responsible for cleaning the generated file.
     """
-    fd, path = tempfile.mkstemp()
+    fd, path = tempfile.mkstemp(prefix=tmpdir_prefix)
```
adamnovak, line 749: It looks like using an absolute path as a prefix actually does work. We might need a comment asserting this; I thought this was wrong when I looked at it, and the docs don't seem to promise that this works.

What if the `--tmpdir-prefix` option is a relative path? It looks like `mkstemp` interprets a non-absolute prefix as relative to its chosen temp directory, not relative to the current directory. Will something turn it into an absolute path before we get here? If so, we should cover that in the docstring.

adamnovak (changes requested):
I'm not sure this will work right for relative-path `--tmpdir-prefix` values; do we know those aren't allowed?

Also, I think we could do better at actually respecting `--tmpdir-prefix` when it is in fact a prefix by creating a temporary directory with that prefix. Maybe take the basename and use it as `prefix` and the dirname and use it as `dir` to `mkdtemp`?

Also also, instead of manually carrying through the prefix to use in particular places where we make temp files on the leader, maybe we should be setting `tempfile.tempdir` on the leader and the worker if we have a tmpdir_prefix?

src/toil/common.py
stxue1: Would we want to prioritize Toil's work dir over the current working directory? If I don't omit the current directory from `gettempdir`, then I think Toil's work dir will never be reached, as gettempdir() should always be returning something (unless toil is somehow installed in a nonwritable directory).

src/toil/bus.py
stxue1: If the prefix passed into `mkstemp` is a relative path, it looks like it will try to prefix that into the filename itself, choosing a valid tmp directory option
```python
>>> tmpdir_prefix = "test"
>>> fd, path = tempfile.mkstemp(prefix=tmpdir_prefix)
>>> path
'/tmp/testmo78tq3y'
```

stxue1: cwltool handles relative paths by creating the tmp file relative to the current working directory. I can probably mimic that.

src/toil/common.py
stxue1: I agree with that. The only downside is there is no easy way without a refactor to just tack on the `workflow_id` onto the prefix path and use that as the coordination directory. The current solution tacks the workflow_id on but then lets `get_local_workflow_coordination_dir` create another subdirectory with the same workflow_id name. This is since the batchsystem cleanup depends on `get_toil_coordination_dir` to return the parent directory of the coordination files, where all the coordination directory logic is. So in the current setup, I think I have to return a full directory path. I think the only way to respect the prefix probably is to either carry arguments down the function call (as I did) or carry return values up (so `get_local_workflow_coordination_dir` would understand when `get_toil_coordination_dir` returns a prefix/is not a full directory path)

src/toil/common.py
adamnovak: It would be nice if we could prioritize the workDir below actual temp directories like the ones from the environment variables and `/tmp`, but above the current working directory as a fallback if `/tmp` is unavailable.

But if the current working directory *is itself `/tmp`*, we still want to use `/tmp`.

adamnovak (changes requested):
I think there must be a simpler way to accomplish:
1. `toil-cwl-runner --tmpdir-prefix whatever` doesn't write to `/tmp`.
2. `workflow.py --workDir /shared/fs` doesn't use `/shared/fs` for the coordination directory.

While preserving the somewhat important security guarantee of:
3. Only secure functions like `tempfile.mkstemp` and `tempfile.mkdtemp` are used to create files in world-writable temp directories.

I think that, rather than trying to solve the hard problem of getting all the workers to agree on startup about what the coordination directory is, given only a prefix for which any given name including it might be taken, we should instead expand [what `toil-cwl-runner` already does for the workDir](https://github.com/DataBiosphere/toil/blob/115be6700e284e90d034b40ac4eea60dff06d526/src/toil/cwl/cwltoil.py#L3615-L3621) and, at workflow start, come up with paths obeying that prefix for the coordination directory and the Python process temp directory. (We could maybe make sure to clean them all up at CWL interpreter shutdown.)

Then we don't have to touch the message bus path generator at all (since it uses the Python process temp directory), and we don't have to let the somewhat fiddly concept of a temp directory prefix leak out of the CWL interpreter implementation to create hard problems in other parts of Toil, that then create complexity when we solve them.

If we do add `tmpdir_prefix` to the config, and we don't do something to teach the `tempfile` module to obey it, we are not going to be able to use the normal functions in `tempfile` to make internal Toil temp files, because they won't obey the prefix. That's a thing we'd have to remember at more or less all points in the Toil codebase, and we'd have to always pass around the config or prefix to anything that might need a temp file, whether CWL-related or not, so if we can avoid it we should.

src/toil/common.py
```diff
@@ -1306,9 +1331,6 @@ def get_toil_coordination_dir(cls, config_work_dir: Optional[str], config_coordi
                     os.path.join(os.environ['XDG_RUNTIME_DIR'], 'toil'))) or
                 # Try under /run/lock. It might be a temp dir style sticky directory.
                 try_path('/run/lock') or
-                # Before trying the workdir, try the /tmp directory as tmp directories
-                # should be local to the node and not shared like a work directoy might be
-                try_path('/tmp/toil_coordination') or
                 # Finally, fall back on the work dir and hope it's a legit filesystem.
```
adamnovak, line 1334: But now we're not actually avoiding using the workdir, for anything that isn't `toil-cwl-runner`.

adamnovak (changes requested):

adamnovak (approved):
OK, I think this will work. The workdir fallback is now I think dead code.

DailyDreaming (approved):
Nice.

Issue #4949, opened 2024-05-23
Support WDL 1.1 standard `gpu: true` Runtime attribute


adamnovak:
Our code for reading GPU requirements in WDL understands `gpuCount` (maybe to try and match Cromwell?):

https://github.com/DataBiosphere/toil/blob/ccf57e6071e32675daabdcbacb91988e871745a9/src/toil/wdl/wdltoil.py#L1582-L1600

But [WDL 1.1 says we need to take `gpu: true`](https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#gpu), which we don't.

We should add support for that flag.

This is somewhat related to #4945.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1579)
┆Issue Number: TOIL-1579

PR #4951, opened 2024-05-23
Grab GPU partition information from Slurm + Support for `gpu` field in WDL

### LINKED ISSUE #4833: Ability to get SLURM partitions.
So that we can send GPU jobs to GPU partitions.

stxue1:
Closes #4833 and #4949
## Changelog Entry

 * Added support for the `gpu` field in WDL
 * Added support to get Slurm partitions and automatically send jobs to GPUs on Slurm

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -398,8 +398,6 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
             # responded to this signal and use the right exit reason for it.
             sbatch_line.append("--signal=B:INT@30")
 
-            if gpus:
-                sbatch_line = sbatch_line[:1] + [f'--gres=gpu:{gpus}'] + sbatch_line[1:]
             environment = {}
             environment.update(self.boss.environment)
             if job_environment:
@@ -444,6 +442,37 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
             if cpu is not None:
                 sbatch_line.append(f'--cpus-per-task={math.ceil(cpu)}')
 
+            if gpus:
+                sbatch_line = sbatch_line[:1] + [f'--gres=gpu:{gpus}'] + sbatch_line[1:]
+
+                lowest_priority_gpus = sorted([i for i in self.boss.partitions if i[1]], key=lambda x: x[4])
+                new_partition_name = lowest_priority_gpus[0][0]  # get name of lowest priority partition that supports gpus
+                if not any(line.startswith("--partition") for line in sbatch_line):
+                    # no partition specified, so specify one
+                    sbatch_line.append(f"--partition={new_partition_name}")
+                else:
+                    # there is a partition specified already, check if the partition has GPUs
+                    for i, line in enumerate(sbatch_line):
+                        if line.startswith("--partition"):
+                            # grab the partition name depending on if it's specified via an "=" or a space
+                            if "=" in line:
+                                separated = False
+                                partition_name = line[len("--partition=")]
+                            else:
+                                separated = True
+                                partition_name = line[i+1]
+
+                            if not any(partition_name == partition for partition in lowest_priority_gpus):
+                                # the specified partition is not compatible, so override with a gpu supported partition
+                                logger.warning(f"Job {jobName} needs {gpus} GPUs, but specified partition {partition_name} is incompatible. Overriding with partition {new_partition_name}.")
+                                # remove old partition from sbatch line
+                                sbatch_line.pop(i)
+                                if separated:
+                                    sbatch_line.pop(i)
+                                # add new partition to sbatch line
+                                sbatch_line.append(f"--partition={new_partition_name}")
+                            break
+
             stdoutfile: str = self.boss.format_std_out_err_path(jobID, '%j', 'out')
             stderrfile: str = self.boss.format_std_out_err_path(jobID, '%j', 'err')
             sbatch_line.extend(['-o', stdoutfile, '-e', stderrfile])
@@ -467,7 +496,25 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                 pass  # slurm may return INVALID instead of a time
             return total_seconds
 
-    def _check_accelerator_request(self, requirer: Requirer) -> None:
+    def __init__(self, config, maxCores, maxMemory, maxDisk):
+        super().__init__(config, maxCores, maxMemory, maxDisk)
+        self._get_partition_info()
+
+    def _get_partition_info(self):
+        sinfo_command = ["sinfo",
+                         "-a",
+                         "-o",
+                         "%P %G %l %p %c %m"]
+
+        sinfo = call_command(sinfo_command)
+
+        self.partitions = []
+        for line in sinfo.split("\n")[1:]:
+            if line.strip():
+                partition_name, gres, time, priority, cpus, memory = line.split(" ")
+                self.partitions.append((partition_name.rstrip("*"), gres != "(null)", time, priority, cpus, memory))
+
+def _check_accelerator_request(self, requirer: Requirer) -> None:
         for accelerator in requirer.accelerators:
             if accelerator['kind'] != 'gpu':
                 raise InsufficientSystemResources(requirer, 'accelerators', details=
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -514,7 +514,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                 partition_name, gres, time, priority, cpus, memory = line.split(" ")
                 self.partitions.append((partition_name.rstrip("*"), gres != "(null)", time, priority, cpus, memory))
 
-def _check_accelerator_request(self, requirer: Requirer) -> None:
+    def _check_accelerator_request(self, requirer: Requirer) -> None:
         for accelerator in requirer.accelerators:
             if accelerator['kind'] != 'gpu':
                 raise InsufficientSystemResources(requirer, 'accelerators', details=
--- a/src/toil/batchSystems/slurm.py
+++ b/src/toil/batchSystems/slurm.py
@@ -457,7 +457,7 @@ class SlurmBatchSystem(AbstractGridEngineBatchSystem):
                             # grab the partition name depending on if it's specified via an "=" or a space
                             if "=" in line:
                                 separated = False
-                                partition_name = line[len("--partition=")]
+                                partition_name = line[len("--partition="):]
                             else:
                                 separated = True
                                 partition_name = line[i+1]
--- a/src/toil/lib/accelerators.py
+++ b/src/toil/lib/accelerators.py
@@ -142,7 +142,7 @@ def get_individual_local_accelerators() -> List[AcceleratorRequirement]:
     accelerator assignment API.
     """
 
-    gpus = [{'kind': 'gpu', 'brand': 'nvidia', 'api': 'cuda', 'count': 1} for _ in range(count_nvidia_gpus())]
+    gpus: List[AcceleratorRequirement] = [{'kind': 'gpu', 'brand': 'nvidia', 'api': 'cuda', 'count': 1} for _ in range(count_nvidia_gpus())]
     gpus.extend([{'kind': 'gpu', 'brand': 'amd', 'api': 'rocm', 'count': 1} for _ in range(count_amd_gpus())])
     return gpus
 
--- a/src/toil/lib/accelerators.py
+++ b/src/toil/lib/accelerators.py
@@ -15,6 +15,7 @@
 """Accelerator (i.e. GPU) utilities for Toil"""
 
 import os
+import string
 import subprocess
 from typing import Dict, List, Set, Union, cast
 from xml.dom import minidom
@@ -106,6 +107,30 @@ def count_nvidia_gpus() -> int:
     # TODO: Parse each gpu > product_name > text content and convert to some
     # kind of "model" that agrees with e.g. Kubernetes naming.
 
+@memoize
+def count_amd_gpus() -> int:
+    """
+    Return the number of amd GPUs seen by rocm-smi, or 0 if it is not working.
+    :return:
+    """
+    try:
+        # see if the amd-smi CLI tool is installed
+        # we believe this is the expected output for amd-smi, but we don't actually have and amd gpu to test against
+        out = subprocess.check_output((["amd-smi", "static"]))
+        gpu_count = len([line for line in out.decode("utf-8").split("\n") if line.startswith("gpu")])
+        return gpu_count
+    except (FileNotFoundError, subprocess.SubprocessError):
+        # if the amd-smi command fails, try rocm-smi
+        # if a different exception is raised, something other than the subprocess call is wrong
+        pass
+    try:
+        out = subprocess.check_output(["rocm-smi"])
+        gpu_count = len([line for line in out.decode("utf-8").split("\n") if len(line)> 0 and line[0] in string.digits])
+        return gpu_count
+    except (FileNotFoundError, subprocess.SubprocessError):
+        pass
+    return 0
+
 
 @memoize
 def get_individual_local_accelerators() -> List[AcceleratorRequirement]:
@@ -117,8 +142,9 @@ def get_individual_local_accelerators() -> List[AcceleratorRequirement]:
     accelerator assignment API.
     """
 
-    # For now we only know abput nvidia GPUs
-    return [{'kind': 'gpu', 'brand': 'nvidia', 'api': 'cuda', 'count': 1} for _ in range(count_nvidia_gpus())]
+    gpus = [{'kind': 'gpu', 'brand': 'nvidia', 'api': 'cuda', 'count': 1} for _ in range(count_nvidia_gpus())]
+    gpus.extend([{'kind': 'gpu', 'brand': 'amd', 'api': 'rocm', 'count': 1} for _ in range(count_amd_gpus())])
+    return gpus
 
 def get_restrictive_environment_for_local_accelerators(accelerator_numbers : Union[Set[int], List[int]]) -> Dict[str, str]:
     """
--- a/src/toil/wdl/wdltoil.py
+++ b/src/toil/wdl/wdltoil.py
@@ -69,6 +69,7 @@ from toil.job import (AcceleratorRequirement,
                       unwrap_all)
 from toil.jobStores.abstractJobStore import (AbstractJobStore, UnimplementedURLException,
                                              InvalidImportExportUrlException, LocatorException)
+from toil.lib.accelerators import count_nvidia_gpus, get_individual_local_accelerators
 from toil.lib.conversions import convert_units, human2bytes, strtobool
 from toil.lib.io import mkdtemp
 from toil.lib.memoize import memoize
@@ -1937,18 +1938,18 @@ class WDLTaskJob(WDLBaseJob):
 
                     extra_flags: Set[str] = set()
                     accelerators_needed: Optional[List[AcceleratorRequirement]] = self.accelerators
+                    local_accelerators = get_individual_local_accelerators()
                     if accelerators_needed is not None:
                         for accelerator in accelerators_needed:
                             if accelerator['kind'] == 'gpu':
-                                if accelerator.get('brand') is None:
-                                    # Base WDL syntax doesn't specify the GPU type/brand
-                                    extra_flags.add('--nv')
-                                elif accelerator['brand'] == 'nvidia':
+                                local_gpus = [accel['brand'] for accel in local_accelerators if accel['kind'] == 'gpu'] or [None]
+                                gpu_brand = accelerator.get('brand') or local_gpus[0]
+                                if gpu_brand == 'nvidia':
                                     # Tell Singularity to expose nvidia GPUs
                                     extra_flags.add('--nv')
-                                elif accelerator['api'] == 'rocm':
+                                elif gpu_brand == 'amd':
                                     # Tell Singularity to expose ROCm GPUs
-                                    extra_flags.add('--nv')
+                                    extra_flags.add('--rocm')
                                 else:
                                     raise RuntimeError('Cannot expose allocated accelerator %s to Singularity job', accelerator)
 
--- a/src/toil/wdl/wdltoil.py
+++ b/src/toil/wdl/wdltoil.py
@@ -1579,7 +1579,7 @@ class WDLTaskWrapperJob(WDLBaseJob):
             total_bytes: float = convert_units(total_gb, 'GB')
             runtime_disk = int(total_bytes)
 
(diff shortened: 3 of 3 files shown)


src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +442,37 @@ def prepareSbatch(self,
             if cpu is not None:
                 sbatch_line.append(f'--cpus-per-task={math.ceil(cpu)}')
 
+            if gpus:
+                sbatch_line = sbatch_line[:1] + [f'--gres=gpu:{gpus}'] + sbatch_line[1:]
+
+                lowest_priority_gpus = sorted([i for i in self.boss.partitions if i[1]], key=lambda x: x[4])
```
adamnovak, line 448: Having a magic `[4]` here suggests to me that the partition representation is not the right one. Maybe it wants to be a `NamedTuple`? What is field 4 and how do I know it's the right field to use here?

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +442,37 @@ def prepareSbatch(self,
...
+                new_partition_name = lowest_priority_gpus[0][0]  # get name of lowest priority partition that supports gpus
+                if not any(line.startswith("--partition") for line in sbatch_line):
```
adamnovak, line 450: I think `sbatch_line` is one (command) "line", so the elements are probably `option` and not `line`.
```suggestion
                if not any(option.startswith("--partition") for option in sbatch_line):
```

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +442,37 @@ def prepareSbatch(self,
...
+                    # no partition specified, so specify one
+                    sbatch_line.append(f"--partition={new_partition_name}")
+                else:
+                    # there is a partition specified already, check if the partition has GPUs
+                    for i, line in enumerate(sbatch_line):
```
adamnovak, line 455: Here also we're getting the lines in the line.

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +442,37 @@ def prepareSbatch(self,
...
+                            if not any(partition_name == partition for partition in lowest_priority_gpus):
+                                # the specified partition is not compatible, so override with a gpu supported partition
+                                logger.warning(f"Job {jobName} needs {gpus} GPUs, but specified partition {partition_name} is incompatible. Overriding with partition {new_partition_name}.")
+                                # remove old partition from sbatch line
+                                sbatch_line.pop(i)
+                                if separated:
+                                    sbatch_line.pop(i)
+                                # add new partition to sbatch line
+                                sbatch_line.append(f"--partition={new_partition_name}")
+                            break
```
adamnovak, line 474: I don't think we need or really even want to override the user's `--partition` option. I might be a user who under no circumstances wants my job to run in the `expensive_gpus` partition, and I might put `--partition=normal` explicitly to make sure that can never happen, and I might then have Toil helpfully ignore what I told it to do and drop stuff in the `expensive_gpus` partition anyway, and then I might have to pay a bunch of money to my university's IT department.

If the user explicitly sets a partition in the Slurm arguments, we probably should drop all the jobs in there even if we think that they will never schedule. We might want to warn the user that we think this isn't going to work (or maybe even fail the run?), but we shouldn't just directly override explicit instructions with auto-detected guesses.

src/toil/batchSystems/slurm.py
```diff
@@ -467,6 +496,24 @@ def parse_elapsed(self, elapsed):
...
+                         "-o",
+                         "%P %G %l %p %c %m"]
+
+        sinfo = call_command(sinfo_command)
+
+        self.partitions = []
+        for line in sinfo.split("\n")[1:]:
+            if line.strip():
+                partition_name, gres, time, priority, cpus, memory = line.split(" ")
+                self.partitions.append((partition_name.rstrip("*"), gres != "(null)", time, priority, cpus, memory))
```
adamnovak, line 515: Yeah, this is a complex enough tuple that it probably should be a `NamedTuple` type with field names.

src/toil/lib/accelerators.py
```diff
@@ -106,6 +107,30 @@ def count_nvidia_gpus() -> int:
...
+        out = subprocess.check_output((["amd-smi", "static"]))
+        gpu_count = len([line for line in out.decode("utf-8").split("\n") if line.startswith("gpu")])
+        return gpu_count
+    except (FileNotFoundError, subprocess.SubprocessError):
+        # if the amd-smi command fails, try rocm-smi
+        # if a different exception is raised, something other than the subprocess call is wrong
+        pass
+    try:
+        out = subprocess.check_output(["rocm-smi"])
+        gpu_count = len([line for line in out.decode("utf-8").split("\n") if len(line)> 0 and line[0] in string.digits])
```
adamnovak, line 128: We should link to the example amd-smi and rocm-smi output that we were looking at, since it was hard to find and also necessary to anyone who wants to understand or adjust this parsing code.

src/toil/wdl/wdltoil.py
```diff
@@ -1579,7 +1580,7 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
             total_bytes: float = convert_units(total_gb, 'GB')
             runtime_disk = int(total_bytes)
 
-        if runtime_bindings.has_binding('gpuType') or runtime_bindings.has_binding('gpuCount') or runtime_bindings.has_binding('nvidiaDriverVersion'):
+        if runtime_bindings.has_binding('gpu') or runtime_bindings.has_binding('gpuType') or runtime_bindings.has_binding('gpuCount') or runtime_bindings.has_binding('nvidiaDriverVersion'):
```
adamnovak, line 1583: We don't actually check the value of the `gpu` key, and we default `gpuCount` to 1 if it is not set. So if I have a task that just has `gpu: false` and none of the other keys, it's going to get 1 GPU.

We probably should check the `gpu` key value, and if it's present and explicitly false we should let it override the other ones, since it's the standard one. I could see someone writing a workflow where `gpuCount` is always 1 and `gpu` is determined by some expression and meant to gate whether GPU mode is actually used.

src/toil/wdl/wdltoil.py
```diff
@@ -1937,15 +1938,20 @@ def patched_run_invocation(*args: Any, **kwargs: Any) -> List[str]:
...
+                                # Tell singularity the GPU type
+                                gpu_brand = accelerator.get('brand') or local_gpus[0]
+                                if gpu_brand == 'nvidia':
                                     # Tell Singularity to expose nvidia GPUs
                                     extra_flags.add('--nv')
-                                elif accelerator['api'] == 'rocm':
+                                elif gpu_brand == 'amd':
                                     # Tell Singularity to expose ROCm GPUs
-                                    extra_flags.add('--nv')
+                                    extra_flags.add('--rocm')
```
adamnovak, line 1954: This won't handle the case where a task requests "2 GPUs" of unspecified types, and the machine has 2 GPUs, but they aren't the same brand, but the task is prepared for this and expects to e.g. poll what brands are available within the task and run two things in parallel that could take either brand of GPU. We'll always look at the first GPU and put the options needed to access it, even if we sort of already consumed it to satisfy the first required accelerator.

I think it probably isn't worth writing the code to support this case; only pedantic people with very unusual setups would try to make that work, and I'm not sure Singularity supports both these options at the same time anyway.

But we should add a note here about how that isn't going to work, so it doesn't look like we think this approach is technically correct.

adamnovak (changes requested):
This looks like it works for autodetecting a GPU partition, but I think the UI for specifying partitions manually is wrong/missing.

Right now it relies on you specifying a partition in `TOIL_SLURM_ARGS`, and then either that partition has GPUs, or when it sees a GPU job it second-guesses that partition and replaces it with an automatically-selected partition that does have GPUs.

I think it needs to be changed so that there's a way for the user to explicitly set the partition(s) to use, probably with a new command line argument on the Toil Slurm batch system. It should only guess partitions when the user hasn't explicitly assigned *any* partitions, and there needs to be a way to run most jobs in a non-GPU partition and the GPU jobs in a partition *other than* whatever the default one it autodetects is.

I'm not sure we need to be able to tell it to put GPU jobs in a partition that it can't tell has GPUs, so probably all we need is a way to feed in a set of partition names for it to pick from. If the set is empty it can guess automatically from all detected partitions, but otherwise it should pick partitions only from the set the user specified.

Then the user will be able to send jobs to partitions without touching `TOIL_SLURM_ARGS`.

src/toil/batchSystems/slurm.py
```diff
@@ -467,6 +489,25 @@ def parse_elapsed(self, elapsed):
...
+    def _get_partition_info(self):
+        sinfo_command = ["sinfo",
+                         "-a",
+                         "-o",
+                         "%P %G %l %p %c %m"]
+
+        sinfo = call_command(sinfo_command)
+
+        self.partitions = []
+        PartitionInfo = namedtuple("PartitionInfo", ["partition_name", "gres", "time", "priority", "cpus", "memory"])
```
adamnovak, line 505: Instead of minting the type at runtime it probably makes more sense to use `typing.NamedTuple` so this can be a `class`: https://docs.python.org/3/library/typing.html#typing.NamedTuple

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +443,29 @@ def prepareSbatch(self,
             if cpu is not None:
                 sbatch_line.append(f'--cpus-per-task={math.ceil(cpu)}')
 
+            if gpus:
+                # This block will add a gpu supported partition only if no partition is supplied by the user
+                sbatch_line = sbatch_line[:1] + [f'--gres=gpu:{gpus}'] + sbatch_line[1:]
+                lowest_priority_gpus = sorted([i for i in self.boss.partitions if i["gres"]], key=lambda x: x["priority"])
+                new_partition_name = lowest_priority_gpus[0][0]  # get name of lowest priority partition that supports gpus
```
adamnovak, line 450: This looks like it is going to cause an IndexError when the user hasn't specified a partition and a GPU job is run but no partition has GPUs, since the list is empty then. It's also going to cause an IndexError when the user *has* specified a partition, and runs a GPU job, but no partition appears to have GPUs, since we try and find a default partition whether we need one or not.

I think in the first case we want to [raise an `InsufficientSystemResources`](https://github.com/DataBiosphere/toil/blob/937d4fe5964a9f5a0faaeb5e5156c58ae9a593df/src/toil/batchSystems/abstractBatchSystem.py#L586) to complain that wee have no GPUs on Slurm, and in the second case we want to just print the warning we have about the selected partition not having GPUs.

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +443,29 @@ def prepareSbatch(self,
             if cpu is not None:
                 sbatch_line.append(f'--cpus-per-task={math.ceil(cpu)}')
 
+            if gpus:
+                # This block will add a gpu supported partition only if no partition is supplied by the user
+                sbatch_line = sbatch_line[:1] + [f'--gres=gpu:{gpus}'] + sbatch_line[1:]
+                lowest_priority_gpus = sorted([i for i in self.boss.partitions if i["gres"]], key=lambda x: x["priority"])
```
adamnovak, line 449: I would move this to happen at startup and split it into two things: a default GPU partition that we select once (or leave None if we don't have any partitions that seem to have GPUs) and a set of partitions we think have GPUs (for fast membership testing for the warning). I don't think "a list of all the GPU partitions sorted so the default one is first" is quite the right representation.

src/toil/batchSystems/slurm.py
```diff
@@ -444,6 +443,29 @@ def prepareSbatch(self,
...
+                        if option.startswith("--partition"):
+                            # grab the partition name depending on if it's specified via an "=" or a space
+                            if "=" in option:
+                                partition_name = option[len("--partition="):]
+                            else:
+                                partition_name = option[i+1]
+
+                            if not any(partition_name == partition for partition in lowest_priority_gpus):
+                                # the specified partition is not compatible, so override with a gpu supported partition
+                                logger.warning(f"Job {jobName} needs {gpus} GPUs, but specified partition {partition_name} is incompatible. This job may not work.")
```
adamnovak, line 466: If we do know of a partition with GPUs, maybe we should name drop it here to help the user.

adamnovak (changes requested):
This isn't quite what I was asking for, but it's probably fine to not build the whole partition-set-specification thing I wanted as part of this PR.

But we do need to handle the case where no partition has any GPUs without crashing in an uncontrolled way, and also I don't think namedtuple is being used in the most natural way.

src/toil/batchSystems/slurm.py
```diff
@@ -68,9 +69,61 @@
...
+        time: str
+        priority: str
+        cpus: str
+        memory: str
+
+    class PartitionSet:
+        """
+        Set of available partitions detected on the slurm batch system
+        """
+        default_gpu_partition: Optional[SlurmBatchSystem.PartitionInfo]
```
adamnovak, line 87: Isn't `PartitionInfo` itself in scope here?

src/toil/batchSystems/slurm.py
```diff
@@ -447,11 +500,15 @@ def prepareSbatch(self,
...
-                new_partition_name = lowest_priority_gpus[0][0]  # get name of lowest priority partition that supports gpus
                 if not any(option.startswith("--partition") for option in sbatch_line):
                     # no partition specified, so specify one
-                    sbatch_line.append(f"--partition={new_partition_name}")
+                    # try to get the name of the lowest priority gpu supported partition
+                    lowest_gpu_partition = self.boss.partitions.default_gpu_partition
+                    if lowest_gpu_partition is None:
+                        # no gpu partitions are available, raise an error
+                        required_gpus = Requirer(dict(accelerator=parse_accelerator(gpus)))
+                        raise InsufficientSystemResources(required_gpus, "accelerators", [], details=["Could not find a Slurm partition that supports GPUs!"])
```
adamnovak, line 510: I don't like that we're having to forge this since we don't actually have the original job anymore here. `InsufficientSystemResources` stringifies the Requirer input and uses that as a job name, so this will give us something like `The job <Requirer instance at wherever> is requesting 1 accelerators...` for our error message. Maybe we can't actually use `InsufficientSystemResources` here and should use a more generic error type.

```suggestion
                       raise RuntimeError(f"The job {jobName} is requesting GPUs, but the Slurm cluster does not appear to have an accessible partition with GPUs")
```

src/toil/batchSystems/slurm.py
```diff
@@ -461,10 +518,11 @@ def prepareSbatch(self,
                                 partition_name = option[len("--partition="):]
                             else:
                                 partition_name = option[i+1]
-
-                            if not any(partition_name == partition for partition in lowest_priority_gpus):
-                                # the specified partition is not compatible, so override with a gpu supported partition
-                                logger.warning(f"Job {jobName} needs {gpus} GPUs, but specified partition {partition_name} is incompatible. This job may not work.")
+                            available_gpu_partitions = self.boss.partitions.gpu_partitions
+                            if not any(partition_name == partition for partition in available_gpu_partitions):
```
adamnovak, line 522: This will not work since the GPU partition collection is actually full of `PartitionInfo`s and not name strings. We should make it a set and test membership I think.

```suggestion
                           if partition_name not in available_gpu_partitions:
```

src/toil/batchSystems/slurm.py
```diff
@@ -68,9 +69,61 @@
...
+        cpus: str
+        memory: str
+
+    class PartitionSet:
+        """
+        Set of available partitions detected on the slurm batch system
+        """
+        default_gpu_partition: Optional[SlurmBatchSystem.PartitionInfo]
+        all_partitions: List[SlurmBatchSystem.PartitionInfo]
+        gpu_partitions: List[SlurmBatchSystem.PartitionInfo]
```
adamnovak, line 89: This should probably be a set of names, since that's how we use it later.
```suggestion
        gpu_partitions: Set[str]
```

src/toil/batchSystems/slurm.py
```diff
@@ -447,11 +500,15 @@ def prepareSbatch(self,
...
                 if not any(option.startswith("--partition") for option in sbatch_line):
                     # no partition specified, so specify one
-                    sbatch_line.append(f"--partition={new_partition_name}")
+                    # try to get the name of the lowest priority gpu supported partition
+                    lowest_gpu_partition = self.boss.partitions.default_gpu_partition
+                    if lowest_gpu_partition is None:
+                        # no gpu partitions are available, raise an error
+                        required_gpus = Requirer(dict(accelerator=parse_accelerator(gpus)))
+                        raise InsufficientSystemResources(required_gpus, "accelerators", [], details=["Could not find a Slurm partition that supports GPUs!"])
+                    sbatch_line.append(f"--partition={lowest_gpu_partition}")
```
adamnovak, line 511: If the lowest GPU partition is a PartitionInfo, this needs to actually get the name.

```suggestion
                    sbatch_line.append(f"--partition={lowest_gpu_partition.name}")
```

src/toil/batchSystems/slurm.py
```diff
@@ -68,9 +69,61 @@
...
+            """
+            Get all available GPU partitions. Also get the default GPU partition.
+            :return: None
+            """
+            self.gpu_partitions = [partition for partition in self.all_partitions if partition.gres]
+            # Grab the lowest priority GPU partition
+            # If no GPU partitions are available, then set the default to None
+            self.default_gpu_partition = None
+            if len(self.gpu_partitions) > 0:
+                self.default_gpu_partition = sorted(self.gpu_partitions, key=lambda x: x.priority)[0]
```
adamnovak, line 105: If we want to make the GPU partitions a set of names we need to change this.

```suggestion
            gpu_partitions = [partition for partition in self.all_partitions if partition.gres]
            self.gpu_partitions = {p.name for p in gpu_partitions}
            # Grab the lowest priority GPU partition
            # If no GPU partitions are available, then set the default to None
            self.default_gpu_partition = None
            if len(gpu_partitions) > 0:
                self.default_gpu_partition = sorted(gpu_partitions, key=lambda x: x.priority)[0]
```

adamnovak (changes requested):
I don't think this actually works right now. Some of the code thinks we're representing partitions as tuples and some seems to expect strings.

src/toil/batchSystems/slurm.py
stxue1: I don't think so, both pycharm and mypy don't think it's within scope.

src/toil/batchSystems/slurm.py
```diff
@@ -344,7 +407,7 @@ def _getJobDetailsFromScontrol(self, job_id_list: list) -> dict:
                     if job_id not in job_id_list:
                         logger.log(TRACE, "%s job %d is not in the list", args[0], job_id)
                         break
-                if job_id not in job_id_list:
+                if job_id is None or job_id not in job_id_list:
```
stxue1, line 410: It's technically possible for job_id to be None if record is empty. I think this is unlikely if not impossible, but I'm not too sure what the expected behavior should be. Currently, I'm assuming it should skip the job_id.

adamnovak (approved):
Good idea to just set up type checking.

We probably want to no longer ignore slurm.py from our regular mypy checks after this.

adamnovak:
It looks like this might [break Slurm in general](https://ucsc-ci.com/databiosphere/toil/-/jobs/70938), though:
```
[31m[1m________ SlurmTest.test_coalesce_job_exit_codes_sacct_raises_job_exists ________[0m
[gw1] linux -- Python 3.12.0 /builds/databiosphere/toil/venv/bin/python
  File "/builds/databiosphere/toil/src/toil/batchSystems/slurm.py", line 236, in _get_job_details
    status_dict = self._getJobDetailsFromSacct(job_id_list)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/databiosphere/toil/src/toil/test/batchSystems/test_slurm.py", line 159, in call_sacct_raises
    raise CalledProcessErrorStderr(1, "sacct: error: Problem talking to the database: "
toil.lib.misc.CalledProcessErrorStderr: <super: <class 'CalledProcessErrorStderr'>, <CalledProcessErrorStderr object>>

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/builds/databiosphere/toil/src/toil/test/batchSystems/test_slurm.py", line 389, in test_coalesce_job_exit_codes_sacct_raises_job_exists
    result = self.worker.coalesce_job_exit_codes(job_ids)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/databiosphere/toil/src/toil/batchSystems/slurm.py", line 208, in coalesce_job_exit_codes
    status_dict = self._get_job_details(job_id_list)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/databiosphere/toil/src/toil/batchSystems/slurm.py", line 238, in _get_job_details
    status_dict = self._getJobDetailsFromScontrol(job_id_list)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/databiosphere/toil/src/toil/batchSystems/slurm.py", line 400, in _getJobDetailsFromScontrol
    job[key] += ' ' + bits[0]
    ~~~^^^^^
KeyError: 'NtasksPerTRES:0'
```

I think the problem is https://github.com/DataBiosphere/toil/pull/4951/files#diff-62547683dfb9de5a967fd3260b78612c1e4fb08470399f06c452bd2c690c526eR399

That loop actually does depend on key being set by one line and then used by other lines later; you can't use the same line's info as both key and value like you are doing.

mr-c:
Will this work for CWL as well?

stxue1:
> Will this work for CWL as well?

The Slurm changes should work on the CWL side, as long as

mr-c:
> > Will this work for CWL as well?
> 
> The Slurm changes should work on the CWL side. The original intention of this PR was to remove the need for a user to supply a partition from `TOIL_SLURM_ARGS` if there was a GPU needed job:
> 
> https://github.com/DataBiosphere/toil/blob/92a90ee8e69acd520e7da52ec8a7559b44b7d4c9/src/toil/batchSystems/slurm.py#L470-L476
> 
> This PR allows Toil to switch between the default partition and a gpu partition by itself, as long as no partition is specified.
> 
> So as long as `toil-cwl-runner` has the machinery to handle GPUs, this should work.

Great, thank you!

Yes, the CWL side has been setting the `gpu`  option for a while, if the `CUDARequirement` CWL extension is used

https://github.com/DataBiosphere/toil/blob/92a90ee8e69acd520e7da52ec8a7559b44b7d4c9/src/toil/cwl/cwltoil.py#L2437-L2447

https://doc.arvados.org/user/cwl/cwl-extensions.html#CUDARequirement

I updated the changelog to make this clear

PR #5001, opened 2024-06-29
Support more methods of WDL task disk specification

### LINKED ISSUE #4995: Allow array of disk specifications for WDL
> If an array of disk specifications is used to specify multiple disk mounts, only one of them is allowed to omit the mount point.

https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L4336

We don't currently support this functionality. The [associated unit test](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L4342-L4359) fails with:
```
	[2024-06-27T20:35:26-0700] [MainThread] [C] [toil.worker] Worker crashed with traceback:
	  File "/home/heaucques/Documents/toil/src/toil/worker.py", line 440, in workerScript
	    job._runner(jobGraph=None, jobStore=job_store, fileStore=fileStore, defer=defer)
	  File "/home/heaucques/Documents/toil/src/toil/job.py", line 2984, in _runner
	                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	  File "/home/heaucques/Documents/toil/src/toil/job.py", line 2895, in _run
	    return self.run(fileStore)
	           ^^^^^^^^^^^^^^^^^^^
	  File "/home/heaucques/Documents/toil/src/toil/wdl/wdltoil.py", line 145, in decorated
	    return decoratee(*args, **kwargs)
	           ^^^^^^^^^^^^^^^^^^^^^^^^^^
	  File "/home/heaucques/Documents/toil/src/toil/wdl/wdltoil.py", line 1778, in run
	    all_specs = disks_spec.split(',')
	                ^^^^^^^^^^^^^^^^
	AttributeError: 'list' object has no attribute 'split'
	
	[2024-06-27T20:35:26-0700] [MainThread] [E] [toil.worker] Exiting the worker because of a failed job on host pop-os
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1607)
┆Issue Number: TOIL-1607

stxue1:
Closes #4995 
## Changelog Entry

 * Support WDL disk specification as per 1.1 spec

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/wdl/wdltoil.py
```diff
@@ -1747,40 +1747,71 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
-                # alone so that the workflow doesn't rely on this weird and
-                # likely-to-change Cromwell detail.
-                if spec_parts[2] == 'LOCAL':
+                if part_mount_point != "local-disk":
+                    # Don't mount local-disk. This isn't in the spec, but is carried over from cromwell
+                    mount_spec[part_mount_point] = int(per_part_size)
+
+                if not os.path.exists(part_mount_point):
+                    # this isn't a valid mount point
+                    raise NotImplementedError(f"Cannot use mount point {part_mount_point} as it does not exist")
```
adamnovak, line 1807: This is what the spec says we're supposed to require (the mount point needs to exist on the "host system"), but I've read the disks section of the spec twice and that requirement doesn't really make any sense, because we're mounting storage into the *container*. The spec doesn't say we actually do anything with this path on the host.

src/toil/wdl/wdltoil.py
```diff
@@ -2204,8 +2236,53 @@ def patched_run_invocation(*args: Any, **kwargs: Any) -> List[str]:
...
+                                Mount(
+                                    abs_mount_point.rstrip("/").replace("{{", '{{"{{"}}'),
+                                    abs_mount_point.rstrip("/").replace("{{", '{{"{{"}}'),
+                                    type="bind",
+                                )
+                            )
+                        return mounts
+                    task_container.prepare_mounts = patch_prepare_mounts_docker  # type: ignore[method-assign]
+                except ImportError:
+                    logger.warning("Docker package not installed. Unable to add mount points.")
```
adamnovak, line 2285: I'm not sure it makes sense to add the ability to make these mounts in Toil as a monkey-patch. There's nothing here that wouldn't make just as much sense in MiniWDL (or more, since MiniWDL actually has a shared filesystem as an assumption), so instead of doing monkey-patches we should PR this machinery to MiniWDL.

Or if we're starting to add multiple monkey-patches to the TaskContainers, maybe we really want to extend them instead?

src/toil/wdl/wdltoil.py
```diff
@@ -2204,8 +2236,53 @@ def patched_run_invocation(*args: Any, **kwargs: Any) -> List[str]:
...
+                    The singularity and docker patch are separate as they have different function signatures
+                    """
+                    # todo: support AWS EBS/Kubernetes persistent volumes
+                    # this logic likely only works for local clusters as we don't deal with the size of each mount point
+                    mounts: List[Tuple[str, str, bool]] = singularity_original_prepare_mounts()
+                    # todo: support AWS EBS/Kubernetes persistent volumes
+                    # this logic likely only works for local clusters as we don't deal with the size of each mount point
+                    for mount_point, _ in self._mount_spec.items():
+                        abs_mount_point = os.path.abspath(mount_point)
+                        mounts.append((abs_mount_point, abs_mount_point, True))
```
adamnovak, line 2256: How sure are you that the spec actually means you're supposed to mount this particular host path through to the container?

It does say that the host path is required to exist. But it also says a lot about mounting "persistent volumes" *at* these mount points, and making them the required size. It doesn't seem to say anything about mounting those host paths specifically into the container.

What if you ask for a 100 GiB `/tmp` mount, and `/tmp` exists on the host, but `/tmp` on the host is only 10 GiB and Toil is actually doing all its work on a much larger `/scratch/tmp`? Shouldn't you get a 100 GiB `/tmp` mount in the container that actually refers to some location in `/scratch/tmp` on the host?

If the mount point feature was really meant to mount particular host paths across, wouldn't it take both a host-side path and a container-side path like the actual underlying mount functionality uses?

adamnovak (changes requested):
I think we might need to go back and get clarification from the WDL folks before we can implement this properly.

The spec talks about "persistent volumes", but doesn't really explain what those are or which tasks would expect to be able to read which other tasks' writes. The implementation here doesn't actually provide any kind of persistence that I can see, unless running somewhere where the worker nodes have persistent filesystems.

It's not really clear to me whether we're meant to be mounting arbitrary host paths into the containers, or doing something more like Docker volumes. There's a requirement for the host-side path to exist but no other evidence that the task would be able to expect to actually access anything *at* that host-side path, and the execution engine is somehow responsible for making the volumes be the right size, which is impossible if it is just meant to mount across whatever's already there.

Can we dig up any workflows that genuinely use the mountpoint feature as more than just a test case, to see how they expect it to behave? Can we find or elicit any more explanation from the spec authors as to what the mount point feature is meant to accomplish?

It also might not really make sense to implement this on our own in Toil without some support from MiniWDL. We rely on MiniWDL for ordering up an appropriate container given the runtime spec, and unless we need to hook it into Toil's job requirements logic or make the batch system do special things with batch-system-specific persistent storage, it would be best if we could just get this feature for free when it shows up in MiniWDL.

adamnovak:
@stxue1 Don't we now have the clarification we need so that we can finish this? I think the WDL spec is being revised to make it clearer that the mounts are meant to request so much storage available at such-and-such a path in the container, and that they are not actually meant to mount specific paths into the container. But we still need the changes in this PR adding array-of-mounts support.

src/toil/wdl/wdltoil.py
stxue1: It probably is a good idea to PR this machinery to MiniWDL. The main reason why I had to monkeypatch this in in the first place is because MiniWDL doesn't actually support the `disks` runtime attribute. Instead of PR-ing the code one-to-one, I think a PR that adds functionality to extend the list of mount points for both docker and singularity is best

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
-            disks_spec: str = runtime_bindings.resolve('disks').value
-            all_specs = disks_spec.split(',')
+            disks_spec: Union[List[WDL.Value.String], str] = runtime_bindings.resolve('disks').value
+            if isinstance(disks_spec, list):
+                # SPEC says to use the first one
+                # the parser gives an array of WDL string objects
+                all_specs = [part.value for part in disks_spec]
+            else:
+                all_specs = disks_spec.split(',')
             # Sum up the gigabytes in each disk specification
```
adamnovak, line 1937:
```suggestion
            # Sum up the space in each disk specification
```

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                specified_mount_point = None
+                for i, part in enumerate(spec_parts):
+                    if part.replace(".", "", 1).isdigit():
+                        # round down floats
+                        part_size = int(float(part))
+                        continue
+                    if i == 0:
+                        # mount point is always the first
+                        specified_mount_point = part
+                        continue
```
adamnovak, line 1958: The mount point isn't always the first; if the first item is all numeric it gets interpreted as a size and not a mount point.

I think we get away with this because it's impossible to have an all-numeric mount point that is valid, since the mount point is specified to be an "absolute" path and thus necessarily contains `/`.

But if someone asks for `001 15 GiB` hoping for a directory named `001` somewhere, we are not going to parse that like they meant it. I think we will interpret that as a root volume size of 15 GiB.

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                    if part_size is not None:
+                        # suffix will always be after the size, if it exists
+                        part_suffix = part
+                        continue
```
adamnovak, line 1962: We don't have anything here to prohibit extraneous pieces. We probably should reject anything that follows neither the spec nor Cromwell's convention, because in that case we know we can't do whatever is being asked for.

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                    # Disk spec did not include a size
+                    raise ValueError(f"Could not parse disks = {disks_spec} because {spec} does not specify a disk size")
 
+
+                if part_suffix == "LOCAL":
+                    # TODO: Cromwell rounds LOCAL disks up to the nearest 375 GB. I
+                    # can't imagine that ever being standardized; just leave it
+                    # alone so that the workflow doesn't rely on this weird and
+                    # likely-to-change Cromwell detail.
+                    logger.warning('Not rounding LOCAL disk to the nearest 375 GB; workflow execution will differ from Cromwell!')
```
adamnovak, line 1974: We should also probably switch the default unit to `GB` here, since that is what the Cromwell syntax expects.

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                    # For cromwell compatibility, assume this means GB in units
+                    # We don't actually differentiate between HDD and SSD
+                    part_suffix = "GB"
+
+                per_part_size = convert_units(part_size, part_suffix)
+                total_bytes += per_part_size
+                if specified_mount_point is not None:
+                    if mount_spec.get(specified_mount_point) is not None:
+                        # raise an error as all mount points must be unique
+                        raise ValueError(f"Could not parse disks = {disks_spec} because the mount point {specified_mount_point} is specified multiple times")
```
adamnovak, line 1985: We never actually add `local-disk` in here, so you are going to be allowed to specify it multiple times. Maybe that's fine? It makes sense to sum them.

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                    # alone so that the workflow doesn't rely on this weird and
+                    # likely-to-change Cromwell detail.
+                    logger.warning('Not rounding LOCAL disk to the nearest 375 GB; workflow execution will differ from Cromwell!')
+                elif part_suffix in ("HDD", "SSD"):
+                    # For cromwell compatibility, assume this means GB in units
+                    # We don't actually differentiate between HDD and SSD
+                    part_suffix = "GB"
+
+                per_part_size = convert_units(part_size, part_suffix)
+                total_bytes += per_part_size
```
adamnovak, line 1981: The total space needed (including any `local-disk` or mount-point-less size) doesn't ever get copied to `runtime_disk` and so isn't actually used for scheduling.

src/toil/wdl/wdltoil.py
```diff
@@ -1915,43 +1921,74 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                if specified_mount_point is not None:
```
adamnovak, line 1982: We're not prohibiting multiple mount-point-less specifications, or using e.g. `25 GiB` and `local-disk 300 SSD` together. And we also don't store either of those in the `mount_spec` dict, which means they don't feed into the `df` check later.

src/toil/wdl/wdltoil.py
```diff
@@ -2210,6 +2248,54 @@ def can_mount_proc(self) -> bool:
...
+        mount_src_mapping = {}
+        # Create one tmpdir to encapsulate all mount point sources, each mount point will be associated with a subdirectory
+        tmpdir = file_store.getLocalTempDir()
+
+        # The POSIX standard doesn't specify how to escape spaces in mount points and file system names
+        # The only defect of this regex is if the target mount point is the same format as the df output
+        # It is likely reliable enough to trust the user has not created a mount with a df output-like name
+        regex_df = re.compile(r".+ \d+ +\d+ +(\d+) +\d+% +.+")
+        try:
+            for mount_target, mount_size in mount_spec.items():
```
adamnovak, line 2275: Rather than doing this as a loop over each requested mount point (except the `local-disk`/no-mount-point ones that set the size of the working directory, whicvh never made it into `mount_spec`), and checking for space for each individually, we should sum up all the space needed and check for at least that much total space. We know all the mount points and also the task working directory are going to come from the same underlying filesystem where the Toil work directory is. So we only need to ask about the free space on that filesystem once.

Then we would no longer have the two-100GB-mount-points problem; we would know that we need 200GB total and only have say 150.

We can probably just read the job's disk requirement (I think `self.requirements.disk`) and make sure `df` shows that much space. Then we don't need two copies of the logic to sum up the total of all the mount points.

adamnovak (changes requested):
I think the total space requirement is not actually getting through to the job's disk requirement field. Also, we're only checking for enough space for each mountpoint individually, when we know they're all going to be fulfilled from the same underlying filesystem and we can just check for the total amount of free space.

src/toil/wdl/wdltoil.py
stxue1: I think it's better to keep the default unit to `GiB` as that is the WDL spec default https://github.com/openwdl/wdl/blob/e43e042104b728df1f1ad6e6145945d2b32331a6/SPEC.md?plain=1#L5082

src/toil/wdl/wdltoil.py
stxue1: I added a check to catch `local-disk` being specified multiple times

src/toil/wdl/wdltoil.py
stxue1: There should be a check now, I also store it into the `mount_spec` dict now

src/toil/wdl/wdltoil.py
```diff
@@ -1979,15 +1984,23 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                    if specified_mount_point is not None:
                         # raise an error as all mount points must be unique
                         raise ValueError(f"Could not parse disks = {disks_spec} because the mount point {specified_mount_point} is specified multiple times")
-
-                    # TODO: we always ignore the disk type and assume we have the right one.
-                    if specified_mount_point != "local-disk":
-                        # Don't mount local-disk. This isn't in the spec, but is carried over from cromwell
-                        mount_spec[specified_mount_point] = int(per_part_size)
+                    else:
+                        if mount_spec.get(specified_mount_point) is not None:
```
adamnovak, line 1992: This condition is inside another `if` with the same condition, so it will always be true.

src/toil/wdl/wdltoil.py
```diff
@@ -1979,15 +1984,23 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
 
                 per_part_size = convert_units(part_size, part_suffix)
                 total_bytes += per_part_size
-                if specified_mount_point is not None:
-                    if mount_spec.get(specified_mount_point) is not None:
+                if mount_spec.get(specified_mount_point) is not None:
```
adamnovak, line 1987: This check doesn't account for how `local-disk` and no mount point mean the same thing, so will still let you specify both.

If there was a test for this validation logic, that would catch the problem.

src/toil/wdl/wdltoil.py
```diff
@@ -1942,24 +1942,29 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+                if len(spec_parts) > 3:
+                    raise RuntimeError(f"Could not parse disks = {disks_spec} because {all_specs} contains more than 3 parts")
                 part_size = None
                 # default to GiB as per spec
-                part_suffix: str = "GiB"
+                part_suffix: str = "GiB"  # The WDL spec's default is 1 GiB
                 # default to the execution directory
                 specified_mount_point = None
+                # first get the size, since units should always be some nonnumerical string, get the last numerical value
                 for i, part in enumerate(spec_parts):
```
adamnovak: The loop visits the parts in order, and with the `break` later `part_size` ends up being set to the *first* numerical value.

adamnovak (changes requested):
This looks better but I think there are still ways to break it.

If the parsing was broken out into its own function with tests for its behavior on both allowed and disallowed disk specification sets, it would be easier to convincingly say that the implementation is correct.

src/toil/wdl/wdltoil.py
stxue1: I mapped `local-disk` and no mount point to the same thing in a conditional branch below. I changed the logic a bit so it's more clear and so there's only one conditional branch that does this duplication check.

src/toil/wdl/wdltoil.py
```diff
@@ -2399,6 +2461,58 @@ def can_mount_proc(self) -> bool:
...
+            # If df somehow isn't available
+            logger.debug("Unable to call df. stdout: %s stderr: %s", e.stdout, e.stderr)
+            logger.warning("Unable to check disk requirements as call to 'df' command failed. Will assume storage is always available.")
+        for mount_target in mount_spec.keys():
+            # Create a new subdirectory for each mount point
+            source_location = os.path.join(tmpdir, str(uuid.uuid4()))
+            os.mkdir(source_location)
+            if mount_target is not None:
+                # None represents an omitted mount point, which will default to the task's work directory. MiniWDL's internals will mount the task's work directory by itself
+                mount_src_mapping[mount_target] = source_location
```
adamnovak, line 2513: This makes a directory for omitted mount points even if it never uses them.

We also sort of assume Docker/Singularity is pulling from the same storage we are for `/`, but never really enforce/check that.

adamnovak (approved):
I think this looks good now.

PR #5002, opened 2024-07-01
Sniff filesystem type and try not to use locks on Ceph

### LINKED ISSUE #4972: Toil can hang when something goes wrong with file lockiong on Ceph
As noted in https://ucsc-gi.slack.com/archives/C05ADDE2TSB/p1718655730933009, Toil can get stuck running a WDL workflow, in a state where it claims a job is running but the job cannot progress. This can happen when the lock file used to control the Singularity image cache is on a Ceph distributed filesystem, and Ceph gets into a bad state.

The bad state is going to be difficult to reproduce on demand, and is also not well-characterized. It *might* be detectable as IO errors when going to do lock/unlock operations, as in #4874. It might be that the "transient" IO errors that #4924 tried to tolerate are (sometimes?) permanent.

If there is some kind of permanent problem with Ceph, Toil needs to eventually fail and not hang, because otherwise the user will not know that Ceph is broken.

adamnovak:
This should fix #4972 by throwing errors before we can lock up Ceph. It accomplishes this by refusing to run if it would need to lock things that appear to be on Ceph. When Ceph fixes its deadlock issue we will have to undo this.

This also adds a subprocess call to global mutex lock attempts, since we can't directly call `statfs` from Python. If that's no good, we can just check the Singularity directories when validating command line options, and not when we would go to lock them. (That might also get us better error messages).

I want to test this manually before merging, so it is a draft.

## Changelog Entry

 * Toil will refuse to proceed if it detects that its coordination directory or a Singularity cache directory it needs to lock is on Ceph, to prevent hanging the Ceph MDS

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -1402,10 +1402,14 @@ class Toil(ContextManager["Toil"]):
 
         # Make it exist
         os.makedirs(subdir, exist_ok=True)
-        # TODO: May interfere with workflow directory creation logging if it's the same directory.
+        # TODO: May interfere with workflow directory creation logging if it's
+        # the same directory.
 
         # Don't let it out if it smells like an unacceptable filesystem for locks
-        ensure_filesystem_lockable(subdir)
+        ensure_filesystem_lockable(
+            subdir,
+            hint="Use --coordinationDir to provide a different location."
+        )
 
         # Return it
         return subdir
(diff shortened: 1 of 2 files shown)


adamnovak:
I made this from the wrong repo so I'm closing it in favor of #5023.

PR #5020, opened 2024-07-15
Add s3 head object function.

DailyDreaming:
We use this functionality already, this just centralizes it into its own function in an s3 lib.

TODO: What is the "Right Way" to invoke boto3 sessions with proper credential caching?

## Changelog Entry

 * Add s3 head object function.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak:
@DailyDreaming The right way to get a global boto3 *session* is not bound out; we only have module-level [client](https://github.com/DataBiosphere/toil/blob/1b4af5442be30e6ac2e3b0b16035371fafe31fb1/src/toil/lib/aws/session.py#L328) and [resource](https://github.com/DataBiosphere/toil/blob/1b4af5442be30e6ac2e3b0b16035371fafe31fb1/src/toil/lib/aws/session.py#L357) getters. So right now you can either hold your own `AWSConnectionManager` for its `session()` function, or you could add something to the `session.py` module to get at the global AWSConnectionManager's `session()`.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1418,15 +1419,13 @@ def readFrom(self, readable):
...
-                                                       str(version))
-                                        info.version = version
-                                        assert info.version is not None
+                                info.version = head_s3_object(
+                                    Bucket=bucket_name,
+                                    Key=compat_bytes(info.fileID),
+                                    **headerArgs
+                                ).get('VersionId', None)
+                                logger.warning('Loaded key for upload with no version and got version %s', str(info.version))
+                                assert info.version is not None
```
adamnovak, line 1428: We probably want to cut these; the warning shouldn't be a warning unless we got the wrong version result, and the assert probably should be a real `raise` if we actually want to stop if it isn't true in a live workflow.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1467,23 +1466,21 @@ def readFrom(self, readable):
...
-                                with attempt:
-                                    headObj = client.head_object(Bucket=bucket_name,
-                                                                 Key=compat_bytes(info.fileID),
-                                                                 **headerArgs)
-                                    info.version = headObj.get('VersionId', None)
-                                    logger.warning('Reloaded key with no version and got version %s', str(info.version))
-                                    assert info.version is not None
+                            resp = head_s3_object(Bucket=bucket_name, Key=compat_bytes(info.fileID), header=headerArgs)
+                            info.version = resp.get('VersionId', None)
+                            logger.warning('Reloaded key with no version and got version %s', str(info.version))
```
adamnovak, line 1482: Here's another warning that probably should be either lower-level or conditional.

src/toil/lib/aws/s3.py
```diff
@@ -0,0 +1,25 @@
...
+import logging
+from typing import Dict, Any, Optional
+from toil.lib.aws import session, AWSServerErrors
+from toil.lib.retry import retry
+
+logger = logging.getLogger(__name__)
+
+
+@retry(errors=[AWSServerErrors])
+def head_s3_object(bucket: str, key: str, header: Dict[str, Any], region: Optional[str] = None):
```
adamnovak, line 23: This should have a docstring. It might have to link to the boto3 docs for all the header values you can have. It's *not* just a dict from HTTP header name to value, right?

src/toil/lib/aws/s3.py
```diff
@@ -0,0 +1,25 @@
...
+    s3_client = session.client("s3", region_name=region)
```
adamnovak, line 24: This is probably a fine way to get ahold of boto3 things.

adamnovak (changes requested):
This looks good in general, but it needs a docstring and I think we need to quiet down some warnings it adds.

src/toil/jobStores/aws/jobStore.py
DailyDreaming: I agree.  This was code unchanged from the old function, and I was trying to make minimal changes.  I'll make that change though, I agree it's not necessary (and would like to get rid of versioned objects altogether in the future).

src/toil/lib/aws/s3.py
DailyDreaming: Good point!

src/toil/jobStores/aws/jobStore.py
DailyDreaming: This was code I copy-pasted over in the refactor, but I will make the changes as I agree that that makes sense.

adamnovak (approved):
This looks OK to me now.

adamnovak:
This is subsumed into #5123.

Issue #4992, opened 2024-06-27
toil-wdl-runner does not handle string to file coercion properly for optional file types


stxue1:
```wdl
version 1.1
workflow testWorkflow {
  input {
  }
  call testTask
  output {
    Array[File?] array_in_output = testTask.array_in_output
    Int len_in_output = testTask.len_in_output
    Array[File?] array_in_body_out = testTask.array_in_body_out
    Int len_in_body_out = testTask.len_in_body_out
    Array[File?] array_in_input_out = testTask.array_in_input_out
    Int len_in_input_out = testTask.len_in_input_out
  }
}

task testTask {
  input {
    Array[File?] array_in_input = ["example1.txt", "example2.txt"]
    Int len_in_input = length(select_all(array_in_input))
  }
  command <<<>>>
  Array[File?] array_in_body = ["example1.txt", "example2.txt"]
  Int len_in_body = length(select_all(array_in_body))
  output {
    Array[File?] array_in_output = ["example1.txt", "example2.txt"]
    Int len_in_output = length(select_all(array_in_output))
    Array[File?] array_in_body_out = array_in_body
    Int len_in_body_out = len_in_body
    Array[File?] array_in_input_out = array_in_input
    Int len_in_input_out = len_in_input
  }
}
```
```
Workflow Progress 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 (2 failures) [00:06<00:00, 0.31 jobs/s]
Traceback (most recent call last):
  File "/home/heaucques/Documents/toil/venv3.12/bin/toil-wdl-runner", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/heaucques/Documents/toil/src/toil/wdl/wdltoil.py", line 143, in decorated
    return decoratee(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/heaucques/Documents/toil/src/toil/wdl/wdltoil.py", line 3401, in main
    output_bindings = toil.start(root_job)
                      ^^^^^^^^^^^^^^^^^^^^
  File "/home/heaucques/Documents/toil/src/toil/common.py", line 940, in start
    return self._runMainLoop(rootJobDescription)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/heaucques/Documents/toil/src/toil/common.py", line 1427, in _runMainLoop
[...]

Issue #5004, opened 2024-07-02
Support filename coercion back to string


stxue1:
The spec expects that filenames can be represented as a string to alter things like filename extensions:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L6521

We virtualize files into our own representation, so coercing back will give us a toilfile URI instead of the filename:
```
data_file = "toilfile:4%3A0%3Afiles%2Ffor-job%2Fkind-WDLTaskJob%2Finstance-kfz8ammc%2Ffile-c91abfb5298d4a83bad4bb00d53beb55%2Ffoo.data/e946ed9a-784f-47d3-bf74-e84a23d9056d/foo.data"
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1609)
┆Issue Number: TOIL-1609

adamnovak:
Yeah, this is going to be tricky to implement. It looks like at least in task output sections they expect to be dealing with the same string that got coerced to a `File`, rather than something processed.

How will we make this work at workflow scope? We might have one decl that turns a `String` into a `File`, then several decls and tasks that depend on that `File` and need to use its virtualized representation, and then another decl that expects to operate on the `File` coerced back to the same original string, possibly in the same expression as working with the `File`'s contents.

We could do something like moving virtualization to the boundaries of tasks, so we virtualize a version of the `File` that we use to pass to tasks, and then we leave a non-virtualized version around for use in decls?

Or we could tack the original non-virtualized path of the `File` onto it with `setattr()` so we can get at it later for string coercion.

Issue #4988, opened 2024-06-24
Defer file virtualization until use for WDL String-to-File coercion


stxue1:
This is for behavioral parity with miniwdl.

Related to https://github.com/openwdl/wdl/issues/667

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1601)
┆Issue Number: TOIL-1601

PR #5028, opened 2024-07-19
Move file virtualization in toil-wdl-runner to task boundaries

### LINKED ISSUE #5031: Symlink pasthrough on WDL is broken again
I think as of #4994, and somewhere in the diff between b3a016c3fd03ee23712145c4b381d34db1f0d407 and fb3b6304932f82df272bb0fc315235bf9e836eac, we added a regression and #4850 came back and WDL tasks can no longer output symlinks to their inputs. Test 72 our conformance tests, which tests this, started failing again, but we didn't notice in CI because the test was never marked passing.

We should fix the symlink passthrough again.

We should also maybe set up CI so that if we have a WDL conformance test marked as failing, it's an error if it *doesn't* fail, to ensure we keep the list up to date when we make fixes.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1620)
┆Issue Number: TOIL-1620

stxue1:
This should solve #5004

This also reverts #4994 and introduces another fix for #4988; there should be behavioral parity with miniwdl

This should also resolve #5031

This is a bit of an overhaul of how we handle files in toil-wdl-runner; instead of immediately virtualizing upon seeing a `File` type (coerced or not), the WDL value is kept as a path until the last second. Only before tasks are sent off, the files will be virtualized. This way, the WDL side functionality should still be the same while allowing for File to String coercion as we no longer replace all File paths with our virtualized representation immediately.

Monkeypatching is removed as we no longer virtualize at the coercion step, but before/after function calls and manually at task boundaries.

I don't think this will increase the amount of IO. One edge case is if the user changes the filename right before being put into a function; toil will virtualize the new filename for that function call and keep both the original and new virtualized file around in the jobstore until the job is completed. It may be worth removing the new virtualized file from the jobstore after use, but I'm not sure where to hook in the behavior.

~~The downside is that all jobs that are marked as local must always be local in the future. I explicitly set the flag in the class constructors where I could.~~ I think the virtualization behavior should ensure local jobs on workers should work.

The function `evaluate_bindings_from_decls` is now used for all decl parsing/evaluation to reduce duplicative code.

## Changelog Entry

 * File virtualization in toil-wdl-runner now only happens at task boundaries
   * File to String coercion should be supported

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/wdl/wdltoil.py
```diff
@@ -1316,95 +1318,6 @@ def add_paths(task_container: TaskContainer, host_paths: Iterable[str]) -> None:
             task_container.input_path_map[host_path] = container_path
             task_container.input_path_map_rev[container_path] = host_path
 
-def import_files(environment: WDLBindings, toil: Toil, path: Optional[List[str]] = None, skip_remote: bool = False) -> WDLBindings:
```
stxue1, line 1319: I don't think this function is needed? At first glance, our virtualization step should do either the same thing or at least something similar.

stxue1: This function actually is useful as it can import files from URLs given a relative path

src/toil/wdl/wdltoil.py
```diff
@@ -911,14 +994,24 @@ def devirtualize_to(filename: str, dest_dir: str, file_source: Union[AbstractFil
...
-            # This is a local file
-            # To support relative paths, join the execution dir and filename
-            # if filename is already an abs path, join() will do nothing
-            if execution_dir is not None:
-                result = os.path.join(execution_dir, filename)
-            else:
-                result = filename
-            logger.debug("Virtualized file %s is already a local path", filename)
+            result = None
+            if wdl_options.get("filename_to_uri") is not None:
```
stxue1, line 998: I think Files coerced from strings will still see these imported paths from the JSON URL. I'm currently not too sure how to avoid that with this implementation; I went with this implementation originally as I *think* this is the simplest solution overall that avoids "hacking" around miniwdl's internals (such as abusing setattr).

src/toil/wdl/wdltoil.py
```diff
@@ -624,26 +610,37 @@ def unpack_toil_uri(toil_uri: str) -> Tuple[FileID, str, str]:
 
     return file_id, parent_id, file_basename
 
-def evaluate_output_decls(output_decls: List[WDL.Tree.Decl], all_bindings: WDL.Env.Bindings[WDL.Value.Base], standard_library: ToilWDLStdLibBase) -> WDL.Env.Bindings[WDL.Value.Base]:
+def evaluate_bindings_from_decls(decls: List[WDL.Tree.Decl], all_bindings: WDL.Env.Bindings[WDL.Value.Base], standard_library: ToilWDLStdLibBase,
```
adamnovak, line 613: Maybe this should be `evaluate_decls_to_bindings`, or `generate_bindings_from_decls`, or even just `evaluate_decls`? I think the decls are what gets evaluated, not the bindings.

src/toil/wdl/wdltoil.py
```diff
@@ -624,26 +610,37 @@ def unpack_toil_uri(toil_uri: str) -> Tuple[FileID, str, str]:
...
-        output_value = evaluate_decl(output_decl, all_bindings, standard_library)
-        drop_if_missing_with_workdir = partial(drop_if_missing, work_dir=standard_library.execution_dir)
-        output_value = map_over_typed_files_in_value(output_value, drop_if_missing_with_workdir)
-        all_bindings = all_bindings.bind(output_decl.name, output_value)
-        output_bindings = output_bindings.bind(output_decl.name, output_value)
-    return output_bindings
+    :param include_previous: Whether to include the existing environment in the new returned environment. This will be false for outputs where only defined decls should be included
+    :param drop_missing_files: Whether to coerce nonexistent files to null. The coerced elements will be checked that the transformation is valid.
+    Currently should only be enabled in output sections, see https://github.com/openwdl/wdl/issues/673#issuecomment-2248828116
+    :return: New bindings object with only the decls
```
adamnovak, line 625: Unless `include_previous` is true in which case the existing bindings are also carried through, right?

src/toil/wdl/wdltoil.py
```diff
@@ -708,6 +705,94 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
         return False
 
+def import_url_files(environment: WDLBindings, file_source: Union[AbstractFileStore, Toil], path: Optional[List[str]] = None, include_self: bool = False) -> Optional[Dict[str, str]]:
+    """
+    Iterate through the environment and import all files from the possible URLs. This will not import if the file is detected as local.
+    Returns a mapping of the original file names to virtualized URIs
+    :param environment: Bindings to evaluate on
+    :param file_source: Context to upload/virtualize files with
+    :param path: List of paths to search through
+    :return:
```
adamnovak, line 715: What does `include_self` do?

src/toil/wdl/wdltoil.py
```diff
@@ -708,6 +705,94 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
                 return True
         return False
 
+def import_url_files(environment: WDLBindings, file_source: Union[AbstractFileStore, Toil], path: Optional[List[str]] = None, include_self: bool = False) -> Optional[Dict[str, str]]:
+    """
+    Iterate through the environment and import all files from the possible URLs. This will not import if the file is detected as local.
+    Returns a mapping of the original file names to virtualized URIs
+    :param environment: Bindings to evaluate on
+    :param file_source: Context to upload/virtualize files with
+    :param path: List of paths to search through
```
adamnovak, line 714: These are for resolving relative paths, and we try them in order?

Also there are multiple things in here, so maybe it should be `paths`? Or `search_paths`?

src/toil/wdl/wdltoil.py
```diff
@@ -708,6 +705,94 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
+        Import a file if from a URL and return a virtualized filename for it.
+        """
+        # Search through any input search paths passed in and download it if found
+        tried = []
+        for candidate_uri in potential_absolute_uris(filename, path if path is not None else []):
+            if not include_self and candidate_uri == filename:
+                # If the candidate uri is unchanged, as in is equal to the original uri
+                # we can discard it if unwanted; for example, miniwdl's backend can handle URLs fine,
+                # and carrying through the url through the workflow won't actually hit the imported file
+                # without more machinery
```
adamnovak, line 730: I don't understand this comment. What does it mean to "hit" the file? What does it mean for the file to be "unwanted"?

src/toil/wdl/wdltoil.py
```diff
@@ -708,6 +705,94 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
                 return True
         return False
 
+def import_url_files(environment: WDLBindings, file_source: Union[AbstractFileStore, Toil], path: Optional[List[str]] = None, include_self: bool = False) -> Optional[Dict[str, str]]:
+    """
+    Iterate through the environment and import all files from the possible URLs. This will not import if the file is detected as local.
```
adamnovak, line 710: I think maybe the concept we want here is "remote files" vs. "local files" and not "URL files" vs. "local files". A `file://` URI is a URI, and unless we think it's not *actually* a "URL" (due to not really locating it on the Internet) and we want to rely on people knowing that, I don't think we want to talk about "URL files" as a category that doesn't include `file://`.

We also might end up having to deal with something like `http://localhost:8000/file.txt`, which is a URL but which is also *really* a local thing we're going to want to be fetching only on the leader.

src/toil/wdl/wdltoil.py
```diff
@@ -767,13 +852,13 @@ def __init__(self, file_store: AbstractFileStore, execution_dir: Optional[str] =
...
-        self._execution_dir = execution_dir
-
-        self._enforce_existence = enforce_existence
+        self._wdl_options = wdl_options
 
     @property
     def execution_dir(self) -> Optional[str]:
-        return self._execution_dir
+        wdl_options = self._wdl_options or {}
+        execution_dir: Optional[str] = wdl_options.get("execution_dir")
```
adamnovak, line 860: Maybe instead of just a dict and a convention we want this to be a TypedDict or even a class? Then MyPy will catch if we e.g. misspell `execution_dir` somewhere, and we can go to they type definition to see all the Toil WDL interpreter options we support and what they mean.

src/toil/wdl/wdltoil.py
```diff
@@ -806,15 +891,69 @@ def _devirtualize_filename(self, filename: str) -> str:
...
+
+        result = self.devirtualize_to(filename, self._file_store.localTempDir, self._file_store, self._wdl_options,
+                                      self._devirtualized_to_virtualized, self._virtualized_to_devirtualized)
         return result
 
     @staticmethod
-    def devirtualize_to(filename: str, dest_dir: str, file_source: Union[AbstractFileStore, Toil], execution_dir: Optional[str],
-                        devirtualized_to_virtualized: Optional[Dict[str, str]] = None, virtualized_to_devirtualized: Optional[Dict[str, str]] = None,
-                        enforce_existence: bool = True) -> str:
+    def _devirtualize_uri(filename: str, dest_dir: str, file_source: Union[AbstractFileStore, Toil]) -> str:
```
adamnovak, line 900: This probably should have a docstring.

src/toil/wdl/wdltoil.py
adamnovak, line 715: Also this empty `:return:` shouldn't be here.

src/toil/wdl/wdltoil.py
adamnovak: I don't understand what it means for a File instance to "see" an imported path that came from the input JSON file's URL.

src/toil/wdl/wdltoil.py
```diff
@@ -911,14 +995,24 @@ def devirtualize_to(filename: str, dest_dir: str, file_source: Union[AbstractFil
...
-            if execution_dir is not None:
-                result = os.path.join(execution_dir, filename)
-            else:
-                result = filename
-            logger.debug("Virtualized file %s is already a local path", filename)
+            result = None
+            if wdl_options.get("filename_to_uri") is not None:
+                # We support importing files relative to a URL when the input JSON itself is downloaded
+                # See import_url_files()
+                filename_uri = wdl_options["filename_to_uri"].get(filename)
```
adamnovak, line 1002: I think there's a missing piece of the explanation here. *Because* we can import workflow inputs by relative URL, we end up with those relative URLs stored somewhere (where they can become `filename` here I guess?), so a filename can *look* like it is a local file but *actually* be meant to be a URL, so we have to go get the URL for it to actually use?

But then what if `inputs.json` has `"WorkflowName.the_file": "file.txt"`, and comes from a URL where we want to fetch the adjacent "file.txt", but then the WDL itself has `File the_other_file = "file.txt"` at workflow scope, which we would expect to read from the current working directory. Won't we not be able to keep them straight?

And since we pass the one `wdl_options` through the whole workflow, won't this case get hit inside a task output section when we want an output that happens to be named "file.txt"?

src/toil/wdl/wdltoil.py
```diff
@@ -911,14 +995,24 @@ def devirtualize_to(filename: str, dest_dir: str, file_source: Union[AbstractFil
...
+                if filename_uri is not None:
+                    # This was previously virtualized, so use the virtualized result
+                    result = ToilWDLStdLibBase._devirtualize_uri(filename_uri, dest_dir, file_source)
```
adamnovak, line 1005: I'm not sure this comment describes what is happening here.

src/toil/wdl/wdltoil.py
```diff
@@ -1578,6 +1587,46 @@ def map_over_typed_files_in_value(value: WDL.Value.Base, transform: Callable[[WD
...
         return value
 
+def map_over_files_in_value_check_null_type(value: WDL.Value.Base, original_value: WDL.Value.Base, expected_type: WDL.Type.Base) -> None:
+    """
+    Run through all nested values embedded in the given value and check that the null values are valid.
+
+    If a null value is found that does not have a valid corresponding expected_type, raise an error
+
+    (This is currently only used to check that coerced File types to null are valid, as in null elements have an equivalent File? type,
+    if this is to be used elsewhere, the error message should be changed to be implementation specific)
```
adamnovak, line 1597:
```suggestion
    (This is currently only used to check that null values arising from File coercion are in locations with a nullable File? type.
    If this is to be used elsewhere, the error message should be changed to describe the appropriate types and not just talk about files.)
```

But if the function is definitely described as being about files specifically, maybe we can cut the explanation of how we might eventually generalize it to other types from the docstring?

src/toil/wdl/wdltoil.py
```diff
@@ -1578,6 +1587,46 @@ def map_over_typed_files_in_value(value: WDL.Value.Base, transform: Callable[[WD
...
+
+    If one of the nested values is null but the equivalent nested expected_type is not optional, this is not allowed
```
adamnovak, line 1599: This is the condition we're checking for, right? This makes it sound like we're not meant to call the function in that case.

src/toil/wdl/wdltoil.py
```diff
@@ -1578,6 +1587,46 @@ def map_over_typed_files_in_value(value: WDL.Value.Base, transform: Callable[[WD
...
+    :return:
+    """
+    if isinstance(value, WDL.Value.File):
+        pass
+    elif isinstance(value, WDL.Value.Array) and isinstance(expected_type, WDL.Type.Array):
+        for elem, orig_elem in zip(value.value, original_value.value):
+            map_over_files_in_value_check_null_type(elem, orig_elem, expected_type.item_type)
+    elif isinstance(value, WDL.Value.Map) and isinstance(expected_type, WDL.Type.Map):
+        for pair, orig_pair in zip(value.value, original_value.value):
+            # The key of the map cannot be optional or else it is not serializable, so we only need to check the value
```
adamnovak, line 1612: Don't we need to check the key in case it is typed as a non-optional File but actually has a null value? Or is File not allowed as a key type?

src/toil/wdl/wdltoil.py
```diff
@@ -1578,6 +1587,46 @@ def map_over_typed_files_in_value(value: WDL.Value.Base, transform: Callable[[WD
         # All other kinds of value can be passed through unmodified.
         return value
 
+def map_over_files_in_value_check_null_type(value: WDL.Value.Base, original_value: WDL.Value.Base, expected_type: WDL.Type.Base) -> None:
```
adamnovak, line 1590: We maybe should cut `map_over_files_in_value` out of the name here, since this doesn't take a function to map. Maybe we want to call it something like `ensure_null_files_are_nullable` or something.

src/toil/wdl/wdltoil.py
```diff
@@ -1578,6 +1587,46 @@ def map_over_typed_files_in_value(value: WDL.Value.Base, transform: Callable[[WD
...
+        map_over_files_in_value_check_null_type(value.value[1], original_value.value[1], expected_type.right_type)
+    elif isinstance(value, WDL.Value.Struct) and isinstance(expected_type, WDL.Type.StructInstance):
+        for (k, v), (_, orig_v) in zip(value.value.items(), original_value.value.items()):
+            # The parameters method for WDL.Type.StructInstance returns the values rather than the dictionary
+            # While dictionaries are ordered, this should be more robust; the else branch should never be hit
+            if expected_type.members is not None:
+                map_over_files_in_value_check_null_type(v, orig_v, expected_type.members[k])
+    elif isinstance(value, WDL.Value.Null):
+        if not expected_type.optional:
+            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), original_value.value)
```
adamnovak, line 1625: Having one version of this traversal logic is what the map function was supposed to be for. But I guess here we're traversing two mirrored structures in parallel so we can't use the other code.

adamnovak (changes requested):
I'm concerned that by leaving relative URLs from the input JSON as *relative* URLs in WDL `File`s and then trying to figure out where they were meant to be relative to later, we've introduced a whole universe of cross-talk that we don't want to deal with. Maybe we can poll for where they ought to come from and turn them into absolute URLs before handing them to the workflow?

I like that we're ripping out the coercion monkey-patch and also `nonexistent:`, which simplifies things. But are we sure that we retain support for using Toil's supported URI schemes when MiniWDL reads a file? `File the_file = "gs://google-bucket-name/filename.txt"` ought to work if Toil has Google Storage support installed, even if MiniWDL doesn't know about reading Google Storage URLs. I think the coercion hook was doing that for us and I can't really tell if something is taking its place. If we're losing that feature and we think it is worth it, we should know we're losing it.

I'd also like to see a new comment at the top describing how the system works now, and what invariants or abstractions it imposes that new code is going to need to follow to keep it working. `File` holds a URL or leader-local path at the workflow level. When we actually read it with `read_lines()` at the workflow level, does MiniWDL fetch it? If we `read_lines()` it twice, do we have a cache? If we load it into the file store when it enters a task, and it is passed out of the task unmodified, do we see the original URL in the task output (because *everything* at workflow level is *always* an original URL) or the imported file store one?

Is talking about "virtualized" (visible to WDL code) and "devirtualized" (visible to Python `open()`) file names still the right way to understand what the code is doing? Or should we be using different terms or concepts to think about the packing and unpacking that happens when taking `Bindings` in and out of tasks?

I also like the idea of `wdl_options` to hold the global settings for the run, kind of like the CWL `RuntimeContext`. But I think it needs a type so we can have a better handle on what's allowed to be in there.

adamnovak:
@stxue1 Another thing I thought of is, if we need to make `local=True` *always* run a job on the leader and *never* run the job elsewhere, as part of this, then we need to remove the command lien option that lets you run local jobs on workers. Otherwise using it will break WDL workflows.

We also might need to revise the batch system documentation, and maybe the inheritance hierarchy, since you would need *all* batch systems, even those coming from plugins, to have the local job logic.

src/toil/wdl/wdltoil.py
stxue1: I don't think an optional file is allowed in a Map type because that has to be serializable to JSON. For example, MiniWDL complains:
```
(test.wdl Ln 6 Col 3) Map[File?,Int] may not be used in input/output a because the keys cannot be written to/from JSON
      Map[File?, Int] a = {"path/to/file": 1}
```

adamnovak:
We think this will also fix #4992.

src/toil/wdl/wdltoil.py
stxue1: I changed the implementation to use setattr so the above comment doesn't really apply anymore. A file isn't going to "see" an imported path anymore as I poll for the remote paths at startup.

src/toil/wdl/wdltoil.py
stxue1: I got rid of `include_self` for a different implementation; now the remote files will be polled for at startup

src/toil/wdl/wdltoil.py
stxue1: Changed to search_paths. The order is determined in the `search_paths` argument. The function is basically an adapted version of the previous `import_files`

src/toil/wdl/wdltoil.py
stxue1: I changed the implementation so this functionality doesn't exist anymore

src/toil/wdl/wdltoil.py
stxue1: Fixed with a new implementation

src/toil/wdl/wdltoil.py
stxue1: Removed

src/toil/wdl/wdltoil.py
stxue1: Made more clear

stxue1:
> @stxue1 Another thing I thought of is, if we need to make `local=True` _always_ run a job on the leader and _never_ run the job elsewhere, as part of this, then we need to remove the command lien option that lets you run local jobs on workers. Otherwise using it will break WDL workflows.

We seem to have a commandline options for CWL `runCwlInternalJobsOnWorkers`, but there doesn't seem to be a WDL equivalent option.

src/toil/wdl/wdltoil.py
```diff
@@ -89,6 +98,14 @@
...
 logger = logging.getLogger(__name__)
 
+# WDL options to pass into the WDL jobs and standard libraries
+#   task_path: Dotted WDL name of the part of the workflow this library is working for.
+#   execution_dir: Directory to use as the working directory for workflow code.
+#   enforce_existence: If true, then if a file is detected as nonexistent, raise an error. Else, let it pass through
+#   share_files_with: If set to an existing standard library instance, use the same file upload and download paths as it.
+WDL_OPTIONS = TypedDict('WDL_OPTIONS', {"execution_dir": NotRequired[str], "container": NotRequired[str],
+                                        "share_files_with": NotRequired["ToilWDLStdLibBase"], "task_path": str})
+
```
adamnovak, line 108: I don't think the design here is right.

If we want this to be "options", it should contain *only* things that are global constants across the workflow execution, which here would be the `execution_dir` (leader working directory) and the `container` (which container runner to use).

I think it does make sense to have it also carry `task_path` so we don't need to keep passing the task path around through everything and manually giving it to the standard library. But then we should probably call this a `wdl_context` rather than `wdl_options`, and it would have a tree structure: you derive a child context that is more specific from a parent context that is more general, and you use it for part of the workflow tree, mirroring the dotted task name structure. Also in that case we would also want to carry `namespace` alongside `task_path`.

I think it makes no sense at all to have either `enforce_existence` or `share_files_with` as part of this. They don't apply to a whole workflow or to a subtree of a workflow; if you actually passed one of them into a job from above there'd be no way to sensibly respect it. They should go back to being separate arguments on the standard library/other file access functions.

If we make those changes, then we can have a nice controlled way in which we work with the options/context information. We always either pass it straight through, or make a copy of it and adjust it to make the `task_path`/`namespace` more specific. We only ever create it once, on the leader. Then we never have to worry about if we are looking at the real options/context passed down from the top or one that was made as a dict literal in a function call.

adamnovak (changes requested):
I think the WDL options setup is not designed quite right, and it also looks like we're able to import but then forget about whole copies of files.

src/toil/wdl/wdltoil.py
```diff
@@ -772,6 +807,115 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
+    Iterate through the environment and convert all files that reference a remote file from the possible URIs to that URI.
+    If the file references a local file, leave it unchanged.
+    :param environment: Bindings to evaluate on
+    :param file_source: Context to search for remote files with
+    :param task_path: Dotted WDL name of the user-level code doing the
+           importing (probably the workflow name).
+    :param search_paths: If set, try resolving input location relative to the URLs or
+           directories in this list.
+    :param skip_remote: If set, don't try to import files from remote
+           locations. Leave them as URIs.
```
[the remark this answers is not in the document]
stxue1: I changed the argument to better reflect what it does. I called it convert since it not only downloads/imports the files, but sets the value in the WDL file value to the new URI.

src/toil/wdl/wdltoil.py
```diff
@@ -772,6 +807,115 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
+
+            if file_basename == "":
+                # We can't have files with no basename because we need to
+                # download them at that basename later.
+                raise RuntimeError(f"File {candidate_uri} has no basename and so cannot be a WDL File")
+
+            if candidate_uri.startswith("file:") or filename == candidate_uri:
+                # Don't replace if the original file was already found
+                # or if the replacement value is the same as the original
+                return file
```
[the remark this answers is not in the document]
stxue1: I removed this, I think it was leftover from a previous implementation

src/toil/wdl/wdltoil.py
```diff
@@ -860,8 +987,19 @@ def __init__(
...
-    def execution_dir(self) -> str:
-        return self._execution_dir
+    def execution_dir(self) -> Optional[str]:
+        execution_dir: Optional[str] = self._wdl_options.get("execution_dir")
+        return execution_dir
+
+    @property
+    def share_files_with(self) -> Optional["ToilWDLStdLibBase"]:
+        share_files_with: Optional["ToilWDLStdLibBase"] = self._wdl_options.get("share_files_with")
+        return share_files_with
```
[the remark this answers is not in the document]
stxue1: Probably not, removed

src/toil/wdl/wdltoil.py
```diff
@@ -870,36 +1008,160 @@ def get_local_paths(self) -> List[str]:
...
+        "generate write_* function implementation based on serialize"
+
+        def _f(
+            v: WDL.Value.Base,
+        ) -> WDL.Value.File:
+            os.makedirs(self._write_dir, exist_ok=True)
+            with tempfile.NamedTemporaryFile(dir=self._write_dir, delete=False) as outfile:
+                serialize(v, outfile)
+                filename = outfile.name
+            from WDL._util import chmod_R_plus
```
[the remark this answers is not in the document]
stxue1: Yeah, I moved it

src/toil/wdl/wdltoil.py
```diff
@@ -89,6 +98,14 @@
 
 logger = logging.getLogger(__name__)
 
+# WDL options to pass into the WDL jobs and standard libraries
+#   task_path: Dotted WDL name of the part of the workflow this library is working for.
+#   execution_dir: Directory to use as the working directory for workflow code.
+#   enforce_existence: If true, then if a file is detected as nonexistent, raise an error. Else, let it pass through
```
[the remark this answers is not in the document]
stxue1: Removed

src/toil/wdl/wdltoil.py
```diff
@@ -916,9 +1178,8 @@ def devirtualize_to(
         should not be added to the cache
 
         :param state: State dict which must be shared among successive calls into a dest_dir.
-        :param enforce_existence: Raise an error if the file is nonexistent. Else, let it pass through.
+        :param wdl_options: WDL options to carry through. If enforce_nonexistent is set to true inside, will raise an error if the file is nonexistent. Else, let it pass through.
```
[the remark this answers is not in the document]
stxue1: Removed, I think I forgot to move enforce_existence into the WDL options, but I think it's best to keep them separate

src/toil/wdl/wdltoil.py
```diff
@@ -1015,39 +1226,72 @@ def _virtualize_filename(self, filename: str) -> str:
...
+            if filename in self._devirtualized_to_virtualized:
+                # Note: this is a little duplicative with the local file path branch, but the keys are different
+                result = self._devirtualized_to_virtualized[filename]
+                logger.debug("Re-using virtualized WDL file %s for %s", result, filename)
+                return result
+            try:
+                imported = self._file_store.import_file(filename)
+            except FileNotFoundError:
+                # todo: raising exceptions inside of this function will be captured in StdLib.StaticFunction._call_eage which always raises an EvalError.
+                #  Ideally, the error raised here should escape to be captured in the wdl runner main loop, but I can't figure out how
```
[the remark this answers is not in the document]
stxue1: Yeah, I'll do that

src/toil/wdl/wdltoil.py
```diff
@@ -1015,39 +1226,72 @@ def _virtualize_filename(self, filename: str) -> str:
...
+                raise RuntimeError("Failed to import URL %s into jobstore." % filename)
+            file_basename = os.path.basename(urlsplit(filename).path)
+            # Get the URL to the parent directory and use that.
+            parent_dir = urljoin(filename, ".")
+            # Pack a UUID of the parent directory
+            dir_id = self._parent_dir_to_ids.setdefault(parent_dir, uuid.uuid4())
+            result = pack_toil_uri(imported, self.task_path, dir_id, file_basename)
+            logger.debug('Virtualized %s as WDL file %s', filename, result)
+            # We can't the Toil URI in the cache because it would point to the URL instead of a file
+            # which we don't have as we haven't written the contents of the URL into a file yet
```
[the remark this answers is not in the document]
stxue1: I meant a local file instead of file in the filestore. I changed the comment to be more clear

src/toil/wdl/wdltoil.py
```diff
@@ -1058,15 +1302,40 @@ class ToilWDLStdLibTaskCommand(ToilWDLStdLibBase):
...
+            os.makedirs(self._write_dir, exist_ok=True)
+            with tempfile.NamedTemporaryFile(dir=self._write_dir, delete=False) as outfile:
+                serialize(v, outfile)
+                filename = outfile.name
+            from WDL._util import chmod_R_plus
+            chmod_R_plus(filename, file_bits=0o660)
+            vfn = self._virtualize_filename(filename)
+            return WDL.Value.File(vfn)
+
+        return _f
```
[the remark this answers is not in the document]
stxue1: I added a comment explaining the rationale

src/toil/wdl/wdltoil.py
```diff
@@ -1015,39 +1226,72 @@ def _virtualize_filename(self, filename: str) -> str:
         File value
```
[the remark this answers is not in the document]
stxue1: Added

src/toil/wdl/wdltoil.py
```diff
@@ -2221,7 +2412,7 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
         # Set up the WDL standard library
         # UUID to use for virtualizing files
         # We process nonexistent files in WDLTaskWrapperJob as those must be run locally, so don't try to devirtualize them
-        standard_library = ToilWDLStdLibBase(file_store, self._task_path, enforce_existence=False)
+        standard_library = ToilWDLStdLibBase(file_store, wdl_options={"task_path": self._wdl_options["task_path"]})
```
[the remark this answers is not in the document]
stxue1: The original code only passed through task_path so I dropped everything else, but I think the entire options dict can be passed through

src/toil/wdl/wdltoil.py
```diff
@@ -2568,64 +2757,68 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+            input_bindings = evaluate_call_inputs(self._node, self._node.inputs, incoming_bindings, standard_library, inputs_mapping)
+
+            # Bindings may also be added in from the enclosing workflow inputs
+            # TODO: this is letting us also inject them from the workflow body.
+            # TODO: Can this result in picking up non-namespaced values that
+            # aren't meant to be inputs, by not changing their names?
+            passed_down_bindings = incoming_bindings.enter_namespace(self._node.name)
+            task_path = self._wdl_options.get("task_path")
+            wdl_options = self._wdl_options
+            wdl_options["task_path"] = f'{task_path}.{self._node.name}'
```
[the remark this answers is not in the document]
stxue1: Changed to only modify copies. Also added the namespace in which will mirror the task path

src/toil/wdl/wdltoil.py
```diff
@@ -2398,14 +2590,16 @@ def patched_run_invocation(*args: Any, **kwargs: Any) -> List[str]:
...
+                if getattr(file, "nonexistent", False) is False:
+                    return WDL.Value.File(task_container.input_path_map[file.value])
+            contained_bindings = map_over_typed_files_in_bindings(bindings, get_path_in_container)
 
             # Make a new standard library for evaluating the command specifically, which only deals with in-container paths and out-of-container paths.
-            command_library = ToilWDLStdLibTaskCommand(file_store, self._task_path, task_container, workdir_in_container)
+            command_wdl_options: WDL_OPTIONS = {"task_path": task_path}
+            if workdir_in_container is not None:
+                command_wdl_options["execution_dir"] = workdir_in_container
+            command_library = ToilWDLStdLibTaskCommand(file_store, task_container, wdl_options=command_wdl_options)
```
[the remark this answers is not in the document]
stxue1: I changed it so the `command_wdl_options` will modify a copy of the task wdl options

src/toil/wdl/wdltoil.py
stxue1: I removed `enforce_existence` and `share_files_with` away from the WDL options so they should be more global now

src/toil/wdl/wdltoil.py
```diff
@@ -1058,15 +1272,48 @@ class ToilWDLStdLibTaskCommand(ToilWDLStdLibBase):
...
         """
 
         # TODO: Don't we want to make sure we don't actually use the file store?
-        super().__init__(file_store, task_path, execution_dir=execution_dir)
+        super().__init__(file_store, wdl_options=wdl_options)
         self.container = container
 
+    # Revert the _read and _write functions to the parent WDL.StdLib.Base implementation
+    # This is because the task command standard library is used in MiniWDL's internals when executing a task
+    # which we don't have much control over (miniwdl will create it's own file objects that represent files within the container)
```
adamnovak, line 1286:
```suggestion
    # which we don't have much control over (miniwdl will create its own file objects that represent files within the container)
```

src/toil/wdl/wdltoil.py
```diff
@@ -89,6 +98,13 @@
 
 logger = logging.getLogger(__name__)
 
+# WDL options to pass into the WDL jobs and standard libraries
+#   task_path: Dotted WDL name of the part of the workflow this library is working for.
+#   execution_dir: Directory to use as the working directory for workflow code.
+#   container: The type of container to use when executing a WDL task. Carries through the value of the commandline --container option
+WDL_Context = TypedDict('WDL_Context', {"execution_dir": NotRequired[str], "container": NotRequired[str],
```
adamnovak, line 105: I wouldn't use an `_` in the name here; it's a different convention than all the other types.

adamnovak (approved):
There are still a couple things I would change, but I think this is fine.

adamnovak:
@stxue1 It looks like conformance test 27 ("Legacy test for type_pair_with_files") is failing because it is trying to devirtualize a file to the same path twice. I think it can't deal with how that test sends the same path twice in the input JSON, so there are two `File` values from the same directory with the same basename (because it's the same file twice).

stxue1:
I changed the function `convert_remote_files`, so that function needs a rereview

src/toil/wdl/wdltoil.py
```diff
@@ -762,6 +801,127 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
+    :param environment: Bindings to evaluate on
+    :param file_source: Context to search for files with
+    :param task_path: Dotted WDL name of the user-level code doing the
+        importing (probably the workflow name).
+    :param search_paths: If set, try resolving input location relative to the URLs or
+        directories in this list.
+    :param import_remote_files: If set, import files from remote locations. Else leave them as URI references.
+    """
+    path_to_id: Dict[str, uuid.UUID] = {}
+    @memoize
```
adamnovak, line 819: I think this memoization gets scoped to the call to the enclosing function, but I also think that's what we want here.

src/toil/wdl/wdltoil.py
```diff
@@ -762,6 +801,127 @@ def is_url(filename: str, schemes: List[str] = ['http:', 'https:', 's3:', 'gs:',
...
+        return None, None
+
+    def convert_file_to_uri(file: WDL.Value.File) -> WDL.Value.File:
+        """
+        Calls import_filename to detect if a potential URI exists and imports it. Will modify the File object value to the new URI and tack on the virtualized file.
+        """
+        candidate_uri, toil_uri = import_filename(file.value)
+        if candidate_uri is None and toil_uri is None:
+            # If we get here we tried all the candidates
+            raise RuntimeError(f"Could not find {file.value} at any of: {list(potential_absolute_uris(file.value, search_paths if search_paths is not None else []))}")
```
adamnovak, line 912: It's odd to re-do the calculation of the possible locations here. If it changes it needs to change in both places. Having a function to compute it helps with that, but then we have a long expression going into the function and it's hard to tell if it is in sync just by looking.

Maybe the list of tried locations really wants to be an output from `import_filename`? Or maybe `import_filename` could raise this, since it is only called in the one place?

adamnovak (approved):
I think the changes are probably fine, but could be slightly better.

src/toil/wdl/wdltoil.py
stxue1: I think that's what we want, it is the same way we did memoization in the previous implementation. I believe file imports happening within the WDL have their own caching system separate from this anyway.

src/toil/wdl/wdltoil.py
stxue1: It might be better to have it inside `import_filename`, though the function wouldn't ever have a base return case, and I'm not sure if that's better or worse behaviorally. My thinking was that even though it is called twice, it will only be called again if a runtime error is hit, which is unlikely.

PR #5047, opened 2024-08-01
Limit reading/mounting from shared filesystem JobStore

### LINKED ISSUE #4673: Allow CWL and WDL workflows to minimize access to a shared-filesystem file job store
We're having some trouble on our Phoenix Slurm cluster, where Toil workflows are putting a high load on the NFS share where the input data and/or file job store are kept.

This is probably due to using symlinks and hardlinks as much as we can by default, and due to the caching file store refusing to enable on Slurm still. So we possibly symlink the input files to the job store, then we follow symlinks to mount files straight from the job store or the original input locations into the containers for jobs, and then the jobs all read directly against the NFS share instead of from local copies, even if many jobs are running at the same time on the same machine.

The NFS implementation in the kernel doesn't do anything smart about caching to local disk for us, so this results in a lot of network traffic and a lot of IO on the NFS server.

We probably could reduce this load and improve performance if we got caching working on Slurm, and allowed multiple jobs on the same node using the same file at the same time to copy it once and work on it from the local filesystem.

We also might get better results, depending on workload structure, if we had a way to lock out links into/mounting from the file job store. In the worst case each job would make a full copy of its input files to read even a single byte, but if the job needs to go through the input file 10 times or does a lot of random access jumping all over it, that could still be more efficient than working over NFS.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1441)
┆Issue Number: TOIL-1441

adamnovak:
This fixes #4673 by adding a `--symlinkJobStoreReads` option.

If you set `--symlinkJobStoreReads=False`, we won't create symlinks pointing into a FileJobStore when jobs go to read from it. We will instead make copies or hardlinks (within a filesystem). So if you have a workload that does more I/O on a file mounted into a container than a single full copy would do, you can use this to make a copy happen instead of having everyone mount straight from the shared filesystem.

If you have caching enabled, you can make the cache actually cache a copy per node instead of symlinking all the way through just because it can.

## Changelog Entry

 * New `--symlinkJobStoreReads=False` option lets you force local-node copies (possibly in the cache) even when reading directly from a FileJobStore is possible, potentially reducing shared filesystem IO.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -286,7 +286,7 @@ Allows configuring Toil's data storage.
                         to use a batch system that does not support
                         cleanup. Set to "true" if caching
                         is desired.
- --symlinkJobStoreReads BOOL
+  --symlinkJobStoreReads BOOL
                         Allow reads and container mounts from a JobStore's
                         shared filesystem directly via symlink. Can be turned
                         off if the shared filesystem can't support the IO load
--- a/docs/running/cliOptions.rst
+++ b/docs/running/cliOptions.rst
@@ -286,6 +286,13 @@ Allows configuring Toil's data storage.
                         to use a batch system that does not support
                         cleanup. Set to "true" if caching
                         is desired.
+ --symlinkJobStoreReads BOOL
+                        Allow reads and container mounts from a JobStore's
+                        shared filesystem directly via symlink. Can be turned
+                        off if the shared filesystem can't support the IO load
+                        of all the jobs reading from it at once, and you want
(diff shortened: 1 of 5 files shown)


DailyDreaming (approved):
Looks good to me, though could something be added to the docs for `--symlinkJobStoreReads` and when to use it?

Also, particularly: https://github.com/DataBiosphere/toil/blob/7d4d0749e9f5ffcc2a4507e13e940e46400a020d/docs/running/cliOptions.rst

PR #5052, opened 2024-08-08
Allow CWL workflows to have jobs use all of a Slurm node's memory

### LINKED ISSUE #4971: toil-cwl-runner used to allow `--defaultMemory 0`, which has special meaning to Slurm, but now no longer does
I recently noticed that it is no longer possible to pass `--defaultMemory 0` to `toil-cwl-runner`. There's a check in place that enforces its value to be 1 or more. AFAICS this poses a problem when using Slurm, because you need to pass `--mem=0` to `sbatch` to instruct Slurm to use all available memory. 

See also: https://matrix.to/#/!sSMDDqTXkxpNofyjTR:gitter.im/$2MQRCwBC00tLV7ODhHvTHt73jEccTQnm0fL3u7BYwPA?via=gitter.im&via=matrix.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1590)
┆Issue Number: TOIL-1590

adamnovak:
This should fix #4971. Instead of `--defaultMemory 0`, to run CWL jobs that lack their own `ramMin` with a full Slurm node's memory, you would now pass `--no-cwl-default-ram --slurmDefaultAllMem=True`.

This might cause some new problems:
- Other internal CWL runner jobs that expected to use the default memory would now use all the memory on their node, if submitted to the cluster.
- Now we use the CWL spec's required default memory for jobs that don't specify a limit, instead of `--defaultMemory` *unless* the user passes `--no-cwl-default-ram`. Previously I *think* we were ignoring the spec and always using the Toil `--defaultMemory`. This might break some workflow runs that used to work because of us giving them more memory than the spec said to.

Also, #4971 says we're supposed to implement a real framework for doing this kind of memory expansion across all batch systems that support it. But I didn't want to add a new bool flag onto `Requirer` for such a specific purpose. Probably if we need it we should combine it with preemptible somehow into a tag/flag system. Or we could implement memory range requirements and allow the top of the range to be unbounded, or treat some threshold upper limit as "all the node's memory" in the Slurm batch system.

## Changelog Entry

* Toil now has a `--slurmDefaultAllMem` option to run jobs lacking their own memory requirements with Slurm's `--mem=0`, so they get a whole node's memory.
* `toil-cwl-runner` now has `--no-cwl-default-ram` (and `--cwl-default-ram`) to control whether the CWL spec's default `ramMin` is applied, or Toil's own default memory logic is used.
* The `--dont_allocate_mem` and `--allocate_mem` options have been deprecated and replaced with `--slurmAllocateMem`, which can be `True` or `False`.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


adamnovak:
I still need to manually test this to make sure it actually does what it is meant to do.

adamnovak:
I wrote a test for this and it does indeed seem to issue jobs that ask for whole Slurm nodes when I use the two new options together.

I also fixed Slurm job cleanup when a workflow is killed; it wasn't doing that before because shutdown() wasn't doing any killing in `AbstractGridEngineBatchSystem`. I needed this for the test to not leave behind pending jobs when there aren't any free entire cluster nodes.

adamnovak:
@DailyDreaming Can you review this?

docs/appendices/environment_vars.rst
```diff
@@ -131,6 +131,9 @@ There are several environment variables that affect the way Toil runs.
 | TOIL_GOOGLE_PROJECTID            | The Google project ID to use when generating       |
 |                                  | Google job store names for tests or CWL workflows. |
 +----------------------------------+----------------------------------------------------+
+| TOIL_SLURM_ALLOCATE_MEM          | Whether to akllocate memory in Slurm with --mem.   |
```
DailyDreaming, line 134: akllocate

docs/appendices/environment_vars.rst
```diff
@@ -143,6 +146,8 @@ There are several environment variables that affect the way Toil runs.
 |                                  | jobs.                                              |
 |                                  | There is no default value for this variable.       |
 +----------------------------------+----------------------------------------------------+
+| TOIL_SLURM_TIME                  | Slurm job time limit, in [DD-]HH:MM:SS format.     |
```
DailyDreaming, line 149: An example might be useful (literally adding `e.g. TOIL_SLURM_TIME="04:00:00" for 4 hours`, assuming that this produces a 4 hour time limit), or a link to SLURM docs if this is a direct alias for a slurm arg.

docs/cwl/running.rst
```diff
@@ -91,9 +91,17 @@ printed to the stdout stream after workflow execution.
...
 
-``--disable-streaming``: Does not allow streaming of input files. This is enabled
-by default for files marked with ``streamable`` flag True and only for remote files
-when the jobStore is not on local machine.
+Extra Toil CWL Options
+++++++++++++++++++++++
+
+Besides the normal Toil options and the options supported by cwltool, toil-cwl-runner adds some of its own options:
+
+  --bypass-file-store   Do not use Toil's file store and assume all paths are accessible in place from all nodes.
```
DailyDreaming, line 99: Maybe a further explanation that files may not be redundantly copied to a potentially remote filestore, and this can speed up the workflow (as a reason for why someone might want this).  Has the downside that the workflow is not restart-able from the filestore if it made progress and failed part-way through.

docs/running/cliOptions.rst
```diff
@@ -237,6 +237,21 @@ levels in toil are based on priority from the logging module:
...
+                        machines than they were designed for, by setting a
+                        value less than 1. (default: 1)
+  --slurmAllocateMem SLURM_ALLOCATE_MEM
+                        If False, do not use --mem. Used as a workaround for
+                        Slurm clusters that reject jobs with memory
+                        allocations.
+  --slurmTime SLURM_TIME
+                        Slurm job time limit, in [DD-]HH:MM:SS format.
+  --slurmPE SLURM_PE    Special partition to send Slurm jobs to if they ask
+                        for more than 1 CPU.
```
DailyDreaming, line 252: Could this be explained further?

src/toil/batchSystems/abstractGridEngineBatchSystem.py
```diff
@@ -417,11 +433,11 @@ def issueBatchJob(self, command: str, jobDesc, job_environment: Optional[Dict[st
             else:
                 gpus = jobDesc.accelerators
```
DailyDreaming, line 434: This code looks like it was maybe left in and is a function now (self.count_needed_gpus)?

src/toil/options/cwl.py
```diff
@@ -357,3 +315,64 @@ def add_cwl_options(parser: ArgumentParser, suppress: bool = True) -> None:
...
+        action="store_true",
+        default=False,
+        help=suppress_help or "Disable file streaming for files that have 'streamable' flag True.",
+        dest="disable_streaming",
+    )
+    parser.add_argument(
+        "--runImportsOnWorkers", "--run-imports-on-workers",
+        action="store_true",
+        default=False,
+        help=suppress_help or "Run the file imports on a worker instead of the leader. This is useful if the leader is not optimized for high network performance."
```
DailyDreaming, line 348: Space needed at the end of this sentence.

src/toil/options/cwl.py
```diff
@@ -357,3 +315,64 @@ def add_cwl_options(parser: ArgumentParser, suppress: bool = True) -> None:
...
+                              "If set to true, the argument --importWorkersDisk must also be set.",
+        dest="run_imports_on_workers"
+    )
+    parser.add_argument(
+        "--importWorkersDisk", "--import-workers-disk",
+        help=suppress_help or "Specify the amount of disk space an import worker will use. If file streaming for input files is not available,"
```
DailyDreaming, line 354: Space needed at the end of this sentence.

src/toil/test/cwl/cwlTest.py
```diff
@@ -451,6 +454,56 @@ def test_glob_dir_bypass_file_store(self) -> None:
...
+            log.debug("Workflow might be waiting for a full node. Stop it.")
+            subprocess.check_call(["toil", "kill", self.jobStoreDir])
+            # Wait another little bit for it to clean up, making sure to collect output in case it is blocked on writing
+            child.communicate(timeout=20)
+            # Kill it off in case it is still running
+            child.kill()
+            # Reap it
+            child.wait()
+            # The test passes
+            return
```
DailyDreaming, line 497: I think that this `return` is a no-op?

DailyDreaming (approved):
LGTM.  Some minor clean-up.

src/toil/test/cwl/cwlTest.py
adamnovak: I think I put it there to make it clear that we intend to go there and end the test. But if you think it makes more sense without the `return` I'll drop it.

PR #5123, opened 2024-10-14
New AWS jobstore.

DailyDreaming:
Rewritten without SDB.

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/lib/aws/s3.py
```diff
@@ -11,21 +11,484 @@
...
+
+@retry(errors=[BotoServerError])
+def copy_s3_to_s3(s3_resource, src_bucket, src_key, dst_bucket, dst_key, extra_args: Optional[dict] = None):
+    if not extra_args:
+        source = {'Bucket': src_bucket, 'Key': src_key}
+        # Note: this may have errors if using sse-c because of
+        # a bug with encryption using copy_object and copy (which uses copy_object for files <5GB):
+        # https://github.com/aws/aws-cli/issues/6012
+        # this will only happen if we attempt to copy a file previously encrypted with sse-c
+        # copying an unencrypted file and encrypting it as sse-c seems to work fine though
```
adamnovak, line 231: The issue linked appears to apply only to the AWS CLI, and [not the API](https://github.com/aws/aws-cli/issues/6012#issuecomment-901416584), and also only describes adding encryption, not removing or changing it. Does it really not work if we do it through the API ourselves? (Or would we need tor each in and fiddle with headers manually?)

src/toil/lib/aws/utils.py
```diff
@@ -346,7 +346,7 @@ def bucket_location_to_region(location: Optional[str]) -> str:
     return "us-east-1" if location == "" or location is None else location
 
 
-def get_object_for_url(url: ParseResult, existing: Optional[bool] = None) -> "S3Object":
+def get_object_for_url(url: Union[ParseResult, str], existing: Optional[bool] = None) -> "S3Object":
```
adamnovak, line 349: I feel like we should have stronger opinions on which of these we ought to have. It's kind of convenient for the caller to just be able to throw whatever in here, but remembering that you *can* do that is harder than remembering that you need one type, I think.

And we're not updating the docstring to say that the URL doesn't *actually* have to be parsed.

src/toil/lib/pipes.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def __enter__(self):
+        self.readable_fh, writable_fh = os.pipe()
+        self.writable = os.fdopen(writable_fh, 'wb' if self.encoding == None else 'wt', encoding=self.encoding, errors=self.errors)
+        self.thread = ExceptionalThread(target=self._reader)
+        self.thread.start()
+        return self.writable
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # Closeing the writable end will send EOF to the readable and cause the reader thread
```
adamnovak, line 121:
```suggestion
        # Closing the writable end will send EOF to the readable and cause the reader thread
```

.gitlab-ci.yml
```diff
@@ -69,7 +69,6 @@ lint:
     - ${MAIN_PYTHON_PKG} -m virtualenv venv && . venv/bin/activate && make prepare && make develop extras=[all]
     - ${MAIN_PYTHON_PKG} -m pip freeze
     - ${MAIN_PYTHON_PKG} --version
-    - make mypy
```
adamnovak, line 72: We probably need to keep the mypy step.

adamnovak (changes requested):
It looks like this probably works as written, but the documentation for how the S3-backed storage works (metadata objects named by file IDs referencing ETag hashes for content-addressed data objects many-to-one) doesn't match what's actually implemented (metadata objects and data objects both named by file ID).

It might be good to actually implement the described content-addressing design, but there are are some big unsolved questions:
- ETags on S3 are all MD5-based and not collision-resistant, so it's easy to get two actual files with the same MD5 and put them in the job store to break it.
- Streaming uploads don't know their hash until they are uploaded, so where should they be uploaded to?
- Without transactions, how do you maintain the consistency of many-to-one relationships between metadata objects and data objects, while still allowing data objects to be deleted when the last referencing metadata object goes away?

It might make sense to not solve this deduplication problem now, and to just admit to using one data copy per file ID as is actually implemented. If we cram executability into the ID somehow, and use the ETag provided by S3 in the response header for integrity checking, it might be possible to not have a metadata file at all.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -301,179 +294,195 @@ def load_job(self, job_id: str):
...
+            extra_args=self.encryption_args
+        )
 
-    def file_exists(self, file_id):
+    def file_exists(self, file_id: str) -> bool:
         return s3_key_exists(s3_resource=self.s3_resource,
                              bucket=self.bucket_name,
-                             key=f'{self.content_key_prefix}{file_id}',
+                             key=self.existing_full_s3_key_from_file_id(file_id),
                              extra_args=self.encryption_args)
```
adamnovak, line 440: This isn't going to work because `file_exists` is meant to be able to return `False`, but `existing_full_s3_key_from_file_id` will always raise if no such key exists.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -511,103 +525,102 @@ def read_shared_file_stream(self, shared_file_name, encoding=None, errors=None):
             if e.response.get('ResponseMetadata', {}).get('HTTPStatusCode') == 404:
                 raise NoSuchFileException(shared_file_name)
 
-    def delete_file(self, file_id):
-        """Only delete the reference."""
-        self.s3_client.delete_object(Bucket=self.bucket_name, Key=f'{self.metadata_key_prefix}{file_id}')
-        self.s3_client.delete_object(Bucket=self.bucket_name, Key=f'{self.content_key_prefix}{file_id}')
+    def delete_file(self, file_id: str) -> None:
+        full_s3_key = self.existing_full_s3_key_from_file_id(file_id)
+        self.s3_client.delete_object(Bucket=self.bucket_name, Key=full_s3_key)
```
adamnovak, line 530: `delete_file` is required to be idempotent:
https://github.com/DataBiosphere/toil/blob/9f0d4580523d8f9f00a76feff252ab450a06c938/src/toil/jobStores/abstractJobStore.py#L1535-L1536
But `existing_full_s3_key_from_file_id` will raise an error if no such key exists, so you won't actually be able to safely call the delete twice.
Probably the error needs to be handled here.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -511,103 +525,102 @@ def read_shared_file_stream(self, shared_file_name, encoding=None, errors=None):
...
-        if issubclass(otherCls, AWSJobStore):
-            src_bucket_name, src_key_name = parse_s3_uri(url)
+        if isinstance(otherCls, AWSJobStore):
+            src_bucket_name, src_key_name = parse_s3_uri(uri.geturl())
             response = head_s3_object(self.s3_resource, bucket=src_bucket_name, key=src_key_name, check=True)
             content_length = response['ContentLength']  # e.g. 65536
-            content_type = response['ContentType']  # e.g. "binary/octet-stream"
-            etag = response['ETag'].strip('\"')  # e.g. "\"586af4cbd7416e6aefd35ccef9cbd7c8\""
+            # content_type = response['ContentType']  # e.g. "binary/octet-stream"
+            # etag = response['ETag'].strip('\"')  # e.g. "\"586af4cbd7416e6aefd35ccef9cbd7c8\""
```
adamnovak, line 554: Probably these can just be cut.
```suggestion
```

src/toil/jobStores/aws/jobStore.py
```diff
@@ -511,103 +525,102 @@ def read_shared_file_stream(self, shared_file_name, encoding=None, errors=None):
...
+                s3_resource=self.s3_resource,
+                src_bucket=src_bucket_name,
+                src_key=src_key_name,
+                dst_bucket=self.bucket_name,
+                dst_key=dst_key,
+                extra_args=self.encryption_args
+            )
+            # TODO: verify etag after copying here?
+
+            # cannot determine exec bit from foreign s3 so default to False
```
adamnovak, line 572: This comment needs to go up where the `0` is now.

src/toil/lib/pipes.py
```diff
@@ -15,64 +15,64 @@
...
-    ...     def readFrom(self, readable):
-    ...         pass
-    >>> with MyPipe() as writable:
-    ...     raise RuntimeError('Hello, world!')
-    Traceback (most recent call last):
-    ...
-    RuntimeError: Hello, world!
-    >>> y = os.dup(0); os.close(y); x == y
-    True
+    # An object-oriented wrapper for os.pipe. Clients should subclass it, implement
```
adamnovak, line 18: I don't think it makes sense to try to have a big comment *inside* a docstring like this.

src/toil/lib/pipes.py
```diff
@@ -117,7 +117,12 @@ def __exit__(self, exc_type: Optional[str], exc_val: Optional[str], exc_tb: Opti
             # thread. To cover the small window before the reader takes over we also close it here.
             if self.readable_fh is not None:
                 # Close the file handle. The reader thread must be dead now.
-                os.close(self.readable_fh)
+                try:
+                    os.close(self.readable_fh)
+                except OSError as e:
+                    # OSError: [Errno 9] Bad file descriptor implies this file handle is already closed
+                    if not e.errno == 9:
```
adamnovak, line 124: This should be `errno.EBADF` and not a magic 9.
```suggestion
                    if not e.errno == errno.EBADF:
```

src/toil/test/__init__.py
```diff
@@ -482,7 +482,7 @@ def needs_aws_batch(test_item: MT) -> MT:
         test_item
     )
     test_item = needs_env_var(
-        "TOIL_AWS_BATCH_JOB_ROLE_ARN", "an IAM role ARN that grants S3 and SDB access"
+        "TOIL_AWS_BATCH_JOB_ROLE_ARN", "an IAM role ARN that grants S3 access"
```
adamnovak, line 485: Should we also revise the role permissions in the provisioner? Or make a new issue for that?

adamnovak (changes requested):
I think this is way better!

I did notice that the "existing" key getter function was used in a couple places where the key won't actually necessarily exist, which I don't think will work.

And I don't think we want to comment out the contents of docstrings.

But those are probably the only things I think really need to be changed.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -110,7 +110,7 @@ class NoSuchFileException(Exception):
     """Indicates that the specified file does not exist."""
 
     def __init__(
-        self, jobStoreFileID: FileID, customName: Optional[str] = None, *extra: Any
+        self, jobStoreFileID: Union[FileID, str], customName: Optional[str] = None, *extra: Any
```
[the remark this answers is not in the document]
adamnovak: I guess some of the user-facing file ID API still accepts file IDs typed as strings, so maybe we do still need to be able to take them here.

PR #5103, opened 2024-09-26
Support importing on workers in WDL

stxue1:
Mirrors #5025 to the WDL side too
## Changelog Entry
To be copied to the [draft changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog) by merger:

* Added support to import files on workers for toil-wdl-runner

## Reviewer Checklist

 * [ ] Make sure it is coming from `issues/XXXX-fix-the-thing` in the Toil repo, or from an external repo.
    * [ ] If it is coming from an external repo, make sure to pull it in for CI with:
        ```
        contrib/admin/test-pr otheruser theirbranchname issues/XXXX-fix-the-thing
        ```
    * [ ] If there is no associated issue, [create one](https://github.com/DataBiosphere/toil/issues/new).
* [ ] Read through the code changes. Make sure that it doesn't have:
    * [ ] Addition of trailing whitespace.
    * [ ] New variable or member names in `camelCase` that want to be in `snake_case`.
    * [ ] New functions without [type hints](https://docs.python.org/3/library/typing.html).
    * [ ] New functions or classes without informative docstrings.
    * [ ] Changes to semantics not reflected in the relevant docstrings.
    * [ ] New or changed command line options for Toil workflows that are not reflected in `docs/running/{cliOptions,cwl,wdl}.rst` 
    * [ ] New features without tests.
* [ ] Comment on the lines of code where problems exist with a review comment. You can shift-click the line numbers in the diff to select multiple lines.
* [ ] Finish the review with an overall description of your opinion.

## Merger Checklist

* [ ] Make sure the PR passes tests.
* [ ] Make sure the PR has been reviewed **since its last modification**. If not, review it.
* [ ] Merge with the Github "Squash and merge" feature.
    * [ ] If there are multiple authors' commits, add [Co-authored-by](https://github.blog/2018-01-29-commit-together-with-co-authors/) to give credit to all contributing authors.
* [ ] Copy its recommended changelog entry to the [Draft Changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog).
* [ ] Append the issue number in parentheses to the changelog entry.

PR #5098, opened 2024-09-19
Allow importing on workers

stxue1:
#5025
## Changelog Entry
To be copied to the [draft changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog) by merger:

 * Added support to import files on workers for toil-cwl-runner
   * `--runImportsOnWorkers` to enable importing files on workers
   * `--importWorkersDisk` to control how much disk space the import worker will use

## Reviewer Checklist

 * [ ] Make sure it is coming from `issues/XXXX-fix-the-thing` in the Toil repo, or from an external repo.
    * [ ] If it is coming from an external repo, make sure to pull it in for CI with:
        ```
        contrib/admin/test-pr otheruser theirbranchname issues/XXXX-fix-the-thing
        ```
    * [ ] If there is no associated issue, [create one](https://github.com/DataBiosphere/toil/issues/new).
* [ ] Read through the code changes. Make sure that it doesn't have:
    * [ ] Addition of trailing whitespace.
    * [ ] New variable or member names in `camelCase` that want to be in `snake_case`.
    * [ ] New functions without [type hints](https://docs.python.org/3/library/typing.html).
    * [ ] New functions or classes without informative docstrings.
    * [ ] Changes to semantics not reflected in the relevant docstrings.
    * [ ] New or changed command line options for Toil workflows that are not reflected in `docs/running/{cliOptions,cwl,wdl}.rst` 
    * [ ] New features without tests.
* [ ] Comment on the lines of code where problems exist with a review comment. You can shift-click the line numbers in the diff to select multiple lines.
* [ ] Finish the review with an overall description of your opinion.

## Merger Checklist

* [ ] Make sure the PR passes tests.
* [ ] Make sure the PR has been reviewed **since its last modification**. If not, review it.
* [ ] Merge with the Github "Squash and merge" feature.
    * [ ] If there are multiple authors' commits, add [Co-authored-by](https://github.blog/2018-01-29-commit-together-with-co-authors/) to give credit to all contributing authors.
* [ ] Copy its recommended changelog entry to the [Draft Changelog](https://github.com/DataBiosphere/toil/wiki/Draft-Changelog).
* [ ] Append the issue number in parentheses to the changelog entry.

PR #5133, opened 2024-10-24
Add filesize sniffing and parallelize importing on jobs

### LINKED ISSUE #5114: Add filesize sniffing to specify the import job's disk space for WDL and CWL
When `--importWorkerDisk` is not specified for WDL and CWL, there should be some sort of file sniffing functionality to detect the minimum amount of disk space needed for the importing job to run. This probably can't take in account whether a file is streamable or not. We can probably try to stream the files in first with a small disk requirement, and then when that doesn't work, start a child job to run the imports without streaming but with more disk space.

This will likely require retooling the import interface so file streaming can be controlled there.

stxue1:
Closes #5114 
## Changelog Entry

 * Add support for parallel file imports
   * Usurps #5103 and #5098
   * `--importWorkersDisk` replaced with `--importWorkersThreshold`. This specifies the threshold where files will begin to be imported on individual jobs. Small files will be batched into the same import job up to this threshold.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+        # Run WDL imports on a worker instead
+
+        individual_files = []
+        batch_files = []
+        # Separate out files that will be imported individually and in a batch
+        for filename, (candidate_uri, parent_id, size) in file_to_data.items():
+            if size < import_workers_threshold:
+                batch_files.append(filename)
+            else:
+                individual_files.append(filename)
```
adamnovak, line 3955: If it gets used like this, it isn't really a threshold for importing on workers (because files under/over it don't get treated differently in terms of whether they are imported on workers). Maybe it should be named more like `max_batch_size`?

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+            filesize = file_to_data[filename][2]
+            if per_batch_size + filesize >= import_workers_threshold:
+                # batch is too big now, store to schedule the batch
+                to_batch.append(per_batch_files)
+                # reset batching calculation
+                per_batch_size = 0
+            else:
+                per_batch_size += filesize
+                per_batch_files.append(filename)
+                batch_files.pop(0)
```
adamnovak, line 3985: If you adjusted this code to be able to handle individual files larger than the target batch size, you could get rid of the concept of `individual_files` and just have everything be (possibly single-file, possibly over-sized) batches.

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+        """
+        Import the workflow inputs and then create and run the workflow.
+        :return: Tuple of a mapping from the candidate uri to the file id and a mapping of the source filenames to its metadata. The candidate uri is a field in the file metadata
+        """
+        import_workers_threshold = self._import_workers_threshold
+        file_to_data = self._file_to_data
+        # Run WDL imports on a worker instead
+
+        individual_files = []
+        batch_files = []
```
adamnovak, line 3949: This sounds more like it ought to hold the files for a particular batch, but really it is the list of all files that need to be batched up.

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+        for file in individual_files:
+            filesize = file_to_data[file][2]
+            candidate_uri = file_to_data[file][0]
+            import_jobs.append(WorkerImportJob([candidate_uri], disk_size=filesize))
+
+        # create multiple batch jobs if necessary to not exceed the threshold
+        per_batch_files = []
+        per_batch_size = 0
+        # This list will hold lists of batched filenames
+        to_batch = []
```
adamnovak, line 3968: And *this* sounds like it holds the list of all files that need to be batched up, but actually it holds the completed batches.

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+
+        # Create individual import jobs
+        import_jobs = []
+        for file in individual_files:
+            filesize = file_to_data[file][2]
+            candidate_uri = file_to_data[file][0]
+            import_jobs.append(WorkerImportJob([candidate_uri], disk_size=filesize))
+
+        # create multiple batch jobs if necessary to not exceed the threshold
+        per_batch_files = []
```
adamnovak, line 3965: This sounds like it would hold a list of lists: for each batch, a list of the files in that batch. The comment above it also suggests it might hold "multiple batch jobs". But actually it holds the current batch under construction.

I would initialize the current thing under construction *after* the list that is going to accumulate all the things constructed.

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+        per_batch_size = 0
+        # This list will hold lists of batched filenames
+        to_batch = []
+        while len(batch_files) > 0 or len(per_batch_files) > 0:
+            if len(batch_files) == 0 and len(per_batch_files) > 0:
+                to_batch.append(per_batch_files)
+                break
```
adamnovak, line 3972: If you move the logic to put a nonempty final batch into the list of batches, and put it *after* the loop, the control flow makes more sense. Then the loop can just be `for filename in batch_files:` and the condition you check after the loop can just be `if per_batch_files:`.

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+
+            filename = batch_files[0]
+            # See if adding this to the queue will make the batch job too big
+            filesize = file_to_data[filename][2]
+            if per_batch_size + filesize >= import_workers_threshold:
+                # batch is too big now, store to schedule the batch
+                to_batch.append(per_batch_files)
+                # reset batching calculation
+                per_batch_size = 0
```
adamnovak, line 3981: Instead of relying on *not* popping the file and coming back around to it, you could start a new batch containing just it.

src/toil/options/runner.py
```diff
@@ -25,14 +25,14 @@ def add_runner_options(
...
-        dest="import_workers_disk",
+        *import_workers_threshold_argument,
+        dest="import_workers_threshold",
         type=lambda x: human2bytes(str(x)),
-        default=None,
-        help="Specify the amount of disk space an import worker will use. If file streaming for input files is not available, "
-        "this should be set to the size of the largest input file. This must be set in conjunction with the argument runImportsOnWorkers."
+        default="1 GiB",
+        help="Specify the file size threshold that determines if the file import will happen in its own job. All files below the threshold "
+        "will go into a batch import job. This should be set in conjunction with the argument --runImportsOnWorkers."
```
adamnovak, line 37: This makes it sound like there's just one batch import job, when really we batch all the imports into batches of about this size.

src/toil/wdl/wdltoil.py
```diff
@@ -1248,51 +1309,168 @@ def _call_eager(
...
+def get_file_sizes(
+    filenames: List[str],
+    file_source: AbstractJobStore,
+    search_paths: Optional[List[str]] = None,
+    import_remote_files: bool = True,
+    execution_dir: Optional[str] = None,
+) -> Dict[str, FileMetadata]:
     """
-    return is_url_with_scheme(filename, REMOTE_SCHEMES)
+    Resolve relative-URI files in the given environment and them then into absolute normalized URIs. Returns a dictionary of WDL file values to a tuple of the normalized URI,
```
adamnovak, line 1332: I don't think `and them then` makes sense.

We might also want to say something different than "WDL file values"; we mean the string values that would appear in the `value` field of a `WDL.Value.File` object. But you could also think that a `WDL.Value.File` object is *itself* a "WDL file value".

src/toil/wdl/wdltoil.py
```diff
@@ -1248,51 +1309,168 @@ def _call_eager(
...
+                # Wasn't found there
+                continue
+            except Exception:
+                # Something went wrong besides the file not being found. Maybe
+                # we have no auth.
+                logger.error(
+                    "Something went wrong when testing for existence of %s",
+                    candidate_uri,
+                )
+                raise
```
adamnovak, line 1386: This duplicates a lot of code with the CWL-side `get_file_sizes()`. Is there a common function that could be extracted here that polls a URL and returns whether it existed and, if so, the size if available, and raises on other errors?

src/toil/wdl/wdltoil.py
```diff
@@ -1248,51 +1309,168 @@ def _call_eager(
...
+
+    return {k: get_filename_size(k) for k in filenames}
+
+
+def convert_files(
+    environment: WDLBindings,
+    file_to_id: Dict[str, FileID],
+    file_to_data: Dict[str, FileMetadata],
+    task_path: str,
+) -> None:
```
adamnovak, line 1428: Instead of altering the `File` objects inside `environment` in place, this function should return a modified copy of `environment`. The `WDLBindings` objects in MiniWDL I think are meant to be immutable.

src/toil/wdl/wdltoil.py
```diff
@@ -5180,8 +5401,8 @@ class WDLStartJob(WDLSectionJob):
 
     def __init__(
         self,
-        target: WDL.Tree.Workflow | WDL.Tree.Task,
-        inputs: WDLBindings,
+        target: Union[WDL.Tree.Workflow, WDL.Tree.Task],
```
adamnovak, line 5404: I think using `|` everywhere instead of `Union` is another pyupgrade change that we want to keep.

src/toil/wdl/wdltoil.py
```diff
@@ -5226,65 +5448,107 @@ def run(self, file_store: AbstractFileStore) -> Promised[WDLBindings]:
...
+        install_imports_job.addFollowOn(start_job)
+
+        return start_job.rv()
 
 
 def make_root_job(
     target: WDL.Tree.Workflow | WDL.Tree.Task,
     inputs: WDLBindings,
-    inputs_search_path: list[str],
+    inputs_search_path: List[str],
```
adamnovak, line 5539: @mr-c also changed to using the new generic support in the base `list`, `dict`, etc. instead of needing to import the versions from `typing`, so we shouldn't undo that.

adamnovak (changes requested):
I think the shape of the import logic is right, but I'm concerned about some variable names and some duplicated code, and about whether the CWL-side import logic actually has all its interdependencies documented sufficiently.

src/toil/cwl/cwltoil.py
```diff
@@ -106,12 +124,21 @@
...
+)
 from toil.jobStores.abstractJobStore import (
     AbstractJobStore,
-    InvalidImportExportUrlException,
-    LocatorException,
     NoSuchFileException,
+    LocatorException,
+    InvalidImportExportUrlException,
     UnimplementedURLException,
 )
```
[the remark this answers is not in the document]
stxue1: It seems like the Makefile code formatter formats quite a few files, except for cwltoil.py. I'll manually undo this change, maybe this was done on accident while I was messing with imports

src/toil/cwl/cwltoil.py
```diff
@@ -1807,17 +1834,68 @@ def path_to_loc(obj: CWLObjectType) -> None:
...
-    import_function: Callable[[str], FileID],
+def extract_files(
+    fileindex: Dict[str, str],
+    existing: Dict[str, str],
+    file_metadata: CWLObjectType,
+    mark_broken: bool = False,
+    skip_remote: bool = False,
+) -> Optional[str]:
+    """
+    Extract the filename from a CWL file record
```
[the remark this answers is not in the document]
stxue1: This function has a bunch of borrowed code from upload_file and write_file. I think if the `file://` scheme exists, some resolving was already done.

src/toil/cwl/cwltoil.py
```diff
@@ -1807,17 +1834,68 @@ def path_to_loc(obj: CWLObjectType) -> None:
...
+        # This is a local file, or we also need to download and re-upload remote files
+        if location not in fileindex:
+            # don't download twice
+            # These dictionaries are meant to keep track of what we're going to import
+            # In the actual import, this is used as a bidirectional mapping from unvirtualized to virtualized
+            # see write_file
+            if not urlparse(location).scheme:
+                rp = os.path.realpath(location)
+            else:
+                rp = location
```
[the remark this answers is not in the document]
stxue1: I borrowed this code from write_file. I believe all files should be resolved into file URIs by this point, so this is likely some edge case. From my limited testing, I can't trigger the realpath branch.

src/toil/cwl/cwltoil.py
```diff
@@ -3445,41 +3523,89 @@ def run(
...
-        super().__init__()
+        basedir: str,
+        skip_remote: bool,
+        bypass_file_store: bool,
+        import_data: Promised[Tuple[Dict[str, FileID], Dict[str, FileMetadata]]],
+        **kwargs: Any,
+    ) -> None:
+        """
+        Job to take the entire CWL object and a mapping of filenames to the imported URIs
+        to convert all file locations to URIs.
```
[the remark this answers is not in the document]
stxue1: Fixed the docstring

src/toil/cwl/cwltoil.py
```diff
@@ -3445,41 +3523,89 @@ def run(
...
         initialized_job_order = unwrap(self.initialized_job_order)
         tool = unwrap(self.tool)
-        root_job, _ = makeJob(
-            tool, initialized_job_order, self.runtime_context, None, None
+
+        def convert_file(filename: str) -> FileID:
+            fileid = candidate_to_fileid[filename]
+            return fileid
+
+        file_convert_function = functools.partial(upload_file, convert_file)
```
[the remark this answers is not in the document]
stxue1: Renamed the functions

src/toil/job.py
```diff
@@ -3758,6 +3826,181 @@ def getUserScript(self):
...
+    For the CWL/WDL runners, this class is only used when runImportsOnWorkers is enabled.
+    """
+
+    def __init__(
+        self,
+        filenames: List[str],
+        disk_size: Optional[ParseableIndivisibleResource] = None,
+        stream: bool = True,
+        **kwargs: Any,
+    ):
```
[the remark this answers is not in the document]
stxue1: Added a docstring. The logic the worker uses to do the import when streaming fails needs to know the amount of disk size it will use. This way, the command line won't report a failed job when streaming fails as I can catch it with a try except block and immediately reschedule. I'm not sure of a foolproof way to move setting the disk requirement to the caller. Then the caller would handle the job failing instead, and would report a failed job, which I don't want.

src/toil/job.py
stxue1: Changed to `file_batches`

src/toil/wdl/wdltoil.py
stxue1: Separated it out into job.py

src/toil/cwl/cwltoil.py
```diff
@@ -3598,18 +3811,22 @@ def visitSteps(
...
         # if they bothered to run the Process __init__.
-        op(cmdline_tool.tool)
+        return op(cmdline_tool.tool)
     else:
         raise RuntimeError(
             f"Unsupported type encountered in workflow "
             f"traversal: {type(cmdline_tool)}"
         )
+    # Satisfy mypy, but this branch should never be reached in practice
+    return []
```
mr-c, line 3829:
```suggestion
    raise RuntimeError(
        f"Unsupported type encountered in workflow "
        f"traversal: {type(cmdline_tool)}"
    )

```

src/toil/cwl/cwltoil.py
```diff
@@ -37,7 +37,24 @@
...
 from tempfile import NamedTemporaryFile, TemporaryFile, gettempdir
 from threading import Thread
-from typing import IO, Any, Callable, Literal, Optional, TextIO, TypeVar, Union, cast
+from typing import (
+    IO,
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
```
adamnovak, line 46: I think we don't need `Dict` and `List` anymore because we can use `dict` and `list` now.

src/toil/job.py
```diff
@@ -3758,6 +3776,374 @@ def getUserScript(self):
...
+    source: str
+    parent_dir: str
+    size: Optional[int]
+
+
+def potential_absolute_uris(
+    uri: str,
+    path: list[str],
+    importer: str | None = None,
+    execution_dir: str | None = None,
```
adamnovak, line 3796: We might actually still need `Optional` on 3.9.

src/toil/job.py
```diff
@@ -3758,6 +3776,374 @@ def getUserScript(self):
...
+
+            # Work out what the basename for the file was
+            file_basename = os.path.basename(urlsplit(candidate_uri).path)
+
+            if file_basename == "":
+                # We can't have files with no basename because we need to
+                # download them at that basename later.
+                raise RuntimeError(
+                    f"File {candidate_uri} has no basename and so cannot be a WDL File"
+                )
```
adamnovak, line 3939: This error message is WDL-specific. I'm not sure if it's OK to impose the constraint on CWL as well, but if we do we should just complain generically that the file has no basename.
```suggestion
            if file_basename == "":
                # We can't have files with no basename because we need to
                # download them at that basename later in WDL.
                raise RuntimeError(
                    f"File {candidate_uri} has no basename"
                )
```

src/toil/job.py
```diff
@@ -3758,6 +3776,374 @@ def getUserScript(self):
...
+        stream: bool = True,
+        **kwargs: Any,
+    ):
+        """
+        Setup importing files on a worker.
+        :param filenames: List of file URIs to import
+        :param disk_size: Designated disk space the worker can use when importing. Disregarded if stream is enabled.
+        :param stream: Whether to stream a file import or not. We don't have machinery to ensure
+        streaming, so if true, assume streaming works and don't give the worker a lot of disk space to work with.
+        If streaming fails, the worker will run out of resources and allocate a child job to handle the import with enough disk space.
```
adamnovak, line 4006: We might need to indent this for the docs to parse it right.
```suggestion
               streaming, so if true, assume streaming works and don't give the worker a lot of disk space to work with.
               If streaming fails, the worker will run out of resources and allocate a child job to handle the import with enough disk space.
```

src/toil/job.py
```diff
@@ -3758,6 +3776,374 @@ def getUserScript(self):
...
+            # the worker enough disk space
+            # OSError 28 is no space left on device
+            if e.errno == 28 and self.stream is True:
+                non_streaming_import = WorkerImportJob(
+                    self.filenames, self.disk_size, stream=False
+                )
+                self.addChild(non_streaming_import)
+                return non_streaming_import.rv()
+            else:
+                raise
```
adamnovak, line 4062: I don't think this is a good approach; we don't always sandbox jobs to keep them within their requested disk space, so if one job goes over its disk request it can make a *different* job fail due to not having enough space for *its* temporary files.

This might honestly be better if we *didn't* handle the case where the attempted streaming import went over its disk space limit, and just made it the user's problem to set a disk space limit big enough if they have any imports that can't actually stream?

Or we could implement a flag on the job store import method to only allow streaming and fail if streaming is not possible, and catch *that* and do the fallback.

Also, if we hit the disk space limit after importing several files already, when do we delete those imported copies? It looks like we will leave them behind in the job store and then re-import the same files non-streaming with more disk space.

src/toil/job.py
adamnovak: We have another empty basename check in the WDL-specific code; is this one maybe redundant?

adamnovak (changes requested):
I still don't think the idea of hoping to notice via exception whether we have exhausted our disk space is going to work. Some backends just kill the job instead of giving you an exception, whereas others let you plow right through your disk limit and interfere with other jobs. (Usually other jobs in the *workflow*, of which there shouldn't really be any of note during file import, but it's still not a thing jobs are meant to knowingly do.)

I am also still dubious of the `extract_file_uri_once` design being the best approach there (who would think to themselves "I want to get the file URI but only if it hasn't already been put in the cache"?). But it kind of has to be the shape it is to be mapped over the files, so maybe it really is the best we can do?

src/toil/job.py
stxue1: I think I'll leave it up to the user to deal with if this is too complicated to support.

adamnovak (approved):
I think that this looks good now.

src/toil/cwl/cwltoil.py
```diff
@@ -3598,20 +3809,23 @@ def visitSteps(
         # For workflows we need to dispatch on steps
         for step in cmdline_tool.steps:
             # Handle the step's tool
-            op(step.tool)
+            ret = op(step.tool)
             # Recures on the embedded tool; maybe it's a workflow.
-            visitSteps(step.embedded_tool, op)
+            recurse_ret = visitSteps(step.embedded_tool, op)
+            ret.extend(recurse_ret)
+            return ret
```
DailyDreaming, line 3816: @stxue1 I'm not sure that this returns a list?  Is this intended to always return the first (and only the first) step?  Maybe this shouldn't be a loop if so, and `cmdline_tool.steps[0]` used, just to make it clear.

DailyDreaming (dismissed):

src/toil/cwl/cwltoil.py
stxue1: To be honest I'm not too certain why this is done in a loop and returned after the first step. `cmdline_tool.steps[0]` would probably work too, except it seems like it can be possible for the cmdline_tool to have no steps and that we should error in that case. As long as the `cmdline_tool` has steps though, the funciton should always return a list though

adamnovak:
It looks like some of the CWL conformance tests fail with this:

```
Test 186 failed: toil-cwl-runner --clean=always --logDebug --statusWait=10 --retryCount=2 --relax-path-checks --eval-timeout=600 --caching=False --bypass-file-store --setEnv=SINGULARITY_DOCKER_HUB_MIRROR --scale=0.1 --outdir=/tmp/tmpcry3fq5m --quiet tests/dynresreq-workflow-tooldefault.cwl tests/empty.json
Within a workflow, test accessing the size attribute of an optional input File as part of a CommandLineTool's ResourceRequirement calculation. The CommandLineTool input has a default value (a local file) and the workflow nor the workflow step does not provide any value for this input.
Returned non-zero
```

src/toil/cwl/cwltoil.py
stxue1: I think I forgot to visit all steps, I moved the return statement down so it should now return the *right* list

PR #5142, opened 2024-10-25
Add FTP support to jobstores

### LINKED ISSUE #5134: Add FTP support to `AbstractJobStore._get_size`
`_get_size` currently doesn't support FTP as it requires special [commands and wizardry](https://stackoverflow.com/questions/22090001/get-folder-size-using-ftplib/22093848#22093848).

This will also need to support extracting username/password from the URL and/or supporting environmental variables to override the username/password

This will allow FTP files to work with #5133

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1660)
┆Issue Number: TOIL-1660

stxue1:
Closes #5134 

Also turns all url_exists requests and size requests to HEAD requests instead, as mentioned here: https://github.com/DataBiosphere/toil/issues/5114#issuecomment-2392527695

In my local testing, this significantly improves FTP performance. (Previously an existence check on a small text file on an insecure FTP server took around 5 seconds, though I'm unsure why it would take so long originally)

## Changelog Entry

 * Added proper FTP support for jobstores
 * URL existence and size gets/checks are now done with HEAD requests

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/jobStores/abstractJobStore.py
```diff
@@ -1958,3 +1986,142 @@ def _get_is_directory(cls, url: ParseResult) -> bool:
     def _list_url(cls, url: ParseResult) -> list[str]:
         # TODO: Implement HTTP index parsing and FTP directory listing
         raise NotImplementedError("HTTP and FTP URLs cannot yet be listed")
+
+
+class FtpFsAccess:
+    """
+    FTP access with upload.
+
+    Taken and modified from https://github.com/ohsu-comp-bio/cwl-tes/blob/03f0096f9fae8acd527687d3460a726e09190c3a/cwl_tes/ftp.py#L8
```
stxue1, line 1995: The license at cwl-tes is also Apache 2.0. Should I add a section to the Toil license that specifically states all code under this class is under cwl-tes's Apache 2.0 license?

docs/appendices/environment_vars.rst
```diff
@@ -217,6 +217,17 @@ There are several environment variables that affect the way Toil runs.
...
 +----------------------------------+----------------------------------------------------+
+| TOIL_FTP_USER                    | The FTP username to override all FTP logins with   |
+|                                  | Example: ``TOIL_FTP_USER=ftp_user``                |
++----------------------------------+----------------------------------------------------+
+| TOIL_FTP_PASSWORD                | The FTP password to override all FTP logins with   |
+|                                  | Example: ``TOIL_FTP_PASSWORD=ftp_password``        |
++----------------------------------+----------------------------------------------------+
+| TOIL_FTP_INSECURE                | Enable or disable connecting to all FTP            |
+|                                  | connections in insecure mode                       |
+|                                  | (``False`` by default).                            |
```
stxue1, line 228: It might be better to leave this as insecure by default instead because that was the default behavior before

src/toil/jobStores/abstractJobStore.py
mr-c: You don't need to repeat the license.

If there are code quotes, then you could mention the copyright of the original authors / copyright holders:

Copyright 2018 EMBL-EBI (Michael R. Crusoe's original implementation, written under contract)
Copyright 2019 adamstruck <adamjstruck@gmail.com>
Copyright 2020 Joris Vankerschaver <jvankerschaver@enthought.com>
Copyright 2020 Dab Brill <dbrill@enthought.com>
Copyright 2020 Manabu ISHII <manabu.ishii.rb@gmail.com>

Otherwise if this implementation is different enough, you don't need to mention the authors.

I suggest this metric: if we wanted to change the license of Toil, would you have to get permission from these copyright holders because the code is essentially the same? Or is the code so adapted or different for other reasons that it is a new work?

adamnovak: One approach is just add their Apache 2.0 license's copyright line to Toil's LICENSE file along with the others.

You could also drop it in a different file and import it to here, and in that file have `Copyright 2017 Oregon Health and Science University` and the copyright boilerplate we use for our files. I think.

Also the `#8` at the end of this link is pointing to the wrong line number.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1960,3 +1988,142 @@ def _get_is_directory(cls, url: ParseResult) -> bool:
...
+    """
+
+    def __init__(
+        self, cache: Optional[dict[Any, ftplib.FTP]] = None, insecure: bool = False
+    ):
```
adamnovak, line 2002: We should have docstrings for this and all the other functions in the class, ideally. It's especially important with a parameter called `insecure`; what is not being secured, and what is it not being secured from? Does it mean we're vulnerable to password sniffing and someone tampering with the FTP data in transit, or are we now executing arbitrary code from every file we download, or what?

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1960,3 +1988,142 @@ def _get_is_directory(cls, url: ParseResult) -> bool:
...
+            ftp = ftplib.FTP_TLS()
+            ftp.set_debuglevel(1 if logger.isEnabledFor(logging.DEBUG) else 0)
+            ftp.connect(host)
+            env_user = os.getenv("TOIL_FTP_USER")
+            env_passwd = os.getenv("TOIL_FTP_PASSWORD")
+            if env_user:
+                user = env_user
+            if env_passwd:
+                passwd = env_passwd
+            ftp.login(user or "", passwd or "", secure=not self.insecure)
```
adamnovak, line 2094: Is this using an undocumented `secure` parameter? What does that do?

docs/appendices/environment_vars.rst
adamnovak: I don't know what "insecure mode" means.

Is the user going to need to set this if their FTP server doesn't support SSL? Or to bypass bad SSL certificates while non-SSL FTP servers still work fine?

adamnovak (changes requested):
I think we need some more documentation about how FTP-over-SSL (which I think is rather unusual) is being used here, and how we're handling non-SSL FTP servers vs. those without certificates (which themselves might actually be downgrade attacks).

We need to write to a user who doesn't necessarily know that FTP can happen over SSL.

docs/appendices/environment_vars.rst
stxue1: insecure in this case means FTP without SSL. Though I'm not sure if bypassing bad SSL certificates is supported

src/toil/jobStores/abstractJobStore.py
stxue1: I think it's easier to do the latter by putting the code in a separate file, so I'll do that

src/toil/jobStores/abstractJobStore.py
stxue1: ftp in this case is an FTP_TLS object, which supports connecting over TLS. It's not on the official FTP_TLS python documentation, nor is there an associated docstring, but the code itself simply sets up an SSL connection to the FTP server

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1872,15 +1874,28 @@ class JobStoreSupport(AbstractJobStore, metaclass=ABCMeta):
     stores.
     """
 
+    ftp = None
+
+    @classmethod
+    def _setup_ftp(cls) -> None:
```
adamnovak, line 1880: It's weird that the `ftp` member is public, but the function that must be called before it is actually used has an underscore prefix.

src/toil/jobStores/ftp_utils.py
adamnovak, line 1: This might be better placed in `src/toil/lib`.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1872,15 +1874,28 @@ class JobStoreSupport(AbstractJobStore, metaclass=ABCMeta):
     stores.
     """
 
+    ftp = None
```
adamnovak, line 1877: If we keep a class-level `FtpFsAccess` instance that is never destroyed, and the `FtpFsAccess` instance caches open FTP connections to servers, when do we close our FTP connections? It might be never, and that's probably not what we want.

src/toil/jobStores/ftp_utils.py
```diff
@@ -0,0 +1,214 @@
...
+    def __init__(
+        self, cache: Optional[dict[Any, ftplib.FTP]] = None, insecure: bool = False
+    ):
+        """
+        FTP object to handle FTP connections. By default, connect over FTP with TLS.
+
+        :param cache: cache of generated FTP objects
+        :param insecure: Whether to connect over FTP with TLS
+        """
+        self.cache = cache or {}
```
adamnovak, line 46: We probably want to be able to turn off using a connection cache at all, if we're going to make an FtpFsAccess that is never destroyed.

src/toil/jobStores/ftp_utils.py
```diff
@@ -0,0 +1,214 @@
...
+            ftp = ftplib.FTP_TLS()
+            ftp.set_debuglevel(1 if logger.isEnabledFor(logging.DEBUG) else 0)
+            ftp.connect(host)
+            env_user = os.getenv("TOIL_FTP_USER")
+            env_passwd = os.getenv("TOIL_FTP_PASSWORD")
+            if env_user:
+                user = env_user
+            if env_passwd:
+                passwd = env_passwd
+            ftp.login(user or "", passwd or "", secure=not self.insecure)
```
adamnovak, line 169: OK, so according to the obviously reliable https://tedboy.github.io/python_stdlib/_modules/ftplib.html#FTP_TLS.login what `secure` does here is call `auth()` if you haven't already set up SSL/TLS on the connection, before doing the login.

So it looks like unless the user sets `TOIL_FTP_USE_SSL=False` themselves, they're going to get an error here whenever they use an `ftp://` URL, unless the server actually supports SSL/TLS.

One server people might want to use is `ftp://ftp-trace.ncbi.nlm.nih.gov`. I've just checked and it doesn't support FTP over SSL/TLS.

Another thing we don't do is call the [`prot_p()` method](https://docs.python.org/3/library/ftplib.html#ftplib.FTP_TLS.prot_p) on the connection to actually secure the data channel. And later we fall back to `urlopen()` and give it the full FTP URL with username and password, and I don't think that `urlopen()` can use or enforce SSL. We'd have to tinker with `urlopen`'s handler as recommended in https://stackoverflow.com/a/73171311.

So I think that by default we *should* allow unencrypted FTP connections without user intervention to accept them. That's the current behavior.

If we want to change Toil to enforce secure FTP by default, then we also need to enforce security on the actual reads. And we probably would need a way to individually allow particular insecure servers?

adamnovak (changes requested):
I think we should cut the feature where we require user approval for an insecure FTP connection. (We might also want to then cut/flag off .netrc support?) If we do implement that, we also need to implement it for the actual requests, not just pass off to `urllib`.

I also think we should address the cached FTP connections that seem to live forever. Maybe we want them to time out somehow? Or maybe we can just bypass the cache or not store a persistent `FtpFsAccess` in a class-level variable.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1872,18 +1874,29 @@ class JobStoreSupport(AbstractJobStore, metaclass=ABCMeta):
     stores.
     """
 
+    @classmethod
+    def _setup_ftp(cls) -> FtpFsAccess:
+        return FtpFsAccess(insecure=strtobool(os.environ.get('TOIL_FTP_USE_SSL', 'False')) is False)
```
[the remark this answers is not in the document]
stxue1: I can leave a comment that this doesn't reuse FTP connections, and that implementing it later may be a good idea

src/toil/lib/ftp_utils.py
```diff
@@ -0,0 +1,230 @@
...
+        Open an FTP url.
+
+        Only supports reading, no write support.
+        :param fn: FTP url
+        :param mode: Mode to open FTP url in
+        :return:
+        """
+        if "r" in mode:
+            host, port, user, passwd, path = self._parse_url(fn)
+            handle = urlopen("ftp://{}:{}@{}:{}/{}".format(user, passwd, host, port, path))
```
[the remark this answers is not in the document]
stxue1: Removed to put into another issue as it appears to be too large of an issue

adamnovak (approved):
Without the SSL enforcement feature I think this looks OK.

PR #5211, opened 2025-02-13
Make tests more independent

mr-c:
Also fixes https://github.com/DataBiosphere/toil/issues/3996 

## Changelog Entry

 * tests now load their data more flexibly
 
## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/test/__init__.py
+++ b/src/toil/test/__init__.py
@@ -13,6 +13,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from contextlib import contextmanager, AbstractContextManager
 import datetime
 import logging
 import os
@@ -29,7 +30,9 @@ import zoneinfo
 from abc import ABCMeta, abstractmethod
 from collections.abc import Generator
 from contextlib import contextmanager
+from importlib.resources import as_file, files
 from inspect import getsource
+from pathlib import Path
 from shutil import which
 from tempfile import mkstemp
 from textwrap import dedent
@@ -49,7 +52,28 @@ from toil.lib.memoize import memoize
 from toil.lib.threading import ExceptionalThread, cpu_count
 from toil.version import distVersion
 
+import pytest
+
 logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def _fallback_get_data(filename: str) -> Generator[Path]:
+    try:
+        yield Path(os.path.dirname(__file__)) / ".." / ".." / ".." / filename
+    finally:
+        pass  # no cleanup needed
+
+
+def get_data(filename: str) -> AbstractContextManager[Path]:
+    """Returns an absolute path for a file from this package."""
+    # normalizing path depending on OS or else it will cause problem when joining path
+    filename = os.path.normpath(filename)
+    try:
+        return as_file(files("toil") / filename)
+    except ModuleNotFoundError:
+        pass
+    return _fallback_get_data(filename)
 
 
 class ToilTest(unittest.TestCase):
@@ -370,6 +394,11 @@ def needs_online(test_item: MT) -> MT:
     return test_item
 
 
+pneeds_online = pytest.mark.skipif(
+    os.getenv("TOIL_SKIP_ONLINE", "").lower() == "true", reason="Skipping online test."
+)
+
+
 def needs_aws_s3(test_item: MT) -> MT:
     """Use as a decorator before test classes or methods to run only if AWS S3 is usable."""
     # TODO: we just check for generic access to the AWS account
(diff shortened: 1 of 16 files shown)


src/toil/test/__init__.py
```diff
@@ -52,6 +55,24 @@
...
+    try:
+        file_manager = ExitStack()
+        atexit.register(file_manager.close)
+        traversable = files("toil") / filename
+        filepath = file_manager.enter_context(as_file(traversable))
+    except ModuleNotFoundError:
+        pass
+    if not filepath or not os.path.isfile(filepath):
+        filepath = Path(os.path.dirname(__file__)) / ".." / ".." / ".." / filename
+    return str(filepath.resolve())
```
adamnovak, line 73: Should `_projectRootPath()` and `ToilTest.rootDir` be removed in favor of this as the One True Way to access test files? (And if this is the One True Way, the docstring should note that so we know nwver to not use this function.)

Should this maybe be a member of ToilTest so it doesn't need a separate import and more sensibly replaces `self.rootDir` references?

Also, if the source tree is laid out so that a CWL file references another file nearby, and I `get_data("whatever.cwl")`, does `get_data()` promise that the path it returns will have those nearby files from the source tree still available, without explicitly being asked for them?

adamnovak: Also, are we sure that the atexit trickery to defeat the context manager is worth it? I don't see many tests using more than 2 `get_data()` calls, so even if it was a context manager you had to be in for each file, the nesting wouldn't be that bad. If we had a way to say `with get_datas(path1, path2, path3) as (file1, file2, file3):` for any number of arguments we could limit the nesting without fighting the library.

adamnovak (changes requested):
I want to know that this isn't able to ship arbitrary files that happen to be on disk in the builds before merging. But other than that, even though I think the design of `get_data()` might be able to be better, I think this is a clear improvement.

src/toil/test/__init__.py
mr-c: > Should `_projectRootPath()` and `ToilTest.rootDir` be removed in favor of this as the One True Way to access test files? (And if this is the One True Way, the docstring should note that so we know nwver to not use this function.)

Sure, eventually

> Should this maybe be a member of ToilTest so it doesn't need a separate import and more sensibly replaces `self.rootDir` references?

No, I'd like to continue migrating away from the `ToilTest` which inheritest from `unitttest.TestCase` and is incompatible with pytest fixtures.

> Also, if the source tree is laid out so that a CWL file references another file nearby, and I get_data("whatever.cwl"), does get_data() promise that the path it returns will have those nearby files from the source tree still available, without explicitly being asked for them?

Only in Python 3.12+, otherwise not when the `toil` module is loaded from a wheel, no. In that case, only the requested file is extracted, and requesting an entire directory doesn't seem to work:

See https://docs.python.org/3/library/importlib.resources.html#importlib.resources.as_file

Python 3.11:
``` python
$ PYTHONPATH=/home/michael/src/toil/dist/toil-8.2.0a1-py3-none-any.whl python3
>>> import importlib.resources
>>> with importlib.resources.as_file(importlib.resources.files("toil") / "test/cwl/") as p1:
...     for c in p.iterdir():
...             print(c)
... 
  File "/usr/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/resources/_common.py", line 80, in _tempfile
    os.write(fd, reader())
                 ^^^^^^^^
  File "/usr/lib/python3.11/zipfile.py", line 2464, in read_bytes
    with self.open('rb') as strm:
         ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/zipfile.py", line 2422, in open
    raise IsADirectoryError(self)
IsADirectoryError: /home/michael/src/toil/dist/toil-8.2.0a1-py3-none-any.whl/toil/test/cwl/
```

Python 3.13:
``` python
$ PYTHONPATH=/home/michael/src/toil/dist/toil-8.2.0a1-py3-none-any.whl python
>>> with importlib.resources.as_file(importlib.resources.files("toil") / "test/cwl/") as p:
...     for c in p.iterdir():
...         print(c)
...         
/tmp/tmp1tvq5n1u/cwl/2.fasta
/tmp/tmp1tvq5n1u/cwl/2.fastq
/tmp/tmp1tvq5n1u/cwl/__init__.py
/tmp/tmp1tvq5n1u/cwl/alwaysfails.cwl
[…]
```

> Also, are we sure that the atexit trickery to defeat the context manager is worth it? I don't see many tests using more than 2 `get_data()` calls, so even if it was a context manager you had to be in for each file, the nesting wouldn't be that bad. If we had a way to say `with get_datas(path1, path2, path3) as (file1, file2, file3):` for any number of arguments we could limit the nesting without fighting the library.

Good point, I updated this PR to not need the atexit trickery

PR #5228, opened 2025-03-06
Delete files that will longer be accessible at the end of workflows and tasks

adamnovak:
This will fix part of #4872 by allowing the WLD interpreter to detect some types of files that can no longer be used, and delete them:
1. Files created in the `inputs` section or body of a workflow that are not output will be deleted when the workflow ends.
2. Files created in the `input:` block of a call (which the enclosing workflow can never get at) will be deleted when the called task or workflow ends.
3. Files created in the `runtime` section of a task (which probably has no use case, but which is technically possible) will be deleted when the task ends.

This isn't a full answer for #4872, because it doesn't delete files as soon as nothing is planning to use them, but I managed to actually write it while my full treatment is seemingly taking forever, so we might be able to get away with just having this and telling users to use subworkflows when they need to control scope or delete large files.

## Changelog Entry

 * Toil now deletes files that go out of scope at the ends of workflows and tasks.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/test/wdl/testfiles/drop_files.wdl
```diff
@@ -0,0 +1,62 @@
+version 1.0
+
+import "drop_files_subworkflow.wdl" as subwf
+
+# Workflow for testing discarding un-output files a tthe end of a workflow.
```
DailyDreaming, line 5: `a tthe end`

src/toil/test/wdl/wdltoil_test.py
```diff
@@ -257,6 +257,84 @@ def test_wait(self):
...
+        assert "This file is created in a runtime section" not in all_file_values
+        assert "This task file is not used" not in all_file_values
+        assert "This file should be discarded" not in all_file_values
+        assert "This file is dropped by a subworkflow" not in all_file_values
+        assert "This file gets stored in a variable" not in all_file_values
+        assert "This file never gets stored in a variable" not in all_file_values 
+        
+        # These are created inside the workflow and output
+        assert "3" in all_file_values
+        assert "This file is collected as a task output twice"
```
DailyDreaming, line 323: This is always True.  I think this needs `in all_file_values` appended.

DailyDreaming (approved):
One minor issue.  I like the test!

PR #5234, opened 2025-03-13
Attempt to fix the CWL conformance test badge

mr-c:
- cwl conformance testing: fix doc string
- cwl conformance tests: run with --bypass-file-store

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -303,7 +303,7 @@ cwl_badge:
     - make download_cwl_spec
     - rm -rf badges1.2
     # We have to run this on one thread to avoid a race condition: https://github.com/common-workflow-language/cwltest/issues/239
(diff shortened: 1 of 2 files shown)


mr-c:
To access the JSON results for a specific branch (but not a specific job ID) via the GitLab API, then the pipeline must succeed

> Download a single file from a job’s artifacts in the latest **successful** pipeline using the reference name. 

https://docs.gitlab.com/api/job_artifacts/#download-a-single-artifact-file-from-specific-tag-or-branch

adamnovak:
Is the idea here that bypassing the file store will make all the tests pass (because there's then no unsupported features), which then would somehow fix badge references?

We have the pipeline *step* allowed to fail when the tests for unsupported features fail, which means the *pipeline* should still succeed and the badges should still be accessible.

mr-c:
When merged, this will fix the CWL conformance badges.

Example: Toil CWL 1.2 Required Conformance Badge

mr-c:
> Is the idea here that bypassing the file store will make all the tests pass (because there's then no unsupported features), which then would somehow fix badge references?

Yes

> We have the pipeline _step_ allowed to fail when the tests for unsupported features fail, which means the _pipeline_ should still succeed and the badges should still be accessible.

It isn't enough, for the particular API you are using, the pipeline/job must succeed

adamnovak:
@stxue1 Can you review this one?

adamnovak (approved):
I'm not sure we need this *and* the test being allowed to fail, which if you're right about the API then doesn't help us at all. But this would get us a 100% on all the optional features I think, which we might like to see.

Issue #4915, opened 2024-05-03
Spike: Determine work necessary to get rid of Docker in Docker


stxue1:
Docker in docker was originally made for Mesos, but most batchsystems nowaday are not as compatible with our current Docker in Docker configuration; there shouldn't be much reason to run the workflow inside a container that is contained inside another container. This is causing some issues with toil-wdl-runner and toil-cwl-runner whenever they try to run an image. If the appliance container isn't given the right permissions, the runners will fail; for example, Funnel wants to run all docker images as read-only with no permissions given for user namespaces, meaning Singularity cannot run. I think this will cause future issues with batch system plugins if they have specific requirements for running docker containers.

@adamnovak How hard do you think it would be to get rid of DinD, and do you think it would be a good idea to implement this?

adamnovak:
We didn't set it up Docker in Docker for Mesos, as far as I know. On Mesos we use the `docker` client in the Toil container, but the actual invoked container runs as a sibling, on the host's Docker daemon. We make sure that paths like `/var/lib/toil` and `/tmp` are mounted into the Toil container form the host at the same paths as they appear on the host, so that when Toil runs `docker -v /tmp/somefile:/some/mount/point`, the `/tmp/somefile` it is talking about that it sees from its own container is also the one that the Docker daemon it is talking to will see and mount.

One approach to get rid of Docker-in-Docker would be to use this sibling container approach in places other than Mesos. This would be fairly straightforward, when it can work. But it can only work when the Toil container can be given direct access to the host's Docker daemon, and it depends on having these paths mounted from the host into the Toil container. I don't think that many Funnel setups will allow you to run a TES task with straight access to the host Docker daemon. It also makes resource accounting and limits difficult: it's no use to limit the memory usage of the Toil container if actually all the memory is used by a sibling Docker container that it started.

Another approach to getting rid of Docker-in-Docker would be to get rid of the Toil container entirely, and only run the container that the workflow is asking to run. This is significantly more challenging; instead of being able to distribute a Toil docker container with all Toil's dependencies installed, we'd need to find a way to inject enough Toil machinery to e.g. download inputs from and upload outputs to a Google job store into basically arbitrary Docker containers. Right now we inject *some* stuff into containers via the command to be run, as does MiniWDL. But it would be challenging to, for example, install Python 3.12 and Toil itself into a (read-only, on TES) Docker container just by tacking stuff on to the front of a Bash script.

[...]

PR #5269, opened 2025-04-03
Add ability to edit security context in Kubernetes

### LINKED ISSUE #5179: Add security context to Kubernetes batch system
Kubernetes clusters are moving to increased security through Pod Security Admission. It would be great, if this could be set as parameters so that K8s Jobs can run even in such restricted environments.

In particular, we need to set the Pod security context:
```
securityContext: 
  fsGroupChangePolicy: OnRootMismatch
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
```

and Container securityContext:

```
securityContext: 
  runAsUser: 1000
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1686)
┆Issue Number: TOIL-1686

stxue1:
Resolves #5179 

This allows the kubernetes pods and containers to have their security context changed, but *doesn't guarantee* that Toil will work with the new specified security context.

I think for certain restrictive security contexts, if we want to run Singularity within Docker, #4915 needs to be resolved first.
## Changelog Entry

 * Added command line options `--kubernetesPodSecurityContext` and `--kubernetesSecurityContext` to change the security context of pods and containers

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/batchSystems/kubernetes.py
```diff
@@ -803,6 +809,26 @@ def _check_accelerator_request(self, requirer: Requirer) -> None:
                     ],
                 )
 
+    class FakeResponse:
```
adamnovak, line 812: It might be nice to have some documentation about how this is part of the hack for deserializing. Maybe we could put the class in the function actually without causing problems? We aren't really letting the objects of the class escape, and it would be more understandable.

src/toil/batchSystems/kubernetes.py
```diff
@@ -803,6 +809,26 @@ def _check_accelerator_request(self, requirer: Requirer) -> None:
...
+    T = TypeVar('T')
+    def _load_kubernetes_object(self, file: str, cls: type[T]) -> T:
+        """
+        Deserialize a YAML representation into a Kubernetes object
+        :param file: Path to YAML file
+        :param cls: Kubernetes API model type for deserialized object
+        :return: Deserialized object
+        """
+        YAML = yaml.YAML(typ='safe')
+        object_def = YAML.load(open('container.yaml').read())
```
adamnovak, line 824: Reading and parsing the file for every pod might not be super efficient; we *could* read the files and make the objects once at startup and just re-use them, since we never mutate them except for setting the privileged flag, which we could also do at startup. I'm not sure how much faster that would be, or whether Kubernetes lets us share parts of the object graph like that.

adamnovak (approved):
I think this will work!

PR #5285, opened 2025-04-18
Removing redundanat mount points from listings when loading inputs to a job

### LINKED ISSUE #5099: toil ignores LoadListingRequirement
When using `toil-cwl-runner` it seems like it always uses `deep_listing` when files or directories are mounted for running within Docker or Singularity.
If I try to add  [LoadListingRequirement](https://www.commonwl.org/v1.2/CommandLineTool.html#LoadListingRequirement) and set it to `shallow_listing` this simply gets ignored. If I use `cwltool` it works as expected.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1649)
┆Issue Number: TOIL-1649

stxue1:
Closes #5099 

I am a little worried in the case where the user would purposefully define some directory with listings and creates a workflow/task that depends on these redundant listings to exist. It seems a little unlikely though.

## Changelog Entry

 * Remove redundant listings when loading inputs to a job

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -193,6 +193,78 @@ class SkipNull:
     """
 
 
+def trim_mounts_op_down(file_or_directory: CWLObjectType) -> None:
+    return
+
+
+def sniff_location(file_or_directory: CWLObjectType) -> Optional[str]:
+    """
+    Return None if we don't have a local path or file URI
+    :param file_or_directory:
+    :return:
+    """
+    # Since we only consider mountable paths, if path is not file URI or bare path, don't consider it
+    path = cast(str, file_or_directory.get('location') or file_or_directory.get('path'))
+    scheme = urlparse(path).scheme
+    if scheme == '' or scheme == 'file':
+        return path
+    else:
+        return None
+
+
+def trim_mounts_op_up(file_or_directory: CWLObjectType, op_down_ret: None, child_results: list[bool]) -> bool:
+    """
+    Nonredundant for something in a directory means its path or location is not within the parent directory or doesn't match its basename
+    Nonredundant for something in a secondary file means its path or location is not adjacent to the primary file or doesn't match its basename
+
+    If on a File:
+    Returns True if anything in secondary files is nonredundant or has nonredundant children to this file, false otherwise
+    If on a Directory:
+    Returns True if anything in top level listing is nonredundant or has nonredundant children, otherwise false.
+    If something in the listing is redundant and all children are redundant, then delete it
+    :param file_or_directory: CWL file or CWL directory type
+    :return: boolean
+    """
+    if file_or_directory.get('location') is None and file_or_directory.get('path') is None:
+        # file or directory is defined by contents or listing respectively, this is not redundant
+        return True
+    own_path = sniff_location(file_or_directory)
+    if own_path is None:
+        return True
+    # basename should be set as we are the implementation
+    own_basename = cast(str, file_or_directory['basename'])
+
+    # If the basename does not match the path, then this is nonredundant
+    if not own_path.endswith("/" + own_basename):
+        return True
+
+    if file_or_directory['class'] == 'File':
+        if any(child_results):
+            # one of the children was detected as not redundant
+            return True
+        for secondary in cast(MutableSequence[MutableMapping[str, CWLOutputType]], file_or_directory.get('secondaryFiles', [])):
+            # secondary files should already be flagged nonredundant if they don't have either a path or location
+            secondary_path = sniff_location(secondary)
+            secondary_basename = cast(str, secondary['basename'])
+            # If we swap the secondary basename for the primary basename in the primary path, and they don't match, then they are nonredundant
+            if os.path.join(own_path[:-len(own_basename)], secondary_basename) != secondary_path:
+                return True
+    else:
+        listings = cast(MutableSequence[MutableMapping[str, CWLOutputType]], file_or_directory.get('listing', []))
+        if len(listings) == 0:
+            return False
+        # We assume child_results is in the same order as the directory listing
+        # iterate backwards to avoid iteration issues
+        for i in range(len(listings) - 1, -1, -1):
+            if child_results[i] is False:
+                if os.path.join(own_path, cast(str, listings[i]['basename'])) == sniff_location(listings[i]):
+                    del listings[i]
+        # If one of the listings was nonredundant, then this directory is also nonredundant
+        if any(child_results):
+            return True
+    return False
+
+
 def filter_skip_null(name: str, value: Any) -> Any:
     """
     Recursively filter out SkipNull objects from 'value'.
@@ -2750,6 +2822,9 @@ class CWLJob(CWLNamedJob):
         logger.debug("Loaded order:\n%s", self.cwljob)
 
         cwljob = resolve_dict_w_promises(self.cwljob, file_store)
+
+        # Deletes duplicate listings
+        visit_cwl_class_and_reduce(cwljob, ["Directory", "File"], trim_mounts_op_down, trim_mounts_op_up)
 
         if self.conditional.is_false(cwljob):
             return self.conditional.skipped_outputs()


adamnovak:
> I am a little worried in the case where the user would purposefully define some directory with listings and creates a workflow/task that depends on these redundant listings to exist.

I think if the task or workflow needs the listings filled in, it needs to have `LoadListingRequirement` set.

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
     """
 
 
+def trim_mounts_op_down(file_or_directory: CWLObjectType) -> None:
```
adamnovak, line 196:
```suggestion
def trim_mounts_op_down(file_or_directory: CWLObjectType) -> None:
    """
    No-op function for mount-point trimming.
    """
```

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
...
+    return
+
+
+def sniff_location(file_or_directory: CWLObjectType) -> Optional[str]:
+    """
```
adamnovak, line 201:
```suggestion
    """
    Get the local bare path or file: URI for a CWL file or directory, or None.
    
```

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
...
+    Return None if we don't have a local path or file URI
+    :param file_or_directory:
+    :return:
```
adamnovak, line 204:
```suggestion
    :return: None if we don't have a local path or file URI
```

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
...
+    scheme = urlparse(path).scheme
+    if scheme == '' or scheme == 'file':
+        return path
+    else:
+        return None
+
+
+def trim_mounts_op_up(file_or_directory: CWLObjectType, op_down_ret: None, child_results: list[bool]) -> bool:
+    """
+    Nonredundant for something in a directory means its path or location is not within the parent directory or doesn't match its basename
```
adamnovak, line 217: The first line in the docstring should summarize the function.

```suggestion
    Remove subtrees of the CWL file or directory object tree that only have redundant stuff in them.

    Nonredundant for something in a directory means its path or location is not within the parent directory or doesn't match its basename
```

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
...
+    if file_or_directory['class'] == 'File':
+        if any(child_results):
+            # one of the children was detected as not redundant
+            return True
+        for secondary in cast(MutableSequence[MutableMapping[str, CWLOutputType]], file_or_directory.get('secondaryFiles', [])):
+            # secondary files should already be flagged nonredundant if they don't have either a path or location
+            secondary_path = sniff_location(secondary)
+            secondary_basename = cast(str, secondary['basename'])
+            # If we swap the secondary basename for the primary basename in the primary path, and they don't match, then they are nonredundant
+            if os.path.join(own_path[:-len(own_basename)], secondary_basename) != secondary_path:
```
adamnovak, line 250: What if one of these is a file: URI and one is a bare path, but both refer to the same place? We won't get a string match but they'll still be equal.

src/toil/cwl/cwltoil.py
```diff
@@ -193,6 +193,78 @@ class SkipNull:
...
+        return True
+    own_path = sniff_location(file_or_directory)
+    if own_path is None:
+        return True
+    # basename should be set as we are the implementation
+    own_basename = cast(str, file_or_directory['basename'])
+
+    # If the basename does not match the path, then this is nonredundant
+    if not own_path.endswith("/" + own_basename):
+        return True
```
adamnovak, line 239: We haven't URL-decoded `own_path` for the case where it's a file: URI, so we will think any filename where there are characters that need to be encoded in file: URIs is nonredundant, even when  it's not.

We might want to make `sniff_location()` handle the conversion to bare, URL-decoded paths, so we can work with only those in here.

adamnovak (changes requested):
I don't think this will work right for files whose names need to be URL-encoded in file URIs (i.e. they contain `#` or `?` or similar characters that URLs reserve).

It would be good to have a bench test for the effects of `visit_cwl_class_and_reduce(cwljob, ["Directory", "File"], trim_mounts_op_down, trim_mounts_op_up)` on some handcrafted edge cases, independent of running any workflow, so we can be confident that it works on secondary files, weird filenames, and cases where we mix URIs and bare paths in the same object tree.

src/toil/cwl/cwltoil.py
```diff
@@ -2751,6 +2750,9 @@ def run(self, file_store: AbstractFileStore) -> Any:
 
         cwljob = resolve_dict_w_promises(self.cwljob, file_store)
 
+        # Deletes duplicate listings
+        visit_cwl_class_and_reduce(cwljob, ["Directory", "File"], trim_mounts_op_down, trim_mounts_op_up)
+
```
adamnovak, line 2755: If we made this into a small function, we could keep the complex concept of "down" and "up" operations for mount trimming from crossing a module boundary, which could make the code more understandable. The down and up functions could even become internal functions and not be floating around at module scope.

src/toil/cwl/utils.py
```diff
@@ -219,3 +221,80 @@ def download_structure(
...
+
+
+def sniff_location(file_or_directory: CWLObjectType) -> Optional[str]:
+    """
+    Get the local bare path or file: URI for a CWL file or directory, or None.
+
+    :return: None if we don't have a local path or file URI
+    """
+    # Since we only consider mountable paths, if path is not file URI or bare path, don't consider it
+    path = unquote(cast(str, file_or_directory.get('location') or file_or_directory.get('path')))
```
adamnovak, line 240: We can't `unquote` the path when it's not a URL, or we won't be able to work on files actually named e.g. `%25.txt`.

We also might want to do the unquote *after* parsing the URL. It really only would matter for cases where e.g. a slash character appears inside a query string value, which can't really apply to files, but I think it's technically correct.

src/toil/cwl/utils.py
```diff
@@ -219,3 +221,80 @@ def download_structure(
...
+    If something in the listing is redundant and all children are redundant, then delete it
+    :param file_or_directory: CWL file or CWL directory type
+    :return: boolean
+    """
+    if file_or_directory.get('location') is None and file_or_directory.get('path') is None:
+        # file or directory is defined by contents or listing respectively, this is not redundant
+        return True
+    own_path = sniff_location(file_or_directory)
+    if own_path is None:
+        return True
```
adamnovak, line 268: The `None` checks look redundant here, but we need the first one to meet the undocumented precondition on `sniff_location` that one of `location` or `path` be set.

It would be neater if we removed that precondition, and made `sniff_location` return `None` in the case where there's neither a `path` nor a `location`. Then we'd only need one check here, and we wouldn't have an undocumented way to crash `sniff_location`.

src/toil/test/cwl/cwlTest.py
```diff
@@ -1906,6 +1908,80 @@ def op_up(thing: "CWLObjectType", down_value: int, child_results: list[str]) ->
...
 
+@needs_cwl
+@pytest.mark.cwl
+@pytest.mark.cwl_small
+def test_trim_mounts_op() -> None:
+    """
+    Make sure our logic fo removing duplicate listings when loading inputs to a job work
+    """
+    structure1 = {"class": "Directory", "basename": "directory", "listing": [{"class": "File", "basename": "file", "contents": "hello world"}]}
+    visit_cwl_class_and_reduce(structure1, ["Directory", "File"], trim_mounts_op_down, trim_mounts_op_up)
```
adamnovak, line 1919: Having a function wrapping this up would save some repeated code here too.

src/toil/cwl/utils.py
```diff
@@ -219,3 +221,80 @@ def download_structure(
...
+    """
+    Get the local bare path or file: URI for a CWL file or directory, or None.
+
+    :return: None if we don't have a local path or file URI
+    """
+    # Since we only consider mountable paths, if path is not file URI or bare path, don't consider it
+    path = unquote(cast(str, file_or_directory.get('location') or file_or_directory.get('path')))
+    scheme = urlparse(path).scheme
+    if scheme == '' or scheme == 'file':
+        return path
```
adamnovak, line 243: It looks like we're letting full URIs but with the encoded characters in them decoded escape from this function.

I don't think that's a good idea; we probably want only fully-decoded bare paths and properly-encoded `file:` URIs to be used as internal formats for communication between functions. If we also start throwing around things that look like URIs but are not properly encoded, it makes it easy to write code that accidentally decodes them twice. There's no way to look at `file:/data/tmp/data-at-100%25/file.txt` and know whether it is a properly encoded URI and refers to the path `/data/tmp/data-at-100%/file.txt`, or whether it is actually a not-encoded URI-like object and actually refers to `/data/tmp/data-at-100%25/file.txt`.

So we probably want to do `urlparse(path)` first, then check the scheme, then if it's not `file` we would return path, and if it is `file` we would `unquote` the path part of the URI and return *that*, and just always have this function return a bare path.

src/toil/cwl/utils.py
```diff
@@ -219,3 +221,80 @@ def download_structure(
...
+    if file_or_directory['class'] == 'File':
+        if any(child_results):
+            # one of the children was detected as not redundant
+            return True
+        for secondary in cast(MutableSequence[MutableMapping[str, CWLOutputType]], file_or_directory.get('secondaryFiles', [])):
+            # secondary files should already be flagged nonredundant if they don't have either a path or location
+            secondary_path = sniff_location(secondary)
+            secondary_basename = cast(str, secondary['basename'])
+            # If we swap the secondary basename for the primary basename in the primary path, and they don't match, then they are nonredundant
+            if os.path.join(own_path[:-len(own_basename)], secondary_basename) != secondary_path:
```
adamnovak, line 285: This still doesn't handle the case of one thing having a URI set and the other having a bare path set, unless `sniff_location()` is changed to normalize to one or the other.

src/toil/test/cwl/cwlTest.py
```diff
@@ -1906,6 +1908,80 @@ def op_up(thing: "CWLObjectType", down_value: int, child_results: list[str]) ->
...
+                "checksum": "sha1$da39a3ee5e6b4b0d3255bfef95601890afd80709"
+            }
+        ],
+        "path": "/home/heaucques/Documents/toil/test_dir"
+    }
+    visit_cwl_class_and_reduce(structure3, ["Directory", "File"], trim_mounts_op_down, trim_mounts_op_up)
+
+    # everything except the nested directory should be removed
+    assert len(structure3['listing']) == 1
+
```
adamnovak, line 1983: The tests here don't address any of the fiddly edge cases regarding URL encoding that I'm worried about. We should have tests that focus on the parts of the code that we're not really sure about.

It would be good to have a test for:
```
{"class": "Directory", "basename": "123", "location": "file:///tmp/%25/123", "listing": [{"class": "File", "path": "/tmp/%/123/456", "basename": "456"}}
```
This should be redundant, because the directory's URL refers to the parent directory of the file's bare path.

Also:
```
{"class": "Directory", "basename": "dir", "path": "/tmp/cat%2Ftag/dir", "listing": [{"class": "File", "path": "/tmp/cat/tag/dir/file", "basename": "file"}}
```
This should be nonredundant, because the file is not under `/tmp/cat%2Ftag`.

And a test for URL encoding normalization wouldn't hurt.
```
{"class": "Directory", "basename": "dir", "location": "file:///tmp/cat%2Ftag/dir", "listing": [{"class": "File", "location": "file:///tmp/cat%2ftag/dir/file", "basename": "file"}}
```
This should be redundant because, despite the difference in capitalization, the file is located inside the directory.

adamnovak (changes requested):
I don't think the handling of URI encoding is correct here.

PR #5323, opened 2025-05-29
Make URL Access Pluggable

### LINKED ISSUE #5302: Make URL access implementations pluggable
We should have plugin support for accessing new URL types, without a full job store implementation

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1748)
┆Issue Number: TOIL-1748

haileyhemmingsen:
Adding URL plugins by changing the existing plugin system to be more generic. From there, we can apply it to the existing batch system plugins and the new URL access implementation. Documentation will be added when example plugin is added. Fixes #5302.
## Changelog Entry

* Deleted deprecated `addBatchSystemFactory`, `BATCH_SYSTEM_FACTORY_REGISTRY`, and `BATCH_SYSTEMS`. Use `add_batch_system_factory` instead.
* Added new unified plugin system
* Toil now supports URL plugins. To provide a plugin, add `toil_url_access_` at beginning of python package name. To register a new URL scheme use `register_plugin("url_access", scheme, implementation_type_factory)`.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/test/lib/url_plugin_test.py
+++ b/src/toil/test/lib/url_plugin_test.py
@@ -25,9 +25,6 @@ from toil.batchSystems.options import OptionSetter
 from toil.batchSystems.registry import add_batch_system_factory
 from toil.common import Toil, addOptions
 from toil.job import JobDescription
-from toil.test import ToilTest
-from toil.lib.plugins import remove_plugin
-
 
 import io
 from urllib.parse import ParseResult


src/toil/lib/plugins.py
```diff
@@ -0,0 +1,66 @@
+import importlib
```
adamnovak, line 1: We probably want to have the copyright header at the top here like we have in the other Toil source files, and also maybe a note (module-level docstring?) about what this file is for.

src/toil/lib/plugins.py
```diff
@@ -0,0 +1,66 @@
...
+PluginType = Literal["batch_system"] | Literal["url_access"]
+plugin_types: list[PluginType] = ["batch_system", "url_access"]
+
+_registry: dict[str, dict[str, Any]] = {k: {} for k in plugin_types}
+
+def register_plugin(
+    plugin_type: PluginType, plugin_name: str, plugin_being_registered: Any
+) -> None:
+    """
+    Adds a plugin to the registry for the given type of plugin.
```
adamnovak, line 16: Since the type annotations don't define it, it might be good to note here what the expected plugin values are meant to be for a given plugin type (i.e. functions that return a class implementing blah interface).

It also might be good to explain what `plugin_name` is used for: for patch systems it is the value the user passes to `--batchSystem` to actually use it, and for URLs it is the URL scheme value that the plugin gets called to handle. It's not *really* the "name of the plugin" like the variable name would suggest; you shouldn't pass `"My Cool Plugin"` or something.

src/toil/lib/plugins.py
```diff
@@ -0,0 +1,66 @@
...
+def get_plugin_names(plugin_type:PluginType) -> list[str]:
+    """
+    Get the names of all the available plugins.
+    """
+    _load_all_plugins(plugin_type)
+    return list(_registry[plugin_type].keys())
+
+def get_plugin(plugin_type: PluginType, plugin_name: str) -> Any:
+    """
+    Get a plugin class by name.
```
adamnovak, line 40: This sounds a little misleading to me because when I read it I think that:
1. The result will be a class, when it's really a function that returns a class when called.
2. The argument is the name of the class (like `toil_plugin_url_access_foobar.lib.plugin.MyCoolClass`), when it's really a user-facing batch system name or URL scheme.

src/toil/test/lib/url_plugin_test.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (C) 2015-2021 Regents of the University of California
```
adamnovak, line 1: This should be bumped up to 2025.

src/toil/test/lib/url_plugin_test.py
```diff
@@ -0,0 +1,108 @@
...
+class TestURLAccess(ToilTest):
+    def setUp(self) -> None:
+        super().setUp()
+        register_plugin("url_access", "fake", lambda: FakeURLPlugin)
+
+    def tearDown(self) -> None:
+        remove_plugin("url_access", "fake")
+        super().tearDown()
+
+    def test_url_exists(self):
```
adamnovak, line 85: MyPy is complaining on the CI run that a lot of these test functions are missing a `-> None` return type annotation.

src/toil/test/lib/url_plugin_test.py
```diff
@@ -0,0 +1,108 @@
...
+    @classmethod
+    def _get_is_directory(cls, url: ParseResult) -> bool:
+        return url.path.endswith("/")
+
+    @classmethod
+    def _list_url(cls, url: ParseResult) -> list[str]:
+        return ["file1.txt", "subdir/"]
+
+    @classmethod
+    def _read_from_url(cls, url: ParseResult, writable: io.BytesIO) -> tuple[int, bool]:
```
adamnovak, line 62: MyPy complained in CI that the `writable` needs to be an `IO[bytes]` like the `URLAccess` has it, and not an `io.BytesIO` which MyPy reads as slightly different.

adamnovak (changes requested):
I think this is pretty good, but it needs to be adjusted to pass tests (and the local `make mypy` which should replicate some of the test failures), and I think some of the docstrings should explain more about how it is meant to be used.

Issue #5316, opened 2025-05-18
Automatically increase Docker container memory limit when below Docker resource constraints


stxue1:
Some WDL workflows made for Cromwell cannot run when the task is given < 4MiB. We should make Toil detect this and up the memory limit.

It seems though the official Docker documentation [lists 6 MB though](https://docs.docker.com/engine/containers/resource_constraints/), are the resource requirements continually changing?

PR #5325, opened 2025-05-30
Update real world example in quick start documentation

### LINKED ISSUE #3534: Real World Example Tutorial is not a Real World Example.
https://toil.readthedocs.io/en/latest/gettingStarted/quickStart.html#a-more-real-world-example

This should be changed to an example of running an RNA seq workflow.  ^

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-847)
┆Issue Number: TOIL-847

stxue1:
Closes #3534

## Changelog Entry

 * toil-wdl-runner will bump container memory if below 4MiB
 * Updated real world example in quick start
 * `--importWorkersThreshold` has been renamed to `--importWorkersBatchsize`

## Reviewer Checklist

        ```
        ```

## Merger Checklist


stxue1:
Also closes #5316

docs/wdl/running.rst
```diff
@@ -84,6 +86,22 @@ the `Cromwell Output Organizer (croo)`_, will default to ``True``.
...
+
+``--runImportsOnWorkers``: Run file imports on workers. This is useful if the leader is not network optimized
+and lots of downloads are necessary. By default, this is false.
+
+``--importWorkersThreshold``: Requires ``--runImportsOnWorkers`` to be true. Specify the target batch size in bytes for batched imports.
+As many files as can fit will go into each batch import job. This also accepts abbreviations, such as ``G`` or ``Gi``.
+
+``--importWorkersDisk``: Requires ``--runImportsOnWorkers`` to be true. Specify the disk size each import worker will get.
+This may be necessary when file streaming is not possible. For example, downloading from AWS to a GCE job store.
+If specified, this should be set to the largest file size of all files to import. By default, this is 1 MiB.
```
adamnovak, line 103: Usually the user won't have to set this at all, because Toil will be able to read the file size from the source and allocate enough disk. The user only has to touch this when that isn't possible. The documentation should reflect that and should be about increasing the disk allocation for downloads that can't work otherwise.

adamnovak: Hm, probably we should make *some* attempt to allocate the right amount of disk at some point.

adamnovak (approved):
I think this is good now.

docs/wdl/running.rst
stxue1: I think the issue then is having some way to figure out if file streaming is supported or not. I had a previous implementation that would attempt to download on a worker with almost no disk space, then catch a crash and allocate the right amount of disk space, but I think we removed it since it was too hacky. This may also still be needed as an argument if we can't sniff file size out of servers.

adamnovak:
We might actually want to show a copy of the error we get without the `--quantCheck=False` flag. Because the error doesn't suggest to use the flag (maybe we need to actually parse the errors and suggest the flag), so we'd want the documentation to come up in search when people search the error.

src/toil/test/wdl/wdltoil_test.py
```diff
@@ -1121,122 +1120,7 @@ def make_string_expr(self, to_parse: str) -> WDL.Expr.String:
         for i in range(1, len(parts), 2):
             parts[i] = WDL.Expr.Placeholder(pos, {}, WDL.Expr.Null(pos))
 
-        return WDL.Expr.String(pos, parts)
-
-    def test_remove_common_leading_whitespace(self) -> None:
```
adamnovak, line 1126: Since we're removing these test that exercise a lot of edge cases, are we sure we have enough tests for the overall behavior (i.e. in workflows in the conformance tests or in Toil) that we'll catch breakage if MiniWDL changes its behavior and stops following the spec again?

adamnovak (approved):
I think this will work. I am a little worried removing all the edge case whitespace tests opens us up to breakage if MiniWDL changes again, but it might not be worth porting them all to e.g. conformance tests.

Issue #2862, opened 2019-11-15
CWL: needs a --minimumMemory to account for site specific overhead


mr-c:
Especially needed for the CWL conformance tests where `singularity pull` will use much more memory than the tests indicate they need

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-456)
┆Issue Number: TOIL-456

adamnovak:
We've run into a similar problem in the Kubernetes batch system, where the memory requirements for a job don't account for the memory needed to start up a whole Kubernetes pod, even if they are fine for e.g. the single machine batch system. Right now we have a hacky patch for it.

It might make sense to make this a feature of Toil in general, probably next to wherever `--defaultMemory` lives.

PR #5336, opened 2025-06-18
Added minimum memory allocation flag to cwl (#2862)

stxue1:
## Changelog Entry

 * Added `--cwl-min-ram` to CWL. This specifies a minimum memory allocation for all jobs. This is different than `--defaultMemory`, which only applies to jobs without explicitly set memory requirements.
## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/options/cwl.py
+++ b/src/toil/options/cwl.py
@@ -423,7 +423,7 @@ def add_cwl_options(parser: ArgumentParser, suppress: bool = True) -> None:
         help=suppress_help
         or "Specify a minimum memory allocation for all tasks ."
         "If --no-cwl-default-ram is passed, this does not apply to tools that do not "
-        "specify a memory requirement; `--defaultMemory` is used for those tools"
+        "specify a memory requirement; --defaultMemory is used for those tools"
         "in that case."
     )
     parser.add_argument(


src/toil/options/cwl.py
```diff
@@ -414,6 +416,15 @@ def add_cwl_options(parser: ArgumentParser, suppress: bool = True) -> None:
...
         dest="cwl_default_ram",
     )
+    parser.add_argument(
+        "--cwl-min-ram",
+        type=human2bytes,
+        action=make_open_interval_action(1),
+        help=suppress_help
+        or "Specify a minimum memory allocation for all tasks "
+        "except those using the Toil --defaultMemory. This applies a minimum memory "
+        "allocation for tools that have an explicit memory requirement.",
```
adamnovak, line 426: This isn't quite right. What really happens is:

1. If the tool specifies a memory requirement, `--cwl-min-ram` affects it.
2. If the tool doesn't specify a memory requirement, but `--no-cwl-default-ram` is not passed, the spec-specified default memory requirement is applied to it, and `--cwl-min-ram` affects it.
3. If the tool doesn't specify a memory requirement, and `--no-cwl-default-ram` *is* passed, then is uses `--defaultMemory` and `--cwl-min-ram` does *not* affect it.

We have to either explain all that in the documentation or option help, or we have to change the behavior to make it more explicable and less astonishing. The natural thing to do would be to remove the `--cwl-default-ram/--no-cwl-default-ram` setting, and provide *only* `--cwl-min-ram` as the escape hatch for deviating from the CWL spec to increase memory, except the reason we had `--no-cwl-default-ram` in the first place was to let jobs get to Slurm with no memory limit assigned so they could interact with another special Slurm setting we have for granting them a whole node's memory. So to really explain this, the docs might need to mention that intended interaction and how to use it.

adamnovak (changes requested):
I think we need to document the complex behavior here better, and maybe justify it by documenting the interaction we're trying to set up between no-requirements jobs and Slurm for giving them full nodes' memory. The short description we're using here in both the help and the docs is not actually correct.

docs/cwl/running.rst
```diff
@@ -100,8 +100,8 @@ Besides the normal Toil options and the options supported by cwltool, toil-cwl-r
   --reference-inputs    Do not copy remote inputs into Toil's file store and assume they are accessible in place from all nodes.
   --disable-streaming   Do not allow streaming of job input files. By default, files marked with ``streamable`` True are streamed from remote job stores.
   --cwl-default-ram     Apply CWL specification default ramMin.
-  --no-cwl-default-ram  Do not apply CWL specification default ramMin, so that Toil --defaultMemory applies.
-
+  --no-cwl-default-ram  Do not apply CWL specification default ramMin, so that Toil --defaultMemory applies. This can help jobs get to Slurm with no memory limit assigned.
+  --cwl-min-ram BYTES   Specify a minimum memory allocation for all tasks except for tools that do not specify a memory requirement and when `--no-cwl-default-ram` is passed, which is when `--defaultMemory` is used instead.
```
adamnovak, line 104: I don't think the top-level structure of this sentence is right; "except for (tools ((that ...) and (when ...)))" doesn't make sense because the "when" condition isn't actually qualifying a set of tools.

```suggestion
  --cwl-min-ram BYTES   Specify a minimum memory allocation for all tasks. If `--no-cwl-default-ram` is passed, this does not apply to tools that do not specify a memory requirement; `--defaultMemory` is used for those tools in that case.
```

src/toil/options/cwl.py
```diff
@@ -411,9 +413,17 @@ def add_cwl_options(parser: ArgumentParser, suppress: bool = True) -> None:
...
+        or "Do not apply CWL specification default ramMin, so that Toil --defaultMemory applies. This can help jobs get to Slurm with no memory limit assigned.",
         dest="cwl_default_ram",
     )
+    parser.add_argument(
+        "--cwl-min-ram",
+        type=human2bytes,
+        action=make_open_interval_action(1),
+        help=suppress_help
+        or "Specify a minimum memory allocation for all tasks "
+        "except for tools that do not specify a memory requirement and when `--no-cwl-default-ram` is passed, which is when `--defaultMemory` is used instead."
```
adamnovak, line 425:
```suggestion
        or "Specify a minimum memory allocation for all tasks ."
        "If --no-cwl-default-ram is passed, this does not apply to tools that do not "
        "specify a memory requirement; `--defaultMemory` is used for those tools"
        "in that case."
```

adamnovak (commented):
I want to reword the help.

adamnovak (approved):
Looks good to me!

PR #5484, opened 2026-04-01
Pretty Filestore

### LINKED ISSUE #5401: Make file store layout pretty using hints
Building on [https://ucsc-cgl.atlassian.net/browse/TOIL-1](https://ucsc-cgl.atlassian.net/browse/TOIL-1|smart-link), we want to be able to lay out files in the file store more like the user expects, if the user is working at the CWL or WDL level and is used to working with Cromwell.

We should be able to send some kind of hints from the workflow to the file store, so it can lay out files for or from a job under e.g. a WDL job name instead of a Toil job store ID.

We should also make this layout consistent across job store implementations, so someone can find workflow files in the same human-readable place in either a local or remote job store.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1789)
┆Issue Number: TOIL-1789

adamnovak:
This will fix #5401.

* [X] It still needs the hints to actually be plugged in to the WDL/CWL runners.

The code here is mostly synthetic, generated with Anthropic Claude Opus 4.6. It looked right at first glance but I also caught it trying to sneak in race conditions, so it needs thorough review.

It also needs some manual testing to run a real WDL or CWL workflow and look at the file store and agree that's what we want.

## Changelog Entry

 * Toil now organizes file store files using hints about e.g. workflow step names.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/fileStores/abstractFileStore.py
```diff
@@ -319,6 +319,11 @@ def writeGlobalFile(self, localFileName: str, cleanup: bool = False) -> FileID:
         :param cleanup: if True then the copy of the global file will be deleted once
                the job and all its successors have completed running.  If not the global
                file must be deleted manually.
+        :param hints: String values such as a workflow names or task names that
+               should be used to store the file at a human-findable location.
+               Large numbers of files stored under the same non-empty hints
+               may be inefficient; hints are intended for human-navigable
+               categorization, not for high-throughput bulk file creation.
```
adamnovak: This sounds like word salad of limited explanatory value.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,8 +455,69 @@ def associate_job_with_file(self, job_id: str, file_id: str) -> None:
 
         ###################################### FILES API ######################################
 
+    def _claim_hinted_s3_slot(self, sanitized_hints: list[str], basename: str) -> str:
```
adamnovak: Everything here is S3; does it need to be in the name?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,8 +455,69 @@ def associate_job_with_file(self, job_id: str, file_id: str) -> None:
...
+        """
+        Atomically claim a numbered slot for a hinted file in S3.
+
+        Uses S3 conditional writes (``IfNoneMatch: *``) so that concurrent
+        writers on different nodes racing for the same hint+basename will
+        each land in a distinct numbered slot.
```
adamnovak: The conditional writes aren't sufficient to provide this by themselves; all they do is avoid clobbering and fail a write instead.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,8 +455,69 @@ def associate_job_with_file(self, job_id: str, file_id: str) -> None:
...
+
+        :param sanitized_hints: Already-sanitized hint path components.
+        :param basename: The file basename.
+        :returns: The file_id (e.g. ``lions/tigers/bears/0``).
```
adamnovak: This shouldn't use test case content as an example, because the reader will not be familiar with the test case.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,8 +455,69 @@ def associate_job_with_file(self, job_id: str, file_id: str) -> None:
...
+        next_n = 0
+        for item in existing:
+            relative = item["Key"][len(list_prefix):]
+            parts = relative.split("/")
+            try:
+                n = int(parts[0])
+                if n >= next_n:
+                    next_n = n + 1
+            except (ValueError, IndexError):
+                continue
```
adamnovak, line 524: This is accomplishing some important stuff around making sure that files with hint lists that are prefixes of each other work properly, which isn't explained in comments.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -450,8 +455,69 @@ def associate_job_with_file(self, job_id: str, file_id: str) -> None:
...
+                    Key=s3_key,
+                    Body=b"",
+                    IfNoneMatch="*",
+                    **encryption_args,
+                )
+                return file_id
+            except ClientError as e:
+                if e.response["Error"]["Code"] in (
+                    "PreconditionFailed",
+                    "412",
```
adamnovak: Is it *really* possible for both symbolic names and numerical code strings to be in the "Code" here?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -483,19 +555,21 @@ def write_file(
...
+            dst_key=f"{prefix}/{basename}",
             extra_args=self._get_encryption_args(),
         )
         return FileID(file_id, size, executable)
 
     def find_s3_key_from_file_id(self, file_id: str) -> str:
         """This finds an s3 key for which file_id is the prefix, and which already exists."""
-        prefix = self._key_in_bucket(identifier=file_id, prefix=self.content_key_prefix)
+        # Trailing / prevents e.g. file_id "path/0" matching key "path/00/x"
+        prefix = self._key_in_bucket(identifier=file_id + "/", prefix=self.content_key_prefix)
```
adamnovak: Shouldn't the trailing slash be `_key_in_bucket()`'s responsibility? It's not really part of an "identifier".

src/toil/jobStores/aws/jobStore.py
```diff
@@ -708,12 +789,39 @@ def read_shared_file_stream(
             else:
                 raise
 
+    def _is_hinted_file_id(self, file_id: str) -> bool:
+        """Check if a file_id was produced by the hinted-file allocation scheme."""
```
adamnovak: We need some class-docstring-level or free-floating-comment-level documentation of the whole two-kinds-of-file-IDs system. The helpers like this one for working with them could go near that.

That would also explain the distinguishing features and make it clear why this sniffing approach is likely to work.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -708,12 +789,39 @@ def read_shared_file_stream(
...
             return
+
+        if self._is_hinted_file_id(file_id):
+            # Write a tombstone BEFORE deleting so the slot can never be
+            # reused, even if we crash between the two operations. This is
+            # necessary because Toil journals deletions and may replay them;
+            # a reused slot would cause a replayed delete to destroy the
+            # wrong file.
+            tombstone_key = self._key_in_bucket(
+                identifier=file_id + "/" + self.TOMBSTONE_SUFFIX,
```
adamnovak: The suffix starting with `.` makes it seem like it's just going to be tacked on as an extension, but we actually add it as a child key.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -907,15 +1015,23 @@ def get_empty_file_store_id(
...
-            bucket=self.bucket_name,
-        )
+        if not basename:
+            basename = "0"
+        sanitized = self._sanitize_hints(hints)
+        if sanitized:
+            # _claim_hinted_s3_slot already creates the empty object.
+            file_id = self._claim_hinted_s3_slot(sanitized, basename)
+        else:
+            file_id = str(uuid.uuid4())
```
adamnovak: This is the third time we have this pattern of do the sanitization, check if any hints are left, and claim a hint slot if using hints. Even though we need to do a write in this particular instance if using a non-hinted slot, we probably should try and roll that general pattern up into a helper function.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1088,18 +1088,60 @@ def jobs(self) -> Iterator[JobDescription]:
     # associated with a given job.
     ##########################################
 
+    # Characters allowed in sanitized hints: these survive quote() unchanged
+    # and are safe in filesystem paths and S3 keys.
+    HINT_SAFE_RE = re.compile(r"[^a-zA-Z0-9_\-.]")
```
adamnovak: `\` is not a *great* character to have in your paths.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1088,18 +1088,60 @@ def jobs(self) -> Iterator[JobDescription]:
...
+        total_length = 0
+        for hint in hints:
+            sanitized = self.HINT_SAFE_RE.sub("", hint)
+            sanitized = sanitized[: self.MAX_HINT_LENGTH]
+            if not sanitized:
+                continue
+            # +1 for the separator between components
+            if total_length + len(sanitized) + (1 if result else 0) > self.MAX_HINTS_PATH_LENGTH:
+                break
+            total_length += len(sanitized) + (1 if result else 0)
```
adamnovak: This is fiddly and repetitive and the comment does not get a the real point here, which is to take hints until the separator and the next hint no longer fit under the limit, then stop.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1221,6 +1295,15 @@ def get_empty_file_store_id(
...
                matching that basename, the file will be detected.
 
+        :param hints: String values such as a workflow names or task names that
+               should be used to store the file at a human-findable location.
+               Two files with the same basename and same hints are still
+               guaranteed to never collide and to have distinct assigned IDs.
+               Large numbers of files stored under the same non-empty hints
+               may be inefficient, as slot allocation scans existing entries;
+               hints are intended for human-navigable categorization, not for
+               high-throughput bulk file creation.
```
adamnovak: This gets repeated a lot, so in addition to this part being of dubious conceptual coherence, it might benefit from the parameter description being shorter overall, with the concept of hints explained in more detail elsewhere.

src/toil/jobStores/fileJobStore.py
```diff
@@ -1267,7 +1267,69 @@ def _stats_archive_directories(self):
...
+        """
+        Get a unique file path for a file stored under a human-readable
+        hint-derived directory hierarchy.
+
+        Each file is placed in a numbered subdirectory (0/, 1/, ...) under
+        the hints directory. Slot claiming is atomic via os.mkdir, so
+        concurrent writers on a shared filesystem will never collide.
+
+        After deletion, the empty numbered directory persists as a tombstone
+        that prevents the slot from being reallocated. This is critical
```
adamnovak: "Critical" is a strong word here. I'd say it's "required".

src/toil/jobStores/fileJobStore.py
```diff
@@ -1267,7 +1267,69 @@ def _stats_archive_directories(self):
 
         return self._walk_dynamic_spray_dir(self.stats_archive)
 
-    def _get_unique_file_path(self, fileName, jobStoreID=None, cleanup=False):
+    def _get_unique_file_path_with_hints(self, fileName, jobStoreID, cleanup, hints):
+        """
+        Get a unique file path for a file stored under a human-readable
+        hint-derived directory hierarchy.
```
adamnovak: Being "hint-derived" isn't a real concept.

adamnovak (commented):
I think the synthetic code suffers from bad style that ought to be fixed.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -92,12 +92,29 @@ class AWSJobStore(AbstractJobStore, URLAccess):
     1. Jobs: These are pickled as files, and contain the information necessary to run a job when unpickled.
        A job's file is deleted when finished, and its absence means it completed.
 
-    2. Files: The inputs and outputs of jobs.  Each file is written in s3 with the file pattern:
-       "files/{uuid4}/{original_filename}", where the file prefix
-       "files/{uuid4}" should only point to one file.
+    2. Files: The inputs and outputs of jobs.  Stored under the ``files/``
+       prefix using one of two file ID schemes:
+
+       * **UUID-based** (default): ``files/{uuid4}/{filename}``.
```
adamnovak, line 98: The fact that this is a UUID4 UUID is an implementation detail that should not escape here.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -92,12 +92,29 @@ class AWSJobStore(AbstractJobStore, URLAccess):
...
+         provided and support an unlimited number of files.
+
+       * **Hinted**: ``files/{hint1}/{hint2}/.../{N}/{filename}``.
+         The file ID is the path up to and including the numbered
+         directory, e.g. ``workflow/task/0``.  The numbered slot is
+         claimed atomically via ``IfNoneMatch`` conditional writes.
+         When a hinted file is deleted, a tombstone key
+         (``.../{N}/.tombstone``) is written *before* the real object is
+         removed so the slot number is never reused — this is required
+         because Toil may journal and replay deletions.
```
adamnovak, line 109: The user of the class doesn't need to be taught this; they just need to know that the tombstone system exists to prevent ID re-use. The stuff about ordering needs to be explained in comments in the functions that actually do that part.

Also, we don't want to use non-typeable characters if it can be avoided. Doesn't RST have `---` for em-dash?

src/toil/jobStores/aws/jobStore.py
```diff
@@ -92,12 +92,29 @@ class AWSJobStore(AbstractJobStore, URLAccess):
...
+
+       The two schemes are distinguished by checking the last path
+       component: numeric means hinted, anything else means UUID-based
+       (since UUID strings never end in a bare integer).
```
adamnovak, line 113: Less-detailed documentation of the file API implementation design should be here, and more detailed documentation should be in the `FILES API` section later.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -562,14 +600,14 @@ def write_file(
 
     def find_s3_key_from_file_id(self, file_id: str) -> str:
         """This finds an s3 key for which file_id is the prefix, and which already exists."""
-        # Trailing / prevents e.g. file_id "path/0" matching key "path/00/x"
-        prefix = self._key_in_bucket(identifier=file_id + "/", prefix=self.content_key_prefix)
+        # Trailing / scopes the listing to exact children of this file_id.
+        prefix = self._key_in_bucket(identifier=file_id, prefix=self.content_key_prefix) + "/"
```
adamnovak, line 604: Then version with examples made this point better; "exact children" sounds like substituting for understanding with emphasis.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -1020,12 +1042,10 @@ def get_empty_file_store_id(
...
-        sanitized = self._sanitize_hints(hints)
-        if sanitized:
-            # _claim_hinted_s3_slot already creates the empty object.
-            file_id = self._claim_hinted_s3_slot(sanitized, basename)
-        else:
-            file_id = str(uuid.uuid4())
+        file_id = self._allocate_file_id(basename, hints)
+        if not self._is_hinted_file_id(file_id):
+            # _allocate_file_id minted a UUID but didn't write anything;
+            # _claim_hinted_slot already wrote the empty placeholder.
```
adamnovak, line 1048: The tense is weird here, and we did not call either of the mentioned functions here. This should be in terms of what `_allocate_file_id` promises.

src/toil/jobStores/aws/jobStore.py
```diff
@@ -504,14 +537,23 @@ def _claim_hinted_s3_slot(self, sanitized_hints: list[str], basename: str) -> st
...
                     continue
                 raise
 
+    def _allocate_file_id(self, basename: str, hints: list[str] | None) -> str:
+        """
+        Choose a file ID for a new file.
+
+        If *hints* are provided and any survive sanitization, claim a
+        numbered hinted slot.  Otherwise mint a fresh UUID.
+        """
```
adamnovak, line 551: It's apparently important later that this function ensures file IDs have empty files created there, while it doesn't do that for UUID file IDs.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -176,6 +176,18 @@ class AbstractJobStore(ABC):
 
     To actually get ahold of a :class:`toil.job.Job`, use
     :meth:`toil.job.Job.loadJob` with a JobStore and the relevant JobDescription.
+
+    .. rubric:: File Hints
```
adamnovak, line 180: This should be a real RST section heading with an anchor that gets referenced by `:ref:`, instead of all the docstrings elsewhere saying to see `AbstractJobStore`-the-class for how hints work.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -176,6 +176,18 @@ class AbstractJobStore(ABC):
...
+
+    File-writing methods accept an optional ``hints`` parameter: a list of
+    strings (e.g. workflow name, task name) that are incorporated into the
+    stored file's path or key so a human browsing the job store can locate
+    it.  Each hint becomes a path component after sanitization.  Two files
+    written with the same hints and basename are guaranteed distinct IDs;
+    disambiguation is handled by appending a numbered directory (``0/``,
+    ``1/``, ...).  Deleted slots are never reused (the empty directory or a
+    tombstone object prevents reallocation), which is required because Toil
+    may journal and replay file deletions.
```
adamnovak, line 190: This is all implementation details and doesn't go here.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -176,6 +176,18 @@ class AbstractJobStore(ABC):
...
     To actually get ahold of a :class:`toil.job.Job`, use
     :meth:`toil.job.Job.loadJob` with a JobStore and the relevant JobDescription.
+
+    .. rubric:: File Hints
+
+    File-writing methods accept an optional ``hints`` parameter: a list of
+    strings (e.g. workflow name, task name) that are incorporated into the
+    stored file's path or key so a human browsing the job store can locate
+    it.  Each hint becomes a path component after sanitization.  Two files
+    written with the same hints and basename are guaranteed distinct IDs;
```
adamnovak, line 186: This shouldn't be double-spaced after periods.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -176,6 +176,18 @@ class AbstractJobStore(ABC):
 
     To actually get ahold of a :class:`toil.job.Job`, use
     :meth:`toil.job.Job.loadJob` with a JobStore and the relevant JobDescription.
+
+    .. rubric:: File Hints
+
+    File-writing methods accept an optional ``hints`` parameter: a list of
+    strings (e.g. workflow name, task name) that are incorporated into the
+    stored file's path or key so a human browsing the job store can locate
```
adamnovak, line 184: "path or key" is also an implementation detail, and "sanitization" is used here like it's something the reader should understand already, when it isn't.

We want to say that the hints are sanitized and used to hierarchically organize the stored files in a human-readable way.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1084,41 +1096,41 @@ def jobs(self) -> Iterator[JobDescription]:
         raise NotImplementedError()
 
     ##########################################
-    # The following provide an way of creating/reading/writing/updating files
-    # associated with a given job.
+    # The following provide a way of creating/reading/writing/updating files
+    # associated with a given job.  See the "File Hints" section in the
+    # class docstring for the hints system.
```
adamnovak, line 1101: We don't need this reference here.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1084,41 +1096,41 @@ def jobs(self) -> Iterator[JobDescription]:
...
     ##########################################
 
-    # Characters allowed in sanitized hints: these survive quote() unchanged
-    # and are safe in filesystem paths and S3 keys.
-    HINT_SAFE_RE = re.compile(r"[^a-zA-Z0-9_\-.]")
+    # Strips everything except characters safe in filesystem paths, S3 keys,
+    # and urllib quote() output: ASCII letters, digits, underscore, dot, hyphen.
```
adamnovak, line 1105: Instead of listing where these are safe, which creates backward references in the architecture, we should not do that.

src/toil/jobStores/abstractJobStore.py
```diff
@@ -1084,41 +1096,41 @@ def jobs(self) -> Iterator[JobDescription]:
...
-            # +1 for the separator between components
-            if total_length + len(sanitized) + (1 if result else 0) > self.MAX_HINTS_PATH_LENGTH:
+            needed = len(cleaned) + (1 if result else 0)
+            if total + needed > self.MAX_HINTS_PATH_LENGTH:
                 break
-            total_length += len(sanitized) + (1 if result else 0)
-            result.append(sanitized)
+            total += needed
+            result.append(cleaned)
         return result
```
adamnovak, line 1134: This is actually interleaving two concerns: sanitizing a single hint so it's safe on a storage backend, and joining hints together in a path-like slug up to a max length with `/`. It's also making the abstract class ultimately in charge of what the implementations can support in their storage backends.

We should have a `hints_to_string()` that takes a separator and a max length as optional arguments, because the different backends might at some point need to deal with different max path lengths or different separator requirements. That should in turn call `_sanitize_hint()` to sanitize each hint, and then handle pasting them together.

We could also assume alphanumeric characters are always "safe", and take a `safe` optional argument for the non-alphanumeric characters we want to keep, kind of like urllib quoting does. Then we can check the regex or membership in the safe string's set. This moves the values from abstract class constants to method default arguments where they make more sense as part of an implementation-building kit.

src/toil/cwl/cwltoil.py
```diff
@@ -2225,6 +2229,9 @@ def __init__(
             # We need something. Put the class.
             name_parts.append(class_name)
 
+        # Keep the structured path for use as file hints.
+        self.task_path = name_parts
```
adamnovak, line 2233: On the CWL side, `task_path` is a list, whereas on the WDL side it is a dotted string.

We should settle on one representation for both (maybe dotted string for now, since that requires less refactoring on the WDL side).

Than we could avoid a separate `unit_name` local and just pass the task path as the unit name to the superclass constructor.

src/toil/jobStores/fileJobStore.py
```diff
@@ -1267,7 +1267,61 @@ def _stats_archive_directories(self):
 
         return self._walk_dynamic_spray_dir(self.stats_archive)
 
-    def _get_unique_file_path(self, fileName, jobStoreID=None, cleanup=False):
+    def _get_unique_file_path_with_hints(self, fileName, jobStoreID, cleanup, hints):
```
adamnovak, line 1270: New code should have type hints on it, even if the rest of the file doesn't.

src/toil/jobStores/fileJobStore.py
```diff
@@ -1267,7 +1267,61 @@ def _stats_archive_directories(self):
...
+            else:
+                root = self._get_job_files_dir(jobStoreID)
+        else:
+            root = self.filesDir
+
+        hints_dir = os.path.join(root, *hints)
+        os.makedirs(hints_dir, exist_ok=True)
+
+        # Each file gets its own numbered subdirectory (0/, 1/, ...).
+        # We use os.mkdir to claim a slot — it is atomic even on NFS, so
```
adamnovak, line 1298: We shouldn't use untypeable characters in code.

src/toil/jobStores/fileJobStore.py
```diff
@@ -1267,7 +1267,61 @@ def _stats_archive_directories(self):
...
+        # concurrent writers on a shared filesystem won't collide.  After
+        # a file is deleted, the empty directory remains as a tombstone
+        # that prevents the slot number from being reallocated.
```
adamnovak, line 1301: It's possible for the numbered subdirectory for one file to overlap with an intermediate directory from a hint in another file, when hints are bare numbers. For example, we can make a file with hints `["a", "b"]` which gets `a/b/0`, and then another file with hints `["a", "b", "0"]` which will get `a/b/0/0`.

We should test this, and we should especially test when a file basename is something like `0` or something that could collide with a hint.

adamnovak (commented):
I have more complaints about how the synthetic code is not exactly the code I would have written. I also think I have a way to break the layout, using hints or basenames that could collide with the numerical directories.

adamnovak:
Now this has failed with a bunch of vague complaints about encrypted AWS job stores:

```
FAILED src/toil/test/jobStores/jobStoreTest.py::EncryptedAWSJobStoreTest::test_banned_hints - toil.lib.aws.s3.AWSBadEncryptionKeyError: Your AWS encryption key is most likely configured incorrectly (HeadObject operation on key 'files/good/files/file.txt': Bad Request).
FAILED src/toil/test/jobStores/jobStoreTest.py::EncryptedAWSJobStoreTest::test_hinted_file_ids_not_reused_after_deletion - toil.lib.aws.s3.AWSBadEncryptionKeyError: Your AWS encryption key is most likely configured incorrectly (HeadObject operation on key 'files/alpha/beta/files/file.txt': Bad Request).
FAILED src/toil/test/jobStores/jobStoreTest.py::EncryptedAWSJobStoreTest::test_hinted_numeric_basename - toil.lib.aws.s3.AWSBadEncryptionKeyError: Your AWS encryption key is most likely configured incorrectly (HeadObject operation on key 'files/a/files/file.txt': Bad Request).
FAILED src/toil/test/jobStores/jobStoreTest.py::EncryptedAWSJobStoreTest::test_hinted_numeric_components - toil.lib.aws.s3.AWSBadEncryptionKeyError: Your AWS encryption key is most likely configured incorrectly (HeadObject operation on key 'files/a/0/files/file.txt': Bad Request).
```

Maybe I'm not passing along all the headers I need to pass when doing things with hinted files?

adamnovak (commented):
This still isn't all in my voice, but I think the code is defensible, and all the bad ideas are ones I came up with because I couldn't think of better ones.

PR #5502, opened 2026-04-23
Pass CWL ToolTimeLimit.timelimit to slurm job submission

### LINKED ISSUE #3037: Walltime for an individual job when running Toil on HPC
Dear developers,
I am using Toil for running a pipeline in a HPC environment that uses Slurm for job scheduling.

Certain jobs of my pipeline are shorter than the others and it would be help me a lot if there was an option for specifying `walltime` for the job. I was not able to find this in the docs. 
Is there a way how to tell the scheduler how long a job is supposed to be running? Any other way than using `$TOIL_SLURM_ARGS`?

Thanks,
Daniel

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-539)
┆Issue Number: TOIL-539

lonbar:
Provides an initial implementation for the option to pass runtimes to slurm's resource allocation. Addresss #3037.

## Changelog Entry

 * Runtimes for slurm jobs can now be set using `--defaultWalltime`
 * Time limits from CWL's `ToolTimeLimit` are used in slurm batch submissions 

## Reviewer Checklist

    * [x] If it is coming from an external repo, make sure to pull it in for CI with:
        ```
        ```

## Merger Checklist


mr-c:
@adamnovak  CI is running using https://github.com/DataBiosphere/toil/tree/issues/3037-wallclock-slurm  (a.k.a `contrib/admin/test-pr lonbar issues/featurebranch issues/3037-wallclock-slurm`)

src/toil/cwl/cwltoil.py
```diff
@@ -2574,6 +2576,14 @@ def __init__(
             # Note: if the job is using the toil default memory, it won't be increased
             memory = max(memory, min_ram)
 
+        # Check if the tool has set a time limit. If yes, use it. Otherwise,
+        # use a None requirement to use the Toil default.
+        tool_max_walltime = tool.get_requirement("ToolTimeLimit")[0] or {}
+        if "timelimit" in tool_max_walltime:
+            walltime = int(tool_max_walltime["timelimit"])
```
mr-c: ToolTimeLimit.timelimit can be a string (a CWL expression), see https://github.com/common-workflow-language/cwl-v1.2/blob/main/tests/timelimit4.cwl


```suggestion
        if (limit_val := tool_max_walltime["timelimit"]) is not None:
            walltime = cast(int, self.builder.do_eval(limit_val)) or None
```

Tested using this, which should be added as its own test in Toil
``` cwl
class: CommandLineTool
cwlVersion: v1.2
inputs: []
outputs: []
requirements:
  ToolTimeLimit:
    timelimit: $(3*4)
  WorkReuse:
    enableReuse: false
baseCommand: [sleep, "3"]
```

src/toil/job.py
```diff
@@ -791,7 +801,11 @@ def requirements_string(self) -> str:
         for k in REQUIREMENT_NAMES:
             v: str | ParsedRequirement | None = self._fetchRequirement(k)
             if v is not None:
-                if isinstance(v, (int, float)) and v > 1000:
+                if k == "walltime":
+                    if v == 0:
+                        continue
+                    v = seconds_to_dhms(v)
```
mr-c:
```suggestion
                    v = seconds_to_dhms(cast(int, v))
```

adamnovak:
Oh excellent, we have wanted this for a while.

lonbar:
@mr-c I have been made aware of [`--maxJobDuration`](https://toil.readthedocs.io/en/latest/autoapi/toil/common/index.html#toil.common.Config.maxJobDuration), which allows toil itself to cancel jobs after a specified time. I was wondering if it makes sense to use this input instead of adding `--defaultWalltime`.

My thinking is that it might make sense to keep them separate, as currently the logic is:

1. If `ToolTimeLimit.timelimit` is set, use that value.
2. If `ToolTimeLimit.timelimit` is not set but `--defaultWalltime` is used, use the value of `--defaultWalltime`.
3. Otherwise, do not set a time limit for slurm jobs.

This means that `--defaultWalltime` provides a baseline wall time for steps that do not specify a one and if, say, a step in a CWL sets `ToolTimeLimit.timelimit: 0`, the resulting slurm submission will not contain a wall time argument even if `--defaultWalltime` is set. `--maxJobDuration` allows a user to put an upper bound to the runtime of such jobs. Happy to hear your thoughts.

mr-c:
> @mr-c I have been made aware of [`--maxJobDuration`](https://toil.readthedocs.io/en/latest/autoapi/toil/common/index.html#toil.common.Config.maxJobDuration), which allows toil itself to cancel jobs after a specified time. I was wondering if it makes sense to use this input instead of adding `--defaultWalltime`.
> 
> My thinking is that it might make sense to keep them separate, as currently the logic is:
> 
>     1. If `ToolTimeLimit.timelimit` is set, use that value.
> 
>     2. If `ToolTimeLimit.timelimit` is not set but `--defaultWalltime` is used, use the value of `--defaultWalltime`.
> 
>     3. Otherwise, do not set a time limit for slurm jobs.
> 
> 
> This means that `--defaultWalltime` provides a baseline wall time for steps that do not specify a one and if, say, a step in a CWL sets `ToolTimeLimit.timelimit: 0`, the resulting slurm submission will not contain a wall time argument even if `--defaultWalltime` is set. `--maxJobDuration` allows a user to put an upper bound to the runtime of such jobs. Happy to hear your thoughts.

I think it makes sense to keep  `--defaultWalltime` and still respect `--maxJobDuration` if set, yes.

src/toil/lib/conversions.py
```diff
@@ -179,6 +180,19 @@ def hms_duration_to_seconds(hms: str) -> float:
...
+def seconds_to_dhms(seconds: int) -> str:
+    """
+    Convert seconds to a days-hours:minutes:seconds string.
+    """
+    if seconds < 0:
+        raise ValueError("Invalid Time, negative value")
+
+    walltime = datetime.timedelta(seconds=seconds)
+    days = walltime.days
+    remainder = str(walltime).split(",")[int(days > 0)].strip()
```
adamnovak: The only documentation about the string format for `timedelta` objects I could find at https://docs.python.org/3/library/datetime.html#timedelta-objects is a claim that it is "confusing". This line here is really dense and seems to depend on the format including the days component and then an H:M:S time, separated by a comma, but not including the days component when the days are zero. And we've created the timedelta object *just* to get access to the string formatting logic that we suspect it has. Are we sure this can't vary by e.g. locale?

I would recommend just doing the modular division ourselves here, or at lease commenting this to explain the approach.

src/toil/options/common.py
```diff
@@ -758,6 +758,7 @@ def __call__(
     )
     cpu_note = "Fractions of a core (for example 0.1) are supported on some batch systems [mesos, single_machine]"
     disk_mem_note = "Standard suffixes like K, Ki, M, Mi, G or Gi are supported"
+    disk_walltime_note = "Values are assumed to be in seconds. A value of 0 does not limit the walltime"
```
adamnovak: There's nothing `disk_` about this.

src/toil/options/common.py
```diff
@@ -777,6 +778,16 @@ def __call__(
             "default", "memory", disk_mem_note, bytes2human(2147483648)
         ),
     )
+    resource_options.add_argument(
+        "--defaultWalltime",
+        dest="defaultWalltime",
+        default="0",
+        type=int,
+        action=make_open_interval_action(1),
```
adamnovak: This means we won't accept a value of 0, but we might want to be able to explicitly take 0.

src/toil/job.py
```diff
@@ -464,7 +465,7 @@ def __init__(self, requirements: Mapping[str, ParseableRequirement | None]) -> N
         Parse and save the given requirements.
 
         :param dict requirements: Dict from string to value
-            describing a set of resource requirments. 'cores', 'memory',
+            describing a set of resource requirments. 'cores', 'walltime', 'memory',
```
adamnovak: Is there any logic to the order we're using here? Sometimes walltime is here and sometimes it's added at the end.

I would recommend always using an order of cores, memory, disk, preemptible, accelerators, walltime, so that we add the new item to the end.

We could instead do cores, memory, disk, walltime, preemptible, accelerators, if we want to keep the numbers together.

adamnovak (changes requested):
This looks pretty good, but I think some of the variable names ought to change and I think supporting `--defaultWalltime=0` would be good.

src/toil/options/common.py
lonbar: Fixed in https://github.com/DataBiosphere/toil/pull/5502/commits/ffa679d0d936393462ab0dbfa6bf96f4607b1abd.

src/toil/job.py
lonbar: That should be fixed in https://github.com/DataBiosphere/toil/pull/5502/commits/a1733848aad5b11abad9a52bd95314278eb47c3f.

src/toil/options/common.py
lonbar: Thanks for catching this! Updated in https://github.com/DataBiosphere/toil/pull/5502/commits/7d734b4f540347e063a8bdf3792e372653ba644e.

src/toil/lib/conversions.py
lonbar: I considered first doing the conversion with module division directly, but thought you would perhaps prefer a standard library. I do agree that what I wrote is a bit cryptic, so I've removed `datetime` in https://github.com/DataBiosphere/toil/pull/5502/commits/59e1c478d2f1f977ea7f3f7485fceed63dc5f611.

src/toil/batchSystems/slurm.py
```diff
@@ -335,6 +335,7 @@ def prepareSubmission(
             self,
             cpu: int,
             memory: int,
+            walltime: int,
```
mr-c, line 355: in CWL, walltime can be a float

lonbar: To clarify, it can be a float because any runtime expressions might evaluate to floats? [TimeToolLimit.timelimit](https://www.commonwl.org/v1.2/CommandLineTool.html#ToolTimeLimit) just says "The time limit, in seconds", with no explicit `float` type.

mr-c: whoops, I was looking at something else; ignore this!

adamnovak (approved):
This could use more testing to ensure that the limit is *really* applied, and I think the WDL interpreter will want to plug into this eventually, and it would be good if `--defaultWalltime` could take formatted durations and not just a number of seconds, but all that can be improved later.

adamnovak:
I pulled this in with `contrib/admin/test-pr lonbar issues/featurebranch issues/3037-wallclock-slurm` for testing.

PR #5507, opened 2026-04-30
Evaluate CWL jobs that should be skipped on the leader

### LINKED ISSUE #3990: CWL Conditional steps that aren't running still get scheduled to a job scheduler.
When I use conditionals, I find that the job still gets scheduled to a batch system.  

conditional-workflow.cwl

```
class: Workflow
cwlVersion: v1.2
inputs:
  message: string
  sleepParam: int 
outputs:
  out1:
    type: File
    outputSource:
      - echo/echoOut
    pickValue: first_non_null

requirements:

steps:
  echo:
    in: 
      message: message
      sleepParam: sleepParam
    run: echo.cwl
    out: [echoOut]
    when: $(inputs.sleepParam > 1)
```

echo.cwl

```
cwlVersion: v1.2
class: CommandLineTool
baseCommand: [echo]
id: "echo"
inputs: 
  message: 
    type: string
    inputBinding:
      position: 1
outputs: 
  echoOut:
    type: stdout
```
param.yaml
```
sleepParam: 1
message: hello
```
When I run this workflow, you can see in the log files that the job gets issued.  For HPC systems, it does look like the job gets scheduled on a HPC cluster.  

[2022-01-07T09:39:12-0500] [MainThread] [I] [toil.leader] Issued job 'CWLJob' echo-sleep-conditional.cwl.echo.echo kind-CWLJob/instance-e9x_ed10 v1 with job batch system ID: 1 and cores: 1, disk: 3.0 Gi, and memory: 2.0 Gi

┆Issue is synchronized with this [Jira Task](https://ucsc-cgl.atlassian.net/browse/TOIL-1121)
┆friendlyId: TOIL-1121

lonbar:
CWL jobs that have a when condition that evaluates to False should not be executed. Currently anything that is not a Workflow or ExpressionTool will always run on a worker node, which means that the check on the Conditional is only done when the worker node is already allocated. If the job is instead run on the leader (local), the step won't be executed.

Checking at the instantiation level makes it possible to determine dynamically if the step should be run on the leader or the worker. This prevents unnecessary overhead in scheduling systems.

Resolves #3990.

## Changelog Entry

 * Toil checks the CWL `when` conditional on the leader.

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/cwl/cwltoil.py
```diff
@@ -2641,15 +2636,23 @@ def __init__(
         # If not using the Toil file store, output files just go directly to
         # their final homes their space doesn't need to be accounted per-job.
 
+        options_dict: dict = {} # type: ignore
```
annagiroti, line 2813: Is the # type: ignore here intentional? As far as I can tell dict = {} is valid Python and shouldn't produce a type error. Would dict[str, Any] be a more precise annotation, and would that remove the need for the ignore comment entirely?

src/toil/cwl/cwltoil.py
```diff
@@ -2641,15 +2636,23 @@ def __init__(
...
+        run_local: bool = self.conditional.is_false(cwljob)
```
annagiroti, line 2814: cwljob may still have unresolved Promise objects at __init__ time if when references an output from an upstream step. Since Conditional.is_false resolves promises without a file store, could this either crash or return the wrong result in that case? The worst case I can think of is is_false incorrectly returning True here, setting local=True with no resources, but then the fully-resolved condition at run() time returning False, meaning real work runs on the leader with no reserved resources. Would wrapping this in a try/except that falls back to run_local = False be a safe way to handle that?

src/toil/cwl/cwltoil.py
```diff
@@ -2641,15 +2636,23 @@ def __init__(
...
-            accelerators=accelerators,
-            preemptible=preemptible,
             tool_id=self.cwltool.tool["id"],
             parent_name=parent_name,
-            local=isinstance(tool, cwltool.command_line_tool.ExpressionTool),
+            local=(
+                isinstance(tool, cwltool.command_line_tool.ExpressionTool)
+                or run_local
+                ),
+            **options_dict,
```
annagiroti, line 2830: When run_local is True, options_dict is empty so cores, memory, disk, accelerators, and preemptible all fall back to Job defaults. CWLJobWrapper, which also runs locally, explicitly passes cores=1, memory="1GiB", disk="1MiB" for its local run. Would it be worth doing the same here for consistency, rather than relying on the defaults being equivalent?

annagiroti (commented):
The overall approach appears to be clean and the options_dict pattern for conditionally passing resources is a nice solution. My main concern is the is_false being called at __init__ time before promises are fully resolved. This is worth making sure that can't cause issues for when conditions that reference upstream step outputs. Would it also be worth adding test cases? For example, one where the when condition is false (verifying the job isn't submitted to the batch system) and one where it references an output from a previous step (to confirm it doesn't crash or mis-schedule).

src/toil/cwl/cwltoil.py
lonbar: Ah, I had hoped that the call to `resolve_dict_w_promises()` in `Conditional.is_false()` would have prevented that from happening. If the step has unresolved promises at init time the workflow will actually crash. I'm a bit embarassed that I did not consider that case.

Hopefully I'll have some time this week to look more into this. Ideally, I think, the conditional is checked when all promises have been resolved. What do you think?

adamnovak:
We decided to take a different approach in #5569, but I think the problem this is solving is fixed!

PR #5508, opened 2026-04-30
Retry on ENOLCK from NFS lockd in fcntl-based locks (#4846)

### LINKED ISSUE #4846: Handle `OSError: [Errno 37] No locks available` in file locking
In https://github.com/ComparativeGenomicsToolkit/cactus/issues/1299 we see Toil failing because sometimes its file locking can throw an exception that locks are not (currently?) available.

I think this might happen if NFS's `lockd` is not responding.

We should handle this and retry for a while, and fail if locks do not become available in a reasonable time.

annagiroti:
When Toil runs on NFS filesystems, `fcntl.flock` can raise `OSError [Errno 37] No locks available` (`ENOLCK`) if the NFS lock daemon (`lockd`) is temporarily unavailable. Previously, this caused jobs to crash immediately. This PR extends the existing retry logic in `safe_lock` (which already handled `EIO` for Ceph) to also retry on `ENOLCK` with exponential backoff. `safe_unlock_and_close` is also updated to swallow `ENOLCK` the same way it does `EIO`. Unit tests are added for both error cases using mocked `fcntl.flock`.

Resolves #4846

## Changelog Entry

 * Retry on ENOLCK from NFS lockd in fcntl-based locks to handle flaky NFS mounts (#4846)

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/lib/threading.py
+++ b/src/toil/lib/threading.py
@@ -165,7 +165,8 @@ def safe_lock(fd: int, block: bool = True, shared: bool = False) -> None:
                         )
                     else:
                         logger.critical(
-                            "Too many IO errors talking to lock file. If using Ceph, check for MDS deadlocks. See <https://tracker.ceph.com/issues/62123>."
+                            "Too many IO errors talking to lock file. If using Ceph, "
+                            "check for MDS deadlocks. See <https://tracker.ceph.com/issues/62123>.",
                         )
                     raise
             else:
@@ -179,11 +180,12 @@ def safe_unlock_and_close(fd: int) -> None:
     try:
         fcntl.flock(fd, fcntl.LOCK_UN)
     except OSError as e:
-        if e.errno not in (errno.EIO, errno.ENOLCK):
-            raise
-        # Sometimes Ceph produces EIO. We don't need to retry then because
-        # we're going to close the FD and after that the file can't remain
-        # locked by us.
+        if e.errno in (errno.EIO, errno.ENOLCK):
+            # We don't need to retry then because we're going to close the FD 
+            # and after that the file can't remain locked by us.
+            pass
+        else:
+            raise  
     os.close(fd)
 
 
--- a/src/toil/lib/threading.py
+++ b/src/toil/lib/threading.py
@@ -141,7 +141,7 @@ def safe_lock(fd: int, block: bool = True, shared: bool = False) -> None:
             if e.errno in (errno.EACCES, errno.EAGAIN):
                 # Nonblocking lock not available.
                 raise
-            elif e.errno == errno.EIO:
+            elif e.errno in (errno.EIO, errno.ENOLCK):
                 # Sometimes Ceph produces IO errors when talking to lock files.
                 # Back off and try again.
                 # TODO: Should we eventually give up if the disk really is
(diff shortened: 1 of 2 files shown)


src/toil/lib/threading.py
```diff
@@ -156,9 +156,17 @@ def safe_lock(fd: int, block: bool = True, shared: bool = False) -> None:
...
+                    if e.errno == errno.ENOLCK:
+                        logger.critical(
+                            "No locks available after %d retries. The filesystem at the "
+                            "lock path may not support POSIX file locking (common on some "
+                            "NFS setups). Use --coordinationDir to point at local storage.",
+                            MAX_ERROR_TRIES,
+                        )
+                    else:
+                        logger.critical(
+                            "Too many IO errors talking to lock file. If using Ceph, check for MDS deadlocks. See <https://tracker.ceph.com/issues/62123>."
```
adamnovak: We might wrap this message like the new one.

src/toil/lib/threading.py
```diff
@@ -171,7 +179,7 @@ def safe_unlock_and_close(fd: int) -> None:
     try:
         fcntl.flock(fd, fcntl.LOCK_UN)
     except OSError as e:
-        if e.errno != errno.EIO:
+        if e.errno not in (errno.EIO, errno.ENOLCK):
             raise
         # Sometimes Ceph produces EIO. We don't need to retry then because
```
adamnovak: This comment now only mentions one of the two cases its branch needs to implement. We probably could drop that and just talk about how we don't need to retry.

src/toil/test/src/threadingTest.py
```diff
@@ -68,6 +70,59 @@ def testLastProcessStanding(self, tmp_path: Path) -> None:
                 assert not filename.startswith(
                     "precious"
                 ), f"File {filename} still exists"
+    
+    # Tests for ENOLCK (toil#4846)
```
adamnovak: Do we need this? Is it better as a docstring?

src/toil/test/src/threadingTest.py
```diff
@@ -68,6 +70,59 @@ def testLastProcessStanding(self, tmp_path: Path) -> None:
...
+    def testSafeLockRetriesOnENOLCK(self) -> None:
```
adamnovak: These new function names ought to be `snake_case`.

src/toil/test/src/threadingTest.py
```diff
@@ -68,6 +70,59 @@ def testLastProcessStanding(self, tmp_path: Path) -> None:
...
+                safe_unlock_and_close(0)
+                mock_close.assert_called_once_with(0)
+ 
+    def testSafeUnlockAndCloseSwallowsEIO(self) -> None:
+        # First call raises EIO, second call succeeds
+        eio = OSError(errno.EIO, "Input/output error")
+        with patch("fcntl.flock", side_effect=eio):
+            with patch("os.close") as mock_close:
+                safe_unlock_and_close(0)
+                mock_close.assert_called_once_with(0)
```
adamnovak: There are two substantially identical sets of 3 tests here, which differ just on the errno value and message used. We should consolidate them, either using inheritance (one base class with an abstract `raise_error` that we implement in IOError and NoLocksError subclasses), or using [pytest subtests](https://pypi.org/project/pytest-subtests/) and a loop over the errno-and-message pairs (or over a constant list of pre-constructed exceptions).

adamnovak (changes requested):
This looks pretty good, but the test code I think should be deduplicated a bit.

adamnovak (approved):

adamnovak:
I like the code now, but if you click on the little red X and then "Details" on the failing Gitlab job, and open up the failing `lint` step and peruse the log, you can see that the type checking is failing because of this:
```
src/toil/test/src/threadingTest.py: note: In member "test_safe_lock_fails_after_max_retries" of class "BaseSafeLockingTest":
src/toil/test/src/threadingTest.py:102:39: error: "Exception" has no attribute
"errno"  [attr-defined]
                        assert e.errno == error.errno
                                          ^~~~~~~~~~~
Found 1 error in 1 file (checked 131 source files)
```

I think the problem is that `get_exception()` is typed to return `Exception`, but for us to check things based on the result's `errno` we need to type it as returning `OSError` instead.

adamnovak:
@annagiroti See if you can fix this up so it passes all the CI tests. You should be able to `make mypy` to run the type checking locally.

PR #5512, opened 2026-05-05
Harden CPU overuse warning checks and add unit tests

avnig05:
Moves CPU-overuse warning evaluation into Job._check_cpu_usage() so the logic is centralized and easier to test. The check skips warnings when wall-clock runtime is non-positive, warns when an invalid core request is encountered (cores <= 0), and only gives a warning when CPU consumption exceeds the configured threshold factor.

## Changelog Entry

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -3334,7 +3334,15 @@ class Job:
         fileStore: AbstractFileStore,
         threshold_factor: float = 1.05,
     ) -> None:
-        """Log a warning when a job consumes more CPU time than its allocation."""
+        """
+        Log a warning when a job consumes more CPU time than its allocation.
+
+        If *job_time* is non-positive, returns immediately (wall-clock time may
+        lack resolution for very short jobs). If *cores* is non-positive, logs
+        a warning that overuse cannot be assessed and returns, avoiding
+        division by zero. Otherwise, compares *job_cpu_time* to
+        ``job_time * cores * threshold_factor``.
+        """
         if job_time <= 0:
             # The job may have run too quickly for wall-clock resolution.
             return
--- a/src/toil/job.py
+++ b/src/toil/job.py
@@ -3304,20 +3304,8 @@ class Job:
             )
             job_time = time.time() - startTime
             job_cpu_time = totalCpuTime - startClock
-            allocated_cpu_time = job_time * self.cores
 
-            if job_cpu_time > allocated_cpu_time and allocated_cpu_time > 0:
-                # Too much CPU was used by this job! Maybe we're using a batch
-                # system that doesn't/can't sandbox us and we started too many
-                # threads. Complain to the user!
-                excess_factor = job_cpu_time / allocated_cpu_time
-                fileStore.log_to_leader(
-                    f"Job {self.description} used {excess_factor:.2f}x more "
-                    f"CPU than the requested {self.cores} cores. Consider "
-                    f"increasing the job's required CPU cores or limiting the "
-                    f"number of processes/threads launched.",
-                    level=logging.WARNING,
-                )
+            self._check_cpu_usage(job_time, job_cpu_time, fileStore)
 
             # Finish up the stats
             if stats is not None:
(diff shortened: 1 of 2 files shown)


src/toil/test/src/jobTest.py
```diff
@@ -35,6 +37,7 @@
 
 
 class TestJob:
+
```
adamnovak, line 40:
```suggestion
```

src/toil/test/src/jobTest.py
```diff
@@ -815,7 +818,51 @@ def cyclic(fNode: int, visited: set[int], stack: list[int]) -> bool | int:
             if cyclic(i, visited, []):
                 return False
         return True
+    
+    def test_cpu_usage(self, subtests: pytest.Subtests) -> None:
+        with subtests.test("no warnings for non-positive runtime"):
+            job = Job()
+            file_store = Mock()
+            job._check_cpu_usage(
+                job_time=0.0, job_cpu_time=100.0, fileStore=file_store)
```
adamnovak: The layout is inconsistent here for no reason I can see.
```suggestion
                job_time=0.0, job_cpu_time=100.0, fileStore=file_store
            )
```

src/toil/test/src/jobTest.py
```diff
@@ -815,7 +818,51 @@ def cyclic(fNode: int, visited: set[int], stack: list[int]) -> bool | int:
...
+            file_store = Mock()
+            job.cores = 0
+            job._check_cpu_usage(
+                job_time=10.0, job_cpu_time=10.0, fileStore=file_store
+            )
+            file_store.log_to_leader.assert_called_once()
+            invalid_cores_message = file_store.log_to_leader.call_args.args[0]
+            invalid_cores_level = file_store.log_to_leader.call_args.kwargs["level"]
+            assert "invalid requested cores value 0" in invalid_cores_message
+            assert invalid_cores_level == logging.WARNING
```
adamnovak: The point that there ought to be a warning for 0 cores is not documented as part of the function's docstring, so it seems odd to me to test it. We end up with a test that just constrains the function to do what it currently happens to do, rather than what it promises to do. In a perfect world, we *ought* to be able to delete the body of a function and reconstruct a version of it that passes all the tests from just the documentation.

I would do something like this, where we make sure we can run the function through with 0 cores but we are agnostic as to whether it complains or not.
```suggestion
        with subtests.test("accepts 0 cores"):
            job = Job()
            file_store = Mock()
            job.cores = 0
            job._check_cpu_usage(
                job_time=10.0, job_cpu_time=10.0, fileStore=file_store
            )
```

We could alternatively add the point that the function promises to warn on 0 cores to its documentation.

src/toil/test/src/jobTest.py
```diff
@@ -815,7 +818,51 @@ def cyclic(fNode: int, visited: set[int], stack: list[int]) -> bool | int:
...
+            threshold_store = Mock()
+            threshold_job._check_cpu_usage(
+                job_time=10.0, job_cpu_time=21.1, fileStore=threshold_store
+            )
+            threshold_store.log_to_leader.assert_called_once()
+            cpu_message = threshold_store.log_to_leader.call_args.args[0]
+            cpu_level = threshold_store.log_to_leader.call_args.kwargs["level"]
+            assert "used " in cpu_message
+            assert "more CPU than the requested 2 cores" in cpu_message
+            assert cpu_level == logging.WARNING
```
adamnovak, line 875: These are testing both the behavior at the threshold and the default threshold value, and we don't promise that the threshold will always be exactly 1.05.

It's also not immediately clear that the magic numbers have the right relationship for these tests to work.

I would do something like:
```suggestion
        TIME = 10
        THRESHOLD = 1.05
        CORES = 2
        with subtests.test("threshold boundary does not warn"):
            threshold_job = Job(cores=CORES)
            threshold_store = Mock()
            threshold_job._check_cpu_usage(
                job_time=TIME,
                job_cpu_time=TIME * CORES * THRESHOLD,
                fileStore=threshold_store,
                threshold_factor=THRESHOLD,
            )
            threshold_store.log_to_leader.assert_not_called()

        with subtests.test("just above threshold warns"):
            threshold_job = Job(cores=CORES)
            threshold_store = Mock()
            threshold_job._check_cpu_usage(
                job_time=TIME,
                job_cpu_time=TIME * CORES * THRESHOLD + 0.1,
                fileStore=threshold_store,
                threshold_factor=THRESHOLD,
            )
            threshold_store.log_to_leader.assert_called_once()
            cpu_message = threshold_store.log_to_leader.call_args.args[0]
            cpu_level = threshold_store.log_to_leader.call_args.kwargs["level"]
            assert str(threshold_job.description) in cpu_message
            assert cpu_level == logging.WARNING
```
(I'm also de-constraining the message text here, to test whether a warning that mentions the job is sent, rather than the specific wording of the message.)

adamnovak (changes requested):
The implementation looks good here.

The tests seem like they were written to pin down exactly what the implementation happens to do right now, rather than what is important about what it promises to do now and in the future. In general we want to test interfaces and not implementation details. For example, if we decide to reword the message to say `used XXX times as much CPU as requested` instead of `used XXXx more CPU than the requested X cores`, we don't want the tests to start to fail.

One way to avoid this is to write the bulk of the tests before writing the implementation. If generating tests, it might make sense to provide the generator only with the documentation/signature for the function, and avoid showing it the body, or at least send along exhortations to ignore it.

src/toil/job.py
```diff
@@ -3337,6 +3325,46 @@ def _executor(self, stats: StatsDict, fileStore: AbstractFileStore) -> Iterator[
...
+        threshold_factor: float = 1.05,
+    ) -> None:
+        """
+        Log a warning when a job consumes more CPU time than its allocation.
+
+        If *job_time* is non-positive, returns immediately (wall-clock time may
+        lack resolution for very short jobs). If *cores* is non-positive, logs
+        a warning that overuse cannot be assessed and returns, avoiding
+        division by zero. Otherwise, compares *job_cpu_time* to
+        ``job_time * cores * threshold_factor``.
```
adamnovak, line 3344: Usually I think we put parameter names in double-backticks or just don't set them off, but this looks more like how the Python official docs do it.

This might be a little more like just the algorithm rewritten in English than we want it to be, but it's probably fine.

adamnovak (approved):

PR #5520, opened 2026-05-19
Fix inaccessible job store errors for grid batch systems (#5007)

### LINKED ISSUE #5007: It's hard to figure out that a Slurm WDL workflow is failing because the default job store isn't on a shared filesystem
I ran this:
```
toil-wdl-runner --batchSystem slurm https://raw.githubusercontent.com/vgteam/vg_wdl/lr-giraffe/workflows/giraffe_and_deepvariant.wdl https://raw.githubusercontent.com/vgteam/vg_wdl/lr-giraffe/params/giraffe_and_deepvariant.json --outputDirectory output/preset/
```

My workflow failed and I got no logs from any of the tasks.

When I manually added `--batchLogsDir logs/` I got a vaguely informative error:

```
[2024-07-03T13:31:46-0700] [MainThread] [W] [toil.leader] The batch system left an empty file logs/toil_5afe8eb3-2e69-47a4-aa37-6802bd7a6bd2.6.3865603.out.log
[2024-07-03T13:31:46-0700] [MainThread] [W] [toil.leader] The batch system left a non-empty file logs/toil_5afe8eb3-2e69-47a4-aa37-6802bd7a6bd2.6.3865603.err.log:
[2024-07-03T13:31:46-0700] [MainThread] [W] [toil.leader] Log from job "kind-WDLTaskJob/instance-g25utj5q" follows:
=========>
      File "/private/home/anovak/workspace/toil/venv/bin/_toil_worker", line 33, in <module>
        sys.exit(load_entry_point('toil', 'console_scripts', '_toil_worker')())
      File "/private/home/anovak/workspace/toil/src/toil/worker.py", line 772, in main
        job_store = Toil.resumeJobStore(options.jobStoreLocator)
      File "/private/home/anovak/workspace/toil/src/toil/common.py", line 1036, in resumeJobStore
        jobStore.resume()
      File "/private/home/anovak/workspace/toil/src/toil/jobStores/fileJobStore.py", line 130, in resume
        raise NoSuchJobStoreException(self.jobStoreDir, "file")
    toil.jobStores.abstractJobStore.NoSuchJobStoreException: The job store 'file:/data/tmp/tmpjn0glxf6/tree' does not exist, so there is nothing to restart.
<=========
```

The error message shouldn't say `there is nothing to restart` if it happens when a worker is tryign to connect to the job store; it should say something else.

Also, we should make it harder for the user to get into this situation where Toil has selected a job store path that can't work. Maybe when the batch system is Slurm or one of the other grid engine ones, `toil-wdl-runner` should pick a default job store in the current directory?

We could also give the worker a special magic exit code to use to complain specifically that it can't reach the job store, prompting a useful error message from the leader.

annagiroti:
Grid batch systems (Slurm, LSF, etc.) now raise an error upfront when no `--jobStore` is specified, directing users to provide a shared filesystem path. Workers that cannot access the job store now exit with a sentinel code, prompting a clear warning from the leader instead of a confusing traceback. 

Resolves #5007

## Changelog Entry

 * Fix inaccessible job store errors for grid batch systems (Slurm, LSF, etc.) (#5007)

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/test/jobStores/jobStoreTest.py
+++ b/src/toil/test/jobStores/jobStoreTest.py
@@ -1835,6 +1835,60 @@ class AWSJobStoreTest(AbstractJobStoreTest.Test):
         delete_s3_bucket(resource, bucket.name)
 
 
+class GenerateDefaultJobStoreTest(ToilTest):
+    """Tests for generate_default_job_store in toil.jobStores.utils."""
+
+    def setUp(self):
+        super().setUp()
+        from toil.lib.io import mkdtemp
+        self.local_dir = mkdtemp()
+
+    def tearDown(self):
+        super().tearDown()
+        shutil.rmtree(self.local_dir, ignore_errors=True)
+
+    def test_single_machine_returns_local_path(self):
+        """Default batch system should return a file job store under the given local directory."""
+        from toil.jobStores.utils import generate_default_job_store
+        result = generate_default_job_store("single_machine", None, self.local_dir)
+        self.assertEqual(result, self.local_dir)
+
+    def test_none_batch_system_returns_local_path(self):
+        """None batch system should fall back to single_machine and return a file job store."""
+        from toil.jobStores.utils import generate_default_job_store
+        result = generate_default_job_store(None, None, self.local_dir)
+        self.assertEqual(result, self.local_dir)
+
+    def test_slurm_raises_no_available_job_store(self):
+        """Slurm batch system should raise NoAvailableJobStoreException."""
+        from toil.jobStores.utils import generate_default_job_store, NoAvailableJobStoreException
+        with self.assertRaises(NoAvailableJobStoreException):
+            generate_default_job_store("slurm", None, self.local_dir)
+
+    def test_slurm_error_mentions_shared_filesystem(self):
+        """Slurm error message should tell the user to use a shared filesystem."""
+        from toil.jobStores.utils import generate_default_job_store, NoAvailableJobStoreException
+        with self.assertRaises(NoAvailableJobStoreException) as cm:
+            generate_default_job_store("slurm", None, self.local_dir)
+        self.assertIn("shared", str(cm.exception).lower())
+        self.assertIn("--jobStore", str(cm.exception))
+
+    def test_grid_engine_batch_systems_raise_no_available_job_store(self):
+        """All grid engine batch systems should raise NoAvailableJobStoreException."""
+        from toil.jobStores.utils import generate_default_job_store, NoAvailableJobStoreException
+        grid_batch_systems = ["slurm", "lsf", "grid_engine", "torque", "htcondor"]
+        for batch_system in grid_batch_systems:
+            with self.subTest(batch_system=batch_system):
+                with self.assertRaises(NoAvailableJobStoreException):
+                    generate_default_job_store(batch_system, None, self.local_dir)
+
+    def test_decoration_included_when_no_local_suggestion(self):
+        """Decoration string should appear in the auto-generated fallback path when no local suggestion is given."""
+        from toil.jobStores.utils import generate_locator
+        result = generate_locator("file", local_suggestion=None, decoration="wdl")
+        self.assertIn("wdl", result)
+
+
 # @needs_aws_s3
 class InvalidAWSJobStoreTest(ToilTest):
     def testInvalidJobStoreName(self):
--- a/src/toil/test/jobStores/jobStoreTest.py
+++ b/src/toil/test/jobStores/jobStoreTest.py
@@ -1835,58 +1835,45 @@ class AWSJobStoreTest(AbstractJobStoreTest.Test):
         delete_s3_bucket(resource, bucket.name)
 
 
-class GenerateDefaultJobStoreTest(ToilTest):
+class TestGenerateDefaultJobStore:
     """Tests for generate_default_job_store in toil.jobStores.utils."""
 
-    def setUp(self):
-        super().setUp()
-        from toil.lib.io import mkdtemp
-        self.local_dir = mkdtemp()
-
-    def tearDown(self):
-        super().tearDown()
-        shutil.rmtree(self.local_dir, ignore_errors=True)
-
-    def test_single_machine_returns_local_path(self):
-        """Default batch system should return a file job store under the given local directory."""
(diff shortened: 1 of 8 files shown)


src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1835,6 +1835,60 @@ def _cleanUpExternalStore(self, bucket):
...
+        for batch_system in grid_batch_systems:
+            with self.subTest(batch_system=batch_system):
+                with self.assertRaises(NoAvailableJobStoreException):
+                    generate_default_job_store(batch_system, None, self.local_dir)
+
+    def test_decoration_included_when_no_local_suggestion(self):
+        """Decoration string should appear in the auto-generated fallback path when no local suggestion is given."""
+        from toil.jobStores.utils import generate_locator
+        result = generate_locator("file", local_suggestion=None, decoration="wdl")
+        self.assertIn("wdl", result)
```
adamnovak: This is not a test for `generate_default_job_store`, but the class claims to be all tests for `generate_default_job_store`.

This could be bumped up a level and named `test_generate_locator_includes_decoration_when_no_local_suggestion`. Or maybe `test_generate_locator_includes_decoration_without_local_suggestion`. Or if we're only going to have the one it could just be `test_generate_locator_includes_decoration`.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1835,6 +1835,60 @@ def _cleanUpExternalStore(self, bucket):
...
+    """Tests for generate_default_job_store in toil.jobStores.utils."""
+
+    def setUp(self):
+        super().setUp()
+        from toil.lib.io import mkdtemp
+        self.local_dir = mkdtemp()
+
+    def tearDown(self):
+        super().tearDown()
+        shutil.rmtree(self.local_dir, ignore_errors=True)
```
adamnovak: Probably we want this to be a new-style test that uses `tmp_path: Path` as an argument to get a temp directory fixture, rather than an old-style test extending `ToilTest`.

If we do extend `ToilTest`, we'd want to use its methods for getting temp directories in tests.

But in neither case do we want to add new code to set up and tear down temp directories.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1835,6 +1835,60 @@ def _cleanUpExternalStore(self, bucket):
...
+        with self.assertRaises(NoAvailableJobStoreException):
+            generate_default_job_store("slurm", None, self.local_dir)
+
+    def test_slurm_error_mentions_shared_filesystem(self):
+        """Slurm error message should tell the user to use a shared filesystem."""
+        from toil.jobStores.utils import generate_default_job_store, NoAvailableJobStoreException
+        with self.assertRaises(NoAvailableJobStoreException) as cm:
+            generate_default_job_store("slurm", None, self.local_dir)
+        self.assertIn("shared", str(cm.exception).lower())
+        self.assertIn("--jobStore", str(cm.exception))
```
adamnovak: Does it really make sense to have one test to check that the error happens and a *different* test to make sure that the error happens with the right message text? It's not possible for the second test to pass if the first test fails.

src/toil/test/jobStores/jobStoreTest.py
```diff
@@ -1835,6 +1835,60 @@ def _cleanUpExternalStore(self, bucket):
...
+
+    def test_grid_engine_batch_systems_raise_no_available_job_store(self):
```
adamnovak: I don't think the test function names here are right.

Mostly in Toil we use short and not-so-descriptive test names; the test class might be `JobStoreGenerationTest` and we'd have things like `test_local()`, `test_slurm()`, etc.

But that's not *really* unit testing; that's writing one test per situation when really you want one test per thing that needs to be true about the unit under test. See for example [this discussion](https://softwareengineering.stackexchange.com/q/7823) on whether tests should always contain exactly one assert.

But then if you're taking a more proper unit testing approach, you want your functions to be named [something like](https://www.py4u.org/blog/python-unit-test-naming-convention-for-module-functions/) `test_<function>_<scenario>_<outcome>`. (You also might not put them in class if the thing they are testing is not itself a class? But also it makes sense to put them in a class to group them.)

The names here are sort of `test_<scenario>_<outcome>`, I think, except when they aren't. `test_decoration_included_when_no_local_suggestion` is `test_<outcome>_<scenario>`, for example. I think that might actually be a more sensible order when the `<function>` has to be implied by being in the test class, because you are not awkwardly missing your subject noun phrase in the name.

The `test_<scenario>_<outcome>` names can kind of read like they might be `test_<unit>_<outcome>`, which I find confusing: `test_single_machine_returns_local_path` parses to me like "This is a test that makes sure that single machine (whatever that is) returns a local path". And `test_grid_engine_batch_systems_raise_no_available_job_store` has "raise" instead of "raises", which also makes it sound like the gird engine batch systems, and not the `generate_default_job_store` unit from the enclosing class, is what's meant to be doing the raising.

I think the test functions should be harmonized to use one naming convention.

src/toil/wdl/wdltoil.py
```diff
@@ -6095,8 +6097,11 @@ def main() -> None:
 
     # Make sure we have a jobStore
     if options.jobStore is None:
-        # TODO: Move cwltoil's generate_default_job_store where we can use it
-        options.jobStore = os.path.join(mkdtemp(), "tree")
+        jobstore = mkdtemp(prefix="toil-wdl-", dir=os.getcwd())
+        os.rmdir(jobstore)
```
adamnovak, line 6101: I *think* the file job store logic insists on actually creating the directory, which means that if someone malicious can write to the directory we're standing in, they can cause our run to fail by creating the directory themselves between when we delete it and when we go to re-create it, but they *can't* force us to use a directory they provide full of malicious files.

But it still would be good to eventually replace the notion of passing a suggested directory to use as the job store with instead an approach where we pass a suggested directory to put the job store in. We could also use a high-entropy UUID that is not created as a directory until we actually go to use it for a job store.

adamnovak (changes requested):
This looks pretty good, but I think the tests need some rationalization.

adamnovak:
@annagiroti Can you also fill in the proposed changelog entry for this?

adamnovak (approved):

PR #5526, opened 2026-05-28
Add resource usage measurement injection to CWL interpreter

avnig05:
CWL and WDL have a shared runtime injection module that samples in-container CPU/memory usage and reports it back to Toil via runtime message files.
## Changelog Entry

 * `toil-cwl-runner --stats` now records resource usage from Docker containers so `toil stats` can read it.
 (Edited by @adamnovak)

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -1105,32 +1105,6 @@ class ToilSingleJobExecutor(cwltool.executors.SingleJobExecutor):
         return super().run_jobs(process, job_order_object, logger, runtime_context)
 
 
-class ToilContainerCommandLineJob(ContainerCommandLineJob):
-    """Container job that collects resource stats from injected in-container code."""
-
-    def _execute(
-        self,
-        runtime: list[str],
-        env: MutableMapping[str, str],
-        runtimeContext: cwltool.context.RuntimeContext,
-        monitor_function: Callable[["subprocess.Popen[str]"], None] | None = None,
-    ) -> None:
-        super()._execute(runtime, env, runtimeContext, monitor_function)
-        handle_injection_messages_from_outdir(self.outdir)
-
-
-class ToilDockerCommandLineJob(ToilContainerCommandLineJob, DockerCommandLineJob):
-    """Docker container job with Toil runtime injection support."""
-
-
-class ToilPodmanCommandLineJob(ToilContainerCommandLineJob, PodmanCommandLineJob):
-    """Podman container job with Toil runtime injection support."""
-
-
-class ToilSingularityCommandLineJob(ToilContainerCommandLineJob, SingularityCommandLineJob):
-    """Singularity container job with Toil runtime injection support."""
-
-
 class ToilTool:
     """Mixin to hook Toil into a cwltool tool type."""
 
@@ -1189,20 +1163,10 @@ class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
     """Subclass the cwltool command line tool to provide the custom ToilPathMapper
     and add the monitoring code to the job's container command line."""
 
-    def make_job_runner(
-        self, runtimeContext: cwltool.context.RuntimeContext
-    ) -> type[cwltool.job.JobBase]:
-        """Use Toil container job classes that collect injected runtime messages."""
-        parent_class = super().make_job_runner(runtimeContext)
-        if parent_class is DockerCommandLineJob:
-            return ToilDockerCommandLineJob
-        if parent_class is PodmanCommandLineJob:
-            return ToilPodmanCommandLineJob
-        if parent_class is SingularityCommandLineJob:
-            return ToilSingularityCommandLineJob
-        return parent_class
-
     def _uses_container(self, runtimeContext: cwltool.context.RuntimeContext) -> bool:
+        """
+        Returns True if this tool will be run inside a container.
+        """
         if not runtimeContext.use_container:
             return False
         docker_req, _ = self.get_requirement("DockerRequirement")
@@ -1212,10 +1176,16 @@ class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
             return runtimeContext.find_default_container(self) is not None
         return runtimeContext.default_container is not None
 
+    # TODO: Why is this on ToilCommandLineTool?
     @staticmethod
     def _file_mounts_from_pathmapper(
         job: ContainerCommandLineJob,
     ) -> list[tuple[str, str]]:
+        """
+        Get all the container mounts that will be used for a containerized job.
+
+        :returns: a list of (host path, container path) tuples, one per mount.
+        """
         file_mounts: list[tuple[str, str]] = []
         for location in job.pathmapper.files():
             ent = job.pathmapper.mapper(location)
@@ -1244,6 +1214,30 @@ class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
                 script = add_injections(script, file_mounts)
                 job.command_line = shell_script_to_command_line(script)
             yield job
+
+     def collect_output_ports(
+        self,
+        ports: CommentedSeq | set[CWLObjectType],
+        builder: cwltool.builder.Builder,
+        outdir: str,
+        rcode: int,
+        compute_checksum: bool = True,
+        jobname: str = "",
+        readers: MutableMapping[str, CWLFileType | CWLDirectoryType] | None = None,
+    ) -> cwltool.command_line_tool.OutputPortsType:
+        """
+        Hook output collection to also collect resource usage statistics.
+        """
+        handle_injection_messages_from_outdir(outdir)
+        return super().collect_output_ports(
+            ports,
+            builder,
+            outdir,
+            rcode,
+            compute_checksum,
+            jobname,
+            readers,
+        )        
 
     def _initialworkdir(
         self, j: cwltool.job.JobBase | None, builder: cwltool.builder.Builder
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -84,6 +84,7 @@ from cwltool.software_requirements import (
     get_container_from_software_requirements,
 )
 from cwltool.stdfsaccess import StdFsAccess, abspath
+from cwltool.udocker import UDockerCommandLineJob
 from cwltool.utils import (
     adjustDirObjs,
     aslist,
@@ -1158,6 +1159,13 @@ class ToilTool:
         """Return string representation of this tool type."""
         return f'{self.__class__.__name__}({repr(getattr(self, "tool", {}).get("id", "???"))})'
 
+# We need to know which cwltool container implementations run the container
+# under the calling process's child tree. Unlisted container implementations
+# are assumed to run it elsewhere through a deamon.
+#
+# We have to list it this way around (and not list the implementations that use
+# a daemon) because some of these extend DockerCommandLineJob.
+CHILD_PROCESS_CONTAINER_JOBS = (PodmanCommandLineJob, SingularityCommandLineJob, UDockerCommandLineJob)
 
 class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
     """Subclass the cwltool command line tool to provide the custom ToilPathMapper
@@ -1206,16 +1214,27 @@ class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
         the job's container command line.
         """
         for job in super().job(job_order, output_callbacks, runtimeContext):
-            if isinstance(job, ContainerCommandLineJob) and self._uses_container(
-                runtimeContext
+            if (
+                isinstance(job, ContainerCommandLineJob) and
+                not isinstance(job, CHILD_PROCESS_CONTAINER_JOBS) and 
+                self._uses_container(
+                    runtimeContext
+                )
             ):
+                # This job is going to run in a container that makes it not be
+                # a descendant of our process. So we need to inject code to
+                # count its resource usage.
+
+                # That code also checks to make sure files mounted on Docker
+                # for Mac are intact, so we need to tell it ablut the file
+                # mounts we are going to use.
                 file_mounts = self._file_mounts_from_pathmapper(job)
                 script = command_line_to_shell_script(job.command_line)
                 script = add_injections(script, file_mounts)
                 job.command_line = shell_script_to_command_line(script)
             yield job
 
-     def collect_output_ports(
+    def collect_output_ports(
         self,
         ports: CommentedSeq | set[CWLObjectType],
         builder: cwltool.builder.Builder,
@@ -1223,7 +1242,7 @@ class ToilCommandLineTool(ToilTool, cwltool.command_line_tool.CommandLineTool):
         rcode: int,
         compute_checksum: bool = True,
         jobname: str = "",
-        readers: MutableMapping[str, CWLFileType | CWLDirectoryType] | None = None,
+        readers: MutableMapping[str, CWLObjectType] | None = None,
     ) -> cwltool.command_line_tool.OutputPortsType:
         """
         Hook output collection to also collect resource usage statistics.
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -31,13 +31,21 @@ import os
 import pprint
 import shutil
 import stat
+import subprocess
 import sys
 import textwrap
 import uuid
 
 # This is also in configargparse but MyPy doesn't know it
 from argparse import RawDescriptionHelpFormatter
-from collections.abc import Callable, Iterator, Mapping, MutableMapping, MutableSequence
+from collections.abc import (
+    Callable,
+    Generator,
+    Iterator,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+)
 from tempfile import NamedTemporaryFile, TemporaryFile, gettempdir
 from threading import Thread
 from typing import IO, Any, Literal, Optional, Protocol, TextIO, TypeVar, Union, cast
@@ -67,6 +75,8 @@ from cwltool.process import (
     fill_in_defaults,
     shortname,
 )
+from cwltool.docker import DockerCommandLineJob, PodmanCommandLineJob
+from cwltool.job import ContainerCommandLineJob
 from cwltool.secrets import SecretStore
 from cwltool.singularity import SingularityCommandLineJob
 from cwltool.software_requirements import (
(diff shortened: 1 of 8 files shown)


adamnovak:
@avnig05 It looks like this isn't passing type checking; if you `make mypy` you should be able to reproduce the problems:
https://ucsc-ci.com/databiosphere/toil/-/jobs/110094#L1195

It looks like mostly you need casts or `#  type: ignore` comments at some places where you're using a fake object or looking in one of these `Expando` objects that are fundamentally dynamically-typed. But it looks like you might not be importing `ResourceMonitor` from the place where it's actually defined?

src/toil/cwl/cwltoil.py
```diff
@@ -1143,7 +1183,64 @@ def __str__(self) -> str:
...
+        if docker_req is not None:
+            return True
+        if runtimeContext.find_default_container is not None:
+            return runtimeContext.find_default_container(self) is not None
+        return runtimeContext.default_container is not None
+
+    @staticmethod
+    def _file_mounts_from_pathmapper(
+        job: ContainerCommandLineJob,
+    ) -> list[tuple[str, str]]:
```
adamnovak: This needs a docstring to at least explain which item in the result tuples is which.

src/toil/cwl/cwltoil.py
```diff
@@ -1088,6 +1102,32 @@ def run_jobs(
...
+    """Docker container job with Toil runtime injection support."""
+
+
+class ToilPodmanCommandLineJob(ToilContainerCommandLineJob, PodmanCommandLineJob):
+    """Podman container job with Toil runtime injection support."""
+
+
+class ToilSingularityCommandLineJob(ToilContainerCommandLineJob, SingularityCommandLineJob):
+    """Singularity container job with Toil runtime injection support."""
+
```
adamnovak: This approach of hooking the `ContainerCommandLineJob` in addition to the `CommandLineTool` is a little awkward. We have to have all these extra classes for each container system, and we already don't have one for `UDockerCommandLineJob` and it's hard to tell whether that's because we don't need one. And the hook here only works in concert with the hook that `ToilCommandLineTool` applies, so we end up with one feature spread out over several classes.

Instead of hooking `ContainerCommandLineJob._execute()`, did you consider hooking `CommandLineTool.collect_output_ports()` instead? It looks like that [has access to the `outdir`](https://github.com/common-workflow-language/cwltool/blob/8949fc2d68efe128ce841739d1190645dbd233bb/cwltool/command_line_tool.py#L1205), and it [gets sent](https://github.com/common-workflow-language/cwltool/blob/8949fc2d68efe128ce841739d1190645dbd233bb/cwltool/command_line_tool.py#L1173-L1174) into `_execute()` and [called inside there](https://github.com/common-workflow-language/cwltool/blob/8949fc2d68efe128ce841739d1190645dbd233bb/cwltool/job.py#L376) after the container has run. And if we did it that way we could keep all the hook logic together inside `ToilCommandLineTool` and not need to worry as much about the container type.

src/toil/cwl/cwltoil.py
```diff
@@ -1143,7 +1183,64 @@ def __str__(self) -> str:
...
+        the job's container command line.
+        """
+        for job in super().job(job_order, output_callbacks, runtimeContext):
+            if isinstance(job, ContainerCommandLineJob) and self._uses_container(
+                runtimeContext
+            ):
+                file_mounts = self._file_mounts_from_pathmapper(job)
+                script = command_line_to_shell_script(job.command_line)
+                script = add_injections(script, file_mounts)
+                job.command_line = shell_script_to_command_line(script)
```
adamnovak: Is this going to inject the monitoring logic even when we're using Singularity (or I think UDocker?) where we should already see CPU and memory usage in the stats because it happens under a child process of Toil? Will that lead to double-counting of CPU usage?

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+from typing import Iterable
+
+from toil.lib.resources import ResourceMonitor
+
+logger = logging.getLogger(__name__)
+
+###
+# Runtime code injection system
+# When a workflow steps runs inside a container, the Toil worker process on the host
+# often cannot see how much CPU and RAM that step actually used. This system allows
```
adamnovak: The word `often` is doing some vigorous handwaving here, and trying to hide that this text does not convey an understanding of how or why (or, more importantly for avoiding double-counting of resource usage, *when*) this will be the case.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+# that can be executed in the container.
+def command_line_to_shell_script(command_line: list[str]) -> str:
+    """
+    Extract or synthesize the inner shell script from a cwltool argv list.
+
+    cwltool uses ``["/bin/sh", "-c", script]`` when ShellCommandRequirement is
+    present, and a flat argv list otherwise.
+    """
+    if (
+        len(command_line) >= 3
```
adamnovak: This might want to be `==` or we'll be able to throw away arguments.

src/toil/lib/interpreter.py
adamnovak, line 1: Rather than having a comment explaining what each function is for and how it works, we should *mostly* convey that information in docstrings. We should only have comments above the functions to convey stuff that can't be put into the docstrings for some reason.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+        and command_line[1] == "-c"
+    ):
+        return command_line[2]
+    return " ".join(shlex.quote(arg) for arg in command_line) # this is the shell script string
+
+
+# Helper function to convert a shell script back into an argv list that can be used by cwltool.
+def shell_script_to_command_line(script: str) -> list[str]:
+    """Wrap a shell script for cwltool/docker (injection requires bash)."""
+    return ["/bin/bash", "-c", script]
```
adamnovak: Since this file isn't about CWL, the behavior contracts for these functions should be explained in general, without specifically talking about what `cwltool` code needs or happens to do.

If we want special code for `cwltool` quirks (like the code to unpack a `bash -c` command when possible instead of adding another level of quoting), we can mention that we're doing this because `cwltool` tends to do such-and-such and we want to handle that well. But whether the functions work overall should be something we can determine without needing to know that `cwltool` even exists. We want to abstract a little more so we no longer need to think about it.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+    command_string: str,
+    file_mounts: Iterable[tuple[str, str]],
+    message_dir: str = INJECTED_MESSAGE_DIR,
+) -> str:
+    """
+    Inject extra Bash code from the Toil runtime into the command for the container.
+
+    Currently doesn't implement the MiniWDL plugin system, but does add
+    resource usage monitoring to Docker containers.
+    """
```
adamnovak: Similarly, if this function is here now, it should no longer be explained in terms of MiniWDL ideas like its plugin system, or the idea that Docker containers have anything to do with Docker Swarm.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+
+    parts = []
+    # We're running on Docker Swarm, so we need to monitor CPU usage
```
adamnovak:
```suggestion
    # We're running on Docker, so we need to monitor CPU usage
```

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+
+    return "\n".join(parts)
+
+# Helper function to parse the injected resource usage monitoring code and update the resource usage stats.
+def handle_message_file(file_path: str) -> None:
+    """
+    Handle a message file received from in-container injected code.
+
+    Takes the host-side path of the file.
+    """
```
adamnovak: It's a little disconcerting that this doesn't need to know anything about the current Toil job and doesn't need to return anything.

It works because it talks to the global per-process `ResourceMonitor`. But maybe the docstring should be more specific and not just say that we "handle" the file, but that we specifically record the resource usage described in it (somewhere) as if it had occurred in a child process.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,194 @@
...
+def add_injections(
+    command_string: str,
+    file_mounts: Iterable[tuple[str, str]],
+    message_dir: str = INJECTED_MESSAGE_DIR,
+) -> str:
+    """
+    Inject extra Bash code from the Toil runtime into the command for the container.
+
+    Currently doesn't implement the MiniWDL plugin system, but does add
+    resource usage monitoring to Docker containers.
```
adamnovak: We don't really cover that this should *only* be used for Docker containers and other daemon-based container systems, and should *not* be used for Singularity containers or in other cases where the container system keeps the contained processes as our descendants. But then in the body we claim that we're definitely running on Docker and not some other container system. That claim in the function body won't be reliably true unless it's documented as a requirement that the caller of the function needs to guarantee to us.

src/toil/test/cwl/cwlTest.py
```diff
@@ -482,6 +484,37 @@ def test_s3_as_secondary_file(self, tmp_path: Path) -> None:
...
+            with get_data("test/cwl/empty.json") as inputs_file:
+                job_store = tmp_path / "jobStore"
+                main_args = [
+                    "--outdir",
+                    str(tmp_path / "output_dir"),
+                    str(cwl_file),
+                    str(inputs_file),
+                    "--jobStore",
+                    str(job_store),
+                    "--stats",
```
adamnovak, line 506: There's nothing in here to make Docker be used, but I looked at the `toil-cwl-runner` options and it has `--singularity`, `--podman`, and `--no-container`, but not a `--docker`. So I guess Docker is the default container engine. Maybe we need a comment here to remind us that the test depends on that?

src/toil/test/cwl/cwlTest.py
```diff
@@ -2283,6 +2281,84 @@ def test_download_structure(tmp_path: Path) -> None:
...
+    # Include malformed and partial lines to exercise parser robustness.
+    message_file.write_text(
+        "CPU\t1000000\n"
+        "Memory\t1024\n"
+        "CPU\t4000000\n"
+        "Memory\t2048\n"
+        "Odd\tline\n"
+        "CPU\t5000000",
+        encoding="utf-8",
+    )
```
adamnovak: One thing that's noticeably absent here is a line with the wrong number of fields, and in particular too few fields.

src/toil/test/cwl/cwlTest.py
```diff
@@ -2283,6 +2281,84 @@ def test_download_structure(tmp_path: Path) -> None:
...
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """
+    Resource monitor messages emitted from a container are translated into
+    Toil's extra CPU and memory accounting.
+    """
+    from toil.lib import interpreter
+
+    message_file = tmp_path / "resources.tsv"
+    # Include malformed and partial lines to exercise parser robustness.
```
adamnovak: A some of what's being tested here (like the fact that final lines without a terminating newline are ignored) seems to be based on the implementation code, rather than anything that the docstring actually promises will be true about the function under test.

If it's an important enough point about the function's behavior that it's worth testing, it's probably worth documenting in the docstring. Otherwise, the next person to touch the function is going to tinker with the internals in a way that doesn't appear to change anything about what the function promises, and then be hit with test failures because there are other secret things that need to be true about the function.

src/toil/test/cwl/cwlTest.py
```diff
@@ -2283,6 +2281,84 @@ def test_download_structure(tmp_path: Path) -> None:
...
+        def files(self) -> list[str]:
+            return ["file://fake-input"]
+
+        def mapper(self, _location: str) -> FakeMapperEntry:
+            return self._entry
+
+    job = object.__new__(cwltoil.ToilDockerCommandLineJob)
+    job.command_line = ["echo", "hello"]
+    job.pathmapper = FakePathMapper(FakeMapperEntry(str(host_input), "/work/input.txt"))
+
```
adamnovak: There aren't any actual assertions in this test, so I don't think it's genuinely testing what its docstring claims it is testing.

src/toil/test/src/jobTest.py
adamnovak, line 875: I think these are leftover changes from #5512. It looks like your new commit c4c69a4 is on top of the commits from the branch for that PR, and not on top of the squashed commit that was created when the PR was merged.

I would recommend rebasing onto the current mainline Toil, and keeping just the commit about this feature in the branch for this PR:
1. Be on the branch for this PR.
2. Do `git fetch origin` (assuming your `origin` remote points to this Github project)
3. Do `git rebase -i origin/master`
4. In the editor, delete all the lines for the commits from the old feature, and only keep the line for the commit that's about the new feature.
5. Save and quit
6. Git should rewrite your history so now your `issues/5477-cpu-usage-docker-cwl` branch just has the new commit for this feature.
7. Do a `git push origin issues/5477-cpu-usage-docker-cwl -f` to force-push your rewritten history to Github and update the PR.

adamnovak (dismissed):
This looks like it should work for the Docker container case.

I think for the Singularity/CWL case, it might be double-counting the resource usage, because it will be counted via having a child process and counted *again* via the command injection.

I think it might be possible to simplify the hooking into the CWL code by overriding another method in `ToilCommandLineTool` that gets to run after the container finishes, which would let us avoid needing `ToilContainerCommandLineJob` and friends and the machinery to attach *them*.

I think the PR should be rebased to not be on top of the pre-squash commits from #5512.

adamnovak:
@avnig05 Since you're gone for the summer, I'm going to take on landing this myself.

src/toil/lib/interpreter.py
```diff
@@ -0,0 +1,238 @@
...
+    return " ".join(shlex.quote(arg) for arg in command_line) # this is the shell script string
+
+
+def shell_script_to_command_line(script: str) -> list[str]:
+    """
+    Wrap a shell script as a list of arguments for launchign a process.
+
+    The resulting command required Bash to be available.
+    """
+    return ["bash", "-c", script]
```
mr-c: Some containers and hosts don't have `bash`

mr-c (commented):
How does this interact with the built-in resource monitoring in `cwltool`?

https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/cwltool/job.py#L522

https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/cwltool/job.py#L861

src/toil/lib/interpreter.py
adamnovak: We only use this when doing the injection, which only has to worry about containers, but indeed we'd run into a problem if the container doesn't have Bash.

[The script we inject](https://github.com/DataBiosphere/toil/pull/5526/changes#diff-2f0ccefef48569ac0f7c0a7cc77943d76a213d80e9f3ce1d86324c51d7576735R107-R138) contains bash-isms, and uses `head`, `cut`, and `awk`, which are not shell built-ins and would have to be available inside the container for the monitoring to actually work.

We could translate the script back to `/bin/sh` (maybe also pulling in a dependency on something else for the math), and add some tests for what happens if a user container doesn't ship standard tools an only ships their user binary (the monitoring can't work, but we need it to fail in a safe way). Or we could figure out how to get something smart enough to do all this injected into the container (maybe a Linux binary we mount?). Or we could try and figure out  how to do the monitoring outside the container, but that probably would require changes in both cwltool and miniwdl.

adamnovak:
> How does this interact with the built-in resource monitoring in `cwltool`?

This duplicates the functionality of the cwltool built-in memory monitoring, and adds CPU usage monitoring.

We could call out to the cwltool memory usage monitoring and only use our own code for CPU usage monitoring, but since Toil needs code for both on the WDL side it seemed simplest to just monitor both the same way on the CWL side as well.

If cwltool grew built-in CPU monitoring we could stop trying to tack on our own logic and use that instead.

adamnovak:
For some discussion about the design on the WDL side, see https://github.com/chanzuckerberg/miniwdl/issues/672. Mike Lin wrote up a plugin for MiniWDL demonstrating within-container monitoring, but that wasn't really quite published as installable and also ran the statistics in-band through the user-visible standard error stream, so I implemented it again with the same basic approach in Toil. If we really want to cut the whole command injection concept from Toil, we'd need to build out CPU monitoring in cwltool, and spruce up that plugin in MiniWDL so Toil could depend on it instead.

adamnovak:
The CWL conformance tests are failing on `initial_workdir_empty_writable_docker`, which might indicate that we're not actually allowed to hide the communications channel for the resource monitoring inside the working directory.

adamnovak:
OK, I have code to fix the need for Bash or other utilities in my Toil fork, but I still haven't worked out how to inject another mount from where I'm working.

adamnovak:
The remaining problem here is the `docker_entrypoint` CWL test.

If I need to inject a script into the container, I need to figure out what the entrypoint is, put it into the script, and override the container's entrypoint with the shell.

(Another complication is that it's entirely possible to make a Docker container without any shell at all, *just* with the application binary. We might end up having to stuff a busybox in there. At which point why am I not sending something more useful? And will user code care if it isn't PID 1?)

mr-c:
For containers you can monitor memory and CPU usage outside the container

adamnovak:
> For containers you can monitor memory and CPU usage outside the container

Only if I do my own integration of the point-in-time CPU percentage from `docker stats`. The total CPU usage in core-seconds is not available except from the Docker API (see https://stackoverflow.com/questions/30271942/get-docker-container-cpu-usage-as-percentage and https://github.com/TheClimateCorporation/docker/blob/master/docs/reference/commandline/stats.md#formatting). And I don't think that the stats samples are very timestamped the way `cwltool`'s`docker_monitor` collects them now: https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/cwltool/job.py#L906-L911. So integrating over time would be hard.

I guess I could try and talk to the Docker API from the Toil CWL interpreter layer. But I don't like having that contact surface (`cwltool` itself never seems to use the Docker API), and then I probably end up with toil-and-cwl-specific code to do it, that I think wouldn't be able to be used on the WDL side because that's all Docker Swarm.

If `cwltool`-side stuff gains support for measuring (or back-calculating) core-seconds used, I could call into it for CWL, and that would be way better than this. But I'm trying to avoid getting into the existing monitoring code. (I'm also trying to put this project down because I am close to being out of available time to work on it, and I'm reluctant to throw out the whole approach I'd initially set Avni on and not merge any of her work.)

src/toil/cwl/cwltoil.py
```diff
@@ -1142,9 +1161,134 @@ def __str__(self) -> str:
...
+#
+# We have to list it this way around (and not list the implementations that use
+# a daemon) because some of these extend DockerCommandLineJob.
+CHILD_PROCESS_CONTAINER_JOBS = (PodmanCommandLineJob, SingularityCommandLineJob, UDockerCommandLineJob)
+
+
+class ToilDockerCommandLineJob(DockerCommandLineJob):
+    """Container job that collects resource stats from injected in-container code."""
+
+    def _execute(
```
mr-c, line 1176: This sub-class of `DockerCommandLineJob` (itself a sub-class of `ContainerCommandLineJob`) doesn't override `ContainerCommandLineJob.run()` which does its own process monitoring: https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/cwltool/job.py#L849-L858

PR #5540, opened 2026-06-23
Revert "Pin back binfmt"

adamnovak:
Reverts DataBiosphere/toil#5536

We're going back to the newer version for vg, and since CI hosts are now shared between jobs and this is scoped to the host, we need to match vg's version.

We probably should kick this out of the project altogether and just make the CI runners responsible for it.

## Changelog Entry

* Toil no longer uses the old version of the multi-arch support container because it causes compiler crashes for other projects on our CI (#5540)


adamnovak:
We also need https://github.com/DataBiosphere/toil/pull/5524 merged or the crash bugs when installing packages will come back.

PR #5542, opened 2026-06-24
Implement cwltool:Loop extension for CWL runner (#4227)

### LINKED ISSUE #4227: CWL: implement the workflow-level loop extension
During development of this, one would need to temporarily depend on https://github.com/common-workflow-language/cwltool/pull/1641 until that is merged and a new release of `cwltool` is made

``` diff
diff --git a/requirements-cwl.txt b/requirements-cwl.txt
index 935907d33..081714c4e 100644
--- a/requirements-cwl.txt
+++ b/requirements-cwl.txt
@@ -1,4 +1,4 @@
-cwltool==3.1.20220913185150
+cwltool @ git+https://github.com/common-workflow-language/cwltool/@loop_prototype#egg=@cwltool
 schema-salad>=8.3.20220913105718,<8.4
 galaxy-tool-util
 ruamel.yaml>=0.15,<=0.17.21
```

```
diff --git a/src/toil/cwl/cwltoil.py b/src/toil/cwl/cwltoil.py
index 6d20984be..de63298ce 100755
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -2689,6 +2689,8 @@ class CWLWorkflow(CWLNamedJob):
                         jobobj: Dict[
                             str, Union[ResolveSource, DefaultWithSource, StepValueFrom]
                         ] = {}
+                        loop = step.get_requirement("http://commonwl.org/cwltool#Loop")
+                        logger.info(f"Loop: {loop}.")
 
                         for inp in step.tool["inputs"]:
                             logger.debug("Takes input: %s", inp["id"])
```

Should we extend the toil-cwl [Conditional object](https://github.com/DataBiosphere/toil/blob/b404bef038d5553621ba78eece762538e3ce06d9/src/toil/cwl/cwltoil.py#L204), or is another approach needed?

annagiroti:
## Changelog Entry
Implement `cwltool:Loop` extension for Toil's CWL runner, adding iterative loop execution, `outputMethod: all` accumulation, and full `LoopInput` feature support (`loopSource`, `linkMerge`, `pickValue`, `default`, `valueFrom`) with a configurable iteration limit via `--cwl-loop-iteration-limit`.

Resolves #4227

## Reviewer Checklist

        ```
        ```

## Merger Checklist


src/toil/test/cwl/single_var_loop.cwl
adamnovak, line 1: I think that, rather than copy-pasting tests from cwltool, we should pull the test files out of the cwltool dependency at runtime, if we can.

It looks like cwltool has a `tests/loop-ext/single-var-loop.cwl` and a corresponding `tests/loop-ext/single-var-loop-job.yml`. Toil should be able to use a YAML instead of a JSON for the inputs just fine. These both [get packaged](https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/setup.py#L148-L149) inside the `cwltool.tests` package, so we can pull them out. You can list all the files with:

```
python3 -c 'import importlib.resources; print(list(importlib.resources.files("cwltool.tests").joinpath("loop-ext").iterdir()))'
```

Then in the code we can do something like Toil's `get_file()`, which gets a Toil test file, but using `cwltool.test` as the base module:
https://github.com/DataBiosphere/toil/blob/4ac2dcaa268618291f344fa06d43ea446ad2ec53/src/toil/test/__init__.py#L81

Then we can let cwltool be responsible for all these test files and not have to worry about fixing bugs in them.

src/toil/test/cwl/cwlTest.py
```diff
@@ -566,6 +566,163 @@ def test_required_input_condition_protection(self, tmp_path: Path) -> None:
...
+        """cwltool:Loop with one rebound input runs to completion (outputMethod=last, P2)."""
+        with get_data("test/cwl/single_var_loop.cwl") as cwl_file:
+            with get_data("test/cwl/single_var_loop.json") as job_file:
+                self._tester(cwl_file, job_file, {"o1": 10}, tmp_path, main_args=["--enable-ext"])
+
+    @needs_cwl
+    @pytest.mark.cwl
+    @pytest.mark.cwl_small
+    def test_loop_iteration_limit_exceeded(self, tmp_path: Path) -> None:
+        """--cwl-loop-iteration-limit aborts runaway loops with a clear error."""
```
adamnovak: This test claims to, but does not, check whether the error message produced is clear.

src/toil/test/cwl/cwlTest.py
```diff
@@ -566,6 +566,163 @@ def test_required_input_condition_protection(self, tmp_path: Path) -> None:
...
+        """cwltool:Loop with outputMethod=all and loopWhen=false on entry yields empty array."""
+        with get_data("test/cwl/all_output_loop_no_iteration.cwl") as cwl_file:
+            with get_data("test/cwl/single_var_loop.json") as job_file:
+                self._tester(cwl_file, job_file, {"o1": []}, tmp_path, main_args=["--enable-ext"])
+
+    @needs_cwl
+    @pytest.mark.cwl
+    @pytest.mark.cwl_small
+    def test_loop_with_all_output_method(self, tmp_path: Path) -> None:
+        """cwltool:Loop with outputMethod=all accumulates each iteration's outputs into a per-port array (P3)."""
```
adamnovak: The per-port arrays are back. Also, this test only has the one output port, so the per-port-ness of the arrays isn't really being tested anyway.

And what's P3?

src/toil/test/cwl/cwlTest.py
```diff
@@ -566,6 +566,163 @@ def test_required_input_condition_protection(self, tmp_path: Path) -> None:
...
+        with get_data("test/cwl/invalid_loop_when.cwl") as cwl_file:
+            with get_data("test/cwl/two_vars_loop_job.json") as job_file:
+                with pytest.raises(ValidationException):
+                    self._tester(cwl_file, job_file, {}, tmp_path, main_args=["--enable-ext"])
+
+    @needs_cwl
+    @pytest.mark.cwl
+    @pytest.mark.cwl_small
+    def test_loop_fail_non_boolean_loop_when(self, tmp_path: Path) -> None:
+        """A loopWhen expression that evaluates to a non-boolean fails at runtime (P5)."""
```
adamnovak: What are P4 and P5?

src/toil/test/cwl/cwlTest.py
```diff
@@ -566,6 +566,163 @@ def test_required_input_condition_protection(self, tmp_path: Path) -> None:
...
+        """cwltool:Loop with multi-source loopSource and pickValue (P4)."""
+        with get_data("test/cwl/multi_source_loop.cwl") as cwl_file:
+            with get_data("test/cwl/single_var_loop.json") as job_file:
+                self._tester(cwl_file, job_file, {"o1": [8, 11, 14, 17, 20]}, tmp_path, main_args=["--enable-ext"])
+
+    @needs_cwl
+    @pytest.mark.cwl
+    @pytest.mark.cwl_small
+    def test_nested_loops(self, tmp_path: Path) -> None:
+        """cwltool:Loop nested inside another cwltool:Loop (last over all) (P5)."""
```
adamnovak: What does "last over all" mean?

src/toil/test/cwl/cwlTest.py
```diff
@@ -566,6 +566,163 @@ def test_required_input_condition_protection(self, tmp_path: Path) -> None:
...
+        from io import StringIO
+        with get_data("test/cwl/invalid_non_boolean_loop_when.cwl") as cwl_file:
+            with get_data("test/cwl/two_vars_loop_job.json") as job_file:
+                rc = cwltoil.main(
+                    ["--enable-ext", f"--outdir={tmp_path}", str(cwl_file), str(job_file)],
+                    stdout=StringIO(),
+                )
+        assert rc != 0
+
+    # TODO: Test `StepInputExpressionRequirement` https://github.com/common-workflow-language/cwltool/blob/8949fc2d68efe128ce841739d1190645dbd233bb/tests/test_loop_ext.py#L271-L279
```
adamnovak: Did you mean to actually add this proposed test? Or do you want to leave it as future work?

src/toil/cwl/cwltoil.py
```diff
@@ -613,6 +617,90 @@ def do_eval(self, inputs: CWLObjectType) -> Any:
...
+        self.output_key = output_key
+
+    def __repr__(self) -> str:
+        """Allow for debug printing."""
+        return f"LoopOutputSource({self.output_key!r})"
+
+    def resolve(self) -> Any:
+        """Extract the named field from the prior iteration's output dict."""
+        outs = cast(dict[str, Any], unwrap(self.prev_iter_outputs))
+        return outs.get(self.output_key)
```
adamnovak: So what this class does is take a Toil promise for a dict, and a key, and when we come back and the Toil promise is fulfilled, we are able to resolve to the value in the dict at that key.

But Toil's core promise system [can already basically do that](https://toil.readthedocs.io/en/latest/autoapi/toil/job/index.html#toil.job.Job.rv). If you pass the key to `.rv()` when getting the Toil promise, you get a Toil promise for the value at that key.

So I think we should drop this class, and anywhere we have a `LoopOutputSource(somejob.rv(), somekey)` (which is just one place---the only usage of this class) we should instead use `JustAValue(somejob.rv(somekey))`. That does the same thing (get you an object where, after the Toil promises are fulfilled, `resolve()` gets that value from the job's returned dict), but it doesn't introduce any new machinery to do it.

src/toil/cwl/cwltoil.py
```diff
@@ -179,6 +179,10 @@ class UnresolvedDict(dict[Any, Any]):
     """Tag to indicate a dict contains promises that must be resolved."""
 
 
+# CWL Jobs have inputs and outputs and sometimes we have to do a process of resolution to figure out
+# what inputs get values from what outputs. Sometimes, collectively we call these inputs and outputs 'ports'.
+
+
```
adamnovak, line 189: OK what's going on is worse than I thought; we have two dueling notions of "promise". [One is `toil.job.Promise`](https://toil.readthedocs.io/en/latest/autoapi/toil/job/index.html#toil.job.Promise), and the other doesn't have a base class but is all the various `resolve()`-able types defined here (`ResolveSource`, `StepValueFrom`, `DefaultWithSource`, and `JustAValue`, to which we're now adding `LoopValueSource` and `LoopStepValueFrom`), plus `UnresolvedDict` and `resolve_dict_w_promises()`.

I think the comment about `a dict mapping to tuple of (key, dict)` is no longer true (it's 11 years old and predates all the `resolve()` stuff), so we should drop it.

I think we want to keep these pieces together as much as possible (maybe by moving `UnresolvedDict` down and quoting any type references to it from higher up, or by moving the other pieces up, or taking all these pieces out into a new `toil.cwl.promises` module or something). And I think we want to expand this comment about resolution to explain that we've defined a new meaning for "promise" in this part of the codebase that differs from what we mean by "promise" in other parts. Ideally we would rename one of these---maybe `Resolveable`? Or  we could somehow actually make them the same concept by making the fancy resolution happen as part of Toil's own promise fulfilment.

As part of this PR I think it only really makes sense to bring the code for the system together and put comments around it so people can tell there's a system here. So this PR should do that since it's making the system even bigger.

src/toil/cwl/cwltoil.py
```diff
@@ -613,6 +617,90 @@ def do_eval(self, inputs: CWLObjectType) -> Any:
...
+        return cwl_utils.expression.do_eval(
+            self.expr,
+            {shortname(k): v for k, v in self.prev_joborder.items()},
+            self.requirements,
+            None,
+            None,
+            {},
+            context=self_value,
+            container_engine=self.container_engine,
+        )
```
adamnovak: [`LoopInput` is defined in terms of `WorkflowStepInput`](https://github.com/common-workflow-language/cwltool/blob/1bf74499ca1c4a5f98e7cffb0ad4aa89aa98cb9e/docs/loop.rst#the-loop-field). And this class is named `LoopStepValueFrom`, which makes it seem like it should be a specialization of `StepValueFrom`.

The net result is *almost* the same flow as in `StepValueFrom` as processed by `resolve_dict_w_promises()`: we don't need to do `eval_prep()` because the loop version can't ask to read files, but we still `resolve()` the source and use the result of that as `context` for `cwl_utils.expression.do_eval()` (except, here, we have the second argument (inputs) set to stuff provided at construction).

But this is a complete independent implementation. We need to justify why that's better than the alternatives:
1. Add a way to feed the CWL expression "inputs" to `StepValueFrom` at construction, and use the existing implementation. Maybe rename it to `ValueFrom` to show it's now more general and covers both steps and loops.
2. Extend `StepValueFrom` with this class, and lock out the parts that can't be used (file loading), and change where the expression "inputs" come from.
3. Make a new base class that implements the simpler functionality used here, and then make `StepValueFrom` extend that with the additional functions of file loading and using the containing dict in `resolve_dict_w_promises()` instead of a pre-provided value as "inputs".
4. Make an abstract base `ValueFrom` class that implements the expression evaluation and defines the interface with `resolve_dict_w_promises()` for things that need to evaluate a `valueFrom`, and then implement a `StepValueFrom` and a `LoopValueFrom` based on that.

I think the simplest approach would be number 1: adding the necessary `inputs` parameter to `StepValueFrom`, letting it override the `do_eval()` argument, and then using that everywhere. That would make the one unified class more complex, but it would better match the spec's idea that this is the same underlying concept held in a slightly different way, and we'd be adding a parameter and a conditional instead of a whole in-these-specific-circumstances class. In loops it's not allowed to use `loadListing`/`loadContents` and needs to be licensed by a different requirement, but that's all enforced at validation time and we don't gain a lot by using an implementation that couldn't do those things if asked.

If you want to do it this way instead, with a new class, you should explain why having a separate class is actually better.

src/toil/cwl/cwltoil.py
```diff
@@ -613,6 +617,90 @@ def do_eval(self, inputs: CWLObjectType) -> Any:
...
+
+    def resolve(self) -> Any:
+        """Extract the named field from the prior iteration's output dict."""
+        outs = cast(dict[str, Any], unwrap(self.prev_iter_outputs))
+        return outs.get(self.output_key)
+
+
+class LoopStepValueFrom:
+    """
+    A `cwltool:Loop` LoopInput carrying a `valueFrom` expression.
```
adamnovak: I guess this does end up being the whole `LoopInput` in a way because it will contain (in `source`) the objects defining the other parts.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        output_keys: list[str],
+    ):
+        """
+        Store the promises needed to extend the accumulator by one iteration.
+
+        :param previous_accumulation: the accumulated outputs from all prior iterations, as a dict mapping output key to list; a plain dict for iteration 0, a Promise thereafter
+        :param iteration_outputs: promise for the current iteration's output dict
+        :param output_keys: short-form names of the output ports to accumulate
+        """
+        super().__init__(cores=1, memory="1GiB", disk="1MiB", local=True)
```
adamnovak: I don't think we want to hardcode disk and memory requirements here. Even though this is copying off of other jobs, one of which it looks like I added.

We should probably make some constants for the internal interpreter job memory and disk and use them everywhere. I'm not sure we shouldn't just be using the Toil defaults. But if we pull them out into constants now it will be easier to change them later if they turn out to be too small or if we decide they need to be user-controllable.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+class CWLLoopAccumulate(Job):
+    """
+    Extend a running cwltool:Loop accumulator with one iteration's outputs.
+
+    Used by `outputMethod: all_iterations` to build the arrays of each output across all iterations.
+    """
+
+    def __init__(
+        self,
+        previous_accumulation: Promised[CWLObjectType] | CWLObjectType,
```
adamnovak: We should type this as `dict[str, list[CWLObjectType]]` or a `Promised[]` that. Then we won't need all the casting any `Any` around it in `run()`, and it will better express what's meant to be in there (a list for each output).

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        :param previous_accumulation: the accumulated outputs from all prior iterations, as a dict mapping output key to list; a plain dict for iteration 0, a Promise thereafter
+        :param iteration_outputs: promise for the current iteration's output dict
+        :param output_keys: short-form names of the output ports to accumulate
+        """
+        super().__init__(cores=1, memory="1GiB", disk="1MiB", local=True)
+        self.previous_accumulation = previous_accumulation
+        self.iteration_outputs = iteration_outputs
+        self.output_keys = output_keys
+
+    def run(self, file_store: AbstractFileStore) -> dict[str, list[Any]]:
```
adamnovak: The result type of `run()` ought to be the non-promised version of the type of `previous_accumulation`, since when we use this we feed the output back in through that input. Then we won't need so many casts.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        """Append the current iteration's outputs to the prior accumulation for each output key."""
+        previous = cast(dict[str, Any], unwrap(self.previous_accumulation))
+        this = cast(dict[str, Any], unwrap(self.iteration_outputs))
+        result: dict[str, list[Any]] = {}
+        for k in self.output_keys:
+            extended = list(previous.get(k, []))
```
adamnovak: We're probably stuck being O(n^2) to run n loop iterations for the accumulate mode, since each iteration ends up copying the accumulated lists around. But I think this copy is not needed. Unless we think we can have the same output key multiple times?

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+    Evaluates `loopWhen` each iteration; if true, spawns the embedded tool
+    as a child and chains a successor CWLLoop as a follow-on. Terminates when
+    `loopWhen` returns false, emitting the last iteration's outputs
+    (outputMethod: last) or the accumulated arrays of each output
+    (outputMethod: all).
+
+    Validation is the responsibility of cwltool's own loop_checker.
+
+    Loops nest naturally because each iteration spawns an independent
+    CWLWorkflow. To combine a loop with scatter, use an intervening
```
adamnovak: The body isn't always a CWLWorkflow object, though. `makeJob()` can return other things. I guess this is talking about each iteration of the outer loop, which can only exist on a workflow?

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        Store context needed to evaluate one iteration of the loop.
+
+        :param step: the WorkflowStep carrying the cwltool:Loop requirement
+        :param cwljob: the input object for this iteration, as an UnresolvedDict or plain dict
+        :param runtime_context: Toil CWL runtime context (filesystem, container engine, etc.)
+        :param parent_name: human-readable name prefix used for child job naming
+        :param iteration_limit: maximum number of iterations before raising an error
+        :param iteration: zero-based index of the current iteration we would execute
+        :param previous_outputs: promise for the previous iteration's output dict; should be None for iteration 0
+        :param previous_accumulation: accumulated outputs from all prior iterations; used for outputMethod: all_iterations; should be None for iteration 0
```
adamnovak: These `:param` definitions can wrap as long as you indent all the subsequent lines for that parameter.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        Each LoopInput record names a step input and describes how its
+        value is updated:
+
+        - `loopSource` pulls from one or more of the just-finished
+          iteration's outputs (with optional `linkMerge` and `pickValue`
+          for merging multiple sources)
+        - `default` supplies a fallback if the source is absent
+        - `valueFrom` runs a CWL expression to compute the final value,
+          with `self` set to the resolved source and `inputs` set to
+          this iteration's joborder
```
adamnovak: Since this gets processed as reStructuredText and not Markdown we probably want double-backticks around the literals.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+            shortname(cast(str, li["id"])): li for li in loop_inputs
+        }
+        step_id_short = shortname(cast(str, self.step.tool["id"]))
+        container_engine = get_container_engine(self.runtime_context)
+
+        result: dict[str, Any] = {}
+        for k, v in cwljob.items():
+            li = loop_by_id.get(k)
+            if li is None:
+                result[k] = JustAValue(v)
```
adamnovak, line 3469: This could use a comment about how we're carrying forward values unchanged here.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                        output_key=shortname(cast(str, raw_sources[0])),
+                    )
+                else:
+                    rs_input: dict[str, Any] = {
+                        "outputSource": [shortname(cast(str, s)) for s in raw_sources],
+                    }
+                    if li.get("linkMerge"):
+                        rs_input["linkMerge"] = li["linkMerge"]
+                    if li.get("pickValue"):
+                        rs_input["pickValue"] = li["pickValue"]
```
adamnovak, line 3491: Why do we need to forge a CWL input object here for `ResolveSource`'s `input` argument instead of giving it the real `li`? Can we get away with not doing that?

If we can't, there should be a comment here explaining why it has to be like this.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                    source = ResolveSource(
+                        name=f"{step_id_short}/{k}",
+                        input=rs_input,
+                        source_key="outputSource",
+                        promises={shortname(cast(str, s)): body_followon for s in raw_sources},
+                    )
+
+            if "default" in li:
+                source = DefaultWithSource(copy.copy(li["default"]), source)
```
adamnovak, line 3501: How do we know we need a shallow copy here? There should be a comment explaining why we can't avoid it.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+
+            if "valueFrom" in li:
+                # TODO: This is a bit of a hack. We should be able to use the same mechanism as CWLJob to resolve valueFrom expressions, but we need to pass in the previous iteration's outputs as well as the current job order. For now, we just call LoopStepValueFrom directly.
```
adamnovak: The part about calling it "directly" doesn't make sense to me; I don't imagine a solution to the problem would somehow involve an *indirect* way of using `LoopStepValueFrom`.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                result[k] = LoopStepValueFrom(
+                    expr=cast(str, li["valueFrom"]),
+                    source=source,
+                    prev_joborder=cwljob,
+                    requirements=self.step.requirements,
+                    container_engine=container_engine,
+                )
+            elif source is not None:
+                result[k] = source
+            else:
```
adamnovak, line 3523: It would be nice if this could somehow share parsing code with https://github.com/DataBiosphere/toil/blame/a491d7b2de2d1e631f9c6c49789fce803f3db24b/src/toil/cwl/cwltoil.py#L3831-L3850, which does a similar process of building up similar objects from similar keys, but that might not be possible.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        follow-on. If false, returns the last iteration's outputs directly
+        (outputMethod: last_iteration) or the accumulated per-port arrays
+        (outputMethod: all_iterations).
+        """
+        cwljob = resolve_dict_w_promises(self.cwljob, file_store)
+
+        if "loop" not in self.step.tool:
+            raise RuntimeError(
+                "CWLLoop scheduled for a step without a cwltool:Loop requirement"
+            )
```
adamnovak, line 3547: Why do we validate the `loop` field and just the `loop` field here, and not e.g. the `when` field we're about to access?

(I'm also confused about how these escaped the extension object and got onto the tool; maybe extensions just work like that.)

Having validated the `loop` field, why do we [later](https://github.com/DataBiosphere/toil/pull/5542/changes#diff-0de692d823906d36705d7d2c02a0bc99e191abc8b48f258401f9611184f339a2R3681) act as if it might not be there by using `get()`?

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+
+        when_expression = cast(str, self.step.tool["when"])
+        output_method = cast(str, self.step.tool.get("outputMethod", "last_iteration"))
```
adamnovak, line 3551: Why aren't we pulling out `loop` right now with the other extension fields?

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+        )
+        if not isinstance(when_val, bool):
+            raise cwl_utils.errors.WorkflowException(
+                f"'loopWhen' expression {when_expression!r} did not evaluate "
+                "to a boolean"
+            )
+
+        if when_val is False:
+            if self.iteration == 0:
+                empty: Any = [] if output_method == "all_iterations" else None
```
adamnovak: This might be better types as `list[Any] | None`.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                output_keys = [shortname(cast(str, o["id"])) for o in self.step.tool["outputs"]]
+                return {k: empty for k in output_keys}
```
adamnovak: Is it going to pose a problem that these can all point to the *same* empty list? I might put the conditional expression inline here. Then we wouldn't need to type it anyway.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+            if output_method == "all_iterations":
+                return cast(dict[str, Any], unwrap(cast(Promised[CWLObjectType], self.previous_accumulation)))
```
adamnovak: I think we need some more comments around here to explain how we're handling the false-on-first-loop case above, and so by the time we get here we know previous_accumulation isn't None, and so the cast is allowed.

It might actually be better to have an `assert self.previous_accumulation is not None`, instead of a cast. MyPy will look at that and narrow the type to exclude the None case, so you'll then be allowed to call `unwrap()` without a cast.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+            return cast(dict[str, Any], unwrap(cast(Promised[CWLObjectType], self.previous_outputs)))
```
adamnovak: Similarly here, it might be good to use a comment and an assert instead of a cast.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                    requirements=self.step.requirements,
+                    container_engine=container_engine,
+                )
+            elif source is not None:
+                result[k] = source
+            else:
+                result[k] = JustAValue(v)
+        return UnresolvedDict(result)
+
+    def run(self, file_store: AbstractFileStore) -> dict[str, Any] | Promised[CWLObjectType]:
```
adamnovak: This should be typed as `CWLObjectType | Promised[CWLObjectType]`. Then we wouldn't need all the casting from `CWLObjectType` unwrap results to `dict[str, Any]`.

src/toil/cwl/cwltoil.py
```diff
@@ -3341,6 +3429,273 @@ def sn(n: Mapping[str, Any] | str) -> str:
...
+                previous_accumulation=prev_acc_arg,
+                iteration_outputs=body_followon.rv(),
+                output_keys=output_keys,
+            )
+            body_followon.addFollowOn(accumulated_loops)
+            next_loop_pred = accumulated_loops
+            next_accumulation: Promised[CWLObjectType] | None = accumulated_loops.rv()
+        elif output_method == "last_iteration":
+            next_loop_pred = body_followon
+            next_accumulation = None
```
adamnovak: It might make more sense to spread the comments over the conditional cases here. The point of the first case is to set up this extra job that we need, and so a comment at the top about how we need to do that would make it clearer that all this stuff about `output_keys` is in service of that. And the second case could explain that in last_iteration mode we don't need any of that and we just need to run the next loop right after the body. And we could also explain `next_accumulation`.

src/toil/cwl/cwltoil.py
adamnovak: If scatter is no longer the first case where these variables are used, it doesn't want to have these long type lists anymore.

adamnovak (changes requested):
This looks pretty good, but I think it could be simplified.

We should be able to pull most of the test files from the packaged data files of our `cwltool` dependency.

And I think we can get away with not adding either `LoopOutputSource` (because Toil promises can already do a dict key lookup) or `LoopStepValueFrom` (because `StepValueFrom` can be generalized to cover both cases by adding an optional parameter and a conditional).

adamnovak (approved):
This looks good to me!

PR #5560, opened 2026-08-03
Add --runDir option; make --workDir/--coordinationDir create-if-missing (#5516)

annagiroti:
## Changelog Entry
Added `--runDir`/`TOIL_RUN_DIR` to derive the job store, work dir, coordination dir, and CWL/WDL image cache locations under one directory. `--workDir`/`--coordinationDir` are now created automatically if missing instead of raising, and Toil logs the resolved run paths at startup.

Known follow-up: the job store path derived from `--runDir` is fixed, so concurrent workflows sharing one `--runDir` will collide. Flagged with TODOs pending a fix.

Addresses #5516 

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -354,7 +354,6 @@ class Config:
 
         # Core options
         set_option("jobStore")
-        # TODO: LOG LEVEL STRING
         set_option("workDir")
         set_option("coordination_dir")
         set_option("runDir")
@@ -1760,8 +1759,9 @@ class Toil(ContextManager["Toil"]):
         Return a path to a writable directory under which per-workflow directories exist.
 
         This directory is always required to exist on a machine, even if the Toil
-        worker has not run yet.  If your workers and leader have different temp
-        directories, you may need to set TOIL_WORKDIR.
+        worker has not run yet, and is created if it does not already exist. If
+        your workers and leader have different temp directories, you may need to
+        set TOIL_WORKDIR.
 
         :param configWorkDir: Value passed to the program using the --workDir flag
         :return: Path to the Toil work directory, constant across all machines
@@ -1773,9 +1773,13 @@ class Toil(ContextManager["Toil"]):
             or tempfile.gettempdir()
         )
         if not os.path.exists(workDir):
-            raise RuntimeError(
-                f"The directory specified by --workDir or TOIL_WORKDIR ({workDir}) does not exist."
-            )
+            try:
+                os.makedirs(workDir, exist_ok=True)
+            except OSError as e:
+                raise RuntimeError(
+                    f"The directory specified by --workDir or TOIL_WORKDIR "
+                    f"({workDir}) does not exist and could not be created: {e}"
+                )
         return workDir
 
     @classmethod
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -396,13 +396,7 @@ class Config:
 
         if self.runDir is not None:
             # A single --runDir was given. Derive defaults for anything the
-            # user didn't set explicitly; explicit flags always win. The
-            # jobStore-derivation branch inside this is only reachable for
-            # direct callers of the flag-based parser (jobstore_as_flag=True)
-            # that leave --jobStore unset. The CWL/WDL runners fill in their
-            # own jobStore default before calling setOptions, and the plain
-            # `toil` entry point requires jobStore as a positional argument,
-            # so neither ever reaches it in practice.
+            # user didn't set explicitly; explicit flags always win.
             self.runDir, self.jobStore, self.workDir, self.coordination_dir = (
                 derive_run_dir_defaults(
                     self.runDir, self.jobStore, self.workDir, self.coordination_dir
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -83,7 +83,7 @@ from toil.lib.misc import StrPath
 from toil.lib.retry import retry
 from toil.lib.threading import ensure_filesystem_lockable
 from toil.lib.url import URLAccess
-from toil.options.common import JOBSTORE_HELP, add_base_toil_options
+from toil.options.common import JOBSTORE_HELP, add_base_toil_options, parse_jobstore
 from toil.options.cwl import add_cwl_options
 from toil.options.runner import add_runner_options
 from toil.options.wdl import add_wdl_options
@@ -121,6 +121,41 @@ def get_default_config_path() -> str:
     The file at the path will not necessarily exist.
     """
     return os.path.join(get_toil_home(), "default.yaml")
+
+
+def derive_run_dir_defaults(
+    run_dir: str,
+    job_store: str | None,
+    work_dir: str | None,
+    coordination_dir: str | None,
+) -> tuple[str, str, str, str]:
+    """
+    Given --runDir and the current values of --jobStore, --workDir, and
+    --coordinationDir, fill in defaults for any of them left unset,
+    derived from --runDir. Explicit values always win.
+
+    Creates the derived work dir and coordination dir; the job store
+    creates itself.
+
+    :return: (run_dir, job_store, work_dir, coordination_dir), with
+             run_dir made absolute and the other three either the
+             original explicit value or the runDir-derived default.
+    """
+    run_dir = os.path.abspath(run_dir)
+    if job_store is None:
+        # TODO: This path is fixed (<runDir>/jobstore), so multiple
+        # concurrent workflows sharing one --runDir will collide trying
+        # to create the same job store. Needs a unique suffix per
+        # invocation, like the create_tmp_dir/mkdtemp fallback used when
+        # --runDir isn't set.
+        job_store = parse_jobstore(os.path.join(run_dir, "jobstore"))
+    if work_dir is None:
+        work_dir = os.path.join(run_dir, "work")
+        os.makedirs(work_dir, exist_ok=True)
+    if coordination_dir is None:
+        coordination_dir = os.path.join(run_dir, "coordination")
+        os.makedirs(coordination_dir, exist_ok=True)
+    return run_dir, job_store, work_dir, coordination_dir
 
 
 class Config:
@@ -360,31 +395,18 @@ class Config:
 
         if self.runDir is not None:
             # A single --runDir was given. Derive defaults for anything the
-            # user didn't set explicitly; explicit flags always win.
-            self.runDir = os.path.abspath(self.runDir)
-            if self.workDir is None:
-                self.workDir = os.path.join(self.runDir, "work")
-                os.makedirs(self.workDir, exist_ok=True)
-            if self.coordination_dir is None:
-                self.coordination_dir = os.path.join(self.runDir, "coordination")
-                os.makedirs(self.coordination_dir, exist_ok=True)
-            if self.jobStore is None:
-                # Only reachable for direct callers of the flag-based parser
-                # (jobstore_as_flag=True) that leave --jobStore unset. The
-                # CWL/WDL runners fill in their own jobStore default before
-                # calling setOptions, and the plain `toil` entry point
-                # requires jobStore as a positional argument, so neither
-                # ever reaches this branch in practice.
-                # TODO: This path is fixed (<runDir>/jobstore), so multiple
-                # concurrent workflows sharing one --runDir will collide
-                # trying to create the same job store. Needs a unique
-                # suffix per invocation, like the create_tmp_dir/mkdtemp
-                # fallback used when --runDir isn't set.
-                from toil.options.common import parse_jobstore
-
-                self.jobStore = parse_jobstore(
-                    os.path.join(self.runDir, "jobstore")
+            # user didn't set explicitly; explicit flags always win. The
+            # jobStore-derivation branch inside this is only reachable for
+            # direct callers of the flag-based parser (jobstore_as_flag=True)
+            # that leave --jobStore unset. The CWL/WDL runners fill in their
+            # own jobStore default before calling setOptions, and the plain
+            # `toil` entry point requires jobStore as a positional argument,
+            # so neither ever reaches it in practice.
+            self.runDir, self.jobStore, self.workDir, self.coordination_dir = (
+                derive_run_dir_defaults(
+                    self.runDir, self.jobStore, self.workDir, self.coordination_dir
                 )
+            )
 
         set_option("noStdOutErr")
         set_option("stats")
--- a/src/toil/common.py
+++ b/src/toil/common.py
@@ -369,9 +369,12 @@ class Config:
                 self.coordination_dir = os.path.join(self.runDir, "coordination")
                 os.makedirs(self.coordination_dir, exist_ok=True)
             if self.jobStore is None:
-                # Only reachable when --jobStore is optional (the CWL/WDL
-                # runners); the plain `toil` entry point requires jobStore
-                # as a positional argument, so it can never be None here.
+                # Only reachable for direct callers of the flag-based parser
+                # (jobstore_as_flag=True) that leave --jobStore unset. The
+                # CWL/WDL runners fill in their own jobStore default before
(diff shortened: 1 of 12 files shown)


src/toil/leader.py
```diff
@@ -262,6 +266,22 @@ def run(self) -> Any:
 
         :return: The return value of the root job's run function.
         """
+
+        # Toil.getToilWorkDir and get_local_workflow_coordination_dir already
+        # create these directories during Toil.__enter__, before the Leader
+        # exists. These calls are a defensive backstop for any code path that
+        # reaches Leader.run() without having gone through Toil.__enter__ first.
+        ensure_dir_exists(self.config.workDir, "--workDir")
+        ensure_dir_exists(self.config.coordination_dir, "--coordinationDir")
```
annagiroti: These should also be included in worker.py

src/toil/cwl/cwltoil.py
```diff
@@ -5029,6 +4748,28 @@ def main(args: list[str] | None = None, stdout: TextIO = sys.stdout) -> int:
...
+            # concurrent workflows sharing one --runDir will collide
+            # trying to create the same job store. Needs a unique suffix
+            # per invocation, like the create_tmp_dir fallback below used
+            # when --runDir isn't set.
+            options.jobStore = "file:" + os.path.join(options.runDir, "jobstore")
+        if options.workDir is None:
+            options.workDir = os.path.join(options.runDir, "work")
+            os.makedirs(options.workDir, exist_ok=True)
+        if options.cachedir is None:
+            options.cachedir = os.path.join(options.runDir, "image-cache")
```
annagiroti, line 4774: Is it necessary to have the code in common.py repeated here for cwltoil.py? If it is, we need to create a function that can be called in cwltoil.py and wdltoil.py instead.

src/toil/wdl/wdltoil.py
```diff
@@ -5973,6 +5985,24 @@ def main() -> None:
...
+        if options.jobStore is None:
+            # TODO: This path is fixed (<runDir>/jobstore), so multiple
+            # concurrent workflows sharing one --runDir will collide
+            # trying to create the same job store. Needs a unique suffix
+            # per invocation, like the mkdtemp fallback below used when
+            # --runDir isn't set.
+            options.jobStore = "file:" + os.path.join(options.runDir, "jobstore")
+        if options.workDir is None:
+            options.workDir = os.path.join(options.runDir, "work")
+            os.makedirs(options.workDir, exist_ok=True)
```
annagiroti: Same issue found in cwltoil.py. Is it necessary to have the code in common.py repeated here for wdltoil.py? If it is, we need to create a function that can be called in cwltoil.py and wdltoil.py instead.

src/toil/test/options/options.py
```diff
@@ -48,3 +50,107 @@ def test_caching_option_priority(self):
...
+        parser = ArgParser()
+        addOptions(parser, jobstore_as_flag=True, wdl=False, cwl=False)
+        job_store = self._getTestJobStorePath()
+        options = parser.parse_args([f"--jobstore=file:{job_store}"])
+        with self.assertLogs("toil.common", level="INFO") as cm:
+            with Toil(options):
+                pass
+        self.assertTrue(
+            any("Resolved Toil run paths" in message for message in cm.output)
+        )
```
annagiroti, line 102: It might be a good idea to update these tests for future code. Make these tests share code to prevent confusion for future coders.

annagiroti: Other suggestion: simplify the tests by making them a test for --rundir function.

src/toil/common.py
```diff
@@ -353,9 +354,37 @@ def set_option(option_name: str, old_names: list[str] | None = None) -> None:
...
+                # (jobstore_as_flag=True) that leave --jobStore unset. The
+                # CWL/WDL runners fill in their own jobStore default before
+                # calling setOptions, and the plain `toil` entry point
+                # requires jobStore as a positional argument, so neither
+                # ever reaches this branch in practice.
+                # TODO: This path is fixed (<runDir>/jobstore), so multiple
+                # concurrent workflows sharing one --runDir will collide
+                # trying to create the same job store. Needs a unique
+                # suffix per invocation, like the create_tmp_dir/mkdtemp
+                # fallback used when --runDir isn't set.
```
annagiroti: The better approach might be utilizing UUIDs for the jobStore directories.

src/toil/test/cwl/cwlTest.py
```diff
@@ -250,6 +250,31 @@ def test_cwl_cmdline_input(self) -> None:
             # If the workflow runs, it must have had options
             cwltoil.main(args, stdout=st)
 
+    def test_cwl_run_dir(self, tmp_path: Path) -> None:
+        """
+        Test that --runDir derives the CWL image cache location. The job
+        store/work dir derivation is covered by
+        commonTests.TestDeriveRunDirDefaults; cachedir is CWL-specific and
+        isn't.
```
adamnovak: Usually you want a docstring to have a one-line summary, then a blank line, and then any further description.

The further description right now is mostly about what *isn't* here and why it shouldn't be here. I don't think that the reader is at risk of thinking that those things should have been here, so trying to disabuse them of that notion seems more likely to confuse them than to help, because they're going to read this and then be confused as to why anyone would think they would have thought that. I would just cut everything that isn't about this test itself.

src/toil/test/cwl/cwlTest.py
```diff
@@ -250,6 +250,31 @@ def test_cwl_cmdline_input(self) -> None:
...
+        with get_data("test/cwl/conditional_wf.cwl") as cwlfile:
+            args = [
+                f"--runDir={run_dir}",
+                "--outdir",
+                str(tmp_path / "out"),
+                str(cwlfile),
+                "--message",
+                "str",
+                "--sleep",
+                "2",
```
adamnovak: This looks like it might wait for 2 seconds that we don't actually need to wait for.

src/toil/leader.py
adamnovak: But `ensure_dir_exists()` is allowed to just exit the whole process if it e.g. doesn't have permission, right? Is someone's Python-based workflow going to be expecting that behavior from inside the `Toil` object, rather than an exception?

I might let `ensure_dir_exists()` propagate its various permission errors upward.

On the other hand, having it handle the exit makes the user-facing behavior in what's now the "normal" case of running a workflow nicer (because there's no stack trace), without us needing to put catches everywhere.

src/toil/leader.py
```diff
@@ -262,6 +266,22 @@ def run(self) -> Any:
 
         :return: The return value of the root job's run function.
         """
+
+        # Toil.getToilWorkDir and get_local_workflow_coordination_dir already
+        # create these directories during Toil.__enter__, before the Leader
+        # exists. These calls are a defensive backstop for any code path that
+        # reaches Leader.run() without having gone through Toil.__enter__ first.
```
adamnovak: Do we think those codepaths are possible or legal? Outside Toil's own testing where we might independently instantiate a Leader?

I think we're not actually allowed to run a `Leader` for a workflow when not inside the `Toil` object context manager, so the thing we're trying to handle isn't possible and we shouldn't handle it (unless it's to complain that it has happened and something has gone terribly wrong).

src/toil/leader.py
```diff
@@ -1655,6 +1675,32 @@ def process_finished_job_description(
...
+                        logger.warning(
+                            "No log file is present, despite job failing: %s. "
+                            "Toil does not retain worker logs by default; rerun with "
+                            "--writeLogs=PATH or --writeLogsGzip=PATH to save failed "
+                            "jobs' logs to disk. Toil looked for the batch system's "
+                            "own logs (see --batchLogsDir) but found none; check the "
+                            "batch system's own tools or logs directly if you are "
+                            "running on a grid engine.",
+                            replacement_job,
+                        )
```
adamnovak: These two similar messages should probably be composed by pasting together shared and variable strings.

Also, if a job fails, and there's a log from the Slurm task but we don't have a log sent back from the Toil worker itself, then something has gone wrong. "Normal" failed jobs due to user error still fail in a way that the worker is able to report in. We still want to log *something* to alert the user that the worker failed to report in like it was supposed to, even if we have some logging about it. We might want to suggest that the user read the logging we do have.

src/toil/common.py
```diff
@@ -353,9 +390,24 @@ def set_option(option_name: str, old_names: list[str] | None = None) -> None:
...
+            # direct callers of the flag-based parser (jobstore_as_flag=True)
+            # that leave --jobStore unset. The CWL/WDL runners fill in their
+            # own jobStore default before calling setOptions, and the plain
+            # `toil` entry point requires jobStore as a positional argument,
+            # so neither ever reaches it in practice.
+            self.runDir, self.jobStore, self.workDir, self.coordination_dir = (
+                derive_run_dir_defaults(
+                    self.runDir, self.jobStore, self.workDir, self.coordination_dir
+                )
+            )
```
adamnovak, line 404: Maybe this really wants to be a method that just gets to use `self`?

adamnovak (changes requested):

src/toil/common.py
annagiroti: It's a free function so cwltoil.py/wdltoil.py can call it too; they don't have a Config object at that point, just an argparse Namespace. Making it a method would mean duplicating the logic there again.

src/toil/common.py
```diff
@@ -123,6 +123,42 @@ def get_default_config_path() -> str:
...
+    work_dir: str | None,
+    coordination_dir: str | None,
+) -> tuple[str, str, str, str]:
+    """
+    Given --runDir and the current values of --jobStore, --workDir, and
+    --coordinationDir, fill in defaults for any of them left unset,
+    derived from --runDir. Explicit values always win.
+
+    Creates the derived work dir and coordination dir; the job store
+    creates itself.
```
adamnovak: This doesn't clearly communicate that `work_dir` and `coordination_dir` will be created, if they didn't exist, *only* when they are defaulted from `run_dir`, and not when they are passed in. I think it *does* mean that (it talks about "derived" directories), but that's weird so if it does that it should mean that harder and more obviously.

Right now the tests test that behavior, but it's not clear to me why that's the right design. The caller is going to need to make sure the directories exist before using them whether they are determined by this function or passed through it, so this design means there need to be at least two places where we worry about directory creation.

If we made this function either always or never create the directories, we'd need to worry about it in fewer places.

src/toil/leader.py
```diff
@@ -262,6 +265,16 @@ def run(self) -> Any:
...
         """
+
+        if isinstance(self.batchSystem, AbstractGridEngineBatchSystem):
+            # The batch system isn't available yet when Toil logs the other
+            # resolved run paths (see Toil._log_resolved_paths), so log this
+            # one here instead, now that it exists. Only grid batch systems
+            # actually write their own logs to this directory.
+            logger.info(
+                "Resolved batch logs dir: %s", self.batchSystem.get_batch_logs_dir()
+            )
```
adamnovak, line 276: We could leave it like this, but it would be nicer if, if only some batch systems use a batch logs directory, we exposed that in the batch system API itself (maybe by making `get_batch_logs_dir()` return None when not used?), instead of needing to know here what implementations actually use it.

src/toil/leader.py
adamnovak: This comment has, independent of the current changes, gone out of date, because the logs can be in `--batchLogsDir` instead now.

src/toil/leader.py
```diff
@@ -1655,6 +1668,37 @@ def process_finished_job_description(
...
+                            "batch system for this job; check the batch "
+                            "system's own tools or logs directly."
+                        )
+                    else:
+                        variable_message = (
+                            "Toil looked for the batch system's own logs "
+                            "(see --batchLogsDir) but found none; check the "
+                            "batch system's own tools or logs directly if "
+                            "you are running on a grid engine."
+                        )
```
adamnovak: Neither of these is really quite applicable for when we're using a batch system like `mesos` (or I think even `single_machine`) which we know in advance don't keep batch logs for us using this system.

When Toil thinks the batch system should have sent batch logs, and it can't find them, it should complain. It can mention `--batchLogsDir`, and maybe suggest making sure it's on shared storage.

When Toil thinks the batch system shouldn't send batch logs for this job (maybe because `batch_system_id` is None, which I think could mean we're looking at a batch system that doesn't use the concept or a local job that didn't go to the cluster), it shouldn't complain if they aren't there.

Encouraging the user to ask the batch system themselves can be useful, because e.g. Mesos will have logs for them somewhere. It won't apply to every batch system, though; I don't think `single_machine` has anything. We could just deal with that, or we could try to figure out when the user has something to find. If we build a system to figure out when it's applicable we might as well make that system let the batch system explain to the user where they need to look ("check the Mesos web UI", etc.).

src/toil/leader.py
```diff
@@ -1655,6 +1668,37 @@ def process_finished_job_description(
...
+                ):
+                    # Alert the user that the worker failed to report in
+                    # like it was supposed to, and say specifically whether
+                    # Toil looked for batch system logs and found nothing,
+                    # or wasn't able to look in the first place.
+                    shared_message = (
+                        "No log file is present, despite job failing: %s. "
+                        "Toil does not retain worker logs by default; rerun "
+                        "with --writeLogs=PATH or --writeLogsGzip=PATH to "
+                        "save failed jobs' logs to disk. "
```
adamnovak: I don't understand how `--writeLogs` is expected to help here. The log we didn't find for the failing job is the one that would get shipped back to the leader for writing to the `--writeLogs` directory. Adding the flag would only get a log to appear here in very weird circumstances involving worker failure after saving logs for a job it thinks has succeeded, and that log would not really help a lot for failures that happen after it is written.

The message here makes it sound like in general failed job logs are not available when `--writeLogs` isn't passed, but that's not true.

src/toil/common.py
```diff
@@ -353,9 +390,24 @@ def set_option(option_name: str, old_names: list[str] | None = None) -> None:
...
+
+        if self.runDir is not None:
+            # A single --runDir was given. Derive defaults for anything the
+            # user didn't set explicitly; explicit flags always win. The
+            # jobStore-derivation branch inside this is only reachable for
+            # direct callers of the flag-based parser (jobstore_as_flag=True)
+            # that leave --jobStore unset. The CWL/WDL runners fill in their
+            # own jobStore default before calling setOptions, and the plain
+            # `toil` entry point requires jobStore as a positional argument,
+            # so neither ever reaches it in practice.
```
adamnovak: I can't tell what aspect of the code here this long description of interactions between other pieces of code is meant to justify. If we need to explain these interactions, I don't think here's the right place.
```suggestion
            # user didn't set explicitly; explicit flags always win.
```

src/toil/common.py
```diff
@@ -1154,11 +1206,32 @@ def __enter__(self) -> "Toil":
...
         signal.signal(signal.SIGTERM, lambda *_: sys.exit(1))
 
         return self
 
+    def _log_resolved_paths(self, config: "Config") -> None:
+        """
+        Log the resolved job store, work dir, and coordination dir paths
+        at INFO, once, so users can find them without tracing flag
+        precedence themselves. Skips the batch logs dir; the batch system
+        isn't built yet here, and grid systems log their own dir on startup.
```
adamnovak: Do grid engine batch systems really log their own batch logs directory on startup? If so, why are we *also* logging it on their behalf in `Leader.run()`?

adamnovak (changes requested):
I think the missing-log-file messages still aren't quite right.

The behavior where `derive_run_dir_defaults()` sometimes promises that the directories it returns exist, and sometimes doesn't, seems weird and under-justified; is there a good reason for this I am missing? Is it just to make sure that a failed directory creation can't print a wrong error message?

I found a few other things I don't like about the comments, that I probably should have found the first time.

PR #5566, opened 2026-08-16
docs: give the WDL quickstart example a version and test it

### LINKED ISSUE #5534: WDL quickstart example in the documentation lacks a `version`, so it won't work (and isn't under test)
In our documentation we have a WDL quickstart with an example WDL, but the example WDL doesn't specify a version, and so Toil won't be willing to run it.

It should be under test so this sort of thing can't happen, and it should be fixed.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1843)
┆Issue Number: TOIL-1843

vshulcz:
## Changelog Entry

 * The WDL quickstart example now specifies a WDL version and is covered by a test.

## Description

Fixes #5534.

The WDL in the quickstart has no `version`, so it is parsed as draft-2 and `toil-wdl-runner` fails on it:

```
(Ln 1, Col 1) MissingVersion, document should declare WDL version; draft-2 assumed
WDL.Error.EvalError: Cannot evaluate no expression for message
```

The inputs JSON also addresses the task input directly (`write_simple_file.write_file.message`), which does not bind.

This ports the example to WDL 1.0 (`input` sections, `~{}` placeholder, the call passing `message` through), so the inputs file uses `write_simple_file.message`.

To keep the docs and a tested file from drifting apart, the example now lives in `src/toil/test/docs/scripts/` next to the other documentation examples and is pulled into `quickStart.rst` with `literalinclude`. `MANIFEST.in` already ships that directory. `test_documentation_quickstart` in `wdltoil_test.py` runs it the same way `test_MD5sum` runs its workflow.

Verified locally on an Apple Silicon macOS host with Docker:

```
$ toil-wdl-runner wdl-helloworld.wdl wdl-helloworld.json -o out --logDebug --retryCount=0
Workflow stopped. Success: True
{"write_simple_file.write_file.test": ".../out/write_simple_file.write_file/wdl-helloworld-output.txt"}
$ cat out/write_simple_file.write_file/wdl-helloworld-output.txt
Hello world!

$ pytest src/toil/test/wdl/wdltoil_test.py -k test_documentation_quickstart -x -q
1 passed, 35 deselected
```

I left `make format` alone: `black --check` already reports `wdltoil_test.py` as unformatted on unmodified `master` here, so running it would bury this change in unrelated reformatting.

### DIFF
--- a/docs/gettingStarted/quickStart.rst
+++ b/docs/gettingStarted/quickStart.rst
@@ -91,22 +91,13 @@ Running WDL workflows using Toil is still in alpha, and currently experimental. 
 
    This installs the ``toil-wdl-runner`` executable.
 
-#. Copy and paste the following code block into ``wdl-helloworld.wdl``::
+#. Copy and paste the following code block into ``wdl-helloworld.wdl``:
 
-        workflow write_simple_file {
-          call write_file
-        }
-        task write_file {
-          String message
-          command { echo ${message} > wdl-helloworld-output.txt }
-          output { File test = "wdl-helloworld-output.txt" }
-        }
+   .. literalinclude:: ../../src/toil/test/docs/scripts/wdl-helloworld.wdl
 
-   and this code into ``wdl-helloworld.json``::
+   and this code into ``wdl-helloworld.json``:
 
-        {
-          "write_simple_file.write_file.message": "Hello world!"
-        }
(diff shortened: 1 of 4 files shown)


adamnovak (approved):
This looks OK to me.

PR #5569, opened 2026-08-20
Evaluate CWL `when` conditionals before scheduling to avoid wasted batch-system submissions (#3990)

### LINKED ISSUE #3990: CWL Conditional steps that aren't running still get scheduled to a job scheduler.

annagiroti:
Steps with a `when` conditional were always scheduled onto the real batch system as a fully-resourced job, even when the condition evaluated false and the step would be skipped. On a busy HPC system, a skipped step that requests large resources can sit queued for hours before discovering it had nothing to do.

`CWLJobWrapper` already solves this for dynamic resource requirements: it runs as a cheap local job that resolves inputs before spawning the real job. This routes `when`-conditional steps through the same wrapper, so the condition is checked with fully-resolved inputs — including promises from upstream step outputs — before the real job is ever created. Skipped steps never touch the batch system.

Resolves #3990 

## Changelog Entry

 * Toil now checks CWL `when` conditionals on the leader before scheduling, so steps that evaluate false are never submitted to the real batch system.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

### DIFF
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -403,9 +403,9 @@ class Conditional:
         self.requirements = requirements or []
         self.container_engine = container_engine
 
-    def is_conditional(self) -> bool:
-        """Whether this has an expression to evaluate."""
-        return self.expression is not None
+    def is_empty(self) -> bool:
+        """Whether this is an empty Conditional."""
+        return self.expression is None
 
     def is_false(self, job: CWLObjectType) -> bool:
         """
@@ -3523,7 +3523,7 @@ def makeJob(
                         has_dynamic_resource_requirement = True
 
         if has_dynamic_resource_requirement or (
-            conditional is not None and conditional.is_conditional()
+            conditional is not None and not conditional.is_empty()
         ):
             # Resource requirements and the `when` conditional can depend on
             # promises from upstream steps that only resolve once the job
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -403,6 +403,10 @@ class Conditional:
         self.requirements = requirements or []
         self.container_engine = container_engine
 
+    def is_conditional(self) -> bool:
+        """Whether this has an expression to evaluate."""
+        return self.expression is not None
+
     def is_false(self, job: CWLObjectType) -> bool:
         """
         Determine if expression evaluates to False given completed step inputs.
@@ -2951,11 +2955,9 @@ class CWLJob(CWLNamedJob):
         cwljob: CWLObjectType,
         runtime_context: cwltool.context.RuntimeContext,
         parent_name: str | None = None,
-        conditional: Conditional | None = None,
     ):
         """Store the context for later execution."""
         self.cwltool = tool
-        self.conditional = conditional or Conditional()
 
         if runtime_context.builder:
             self.builder = runtime_context.builder
@@ -3159,9 +3161,6 @@ class CWLJob(CWLNamedJob):
 
         # Deletes duplicate listings
         remove_redundant_mounts(cwljob)
-
-        if self.conditional.is_false(cwljob):
-            return self.conditional.skipped_outputs()
 
         fill_in_defaults(
             self.step_inputs, cwljob, self.runtime_context.make_fs_access("")
@@ -3524,7 +3523,7 @@ def makeJob(
                         has_dynamic_resource_requirement = True
 
         if has_dynamic_resource_requirement or (
-            conditional is not None and conditional.expression is not None
+            conditional is not None and conditional.is_conditional()
         ):
             # Resource requirements and the `when` conditional can depend on
             # promises from upstream steps that only resolve once the job
@@ -3544,7 +3543,6 @@ def makeJob(
             jobobj,
             runtime_context,
             parent_name=parent_name,
-            conditional=conditional,
         )
         return job, job
 
--- a/src/toil/cwl/cwltoil.py
+++ b/src/toil/cwl/cwltoil.py
@@ -2890,13 +2890,11 @@ class ResolveIndirect(CWLNamedJob):
 
 class CWLJobWrapper(CWLNamedJob):
     """
-    Wrap a CWL job that uses a dynamic resources requirement, or that may be
-    skipped by a `when` conditional that can't be safely evaluated until the
-    step's inputs are resolved.
+    Determines how and whether to run the wrapped CWL job.
 
-    When executed, this runs on the leader with minimal resources, resolves
-    the job's inputs, and then either reports the step as skipped or creates
-    a new child job which has the correct resource requirement set.
+    Wraps a CWL job that uses a dynamic resource requirement or has a
+    conditional. This job is responsible for creating a CWLJob child with
+    the right resource requirements, when the job should not be skipped.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


adamnovak:
This closes #5507 by replacing it, I think.

adamnovak:
@annagiroti Can you fill in the changelog entry (`PR submitter writes their recommendation for a changelog entry here` above)?

src/toil/test/cwl/conditional_step_depends_on_step.cwl
```diff
@@ -0,0 +1,37 @@
...
+class: Workflow
+requirements:
+  InlineJavascriptRequirement: {}
+inputs:
+  sleep: int
+outputs: []
+steps:
+  produce:
+    in:
+      sleep: sleep
```
adamnovak: We should rename this input to something other than `sleep` (maybe `number`?), because other test workflows will have `sleep` inputs that make them actually wait, and it's weird that this one doesn't.

src/toil/cwl/cwltoil.py
```diff
@@ -2890,10 +2890,13 @@ def run(self, file_store: AbstractFileStore) -> CWLObjectType:
 
 class CWLJobWrapper(CWLNamedJob):
     """
-    Wrap a CWL job that uses dynamic resources requirement.
+    Wrap a CWL job that uses a dynamic resources requirement, or that may be
```
adamnovak: Ideally we want a 1-line summary at the top of our docstring. Maybe say that this job determines how and whether to run the wrapped job?

src/toil/cwl/cwltoil.py
```diff
@@ -2890,10 +2890,13 @@ def run(self, file_store: AbstractFileStore) -> CWLObjectType:
...
+    skipped by a `when` conditional that can't be safely evaluated until the
```
adamnovak: Saying it that way makes it sound like there are some conditionals that can be safely evaluated without the step's inputs, and others that can't, and our code actually knows the difference and acts on it to evaluate only the conditionals that need the inputs with this mechanism.

That's not true, so we can probably just say "a conditional" instead of the long mis-explanatory phrase.

src/toil/cwl/cwltoil.py
```diff
@@ -2890,10 +2890,13 @@ def run(self, file_store: AbstractFileStore) -> CWLObjectType:
...
+    step's inputs are resolved.
 
-    When executed, this creates a new child job which has the correct resource
-    requirement set.
+    When executed, this runs on the leader with minimal resources, resolves
+    the job's inputs, and then either reports the step as skipped or creates
+    a new child job which has the correct resource requirement set.
```
adamnovak: This whole piece should probably be cut, because it's a summary rather than an explanation.

If we do want to write down the responsibility, we should say something like "This job is responsible for creating a CWLJob child with the right resource requirements, when the job should not be skipped."

src/toil/cwl/cwltoil.py
adamnovak: We're still passing the conditional to `CWLJob`, and `CWLJob` hasn't had any conditional-checking code removed from it. If `CWLJobWrapper` is now responsible for evaluating all the conditionals on jobs, `CWLJob` should probably forget how to do it itself.

src/toil/cwl/cwltoil.py
```diff
@@ -3507,29 +3510,37 @@ def makeJob(
...
-                            runtime_context,
-                            parent_name=parent_name,
-                            conditional=conditional,
-                        )
-                        return job_wrapper, job_wrapper
-        # Otherwise, all requirements are known now.
+        if has_dynamic_resource_requirement or (
+            conditional is not None and conditional.expression is not None
+        ):
+            # Resource requirements and the `when` conditional can depend on
```
adamnovak, line 3528: It probably makes more sense to talk about conditionals in the code's terms, where we have a `Conditional` type, rather than in CWL language terms where we have a `when` field in various places in YAML.

adamnovak (changes requested):
This looks like it should work OK, but:

* The test has a `sleep` input, presumably copied from a different test and not changed, which does not control a sleep
* The docstrings are following the Way of Claude rather than [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings), and in particular are not leveraging a theory of mind to make the reader reach true conclusions about the code.

src/toil/cwl/cwltoil.py
```diff
@@ -403,6 +403,10 @@ def __init__(
         self.requirements = requirements or []
         self.container_engine = container_engine
 
+    def is_conditional(self) -> bool:
+        """Whether this has an expression to evaluate."""
+        return self.expression is not None
```
adamnovak: I'm now realizing `Conditional` is our own type, not the type `cwltool` uses to represent the value in one of its `when` fields.

Although we've been passing around `Conditional`s without an actual expression since we added `Conditional`, I don't think it makes a lot of sense for a non-`is_conditional()` `Conditional` to exist.

In the long term we should make `Conditional` not constructable without an actual condition, and use `None` and not a no-argument `Conditional()` to represent not having a condition.

In the short term, it's probably least ugly to call this something more like `is_empty()` and invert the sense.

```suggestion
    def is_empty(self) -> bool:
        """Whether this is an empty Conditional."""
        return self.expression is None
```

src/toil/cwl/cwltoil.py
```diff
@@ -3507,35 +3507,42 @@ def makeJob(
...
-                            cast(ToilCommandLineTool, tool),
-                            jobobj,
-                            runtime_context,
-                            parent_name=parent_name,
-                            conditional=conditional,
-                        )
-                        return job_wrapper, job_wrapper
-        # Otherwise, all requirements are known now.
+        if has_dynamic_resource_requirement or (
+            conditional is not None and conditional.is_conditional()
```
adamnovak:
```suggestion
            conditional is not None and not conditional.is_empty()
```

adamnovak (changes requested):
I think we had some preexisting bad design around `Conditional` being able to not actually have a condition, and being able to carry a set of resource requirements that don't seem to be respected. I think this is mostly fine but I'm going to rename/flip one of the methods to avoid introducing a cursed concept of a non-conditional conditional.

adamnovak (approved):

PR #5583, opened 2026-09-21
Slurm: skip scontrol records that do not start with JobId= instead of crashing the leader

### LINKED ISSUE #5553: Slurm batch system: RuntimeError: Unknown jobID, could not be converted
When using Toil with `--batchSystem=slurm` some workflows triggered the following error: "RuntimeError: Unknown jobID, could not be converted"

Logs:

```txt
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 329, in run
    while self._runStep():
          ^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 316, in _runStep
    if self.checkOnJobs():
       ^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 256, in checkOnJobs
    statuses = self.boss.with_retries(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 655, in with_retries
    return operation(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/slurm.py", line 423, in coalesce_job_exit_codes
    status_dict = self._get_job_details(job_id_list)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/slurm.py", line 466, in _get_job_details
    status_dict.update(self._getJobDetailsFromScontrol(job_id_list))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/slurm.py", line 810, in _getJobDetailsFromScontrol
    job[key] += " " + bits[0]  # type: ignore[has-type]  # we depend on the previous iteration to populate key
        ^^^
UnboundLocalError: cannot access local variable 'key' where it is not associated with a value

The above exception was the direct cause of the following exception:

  File "/tmp/venv/lib/python3.12/site-packages/toil/common.py", line 1308, in start
    return self._runMainLoop(rootJobDescription)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/common.py", line 1913, in _runMainLoop
    ).run()
      ^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/leader.py", line 299, in run
    self.innerLoop()
  File "/tmp/venv/lib/python3.12/site-packages/toil/leader.py", line 987, in innerLoop
    update: UpdatedBatchJobInfo = self.batchSystem.getUpdatedBatchJob(maxWait=2)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 583, in getUpdatedBatchJob
    raise self.GridEngineThreadException(
toil.batchSystems.abstractGridEngineBatchSystem.AbstractGridEngineBatchSystem.GridEngineThreadException: Unexpected GridEngineThread failure

During handling of the above exception, another exception occurred:

  File "/tmp/venv/bin/toil-cwl-runner", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/cwl/cwltoil.py", line 4572, in main
    outobj = toil.start(wf1)
             ^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/common.py", line 1310, in start
    self._shutdownBatchSystem()
  File "/tmp/venv/lib/python3.12/site-packages/toil/common.py", line 1919, in _shutdownBatchSystem
    self._batchSystem.shutdown()
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 621, in shutdown
    self.background_thread.killJobs()
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 222, in killJobs
    batchJobID = self.getBatchSystemID(jobID)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/venv/lib/python3.12/site-packages/toil/batchSystems/abstractGridEngineBatchSystem.py", line 116, in getBatchSystemID
    raise RuntimeError("Unknown jobID, could not be converted")
RuntimeError: Unknown jobID, could not be converted
```

It seems the problem is raised in [`_getJobDetailsFromScontrol`](https://github.com/DataBiosphere/toil/blob/a5f22682d5cd7f76825049cfaabcedc080ad7d9b/src/toil/batchSystems/slurm.py#L762) when parsing scontrol stdout. Unfortunately I could not obtain the stdout that triggered this error.

I implemented a fix for this method here: https://github.com/csgroup-oss/toil/commit/4ffb655af991b4713afb91e10f7f1597991c7e6c. It's simple python logic to fix `UnboundLocalError` by ensuring the variable cannot be unbound, I don't know if it will fix my problem but it should prevent the crash for this reason at least.

Is this fix interesting for Toil, should I open a PR?

## Environment

- toil (`toil[cwl]`): 9.5.0
- python: 3.12
- slurm: 23.11.8-BullSequana.1.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TOIL-1853)
┆Issue Number: TOIL-1853

glennhickey:
## Changelog Entry

 * The Slurm batch system no longer crashes the leader when `scontrol show job` output contains a record that does not start with `JobId=` (for example another user's job with a multi-line field); such records are skipped with a warning.

## What was happening

With more than one job outstanding, `_getJobDetailsFromScontrol` runs `scontrol show job` with **no job id**, which dumps every job on the cluster (all users). The output is split on blank lines and each chunk gets `int(job["JobId"])`. Any chunk whose first line is not `JobId=…` blows up, and `with_retries` only retries `CalledProcessErrorStderr`, so the leader dies:

```
Got a <class 'KeyError'>: 'JobId' which is not retriable according to <function AbstractGridEngineBatchSystem.with_retries.<locals>.<lambda> ...>
GridEngine like batch system failure: 'JobId'
...
RuntimeError: Unknown jobID, could not be converted
```

Depending on the shape of the junk chunk the first exception is `KeyError: 'JobId'` (first line has no `JobId=` token, e.g. a whitespace-only line), `KeyError: '<previous key>'` (continuation text after a record that was skipped) or `UnboundLocalError: cannot access local variable 'key'` (continuation text as the very first chunk) — the last is #5553. The trigger is something in *someone else's* job record (a `Comment=` spanning a blank line, say), so it comes and goes with that job: on one Slurm cluster it killed four multi-day cactus-pangenome runs in three weeks (Toil 9.5.0 and 9.6.0a1, after 1.4k–24k jobs each), and reappeared within 8 minutes of a restart while the offending job was still queued.

## The fix

Skip chunks that do not yield an integer `JobId` (logged at warning level with the first 200 characters of the chunk), and initialise `key` so continuation text can't reference it before assignment. Real records parse exactly as before; the existing scontrol tests are unchanged.

## Test

`test_getJobDetailsFromScontrol_many_unparseable_records`: the all-jobs dump carries text before the first record, a foreign job whose `Comment` spans a blank line, and a whitespace-only line between records. On master it fails with `UnboundLocalError`; with the fix it returns the same statuses as `test_getJobDetailsFromScontrol_many_all_exist`.

Fixes #5553.

## Reviewer Checklist

        ```
        ```

## Merger Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013YWGkqE3ND862JF8QRZERV


src/toil/batchSystems/slurm.py
```diff
@@ -814,18 +815,33 @@ def _getJobDetailsFromScontrol(
...
-                        break
+                    if job_id is None:
+                        try:
+                            job_id = int(job["JobId"])
+                        except (KeyError, ValueError):
+                            # Not a job record.  With no job id on the command line scontrol dumps
+                            # every job on the cluster, and a foreign job whose field (Comment,
+                            # JobName, ...) spans a blank line, or a whitespace-only line between
+                            # records, yields a chunk that does not start with JobId=.  Skip it
+                            # rather than take the whole leader down with it.
```
adamnovak, line 835: This is not an option worth mentioning.

src/toil/test/batchSystems/test_slurm.py
```diff
@@ -247,6 +247,30 @@ def call_scontrol(args, **_) -> str:
     return stdout
 
 
+def call_scontrol_unparseable(args, **kwargs) -> str:
+    """
+    Like `call_scontrol`, but the dump of all jobs also carries the kinds of chunk seen on real
+    clusters that do not start with ``JobId=``: text before the first record, a job belonging
```
adamnovak, line 253: Can a real cluster ever have text before the first record?

src/toil/test/batchSystems/test_slurm.py
```diff
@@ -247,6 +247,30 @@ def call_scontrol(args, **_) -> str:
...
+    to another user whose Comment spans a blank line, and a whitespace-only line between two
+    records.  Each is a separate blank-line-separated chunk to the parser.
+    """
```
adamnovak, line 256: Why don't we include an edge-case newline-containing record in our main set of example records, instead of having this function? I don't think it make sense to come up with a concept of "unparseable chunks" that we then test, since the idea of a "chunk" is something internal to the parser. It makes it sound like we think we're testing that the parser handles the cases that it can't handle.

src/toil/test/batchSystems/test_slurm.py
```diff
@@ -433,6 +457,25 @@ def test_getJobDetailsFromScontrol_many_none_exist(self):
         result = self.worker._getJobDetailsFromScontrol(list(expected_result))
         assert result == expected_result, f"{result} != {expected_result}"
 
+    def test_getJobDetailsFromScontrol_many_unparseable_records(self):
+        """
+        Asked about more than one job, scontrol is run with no job id and dumps every job on
+        the cluster, so the output can carry chunks that do not start with ``JobId=``: text
+        before the first record, another user's job with a field spanning a blank line, a
+        whitespace-only line between records.  They must be skipped, not crash the leader
+        with KeyError or UnboundLocalError (#5553).
```
adamnovak, line 466: This should be exclusively about the properties of the jobs that we are testing that we tolerate (i.e. that they contain arbitrary text with newlines in field values), and the desired correct behavior (which is that we can correctly understand all other jobs).

src/toil/batchSystems/slurm.py
```diff
@@ -814,18 +815,33 @@ def _getJobDetailsFromScontrol(
...
-                        logger.log(
-                            TRACE, "%s job %d is not in the list", args[0], job_id
-                        )
-                        break
+                    if job_id is None:
+                        try:
+                            job_id = int(job["JobId"])
+                        except (KeyError, ValueError):
+                            # Not a job record.  With no job id on the command line scontrol dumps
+                            # every job on the cluster, and a foreign job whose field (Comment,
```
adamnovak, line 832: Why would the reader think we might have passed a job ID?

adamnovak (commented):
If `scontrol show jobs` is using in-band signaling to delimit field values, we can't *safely* parse its output, because somebody could put a comment in their job that we would parse as a job record saying that our job failed.

I've asked on the slurm-users mailing list what we might be expected to do about this. Since the comment appears so late in the output for a job, and everything we care about appears pretty early, we can sort of get away with throwing away everything after a field containing a newline, as long as the field contents never look like real records and our own Toil jobs don't have any newlines. Even if we're in a directory containing a newline, I think all the directory fields come late enough.

But the code here is saying some worrying things that makes it sound like there can be material in the stream that isn't part of a job record. It's not that there's garbage in the `scontrol` output that isn't part of a record or that is a malformed record, it's that the records are not in fact nicely delimited and unambiguously parseable like we thought they were.

Probably what we *really* want to do, if we're stuck parsing this format, is to change they way we recognize record boundaries and use the `JobID=#####` at the start of the line to recognize a start, and the indent and`<key>=` at the start of an internal line to recognize a new field, instead of trying to chunk it up and *then* parse it. Then we would be able to parse all but the most adversarial comments.

But if we don't want to implement that, we could stick with this general approach as long as we explain better how it can't work in theory and only works in practice.

glennhickey:
@adamnovak thanks for looking at this, and sorry for the 100%-claude-made PR.   But it does raise a real (and annoying issue) that I ran into the other day:  it seems like someone (I wasn't able to trace who) can launch a slurm job that kills my toil runs, jsut by virtue of how their job is formatted

adamnovak:
Yeah, I think we do need something like this to sort out our own work, even if it's a hack.

I heard back from the Slurm mailing list, and in newer versions of Slurm (it looks like 24.05 and later), there's a `--json` option for `scontrol show job` that lets you get the information in a machine-readable way. We're still on 21.08, but Toil should probably support the new way first.

There's also a `--oneline` option, which we *do* have, which might simplify the parsing. As I understand it it doesn't *actually* remove newlines from fields, though.

It was also suggested that we get our admins to attach a Lua script in Slurm's `job_submit.lua` to remove special characters from job names and job comments, if we don't like them.

adamnovak:
OK, I've changed this to just delimit the records with the `JobId=` lines, instead of trying to use a concept of partial or unparseable records. Now we can parse the multi-line values (though we replace the newlines with other whitespace). As long as nobody includes key=value stuff *inside* a field, everything should work.

adamnovak:
I've also pulled this in for testing, on the `issues/5553-scontrol-unparseable-records` branch.

adamnovak (approved):

